PR #658, opened 2014-07-26
Support Bower's main-files feature

kpullen:
Bower packages can provide a list of files that are important to
projects using this package. Some build tools use this list to copy,
concatenate, or munge the contents of the files in some way without
touching the files in `bower_components`.

This commit adds the Polymer files to the list of files that should be
delivered to a build tool.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -4,5 +4,11 @@
   "dependencies": {
     "platform": "Polymer/platform#master",
     "core-component-page": "Polymer/core-component-page#master"
-  }
+  },
+  "main": [
+    "layout.html",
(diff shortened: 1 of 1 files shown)


adewinter:
:+1: 
I was about to create a PR exactly like this myself.  Please accept! :)

razvanz:
+1   This should also happen in any other polymer components.

tjsavage:
This makes sense to me. We recently made a major change to the `master` branch of this repo with the 0.9 release - would you mind re-submitting a PR here rebasing from the latest `master`? Thanks!

PR #985, opened 2014-12-01
[0.8] more-feature supporting WebAnimations

nevir:
Hopefully makes web animations a little easier to declare within polymer elements.

Note that this gives us all the same shorthand behavior as `element.animate` (i.e. groups). For the next iteration, I want to tackle shorthands for groups and sequences (which potentially have different target selectors)

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,7 +18,8 @@
   },
   "devDependencies": {
     "web-component-tester": "*",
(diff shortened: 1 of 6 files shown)


addyosmani:
Unsure if the intention is to still pull this in. Needs rebasing and a test re-check :)

tjsavage:

PR #1077, opened 2015-01-07
Add [extend] option for layout

YarmoM:
The extend tag allows (for example) paper-shadow to extend to the full
width of the parent div when the content is smaller than the parent.

When the parent does not have a fixed width or height, the fix tag
breaks the layout and in that case, the extend tag can be used.

### DIFF
--- a/layout.html
+++ b/layout.html
@@ -260,6 +260,11 @@ html /deep/ [fit] {
   left: 0;
 }
 
+html /deep/ [extend] {
+  width: 100%;
+  height: 100%
+}
+
 body[fullbleed] {
   margin: 0;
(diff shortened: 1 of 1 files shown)


tjsavage:
Thanks for the PR! We recently made a major change to the `master` branch of this repo with the 0.9 release - would you mind re-submitting a PR here rebasing from the latest `master`? For this change in particular, layout has been separated out to https://github.com/polymer/layout

arthurevans:
@tjsavage I think the latest greatest version of the layout is actually in https://github.com/PolymerElements/iron-flex-layout ... should use that going forward.

PR #1086, opened 2015-01-10
Grunt "retire" task added

jesusprubio:
All dependencies are fine now but I think it could be a good idea to automate the check process.

polymerretire

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -86,6 +86,16 @@ module.exports = function(grunt) {
         }
       }
     },
+    retire : {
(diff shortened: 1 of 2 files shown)


jesusprubio:
CLA signed! :)

samccone:
Hey @jesusprubio Thans for the super cool PR ... TIL http://retirejs.github.io/retire.js/

I think this task is best left up to someone consuming polymer (as in part of their deployment / dev script) as compared to being part of the core build process.

Does anyone else have thoughts on this? 

cc @ebidel

kevinpschaaf:
Thanks for the PR.  After review, we decided this is best left to Polymer consumers and we won't merge this to Polymer.

Issue #1127, opened 2015-01-27
Reference to `HTMLLinkElement` in `Polymer.import` callback


tomalec:
I'm trying to use `Polymer.import` helper to dynamically load some documents with Polymer Elements nested deeper inside. With native `HTMLLinkElement#onload` I could easyli access imported content to use it:
```js
link.onload =  function(){
  // Insert template into DOM.
  var template = this.import.querySelector("template");
```
live example [here](http://plnkr.co/edit/EfNg4Lgb7pHf2oKjwRNg?p=info)
And it works like charm, except the problem with 
> Attributes on my-element were data bound prior to Polymer upgrading the element. This may result in incorrect binding types.

As `onload` may be called before Polymer upgrades the polymer elements. So I need either to hack it with Polymer's internal `queue` or at least with `Polymer.whenReady` like [here](http://plnkr.co/edit/eknhZINbrxdoEKmgV9oz) or finally use `Polymer.import` which seems nice and clean.
However it does not give a reference to imported document neither with `this` nor via `arguments`.
So I need to write every time:
```js
Polymer.import(["/my/path/here"], function(){
  var link = document.querySelector('link[href="/my/path/here"]');
  var template = link.import.querySelector("template");
```
[live](http://plnkr.co/edit/Wh7BCEPvKZ5ut3UNVIcu)
What doesn't look that nice, and may hit the performance.

I suggest adding array of corresponding `HTMLLinkElement`s as callback argument:
```js
Polymer.import(["/my/path/here"], function(links){
  var template = links[0].import.querySelector("template");
```
So `Polymer.import` could be used as and alternative to native `.onload`.

ebidel:
+1. In a current project, I've also been using `document.querySelector` to get back the import that was added but `Polymer.import` should just pass this info in the callback.

@sorvell @sjmiles is this something we can add?

tomalec:
Thanks, Should I create PR (#1128) also for [Polymer/polymer-dev](https://github.com/Polymer/polymer-dev)

PR #1128, opened 2015-01-27
Add array of `HTMLLinkElement`s to `Polymer.import` callback

tomalec:
That is my proposed solution for suggested feature #1127


ebidel:
@sjmiles @kevinpschaaf @sorvell Is this good to add? We're doing the workaround for the IO site and would like to use this instead.

ebidel:
This has gone stale with the 1.0 rewrite. Closing as a result. If you'd like to see this added, please resubmit a fresh PR.

PR #1152, opened 2015-01-31
[0.8] Fix bespoke element constructors

### LINKED ISSUE #1151: [0.8-preview] Bespoke element constructors
http://jsbin.com/puvaze/3/edit?html,output

Bespoke element constructors, mentioned in the [PRIMER.md](https://github.com/Polymer/polymer/blob/0.8-preview/PRIMER.md#bespoke-constructor-support), don't seem to work because `this` doesn't seem to be correctly bound.  As such `this.is` is incorrect and nothing good gets passed to `document.createElement()` :(

peterwmwong:
Fixes #1151

### DIFF
--- a/src/features/micro/constructor.html
+++ b/src/features/micro/constructor.html
@@ -43,8 +43,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // registration-time
 
       _prepConstructor: function() {
+        var ctor;
         if (!this.hasOwnProperty('constructor')) {
-          var ctor;
           if (this.extends) {
             ctor = function() {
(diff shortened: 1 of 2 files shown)


addyosmani:
cc @sjmiles @kevinpschaaf

kevinpschaaf:
Fixed in https://github.com/Polymer/polymer/commit/24a29141ca7001ba2f0e4f97795a82bbc9e77620 (didn't take PR #1152 as the implementation has changed a bit). Note the API change, such that the custom constructor need not call `document.createElement` and return the element.  This will be done automatically; the user constructor should only configure the instance (`this`).

PR #1156, opened 2015-02-02
Fixed type annotations for primitives.

jayphelps:
JavaScript primitive types are not actually instances of their own constructor and so technically these type declarations were incorrect. e.g. if you actually provided a `Boolean` instance, it does not act like the primitive `true|false` as one might think.

```javascript
var value = new Boolean(false);
!!value;
// true, because it is an Object, which is always true...
```

Obviously...this is nitpicking since no one really does this..but it's a simple fix so...why not. :beers:

### DIFF
--- a/src/declaration/path.js
+++ b/src/declaration/path.js
@@ -25,10 +25,10 @@
   *     <a href="{{resolvePath(path)}}">Resolved</a>
   * 
   * @method resolvePath
-  * @param {String} url Url path to resolve.
-  * @param {String} base Optional base url against which to resolve, defaults
+  * @param {string} url Url path to resolve.
+  * @param {string} base Optional base url against which to resolve, defaults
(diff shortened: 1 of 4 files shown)


jayphelps:
CLA signed.

samccone:
@kevinpschaaf  is this PR still relevant?

jayphelps:
@samccone the idea, yes; there are still places where `{String}` is used instead of `{string}`...but this PR is obviously outdated as the code changed significantly after 0.5.

jayphelps:
I do not see any incorrect `Boolean` annotations though :tada:

samccone:
awww yess thanks @jayphelps also, any future PRs to resolve the String issue are more than welcome, thanks a bunch!!! :grapes:

PR #1186, opened 2015-02-13
Stores args for import callback

### LINKED ISSUE #1127: Reference to `HTMLLinkElement` in `Polymer.import` callback

### LINKED ISSUE #1184: Polymer.Import - handle 404's
I use polymer.import to load content on my page that is based of the url. When a user goes to a page that does not exist I need to be able to handle the 404 response my server sends. I am unable to do this currently with Polymer.import. 

This is a feature request, not a bug report, but I wasn't sure where to make the request. If there is a place please let me know.

thanks
Tim

pflannery:
fixes #1184 
fixes #1127 

requires PR https://github.com/webcomponents/webcomponentsjs/pull/199

Example usage:
```js
Polymer.import(["http://url1", "http://url2", "http://etc.."], function(detail) {
  var hadPriorImports = detail.priorImports.length > 0;
  var hasLoadedImports = detail.loadedImports.length > 0;
  var hasErrorImports = detail.errorImports.length > 0;
}
```

### DIFF
--- a/src/declaration/queue.js
+++ b/src/declaration/queue.js
@@ -146,18 +146,18 @@
       requestAnimationFrame(this.flushReadyCallbacks);
     },
 
-    addReadyCallback: function(callback) {
+    addReadyCallback: function(callback, args) {
       if (callback) {
-        readyCallbacks.push(callback);
+        readyCallbacks.push({fn: callback, args: args});
       }
     },
 
     flushReadyCallbacks: function() {
       if (readyCallbacks) {
-        var fn;
+        var callback;
         while (readyCallbacks.length) {
-          fn = readyCallbacks.shift();
-          fn();
+          callback = readyCallbacks.shift();
+          callback.fn(callback.args);
         }
       }
     },
@@ -198,8 +198,8 @@
   function whenReady(callback) {
     queue.waitToReady = true;
     Polymer.endOfMicrotask(function() {
-      HTMLImports.whenReady(function() {
-        queue.addReadyCallback(callback);
+      HTMLImports.whenReady(function(args) {
+        queue.addReadyCallback(callback, args);
         queue.waitToReady = false;
         queue.check();
     });


src/declaration/queue.js
```diff
@@ -198,8 +198,8 @@
   function whenReady(callback) {
     queue.waitToReady = true;
     Polymer.endOfMicrotask(function() {
-      HTMLImports.whenReady(function() {
-        queue.addReadyCallback(callback);
+      HTMLImports.whenReady(function(e) {
+        queue.addReadyCallback(callback, e);
```
ebidel: e -> args?

pflannery: haha yeah your right, I will make it ```detail```

addyosmani:
> requires PR webcomponents/webcomponentsjs#199

The polyfills portion of this work has since been reviewed, tested and merged.

pflannery:
@ebidel renamed e -> args PTAL

When satisfied I will squash to one commit again

ebidel:
LGTM from me. @sorvell should take a look.

tjsavage:

Issue #1192, opened 2015-02-13
[0.8-preview] x-repeat standalone issue


frankiefu:
```html
<!doctype html>
<html>
<head>
  <title></title>
  <meta charset="utf-8">
  <link rel="import" href="../../neoprene/components/polymer/polymer.html">
</head>
<body>

<template is="x-repeat" id="repeat">
  <div>{{id}}</div>
</template>

<script>

  var t = document.querySelector('#repeat');
  t.items = [{id: 0}, {id: 1}, {id: 2}];

</script>

</body>
</html>
```

got error: Uncaught TypeError: Cannot set property '_ctor' of undefined templatizer.html:54

diddledan:
diggin-up an old issue. I am getting this specific error on a Universal Windows App that I'm attempting to build - it seems that even with the check on line 103 thru 105 that `template.content` is also `undefined` so the test is assigning the `undefined` `template.content` instead of the `undefined` `template._content` - as both are undefined then _ctor still cannot be accessed when using the code as it stands. I am unclear why `template.content` and `template._content` are BOTH `undefined` though so I can't really help with that bit of diagnosis.

diddledan:
ok I think it's safe to ignore me - I have got further by changing my polyfill reference from `webcomponents-lite.js` to `webcomponents.js`

PR #1199, opened 2015-02-17
Basic test for x-template stamping

nevir:
See https://github.com/Polymer/polymer/issues/1192

https://github.com/Polymer/polymer/commit/a292fd7d8a12846be70ae4b0a23996927883fc7a does cause the test to pass


nevir:
Oh, sorry, I didn't realize that I sucked in 0.8-template-fix in the history for this! :+1: though

PR #1243, opened 2015-03-02
Unique comp props

ssorallen:
If multiple computed properties call the same method but with different
arguments, calls to the method are debounced because they pass the same
value as the task name. Rather than name the task only the method name,
use the full expression to let the developer ensure uniqueness.

Example:

```js
computed: {
  foo: 'computeClassName(bar)',
  baz: 'computeClassName(xyz)'
}
```

Currently they both create property effects wrapped in debounce calls
with task names of '_computeClassName':

```js
this.debounce('_computeClassName', function() { ... })
this.debounce('_computeClassName', function() { ... })
```

By passing the full expression rather than the method name, they result
in unique debounce task names:

```js
this.debounce('_computeClassName(bar)', function() { ... })
this.debounce('_computeClassName(xyz)', function() { ... })
```


peterwmwong:
+1

sorvell:
LGTM Thx! (will merge after conflicts are resolved)

PR #1256, opened 2015-03-05
0.8 property config

kevinpschaaf:
* Renames `published` configuration object to `properties`
   * Adds `computed` field to `properties`. Replaces use of general `computed` block
   * Adds `observer` field to `properties`. Replaces use of general `bind` block for single-property observers
   * Adds `value` field to `properties`. Each property in `properties` accepts a `value` field to define a default value; may either be a static value or function that returns a value (e.g. for initializing objects/arrays); replaces use of general `configure` function (although general function still exists for exotic use cases)
* Removes general `computed` configuration object (these must move to `properties` per property)
* Renames `bind` to `observers` (general `observers` configuration object remains for path and multiple-dependency observers; will still work for single-property observers, although these should generally go in `properties`)
* Adds multiple-dependency observer support to general `observers` object
* Fixes attribute deserialization loop when setting `reflect: true` properties
* Adds attribute deserialization tests

Minimal element changes required for compatibility with this API update:
* Rename `published` to `properties`
* Move all `computed` properties into `properties` object
* Rename `bind` to `observers`

Not strictly required, but should update to match style going forward:
* Move all single-property `observers` (formerly `bind`) into `observer` field of `properties` object
* Move all default values defined in `configure` to `value` field of `properties` object (and remove `configure` function)


property-config.html
```diff
@@ -0,0 +1,159 @@
+<script>
```
sorvell: delete this file.

polymer-micro.html
```diff
@@ -12,7 +12,7 @@
 <link rel="import" href="src/features/micro/mixins.html">
 <link rel="import" href="src/features/micro/extends.html">
 <link rel="import" href="src/features/micro/constructor.html">
-<link rel="import" href="src/features/micro/published.html">
+<link rel="import" href="src/features/micro/property-config.html">
```
sorvell: let's name this properties.html

sorvell:
LGTM with minor comments.

Issue #1241, opened 2015-03-02
[0.8] value binding not working in samples.html


dhubler:
If you open
  bower_components/polymer/explainer/samples.html
in your web browser you see
   Hello World, my user is nobody.
when you should see
   Hello World, my user is Scott.
based on the example source.

PR #1261, opened 2015-03-05
Fixes more #1241

Pofigizm:
I haven't `ready` function when i use polymer-micro.html
I think, it's better to use `attached` function, because it is basic lifecycle callbacks.

### DIFF
--- a/PRIMER.md
+++ b/PRIMER.md
@@ -235,8 +235,8 @@ Polymer({
     }
   },
 
-  ready: function() {
-    this.innerHTML = 'Hello World, I am a <b>Custom Element!</b>';
+  attached: function() {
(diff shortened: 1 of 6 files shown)


Pofigizm:
My CLA has been submitted.

arthurevans:
@sorvell or @kevinpschaaf should probably review this one.

Pofigizm:
In PRIMER.md:274 also wrong example. Need change to attached:

Pofigizm:
I don't want include https://github.com/Pofigizm/polymer/commit/7d2398888f4d6049d59d82931cf678f8636d7317 to this PR. I don't understand why it came here.

PR #1264, opened 2015-03-06
0.8 demodulate

sjmiles:
Replaces `using` and `modulate` with Polymer.<module name>.

We like modules, but until native modules are supported we would have to rely on some polyfill or other placeholder. Occam says: instead just use Polymer namespace.


sorvell:
LGTM

sorvell:
Polymer.dom = {dom: 'shadow'} is not working with these changes, but we'll correct that in a follow on commit.

PR #1265, opened 2015-03-07
[0.8] Fix x-repeat: demodulate ArrayObserve, fix references in Collection and x-repeat

peterwmwong:


peterwmwong:
[nvm](https://github.com/Polymer/polymer/commit/ecd35bd30ac227c16493534322ae883a91f4c1c7)

PR #1295, opened 2015-03-15
Bugfix: ReferenceError: c is not defined

atotic:
Oversight: c needs to be declared. Fixed by looking at what c was in other routines.


kevinpschaaf:
Thanks!

PR #1296, opened 2015-03-15
Separate attributes function from annotations

Pofigizm:
It is repeated pull request, after rebase upstream branch.

Original vulcanized polymer-micro and polymer-mini has size 10 kb and 30 kb. After merge my pull request vulcanized polymer-micro and polymer-mini will have size 6 kb and 26 kb.

It is not good idea include all `annotations` features in polymer-micro and polymer-mini, i think.

@kevinpschaaf, @sorvell, @nevir, @sjmiles, @ebidel, can you comment?


kevinpschaaf:
Great catch!  These commits were cherry-picked into https://github.com/Polymer/polymer/issues/1300 and merged there.  Thanks again.

PR #1307, opened 2015-03-18
Make webcomponentsjs a bower dependency

robdodson:
This change would make it so webcomponentsjs is installed when I bower install polymer

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -16,9 +16,11 @@
     "type": "git",
     "url": "https://github.com/Polymer/polymer.git"
   },
-  "devDependencies": {
-    "web-component-tester": "*",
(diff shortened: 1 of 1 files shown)


nevir:
We should probably point it to the latest released copy (drop the `#master`)

robdodson:
@nevir I'm cool with that. @garlicnation is 0.6 rolling out soon?

nevir:
I think the latest release of it is safe - @sorvell do you know what the
earliest supported version of the polyfill is?

addyosmani:
Pinging @sorvell :)

sorvell:
LGTM

PR #1309, opened 2015-03-19
0.8 gestures

frankiefu:


arthurevans:
What did we just add here? And is this in 0.8, or post-0.8?

frankiefu:
This is wip gesture system (currently on-tap and on-track are somewhat supported).  @azakus will continue to work it next week.  My understanding is this is not part of the 0.8 official release.  @sorvell can confirm.

peterwmwong:
FYI - I noticed that the track gesture introduced in this PR breaks touch scrolling.  [PolymerElements/paper-drawer-panel](https://github.com/polymerelements/paper-drawer-panel) content cannot be scrolled on mobile devices because the track gesture is calling `preventDefault()` on touch move [here](https://github.com/Polymer/polymer/pull/1309/files#diff-051391d48752db6ed28755b4c2fdf5c0R208).

Reduction: http://jsbin.com/wocoba/12/edit?html,output vs http://jsbin.com/wocoba/13/edit?html,output

azakus:
This should be addressed by the complete gestures rewrite in #1486

PR #1320, opened 2015-03-20
Shady DOM optimizations: (1) fast path distribution when no insertion po...

sorvell:
...ints exist in a root, (2) avoid tree walking root for insertion points by keeping an ordered list of them.


frankiefu:
lgtm

PR #1335, opened 2015-03-25
0.8 custom notify event

kevinpschaaf:
Two-way binding interop support for elements that don't notify property changes using `property-change` convention.  Event to trigger upward propagation specified in binding, delimited by `::`.

```html
<input value="{{myValue::change}}">

<video currenttime="{{time::timeupdate}}">
```


sorvell:
LGTM

PR #1351, opened 2015-03-31
Fix undeclared variable exception in _distributeInsertionPoint.

icetraxx:
- Fix `children` is an undefined variable in `_distributeInsertionPoint`.
  - This caused an exception to be thrown when you instantiate a polymer element that uses localdom without any distributed childNodes.
- Removed trailing whitespace and removed unnecessary variable redeclarations.


abi:
:+1:

sorvell:
LGTM... 

It's unclear how this generated an exception (http://jsbin.com/tumubuligo/1/edit).

PR #1369, opened 2015-04-03
Update PRIMER.md

ragingwind:
I've updated the document cause of below reasons. Please let me know If I have missed

- Code consistencies
- Fix intent
- Fix some errors
- Fix typo

### DIFF
--- a/PRIMER.md
+++ b/PRIMER.md
@@ -114,7 +114,7 @@ Because the vast majority of users will always want to register the custom eleme
 Example:
 
 ```js
-MyElement = Polymer({
+var MyElement = Polymer({
 
   is: 'my-element',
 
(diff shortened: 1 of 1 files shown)


addyosmani:
@arthurevans With your in-flight docs, do you see a need for the primer to continue being evolved in the repo? Probably worth looping in @kevinpschaaf too as it appears to continue to be a good dumping ground for experimental APIs that are in flux.

arthurevans:
I think we're planning to keep the primer up for the near future at least, so the core team can record code changes as they happen.

ragingwind:
I close this PR. I think, It seems not necessary to be updated now. Separated from else an issue

PR #1372, opened 2015-04-04
Clarify concept of host to mean dom host. Related:

sorvell:
* remove configuration element list after configuration.
* shady dom dynamically re-renders only dirty roots.
* x-scope styling styles elements found via dom.
* make template stamped element hosted by templateInstance
* marshal templates at config time and make x-repeat render in ready.


sorvell:
@kevinpschaaf gave LGTM.

PR #1418, opened 2015-04-16
Update documentation for `hostAttributes` and friends to match the latest

nevir:
screen shot 2015-04-16 at 4 57 12 pm

### DIFF
--- a/src/micro/attributes.html
+++ b/src/micro/attributes.html
@@ -12,50 +12,37 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 <script>
 
-  /**
-   * Support for `hostAttributes` property.
-   *
-   *     hostAttributes: 'block vertical layout'
-   *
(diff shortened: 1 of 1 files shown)


src/micro/attributes.html
```diff
@@ -12,51 +12,38 @@
...
+     * `serialize()`.
+     *
+     * For example:
+     *
+     *     Polymer({
+     *       is: 'x-thing',
+     *       hostAttributes: {
+     *         tabindex: 0,
+     *         role: 'button',
+     *         'aria-disabled': true
```
arthurevans: Note that in the case of ARIA, this should actually be a string-valued attribute, `'aria-disabled': 'true'`.

tjsavage:
We recently rolled the 0.9 branch into the latest `master` - would you mind re-submitting a PR here rebasing from the latest `master`? Thanks!

PR #1422, opened 2015-04-18
0.8 path effects

kevinpschaaf:
Adds support for path arguments (including wildcards) to multiple-property observers and computed functions.  Adds docs & tests for recent additions.


sorvell:
LGTM

PR #1436, opened 2015-04-22
No more circular dependencies, because bower.

nevir:


sorvell:
This is a temporary measure until we can fix the testing infrastructure to not require this.

LGTM.

PR #1439, opened 2015-04-22
Fixes style scoping when elements are stamped inside repeats.

sorvell:
Removes `element.host` which was used by the platform (e.g. <a>.host), in favor of `element.domHost`. This is just a shorthand for `Polymer.dom(element).getOwnerRoot().host`


kevinpschaaf:
LGTM

PR #1448, opened 2015-04-24
[0.8] ready isn't bottom-up for elements in the same scope

nevir:
We [advertise](https://github.com/Polymer/polymer/blob/0.8-preview/PRIMER.md#ready-callback) the `ready` callback as being bottom up. And it is, for every shadow/shady root. However, for elements declared in the same scope, `ready` fires top-down. The tests indicate this too.

Is this WAI? It certainly seems a bit wonky. The primer does seem to word things to support top-down inside each element template (we might want to add a caveat if it is WAI)

I've updated the tests to what I would _expect_ to be the order (though `'x-zot#d', 'x-zot#b', 'x-zot#c', 'x-zot#a'` would also be proper)

### DIFF
--- a/test/unit/ready.js
+++ b/test/unit/ready.js
@@ -1,7 +1,7 @@
 suite('ready', function() {
 
   var configure = ['x-ready', 'x-zot#a', 'x-zot#b', 'x-zot#c', 'x-zot#d', 'x-foo#foo', 'x-bar#bar1', 'x-zot', 'x-bar#bar2', 'x-zot'];
-  var ready = ['x-zot#a', 'x-zot#b', 'x-zot#c', 'x-zot#d', 'x-zot', 'x-bar#bar1', 'x-zot', 'x-bar#bar2', 'x-foo#foo', 'x-ready'];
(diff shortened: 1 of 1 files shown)


sorvell:
This is WAI. Elements in the same scope ready in document order. Elements in a scope are all ready before the host of the scope is ready. Depending on the ordering of ready between scope-siblings is generally problematic, and instead the scope host should be used as a mediator. 

Is there a use case that led to this issue?

nevir:
re: use case: yeah, in more-routing I support nested routes, and I need to be able to guarantee that parent elements have set up their state before children attempt to read it (came up when porting it). It sounded like @morethanreal needed something similar for resizing too.

The code to do it isn't awful (in `ready`, walk up the tree and activate your parents first, then do work), just fiddly. It might be generalizable as a behavior, actually

arthurevans:
I would guess this is more of a doc issue then? The phrase "bottom-up"
seems misleading here.

sorvell:
In the same scope, it's best for children to inform parents of their
existence. We need to promote/explain this pattern better.

PR #1453, opened 2015-04-25
[0.8] Fix argument order call to transform() from translate3d()

peterwmwong:
PR #1440 changed the order of arguments for `Polymer.Base.transform` from `node, transform` to `transform, node` (https://github.com/Polymer/polymer/pull/1440/files#diff-0fa4a3dd1b881640e771f03a86c32de5L96).

Unfortunately, this introduced a regression as `Polymer.Base.translate3d`'s call to `Polymer.Base.transform` was not updated.

This breaks paper-elements, like `<paper-ripple>`, which use `Polymer.Base.translate3d()` to render ripples.

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -219,14 +219,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * property.
      *
      * @method translate3d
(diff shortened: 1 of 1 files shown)


peterwmwong:
Fixed by #1478

PR #1472, opened 2015-04-30
[0.8] Fixes #1471

### LINKED ISSUE #1471: [0.8] Binding the value of an input box with {{value::input}} loses caret index
When binding a value to an input box using the custom notify event syntax, you lose the caret index when adding a character. The selectionStart always jumps to the end.

```html
<input value="{{value::input}}" />
```

http://jsbin.com/situwa/2/edit?html,output

When digging in the ***effects.html***, the value of the input is set, even though it has the same value already.

```js
_applyEffectValue: function(value, info) {
    var node = this._nodes[info.index];
    // TODO(sorvell): ideally, the info object is normalized for easy
    // lookup here.
    var property = info.property || info.name || 'textContent';
    // special processing for 'class' and 'className'; 'class' handled
    // when attr is serialized.
    if (info.kind == 'attribute') {
        this.serializeValueToAttribute(value, property, node);
    } else {
        // TODO(sorvell): consider pre-processing this step so we don't need
        // this lookup.
        if (property === 'className') {
          value = this._scopeElementClass(node, value);
        }
        return node[property] = value;
    }
}
```

At the end of this method ```return node[property] = value;``` causes the input's value to be overwritten with the same value which moves the caret to the end.

Can we maybe change it to the following?

```js
if(node[property] !== value)
    node[property] = value;

return value;
```

Sleeckx:


kevinpschaaf:
Polymer elements perform that dirty check inside the setter for properties, so this change would effectively result in double work for all binding propagation.

Instead, we'll do a similar test only in the case that we use a custom event (which will generally only be for non-Polymer elements), per https://github.com/Polymer/polymer/commit/275c41a82b80260c20a4368e6ea394f2c9445802.

kevinpschaaf:
Thanks for reporting the issue and helping us find the right solution!

PR #1474, opened 2015-05-01
don't trap `id` for marshalling if it's a binding directive + test

sjmiles:
Also factored out `_testEscape` method which unifies test for `{{` and `[[`.


frankiefu:
lgtm

PR #1477, opened 2015-05-01
0.8 array notification

kevinpschaaf:
Removes use of ArrayObserve.observe.  Instead, users must use `this.push()`, etc. on elements to mutate arrays.  Collection is also no longer observable; simply stands as a store of key->item and item->key mappings for arrays.

Renames `getPathValue` and `setPathValue` to `get` and `set`.

`get` and `set` can now take an array of path parts, to avoid needing to concat, e.g. `set([items', 4, 'name'], 'bob);`, and also does automatic array index-to-key mapping when notifying path (`4` in example would be replaced with the key for `items[4]` when notifying).

Renames `_data` to `__data__` to avoid collision with user data.


arthurevans:
What's the motivation for this change? Is Array.observe too heavy-weight, or just not working right?

kevinpschaaf:
@arthurevans Prior to this change, we used Array.observe (on Chrome) but on all other browsers overrode array instances with patched push/pop/etc. methods to be able to observe when changes occurred (remember, we're avoiding the observe-js polyfill overhead for Polymer proper).  The patching approach was not only intrusive, but led to a hard-to-explain API w.r.t. making changes to structured data (if you want to set set `this.items[4]` and have Polymer see it, you need to use `this.set`, but if you want to push onto `this.items`, you could just push onto the array).  Now, `all` mutation of objects/arrays should be done through Polymer.Base API's:
* get
* set
* push
* pop
* shift
* unshift
* splice

sorvell:
LGTM

PR #1478, opened 2015-05-01
Fix translate3d call after transform refactor

azakus:


src/standard/utils.html
```diff
@@ -219,14 +219,14 @@
      * property.
      *
      * @method translate3d
-     * @param {HTMLElement} node Element to apply the transform to.
      * @param {number} x X offset.
      * @param {number} y Y offset.
      * @param {number} z Z offset.
+     * @param {HTMLElement} node Optional Element to apply the transform to.
```
nevir: For optional arguments, you can call that out via the type expression:

    @param {HTMLElement=} node Element to apply the transform to.

azakus: We should go over the whole codebase. There are plenty of "Optional in description only" type expressions. I'll fix them for this PR, just noting for future reference.

PR #1482, opened 2015-05-04
0.8 x style

sorvell:
* rename `x-style` --> `custom-style`
* rename `mixin` --> `@apply`
* support `:root` selector
* variable defaults` var(--a, red)`
* fixes bug: `c--foo:` is seen as a custom property
* consume variable, mixin inside mixin

   ```css
   --foo: {
      color: var(--my-color);
      @apply(--my-theme);
   };
   ```
* consume custom properties in `custom-style` (short-term workaround to support `@extend`-use case style sharing

### DIFF
--- a/polymer.html
+++ b/polymer.html
@@ -86,7 +86,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 </script>
 
-<link rel="import" href="src/lib/x-style.html">
+<link rel="import" href="src/lib/custom-style.html">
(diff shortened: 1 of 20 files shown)


kevinpschaaf:
LGTM

arthurevans:
What's a :root when it's at home?

sorvell:
We'll add info to the primer.

It's just this (http://www.w3schools.com/cssref/sel_root.asp) and it's a
way to define custom properties for the entire root scope (where previous
we used '*')

Issue #1444, opened 2015-04-23
[0.8] behaviors override properties on elements


nevir:
```js
Polymer({
  is: 'sad-face',
  behaviors: [{mood: 'sad'}],
  mood: 'happy',
});
console.log(document.createElement('sad-face').mood);
// logs "sad", expected "happy"
```

jklein24:
+1. This seems really backwards to me.

sorvell:
Agreed, we will be changing this.

sorvell:
Fixed via https://github.com/Polymer/polymer/commit/bcd2923849777b88fbed446234f2b5222c6e8f1b

nevir:
:cake:

PR #1483, opened 2015-05-04
Update the Behaviors section to reflect the change in #1444

MatrixFrog:


jklein24:
cc @arthurevans

PRIMER.md
```diff
@@ -363,9 +363,9 @@ Results in:
 
 Polymer supports extending custom element prototypes with shared code modules called "behaviors".
 
-A behavior is simply an object that looks very similar to a typical Polymer prototype.  It may define lifecycle callbacks, `properties`, `hostAttributes`, or other features described later in this document like `observers` and `listeners`.  To add a behavior to a Polymer element definition, include it in a `behaviors` array on the prototype.
+A behavior is simply an object that looks very similar to a typical Polymer prototype.  It may define lifecycle callbacks, `properties`, `hostAttributes`, or other features described later in this document like `observers` and `listeners`.  To add a behavior to a Polymer element definition, include it in a `behaviors` array on the prototype.  If multiple behaviors for the same element define the same function, only the last behavior's function is used.
```
arthurevans: This sentence is before the sentence that describes adding arbitrary (non-lifecycle callback) functions.

If it's specifically about these functions, it should move to the next paragraph. If it's talking about the `properties` object, etc., should be rephrased.

(Also, @sjmiles @kevinpschaaf -- I don't know how this works. We should clarify what happens if multiple behaviors include a `properties` object, etc. Are each of these merged into the host element's `properties` object, etc.?

kevinpschaaf: There is no merging; only "set if not already defined."

arthurevans: Does that mean that if the element has a properties object, the behavior can't define any properties? That doesn't seem right. Or that if the host defines a specific property, the behavior's definition of that same property is ignored?

tjsavage:
We recently made a major change to the `master` branch of this repo with the 0.9 release - would you mind re-submitting a PR here rebasing from the latest `master`? Thanks!

MatrixFrog:
I'm unclear on what the exact behavior is, and it sounds like some of the Polymer team are as well, given that there are some unanswered questions in this thread. Maybe someone on the team can make the change, since they will know how best to phrase things.

PR #1518, opened 2015-05-12
Implement a minimum track distance

### LINKED ISSUE #1496: [0.9] touch track fails on iPhone, .touchIdentifier vs .identifier
In my tests `ev.changedTouches[0].touchIdentifier` in https://github.com/Polymer/polymer/blob/0.8-preview/src/standard/gestures.html#L133 is always `undefined`

But it seems that `ev.changedTouches[0].identifier` returns the expected 'finger'.

This causes for me that touch doesn't trigger any track events for me both on iOS and Chrome. 

Wondering if I'm doing anything wrong?

azakus:
- Allows for "sloppy" taps
- 5px for now
- Fixes #1496 [0.9] touch track fails on iPhone, .touchIdentifier vs .identifier
- Fixes a bug with "nested" event listeners with touch events


azakus:
@frankiefu PTAL as well

frankiefu:
lgtm

PR #1521, opened 2015-05-12
Fix typo, remove useless operator

ragingwind:


tjsavage:
Thanks for this PR as well - same as the later one, would you mind re-basing to the latest `master`?

PR #1556, opened 2015-05-18
Xstyle

sorvell:
* Improvements to cross-scope styling.
   - Support for rules using calc, e.g. calc(var(--a) + var(--b))
   - performance improvments
* support for @extends (https://tabatkins.github.io/specs/css-extend-rule/)


sorvell:
@kevinpschaaf PTAL

PR #1562, opened 2015-05-18
Update README.md

bananenmannfrau:
Fixes few lexical errors in README


kevinpschaaf:
The text in that README was removed recently, and replaced with a new landing page.  Thanks for your help here though.

PR #1570, opened 2015-05-19
X style fixes

sorvell:
* ensure custom-style rules are all available to elements
* ensure elements do not receive improper x-scope styling due to incorrect scope cache hit


kevinpschaaf:
LGTM

PR #1576, opened 2015-05-20
allow behavior arrays to contain nested behavior arrays

sjmiles:


sorvell:
LGTM... nit: would be nice to have a 3x nested test.

PR #1577, opened 2015-05-20
Literal args comma fix

sjmiles:


src/standard/effectBuilder.html
```diff
@@ -1,214 +1,269 @@
...
+          effect.static = false;
+        }
+        return arg;
+      }, this);
+      return effect;
+    },
+
+    _parseArg: function(rawArg) {
+      // clean up whitespace
+      var arg = String(rawArg).trim()
```
kevinpschaaf: Seems like `rawArg` is guaranteed to be a string (comes from `split`)...  seems safe to just do `rawArg.trim(). ...`

src/standard/effectBuilder.html
```diff
@@ -1,214 +1,269 @@
...
+        // replace comma entity with comma
+        .replace(/&comma;/g, ',')
+        // repair escape sequences
+        .replace(/\\(.)/g, '\$1')
```
kevinpschaaf: We couldn't figure out why this is necessary.

Is this related to your comment in chat?
> you can use \' inside of an attribute, like foo='foo\'bar'
> [2:39 PM] Scott Miles: but the value comes out like: foo\\'bar

sjmiles: Yes, that is why. If you do 'foo\'bar', the parser will accept it, but it produces a string like 'foo\\'bar', which is an outright bug in the platform IMO. This regex attempts to fix that problem.

miztroh:
From my tests, I'm still having issues with computed expressions that have only literal arguments.  In the example below, node.foobar returns undefined.

```
foobar: {
   computed: "computeFooBar('foo', 'bar')"
}
```

```
computeFooBar(foo, bar) {
   return foo + ' ' + bar;
}
```

Issue #1578, opened 2015-05-20
Binding text remains in <input value="{{value::input}}"> on IE10


morethanreal:

Sleeckx:
See also #1491

PR #1589, opened 2015-05-22
Clear input.value attribute before removing for IE. Fixes #1491. Fixe…

### LINKED ISSUE #1491: [0.8] Binding a property with value 'undefined' to an input value on IE11 shows the raw binding
Opening the following test case on IE11 will show an input box with value "{{value1::input}}". Setting the  initial value of the property to an empty string fixes the issue.

http://jsbin.com/wadasa/1/edit?html,output

Above output on IE11:
image

kevinpschaaf:
…s #1578.


kevinpschaaf:
Reviewed by @sorvell

PR #1591, opened 2015-05-22
Perform dom-bind work in attached/detached. Add tests.

### LINKED ISSUE #1515: [0.9] dom-bind not working with document.createElement
Tried to create the dom-bind element using `document.createElement('template', 'dom-bind')`. Throws the following error: "Uncaught NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node."

kevinpschaaf:
Fixes #1515.

### DIFF
--- a/test/unit/dom-bind.html
+++ b/test/unit/dom-bind.html
@@ -1,0 +1,241 @@
+<!doctype html>
+<!--
+@license
+Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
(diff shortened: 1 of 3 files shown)


test/unit/dom-bind.html
```diff
@@ -0,0 +1,241 @@
...
+      });
+
+      test('move dom-bind', function() {
+        domBind.parentElement.removeChild(domBind);
+        // TODO(kschaaf): detached/attached not called if takeRecords isn't
+        // called between remove & insert on polyfill... seems bad?
+        CustomElements.takeRecords();
+        container.appendChild(domBind);
+        CustomElements.takeRecords();
+        assert.equal(container.firstElementChild, el1);
```
kevinpschaaf: @sorvell If I don't call `CustomElements.takeRecords()` between `remove` and `insert`, I don't get detached/attached callbacks when moving a node in the DOM on polyfill... expected?

PR #1594, opened 2015-05-22
Wait until imports resolve to stamp. Fixes #1500

### LINKED ISSUE #1500: [0.9] Annotated computed properties don't work on autobinding template
Not sure if this a bug or not, but because we rely on A. C. P. to replace expressions, it would be nice to be able to use them on a `dom-bind` template.

Currently, it's failing with the error:

```
"TypeError: Cannot read property 'apply' of undefined
    at function-test.Polymer.Base.extend._annotatedComputationEffect (https://rawgit.com/polymer/polymer/0.8-preview/src/lib/bind/effects.html:73:38)
```

Repro:

http://jsbin.com/hijise/1/edit?html,console,output

kevinpschaaf:
Depends on https://github.com/webcomponents/webcomponentsjs/pull/310

### DIFF
--- a/src/lib/imports-status.html
+++ b/src/lib/imports-status.html
@@ -14,21 +14,23 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     Polymer.ImportStatus._importsLoaded();
   });
 
+  if (window.HTMLImports) {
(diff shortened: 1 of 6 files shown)


sorvell:
LGTM

PR #1597, opened 2015-05-22
1565,1582,1584 fix

### LINKED ISSUE #1565: [0.9.0] Problem putting a dom-if template in a light DOM when the component's <content> itself is wrapped in a dom-if
If an inner-component wraps it's <content> in a <template is=dom-if> there is a problem if the user of that component also puts a <template is=dom-if> in the light DOM.  Suspicion is that someone is still trying to populate or deal with the <content> even though it isn't stamped in the DOM?

Anyway, the end result is that things appear to more-or-less work except that when run with the Chrome debugger enabled there is an unhandled exception thrown:

    polymer-mini.html:499 - nativeInsertBefore.call(container, node, ref_node);`  `DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.

I wasn't able to exactly reproduce my situation.   However, here is a test case to reproduce a similar issue: http://jsbin.com/nuxupejije/1/edit (Note, the expected output is only "inner_before inner_after", the problem is the unhandled exception.  Though, this test case has the exception in dom-api.html:110 instead.

### LINKED ISSUE #1582: Under native ShadowDOM "dom-if" doesn't stamp out the content even when "if" property is true
Tapping on the `paper-tab` should show the ripple effect but it doesn't under native ShadowDOM.  In `paper-tab`'s template it has `paper-ripple` inside a `<template is="dom-if">` and even when `if` property is true the `paper-ripple` still not getting stamped.

Note: this is on polymer#master

paper-tab.html
```html
...
<template>

  <div class="tab-content flex auto center-center horizontal layout">
    <content></content>
  </div>

  <template is="dom-if" if="{{!noink}}">
    <paper-ripple id="ink" initial-opacity="0.95" opacity-decay-velocity="0.98"></paper-ripple>
  </template>

</template>
...
```

app.html
```html
<!doctype html>
<html>
<head>

  <title></title>
  <meta charset="utf-8">
  <script src="../../../polymer/components/webcomponentsjs/webcomponents-lite.js"></script>
  <link rel="import" href="x-app.html">

</head>
<body>

  <x-app></x-app>

</body>
</html>
```

x-app.html
```html
<link rel="import" href="../../../polymer/components/polymer/polymer.html">
<link rel="import" href="../../../polymer/components/paper-tabs/paper-tabs.html">
<link rel="import" href="../../../polymer/components/paper-tabs/paper-tab.html">

<dom-module id="x-app">

  <style>

    paper-tabs {
      background: #00bcd4;
    }

  </style>

  <template>

    <paper-tabs>
      <paper-tab>Foo</paper-tab>
      <paper-tab>Bar</paper-tab>
      <paper-tab>Zot</paper-tab>
    </paper-tabs>

  </template>

  <script>

    Polymer({

      is: 'x-app'

    });

  </script>

</dom-module>
```

### LINKED ISSUE #1584: Content tag does not work inside dom-if template
Content is not working inside the conditional template tag. It used to work in Polymer 0.5. 

 <template is="dom-if" if="{{!isIconOnly}}">
          <paper-button raised="{{raised}}" disabled="{{disabled}}" toggles="{{toggles}}" >
               <content ></content>
              <iron-icon icon="{{icon}}" hidden$="{{rightIcon}}"></iron-icon>
          </paper-button>
      </template>

sorvell:
Fixes #1565.
Fixes #1582.
Fixes #1584.

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -124,6 +124,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           this._removeNodeFromHost(node);
         }
         if (!distributed) {
+          if (factory(node).parentNode !== this.node) {
+            console.warn('The node to be removed is not a child of this node');
+          }
           // if removing from a shadyRoot, remove form host instead
           var container = this.node._isShadyRoot ? this.node.host : this.node;
           // not guaranteed to physically be in container; e.g. 
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -117,6 +117,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         This method also performs dom composition.
       */
       removeChild: function(node) {
+        if (factory(node).parentNode !== this.node) {
+          console.warn('The node to be removed is not a child of this node', 
+            node);
+        }
         var distributed;
         if (this._nodeIsInLogicalTree(this.node)) {
           var host = this._hostForNode(this.node);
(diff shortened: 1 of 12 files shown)


kevinpschaaf:
LGTM

PR #1600, opened 2015-05-23
Notify array.length changes.  Fixes #1573.

### LINKED ISSUE #1573: Array helpers don't notify length change
If I use the built-in array mutation helpers, I would expect that they would properly notify not just on the array itself but also on its length. Right now I have an app with a counter and I have to manually `notifyPath` for the length each time I mutate the array.

kevinpschaaf:


sorvell:
LGTM

PR #1614, opened 2015-05-25
Update docs.  Add warnings.

### LINKED ISSUE #1206: [0.8] better error when bound change handler is missing
Currently, if you `bind: {'foo': 'notPresent'}` you get a rather unhelpful error:
```
TypeError: undefined is not a function
```

### LINKED ISSUE #1538: Improve error message if observer is missing
If I define an observer on my properties object but don't actually implement the method, I get a strange error in effects.html: `Uncaught TypeError: this[effect.method] is not a function`. It's really not clear that the issue is a missing observer, I had to track down @cdata who could point me in the right direction. Any chance to handle and send out a helpful error message in this scenario?

kevinpschaaf:
Fixes #1538.  Fixes #1206.


src/micro/behaviors.html
```diff
@@ -40,6 +40,22 @@
...
+    /**
+     * Array of objects to extend this prototype with.
+     *
+     * Each entry in the array may specify either a behavior object or array
+     * of behaviors.
+     *
+     * Each behavior object may define lifecycle callbacks, `properties`,
+     * `hostAttributes`, `observers` and `listeners`.
+     *
+     * Lifecycle callbacks will be called on the base prototype first, then
```
sorvell: ordering is wrong here.

sorvell:
LGTM

PR #1618, opened 2015-05-26
Initialize _config with values set before creating accessors.

### LINKED ISSUE #1615: [0.9.2] dom-repeat issues in 0.9.2
Somewhere between 0.9.0 and 0.9.2, it seems like a bug's been introduced in `<dom-repeat>`. Here are two JSBins illustrating the problem, nested inside of a `<dom-bind>` parent:

Working 0.9.0: http://jsbin.com/topize/3/edit
Broken 0.9.2: http://jsbin.com/fuxode/2/edit

In 0.9.2, the <template>'d content isn't being stamped out to the DOM. By the time the `WebComponentsReady` is fired, the property on the parent `<dom-bind>` that is being repeated is set to undefined.

kevinpschaaf:
Fixes #1615.

http://jsbin.com/wubihisada/2/edit

### DIFF
--- a/src/lib/template/dom-bind.html
+++ b/src/lib/template/dom-bind.html
@@ -105,12 +105,22 @@ elements to the template itself as the binding scope.
       }
     },
 
+    _prepConfigure: function() {
+      var config = {};
+      for (var prop in this._propertyEffects) {
+        config[prop] = this[prop];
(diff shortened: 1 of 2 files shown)


sorvell:
LGTM

PR #1626, opened 2015-05-26
Add back Polymer.version string

### LINKED ISSUE #1625: Polymer.version undefined in 0.9
We misplaced this property sometime between 0.5 and now. We should dig it up.

azakus:
Replace version string when releasing
Fixes #1625


kevinpschaaf:
LGTM

PR #1564, opened 2015-05-19
Remove unused keyCodes enum

jklein24:
This depends on https://github.com/PolymerElements/iron-behaviors/pull/11 which removes the only usage of this enum.

PR #1627, opened 2015-05-26
Remove deprecated `keyCodes` object

azakus:


azakus:
Whoops, duplicate of #1564

PR #1629, opened 2015-05-27
Re-insert rows when re-attaching. Fixes #1498. Fixes #1714.

### LINKED ISSUE #1498: [0.9] when I move a dom-repeat element from one parent to another, the items will gone
```html
<dom-module id="custom-element">
  <template>
    <template is="dom-repeat">{{users}}</div>
  </template>
</dom-module>
```

In the <body>

```html
<div id="first"><custom-element id="my-list"></custom-element></div>
<div id="second"></div>
```

When I use `Polymer.dom(element("#second")).appendChild( element("#my-list") )`, the whole list inside it will be removed.

### LINKED ISSUE #1714: Uncaught TypeError When splicing an array into emptiness
This appears to happen when the last element of an array is spliced out using the array manipulation helper 'splice'.

    Uncaught TypeError: Cannot read property '0' of null
        Polymer._detachRow @ polymer.html:3425
        (anonymous function) @ polymer.html:3409
        Polymer._applySplicesArraySort @ polymer.html:3407
        Polymer._render @ polymer.html:3278
        Polymer.Templatizer._flushTemplates @ polymer.html:2900
        Polymer.Debounce.Debouncer.complete @ polymer.html:847
        atEndOfMicrotask @ polymer.html:808

kevinpschaaf:
Note, this fix suffers from the same problem as `dom-bind` (and `dom-if`) on polyfill, in that the user must manually do `remove`, `CustomElements.takeRecords()`, `append` when moving in order for the element to actually be notified via `attached` of its change in DOM position (see https://github.com/webcomponents/webcomponentsjs/issues/311).  May need to consider another approach for detecting movement.

### DIFF
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -205,7 +205,15 @@ Then the `observe` property should be configured as follows:
           this._detachRow(i);
         }
       }
-      this.rows = null;
+    },
+
+    attached: function() {
+      if (this.rows) {
+        var parentNode = Polymer.dom(this).parentNode;
+        for (var i=0; i<this.rows.length; i++) {
+          Polymer.dom(parentNode).insertBefore(this.rows[i].root, this);
+        }
+      }
     },
 
     ready: function() {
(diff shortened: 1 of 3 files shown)


sorvell:
LGTM.

PR #1635, opened 2015-05-27
dom-if hidden state is (this._hideTemplateChildren || !this.if).

### LINKED ISSUE #1632: [0.9.4] Nested dom-if templates show invalid content
Consider the following case: http://jsbin.com/biquxakiqi/1/edit?html,output

After clicking the steps one by one, you will see that *WORLD* is shown even though the value of the *world* property, used in the inner dom-if template, is still false.

Result (tested on Chrome):
image

Adding **restamp** to the inner dom-if template fixes the issue, as seen in this sample: http://jsbin.com/ducakepone/1/edit?html,output

image

kevinpschaaf:
Fixes #1632.

### DIFF
--- a/src/lib/template/dom-if.html
+++ b/src/lib/template/dom-if.html
@@ -77,7 +77,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // TODO(kschaaf): add logic to re-stamp in attached?
       this._teardownInstance();
     },
-    
+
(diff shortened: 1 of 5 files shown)


sorvell:
We don't handle the case of a `dom-repeat` stamping into a hidden `dom-if`. Will handle this via a separate issue.

sorvell:
LGTM.

PR #1636, opened 2015-05-28
Pass the index into the user defined filter function

filaraujo:
This patch passes the index into the user defined filter function, this allows the user to define any function and receive the similiar signature that they would expect in a Array.filter call

### DIFF
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -370,7 +370,7 @@ Then the `observe` property should be configured as follows:
       // Apply user filter to keys
       if (this._filterFn) {
         this._orderedKeys = this._orderedKeys.filter(function(a) {
-          return this._filterFn(c.getItem(a));
+          return this._filterFn(c.getItem(a), a);
         }, this);
       }
       // Apply user sort to keys


filaraujo:
how can i rerun these tests?

src/lib/template/dom-repeat.html
```diff
@@ -370,7 +370,7 @@
       // Apply user filter to keys
       if (this._filterFn) {
         this._orderedKeys = this._orderedKeys.filter(function(a) {
-          return this._filterFn(c.getItem(a));
```
samccone: to be totally inline with the filter api we should also pass the original array as well, thoughts?

ebidel:
@filaraujo Can you rebase and add the other `Array.filter` args? We should be 1:1 with the native api.

filaraujo:
Looking at the latest code, index and items would be troublesome and misleading.  

Unsure which index to pass. Do I pass the virtual index ( the one that references the internal collection) or do I pass the current index of the filter loop.  Passing the filter iteration loop doesn't really provide much use.  

As for the items, I can pass the original collection or the references to the weak map collection.  

Thoughts?

kevinpschaaf:
Due to the API inconsistency that would result due to differences in when we can efficiently pass the index and when it would be prohibitive to do so (because the array is operating on sorted keys, see https://github.com/Polymer/polymer/issues/2187#issuecomment-127452542), we're electing to not merge this now.  Thanks for the efforts here, this is hard problem.

PR #1638, opened 2015-05-28
Custom style fix

### LINKED ISSUE #1637: custom-style sometimes does not apply variables
It's possible for `custom-style` to upgrade when its textContent is not yet available, for example due to the browser's html parser yielding. In this case, any custom properties defined in a custom-style will not apply. To address this, custom-style should defer acting until it has content.

sorvell:
Fixes #1637


kevinpschaaf:
Reviewed. LGTM.

PR #1717, opened 2015-06-02
Fix for IE.

achimnol:
My website was not working on IE because the styles were missing -- and the IE javascript console said that it could not read "cssText" property. I dig down a little bit and found that this simple two-line patch makes it working perfectly on IE 11 as well as Chrome and Firefox. I haven't tested this on other versions of IE.

 * Ensure `rule.propertyInfo` is correctly set in style-properties.html.  It was checking a wrong attribute: `rule.properties` instead of `rule.propertyInfo` before calling `decorateRule()`.
 * Fix a condition check in custom-style.html when `rule.propertyInfo` is null or undefined.

### DIFF
--- a/src/lib/custom-style.html
+++ b/src/lib/custom-style.html
@@ -107,7 +107,7 @@ Note, all features of `custom-style` are available when defining styles as part 
             rule.selector = 'body';
           }
           var css = rule.cssText = rule.parsedCssText;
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
@sorvell PTAL

sorvell:
LGTM... thx!

achimnol:
Thanks! :D

PR #1724, opened 2015-06-02
Move non-webcomponents script unresolved case to `load` instead of synchronous

### LINKED ISSUE #1723: When not including `webcomponentsjs`, a script in `<head>` after imports will break `unresolved` attribute
This is because we synchronously try to remove the attribute in this case, but `document.body` does not exist yet.

azakus:
Fixes #1723

### DIFF
--- a/src/lib/unresolved.html
+++ b/src/lib/unresolved.html
@@ -9,7 +9,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 -->
 <script>
 (function() {
-  
+
   // Ensure that the `unresolved` attribute added by the WebComponents polyfills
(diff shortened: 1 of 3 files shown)


src/lib/unresolved.html
```diff
@@ -22,7 +22,7 @@
   if (window.WebComponents) {
     addEventListener('WebComponentsReady', resolve);
   } else {
-    resolve();
+    addEventListener('load', resolve);
```
sorvell: Let's use `document.readyState` iff not at least at `interactive` and then resolve on `interactive`. `load` is overly long.

azakus:
@sorvell PTAL with the new patch

sorvell:
LGTM, thx.

PR #1740, opened 2015-06-04
Add `unlisten` function

### LINKED ISSUE #1639: Need Polymer.Base.unlisten to remove the event listener
Currently we have `Polymer.Base.listen` to add an event listener e.g. `tap`.  But we also need `Polymer.Base.unlisten` to remove the event listener.

azakus:
Keep generated event handlers, attach with WeakMap to host element
Add tests for events

Fixes #1639

### DIFF
--- a/src/standard/events.html
+++ b/src/standard/events.html
@@ -84,9 +84,40 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         this._createEventHandler(node, eventName, methodName));
     },
 
+    _boundListenerKey: function(eventName, methodName) {
+      return (eventName + ':' + methodName);
+    },
+
+    _recordEventHandler: function(host, eventName, target, methodName, handler) {
+      var hbl = host.__boundListeners;
+      if (!hbl) {
+        hbl = host.__boundListeners = new WeakMap();
+      }
+      var bl = hbl.get(target);
+      if (!bl) {
+        bl = {};
+        hbl.set(target, bl);
+      }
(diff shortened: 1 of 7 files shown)


src/standard/events.html
```diff
@@ -84,9 +84,34 @@
...
+      var hbl = host.__boundListeners;
+      if (!hbl) {
+        hbl = host.__boundListeners = new WeakMap();
+      }
+      var bl = hbl.get(target);
+      if (!bl) {
+        bl = {};
+        hbl.set(target, bl);
+      }
+      bl[eventName] = handler;
```
sorvell:
```
var be = bl[eventName];
if (!be) {
  be = {};
}
be[methodName] = handler;
```

src/standard/events.html
```diff
@@ -84,9 +84,34 @@
...
+    },
+
+    _recallEventHandler: function(host, eventName, target) {
+      var hbl = host.__boundListeners;
+      if (!hbl) {
+        return;
+      }
+      var bl = hbl.get(target);
+      if (!bl) {
```
sorvell:
```
var be;
return bl && (be = bl[eventName]) && be[methodName];
```

azakus:
Not sure if `be` is necessary. Since both `methodName` and `eventName` are strings, we can just concatenate them with into one key in `bl`.

azakus:
PTAL @sorvell, multidispatch is supported now.

kevinpschaaf:
LGTUs

PR #1741, opened 2015-06-04
WIP: Discussion on how to trigger array splices for external changes

solsson:
We're using Polymer with existing client side business logic that is quite elaborate, involving for example https://github.com/simonsoft/authormodel. We would like such logic to stay independent of the presentation layer.

We naturally manipulate collections, but it is infeasible to call Polymer elements' push/pop/splice for that. Instead I implemented https://github.com/simonsoft/backbone-polymer -- we use Backbone only for the data strucutres, not the UI part -- that uses `notifyPath`. Trivial for key-value changes but not for array splices.

I actually managed to do this with test coverage in https://github.com/simonsoft/backbone-polymer/commit/cb496ce3b56c8e913fc154e8210d85d18d485b54.

BUT that solution looks messy and also Polymer docs state that arrays must be manipulated through element push/pop/splice. Is there a path forward for such notifications from externally changed arrays?


solsson:
https://github.com/simonsoft/polymer/commit/9b2ef7f7fad3afa9d2f8107d327e64e4a88efcf1 clarifies what backbone-polymer is doing atm.

kevinpschaaf:
We will likely re-introduce an optional [`observe-js`](https://github.com/polymer/observe-js)-based observation layer that you could import for your app that ties into Polymer 1.0's data notification system to allow "action-at-a-distance" data observation use cases like the one you described.  We teased this on our roadmap here: https://www.polymer-project.org/1.0/docs/#binding-features.  Stay tuned for that.

solsson:
Sounds interesting and would most likely solve this case.

It adds a layer of complexity and a likely performance penalty though, for a case like this where the at-a-distance lib already has change notifications. I'd just need a `notifyPath` API extension that can be called from `add` and `remove` event listeners (I guess Backbone is pretty generic here).

I realize that the .splices change record needs to be a complex object, but as long as it is a stable API it could serve as the distant trigger.

I think the Polymer team made a wise choice to not magically listen to array changes like some other frameworks do. A notification API feels more robust.

solsson:
Rebased onto 1.0.3 and restructured the tests.

I called this notify-path-api because I think there's a gap to fill with tests on externally triggered notifyPath overall (https://www.polymer-project.org/1.0/docs/devguide/data-binding.html#set-path). From a downstream perspective you want to know if and how these notifications render. For example you pass a value along, but what if that value differs from the current value of the property.

jscissr:
Maybe you're insterested in this; I tried to create a simple [Polymer element](https://github.com/jscissr/polymer-observe/) which uses Object.observe to call `notifyPath`.

kevinpschaaf:
Closing as this PR & discussion has grown stale.

We are actively pursuing improvements to the data system per https://github.com/Polymer/polymer/issues/3343 which will likely be a good path forward for applications such as the one described here.  Please track that issue for details.

Issue #1747, opened 2015-06-04
Should automatic node finding convert id names with dashes?


dotDeeka:
When mapping attribute names to property names, Polymer will convert the attribute some-cool-attr to someCoolAttr when assigning the property.

However, when using automatic node finding if the id has a dash in it Polymer does nothing. I understand these are two different things, but its not how I expected it to work. I was expecting more like

```html
<template>
    <div id="cool-div"></div>
</template
```
to be referenced
```javascript
Polymer({
   // ... polymer code
    ready: function () {
        // I expected the following
        this.$.coolDiv;
        // but instead I would have to do the following
        this.$['cool-div'];
    }
});
```

Have there already been discussions around this sort of thing I'm not aware of? Should I just suck it up and name my ids as camelCase?

Unfortunately this would be a breaking change of course.

zishanj:
+1 for this enhancement

jklein24:
+1. It would be nice to lose the brackets and still use dashes in id names.

danbeam:
see also #150

jklein24:
That's already done in Polymer 1.0 for attributes.

pensierinmusica:
+1 agree with @dotDeeka

NemoD503:
+1 for this feature

kosiasik:
It would be nice +1

pensierinmusica:
@kosiasik if I understand correctly HTML attributes [are case insensitive](http://www.w3.org/TR/html-markup/documents.html#case-insensitivity), which is why Polymer converts camelCase JS properties in the first place.

So, in your example the outer div has a property `id` with value `cooldiv`, whereas the inner div has a property `id` with value `cool-div`. Hence they're easily distinguishable by the interpreter.

Does this make sense?

zoechi:
`id` (the name) is case insensitive `coolDiv` the value is not.

pensierinmusica:
I thought so, and was actually looking for an official reference to it but couldn't find it. The closest match I've found is this (which is an old document though, and suggests to to treat values as case-insensitive anyways):
https://developer.mozilla.org/en-US/docs/Case_Sensitivity_in_class_and_id_Names

@zoechi do you have a link to official HTML standard docs where is stated that attribute values are case sensitive, or is it left to browser implementations?

Thanks, cheers!

zoechi:
[...]

PR #1750, opened 2015-06-04
Should camelize id names when creating this.$ automatic node finding object.

dotDeeka:
I mentioned this in issue #1747 .

This PR would not make a breaking change. In order for ids with dashes to be camelized you would have to opt in by setting camelizeIds: true in your element.

```html
  <dom-module id="my-cool-element">
    <template>
      <div id="foo-element"></div>
    </template>
  </dom-module>
```
```javascript
Polymer({
  is: 'my-cool-element',
  camelizeIds: true,
  ready: function () {
    this.$.fooElement.innerHtml = 'foo';
  }
});
```
I wanted to figure out a way to set this config globally but I'm not sure if that makes sense.

This is the first PR I'm proposing to Polymer so please let me know if I should be doing anything differently.

Feedback appreciated.

### DIFF
--- a/src/standard/annotations.html
+++ b/src/standard/annotations.html
@@ -226,11 +226,20 @@ TODO(sjmiles): this module should produce either syntactic metadata
     // construct `$` map (from id annotations)
     _marshalIdNodes: function() {
       this.$ = {};
-      this._notes.forEach(function(a) {
-        if (a.id) {
-          this.$[a.id] = this._findAnnotatedNode(this.root, a);
-        }
-      }, this);
+
+      if (this.camelizeIds === true) {
+        this._notes.forEach(function(a) {
+          if (a.id) {
+            this.$[Polymer.CaseMap.dashToCamelCase(a.id)] = this._findAnnotatedNode(this.root, a);
+          }
+        }, this);
+      } else {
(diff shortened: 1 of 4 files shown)


dotDeeka:
I signed it!

samccone:
hey @dotDeeka great call on this one and great call on putting it behind a flag until we bump to 2.0, I agree this is totally confusing at the moment, given how we automagically do this with attributes I think that this falls totally inline :+1: .

cc @ebidel for :eyes:

dotDeeka:
@samccone Other than the fact that 2 tests are failing at the moment, does this implementation look ok, or is there something different you guys would like to see? Thanks.

samccone:
@dotDeeka The final :+1: :-1: is actually not up to me, going to defer to @tjsavage and team for final decision

dotDeeka:
Cool.

kevinpschaaf:
We use dash-case in _attribute names_ in which case information is lost by the parser, but since node id's are case-sensitive, there is no technical reason this is required (as spelled out in https://github.com/Polymer/polymer/issues/1747#issuecomment-150313046).  From a simplicity point of view we are electing to close this.

PR #1771, opened 2015-06-07
Allow function literals instead of method names in listeners

DmitryTrigonis:
My English so bad, so i think you get an idea by youself

### DIFF
--- a/src/standard/events.html
+++ b/src/standard/events.html
@@ -88,7 +88,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var host = this;
       return function(e) {
         var type = typeof methodName;
-        if(type === 'function' || type === 'object' && !!obj){
+        if(type === 'function'){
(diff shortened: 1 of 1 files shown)


DmitryTrigonis:
Now we can write something like
listeners: {
			'tap': function(){console.log(this.name)}
		},

instead of
listeners: {
			'tap': 'somehandler'
		},

somehandler: function(){
console.log(this.name)
}

DmitryTrigonis:
#318cd1b

sjmiles:
Here are a few of the reasons why it's a good idea to restrict the syntax of `listeners` (and `observers`, `computed`, and so on) not to allow inline functions:

1. more likelihood for reuse of functions
2. better separation of declarative and imperative code
3. simplified overall syntax

for these reasons, we are not at this time wanting to expand the syntax in this way. 

But thanks for your efforts, we may reconsider this issue in the future.

Issue #651, opened 2014-07-21
Need consistent path behavior


robdodson:
Original thread: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/polymer-dev/JcBQZCPamcY/OMKhLetXg-4J

If I create an element that looks like this:

``` html
<polymer-element name="x-foo">
  <template>
    <template repeat="{{link in links}}">
      <div><a href="#ok{{link}}">Ok Hash Link {{link}}</a></div>
    </template>
    <div><a href="#bad">Bad Hash Link</a></div>
  </template>
  <script>
    Polymer('x-foo', {
      links : [2,3,4]
    });
  </script>
</polymer-element>
```

and then import it into my project, it will treat the `href`s differently. The `href`'s inside of the repeating template will look like this:

```
<a href="#ok2">
<a href="#ok3">
<a href="#ok4">
```

but the `href` outside of the repeating template will look like this:

```
<a href="x-foo.html#bad">Bad Hash Link</a>
```

ssorallen:
@sorvell, was it intentional to rewrite `href` for anchor elements? It doesn't seem like those should be rewritten at all.

sorvell:
Yes, the idea is that you should always be able to link resources with paths relative to the importing document. If it's an absolute url or a # based one we should bail.

azakus:
The rewriter doesn't check that the node is an anchor, it just rewrites href. The simplest fix is to ignore hash-only hrefs for rewriting.

PR #1780, opened 2015-06-08
Fixes #1757, do not resolve hash-only urls used for routing

### LINKED ISSUE #1757: [1.0.3] Do not resolve hash-only urls used for routing
I think the error #651 that was fixed in 0.5 is back in 1.0.

I have an `my-element.html` that defines: 
```
<a href="#del"></a>
```
but in the browser it's converted to `my-element.html#del`.

nippur72:
Fixes #1757 by applying the same fix that was used in 0.5 in #651:

https://github.com/Polymer/polymer/blob/c7169640c567fb9bc709207ecbd9f747f8df48d5/src/lib/url.js#L103-L106

### DIFF
--- a/src/lib/resolve-url.html
+++ b/src/lib/resolve-url.html
@@ -41,6 +41,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
 
     function resolve(url, ownerDocument) {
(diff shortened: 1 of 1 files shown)


azakus:
I'll merge this and add a test. Thanks!

PR #1793, opened 2015-06-09
Use gulp-vulcanize concurrently for faster builds

azakus:
Remove mkdir task, gulp.dest does that for us
Make clean an optional task, it is not necessary for most things


kevinpschaaf:
LGTM

PR #1795, opened 2015-06-09
Do not apply/notify keySplices if array has not been Collectionified. Fixes #1744

### LINKED ISSUE #1744: nested dom-repeat with sort attribute shows duplicate entries when adding new items.
When pushing new items into a nested array rendered by a dom-repeat with a sort attribute, duplicate items are rendered: 

http://jsbin.com/fazazujamu/2/edit

(Click 'go' once, you should see items 11 and 21 duplicated in the output)

from my testing, this requires the array to be  nested inside another array which is rendered with dom-repeat, and it requires the items to be added one by one, AFTER the parent item is already rendered, and must be added using this.push.

kevinpschaaf:


sorvell:
LGTM.

PR #1797, opened 2015-06-09
Fix 1752

sorvell:
Do not allow properties to be late bound via inheritance in custom-styles. This appears to be correct spec behavior and matches the implementation in Firefox.


kevinpschaaf:
LGTM

PR #1798, opened 2015-06-09
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "polymer",
-  "version": "1.0.3",
   "main": [
     "polymer.html"
   ],


kevinpschaaf:
We may use this for tooling or other future uses.

kkirsche:
Interesting choice. Not sure why you'd want to rely on something which they say to leave out and isn't even used, but to each their own. Best of luck

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

IntranetFactory:
We use it in our build process, so I'd prefer it it stays there and is maintained.

PR #1820, opened 2015-06-10
Fix jsdoc for splice

jscissr:
`Array.prototype.splice` returns an Array, not a Number. You could also change the code to do what the docs say instead.


kevinpschaaf:
Thanks!  Yes, the intention is to match the native semantics.

PR #1837, opened 2015-06-11
No implicit template (fixes #1827)

### LINKED ISSUE #1827: Remove the undocumented "find nearest template" feature when registering
If you have
```html
<template is="dom-bind">
<template is="dom-repeat">
</template>
</template>
```
in index.html,  and use an external script for polymer via crisper, dom-repeat and dom-bind will find the top level dom-bind template as their shadowdom templates and recursively expand until the stack explodes.

sjmiles:


arthurevans:
... and there was much rejoicing.

sorvell:
LGTM

PR #1838, opened 2015-06-11
Use var keyword when declaring local variable so it doesn't leak to global scope.

trevordixon:
It looks like `var` was forgotten in a couple places, causing `v` and `style` to leak to the global scope.


sorvell:
LGTM, Good catch, we'll also add `'use strict';` to avoid this in the future.

PR #1841, opened 2015-06-11
Fix incorrect test for `toggle`. Fixes #1810.

### LINKED ISSUE #1810: array-selector doesn't work with `multi` unless `toggle` is specified
## Problem 1
In the [section](https://www.polymer-project.org/1.0/docs/devguide/templates.html#array-selector) regarding `array-selector` in the developer docs, it is unexplained the importance & relevance of `toggle` being set for the code snippet provided: 

```html
<dom-module id="employee-list">

  <template>

    <div> Employee list: </div>
    <template is="dom-repeat" id="employeeList" items="{{employees}}">
        <div>First name: <span>{{item.first}}</span></div>
        <div>Last name: <span>{{item.last}}</span></div>
        <button on-click="toggleSelection">Select</button>
    </template>

    <array-selector id="selector" items="{{employees}}" selected="{{selected}}" multi toggle></array-selector>

    <div> Selected employees: </div>
    <template is="dom-repeat" items="{{selected}}">
        <div>First name: <span>{{item.first}}</span></div>
        <div>Last name: <span>{{item.last}}</span></div>
    </template>

  </template>

  <script>
    Polymer({
      is: 'employee-list',
      ready: function() {
        this.employees = [
            {first: 'Bob', last: 'Smith'},
            {first: 'Sally', last: 'Johnson'},
            ...
        ];
      },
      toggleSelection: function(e) {
        var item = this.$.employeeList.itemForElement(e.target);
        this.$.selector.select(item);
      }
    });
  </script>

</dom-module>

```

If `toggle` is removed, the element's behavior doesn't work as expected. I & many perhaps thought the removal of `toggle` means that that a selection canbe added to `{{selected}}` but not removed. Unfortunately, that isn't the case. This seems like a bug. 

An example of this can be found [here](http://jsbin.com/sikide/8/edit); remove toggle & see what happens.

## Problem 2
Another problem beyond the problem(s) above is showing `this.$.selector`.select being used outside of  an event handler. It  was obvious to me inside things like the `ready` callback but there may be value if it was demonstrated the following ways for developers:
- With `filter` &/or `sort`
- Doing something with the `selected` contents when the element is detached
- `{{selected}}` being used outside the app with an outside listener by the forsight of the code defining the property name put inside the `selected` attribute being available to the outside world with an event listner

arthurevans:
Testing `toggle` on the incorrect branch of the `if` statement meant that  multi-select `array-selector` didn't work unless `toggle` was enabled. 

There don't appear to be any unit tests for `array-selector`, but with this fix the sample shown in #1810 works correctly without `toggle`.

### DIFF
--- a/src/lib/template/array-selector.html
+++ b/src/lib/template/array-selector.html
@@ -153,8 +153,10 @@ is false, `selected` is a property representing the last selected item.  When
         var scol = Polymer.Collection.get(this.selected);
         var skey = scol.getKey(item);
         if (skey >= 0) {
-          this.deselect(item);
-        } else if (this.toggle) {
+          if (this.toggle) {
+            this.deselect(item);
+          }
+        } else {
           this.push('selected', item);
           // this.linkPaths('selected.' + sidx, 'items.' + skey);
           // skey = Polymer.Collection.get(this.selected).add(item);


kevinpschaaf:
LGTM.  Thanks, Arthur!

PR #1845, opened 2015-06-12
Clear composedNodes when an element upgrades without an insertion point

### LINKED ISSUE #1804: Unwrapped dom-if causes DOMException
The following case throws a **DOMException: NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.**

http://jsbin.com/gujenocexo/2/edit?html,output

Wrapping the dom-if template in a ```div``` fixes the error.

Could this be related to #1631?

sorvell:
Fixes #1804 

A lazily upgraded element which subsequently adds an insertion point can generate an error. This occurs because composedNode tracking gets out of whack. Explicitly reset composedNodes when upgrading to avoid the problem.

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -59,8 +59,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // 3. node is <content> (host of container needs distribution)
       appendChild: function(node) {
         var handled;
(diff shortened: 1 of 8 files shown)


kevinpschaaf:
Reviewed.

PR #1861, opened 2015-06-13
simplify splice

jscissr:
After #1820, I now noticed that `rem` and `ret` are actually the same.


kevinpschaaf:
Awesome!  Nice catch.

PR #1889, opened 2015-06-16
dom-template: parentProps should not override argument based props

filaraujo:
found an issue in dom-template in which parentProps overrides the model that is provided in arguments

### DIFF
--- a/src/lib/template/templatizer.html
+++ b/src/lib/template/templatizer.html
@@ -277,7 +277,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       model = model || {};
       if (this._parentProps) {
         for (var prop in this._parentProps) {
-          model[prop] = this['_parent_' + prop];
+          if (!model[prop]) {
+            model[prop] = this['_parent_' + prop];
+          }
         }
       }
       return new this.ctor(model, this);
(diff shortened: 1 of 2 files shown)


gazal-k:
@filaraujo I'd found the same issue when trying to get recursive templates working and fixed it here: https://github.com/Polymer/polymer/pull/2196. Just noticed this PR now :smile:

kevinpschaaf:
Sorry for the delay on this.  In general the idea is that `Templatizer` clients can define an `_instanceProps` list that indicate which properties are provided by managed by the client (and thus should not be forwarded to the outer scope).  By doing so, they are excluded from `_parentProps`, and wouldn't hit this code.

That said, implementing the outer-scope forwarding hooks is not compulsory, so this change probably makes sense.  If you can update the PR so it can merge to master that'd be great, otherwise one of us can fork and merge it soon.  Thanks.

kevinpschaaf:
@filaraujo I updated your branch to master and integrated the new test with the test runner in a new PR here: https://github.com/Polymer/polymer/pull/1889, so closing this one.  Thanks for your effort here!

FYI, I also made a slight [fix](https://github.com/Polymer/polymer/pull/3373/files#diff-65d1593d7f79b51a64ac3cfc9465a1f4R419) to use an `undefined` check, since just a falsy check could still overwrite inner scope variables with `0` or `false`.

filaraujo:
thank you @kevinpschaaf

PR #1891, opened 2015-06-16
Includes element defaults in the list of own properties by which elements are styled.

### LINKED ISSUE #1752: overriding the custom css variables only works for the first dom element on the page
tldr: http://plnkr.co/edit/PxEdsV?p=preview

I was playing around different material design themes, and it is not completely clear to me in which way I can override the custom css style variables.

For example, let's say that I would have two themes:
```
<style is="custom-style">
  .amber {
    --dark-primary-color: #FFA000;
    --default-primary-color: #FFC107;
    --light-primary-color: #FFECB3;
    --text-primary-color: #212121;
    --accent-color: #FFC107;
    --primary-text-color: #212121;
    --secondary-text-color: #727272;
    --divider-color: #B6B6B6;
  }
  
  .blue {
    --dark-primary-color: #1976D2;
    --default-primary-color: #2196F3;
    --light-primary-color: #BBDEFB;
    --text-primary-color: #FFFFFF;
    --accent-color: #448AFF;
    --primary-text-color: #212121;
    --secondary-text-color: #727272;
    --divider-color: #B6B6B6;
  }
</style>
```

Now if a dom element has such a class, and if it contains paper-elements, I would think that those paper-elements would drop their default styling and instead uses the variables defined in the class:

```
<paper-sample-app class="amber"></paper-sample-app>
<paper-sample-app class="blue"></paper-sample-app>
```

This does happen for the first dom element on the page. But the second dom element, actually get the styling of the first dom element. So, that is really strange, I would think it would maybe do nothing, and it gets the default styling, but the second element inherits the custom styling of the first dom element...

see: http://plnkr.co/edit/PxEdsV?p=preview

If I do this workaround, it works as I would expect:
```html
<dom-module id="paper-sample-app">
  <style>
    :host {
      /* I would think that this lines shouldn't be necesarry
         as this is the default value for --paper-toolbar-background */
      --paper-toolbar-background: var(--default-primary-color);
    }
  </style>
```

If I have a custom element with lots of different paper-elements, it feels bit redundant if I would have to define all the default styling of the paper elements, again, especially, as it is already somewhere defined.

sorvell:
Fixes #1752 where an element that uses only values supplied by variable defaults can be incorrectly styled.

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -98,6 +98,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var m;
         while (m = this.rx.VAR_CAPTURE.exec(cssText)) {
           props[m[1]] = true;
+          var def = m[2];
+          if (def && def.match(this.rx.IS_VAR)) {
+            props[def] = true;
+          }
         }
       },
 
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
LGTM

PR #1895, opened 2015-06-17
Basic support for host-context

sshumaker:
Support for transforming a host-context rule inside a custom style block. Example usage:

```HTML
<dom-module is="foo">
  <style>
    :host-context(.bar) .item {
      background-color: red;
    }
  </style>
  <template>
    <div class="item">Hello</div>
  </template>
</dom-module>

<body>
  <div class="bar">
    <foo></foo>
  </div>
</body>
```


sshumaker:
I signed it!

kevinpschaaf:
Thanks!  This was fixed along with some other fixes in https://github.com/Polymer/polymer/pull/1948.

PR #1916, opened 2015-06-19
Make sure updateStyles works based on changes to customStyles and enhance api.

### LINKED ISSUE #1851: CustomStyle change has no effect
Looking at https://www.polymer-project.org/1.0/docs/devguide/styling.html#style-api I thought that this example should work:

http://jsbin.com/ciyomuruha/1/edit?html,output

In the example I expected that the background turns blue after clicking on that button.

### LINKED ISSUE #1915: Provide a convenience method for setting `customStyle` and calling `updateStyles`
Polymer provides a `customStyle` object which can be used to set css custom properties. To apply these values, `updateStyles` must be called. Since this is typically done in lockstep, there should be a 1 step option, e.g.:

```
this.updateStyles({'-foo': 'red'});
// short for
this.customStyles['--foo'] = 'red';
this.updateStyles();
```

sorvell:
Fixes #1851: Include `customStyle` in the style host cache key so that changes to customStyle are recognized.

Fixes #1915: Include a `customStyleProperties` argument to `updateStyles` for convenience.

### DIFF
--- a/src/lib/style-defaults.html
+++ b/src/lib/style-defaults.html
@@ -8,18 +8,21 @@ Code distributed by Google as part of the polymer project is also
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 -->
 <link rel="import" href="style-util.html">
+<link rel="import" href="style-cache.html">
 <script>
 
   Polymer.StyleDefaults = (function() {
 
     var styleProperties = Polymer.StyleProperties;
     var styleUtil = Polymer.StyleUtil;
+    var StyleCache = Polymer.StyleCache;
 
     var api = {
 
       _styles: [],
       _properties: null,
       customStyle: {},
+      _styleCache: new StyleCache(),
 
       addStyle: function(style) {
         this._styles.push(style);
--- a/src/lib/style-defaults.html
+++ b/src/lib/style-defaults.html
@@ -19,6 +19,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _styles: [],
       _properties: null,
+      customStyle: {},
 
       addStyle: function(style) {
         this._styles.push(style);
@@ -36,6 +37,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           this._styles._scopeStyleProperties = null;
           this._properties = styleProperties
             .scopePropertiesFromStyles(this._styles);
+          // mixin customStyle
+          styleProperties.mixinCustomStyle(this._properties, this.customStyle);
           styleProperties.reify(this._properties);
         }
         return this._properties;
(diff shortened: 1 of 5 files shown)


src/lib/style-defaults.html
```diff
@@ -47,7 +50,25 @@
...
+       * shortcut for setting `customStyle` and then calling `updateStyles`.
+      */
+      updateStyles: function(properties) {
+        // force properties update.
+        this._properties = null;
+        if (properties) {
+          Polymer.Base.mixin(this.customStyle, properties);
+        }
         // invalidate the cache
         this._styleCache.clear();
```
kevinpschaaf: Users have reported on https://github.com/Polymer/polymer/issues/1851#issuecomment-112220994 that `Uncaught TypeError: Cannot read property 'clear' of undefined` can be thrown if `updateStyles` is called too early.  Should this just be guarded on `this._styleCache` being defined?  Or is this handled elsewhere?

src/lib/style-properties.html
```diff
@@ -330,6 +330,20 @@
         return style;
       },
 
+      // customStyle treats null property values as deletes. This enables
+      // properties to be reset to inherited values.
+      mixinCustomStyle: function(props, customStyle) {
+        var v;
+        for (var i in customStyle) {
+          v = customStyle[i];
+          if (v === null) {
```
kevinpschaaf: Worth changing this to `v == null || v == ''`, to be more similar with native `style`, which deletes on empty string?

src/standard/x-styling.html
```diff
@@ -94,13 +94,15 @@
           var cacheableStyle = style;
           if (nativeShadow) {
             cacheableStyle = style.cloneNode ?
-            style.cloneNode(true) : Object.create(style || null);
+            style.cloneNode(true) : null;
```
kevinpschaaf: wat was (?) style that didn't have`cloneNode`?

src/standard/x-styling.html
```diff
@@ -129,7 +131,7 @@
         // finally mixin properties inherent to this element
         this.mixin(props,
           propertyUtils.scopePropertiesFromStyles(this._styles));
-        this.mixin(props, this.customStyle);
+        propertyUtils.mixinCustomStyle(props, this.customStyle);
```
kevinpschaaf: Actually, it seems like `mixinCustomStyle`'s behavior on `null` should not delete properties that came from the scope... i.e. if you set `this.customStyle['--foo'] = 'red';`, that should override `:root`, but `this.customStyle['--foo'] = null` should just let any props in `:root` through without overriding with `null`.

src/standard/x-styling.html
sorvell: Good catch. This needs more cleanup. Previously a fake style element was returned to cache. This is no longer necessary since a style info object is cached.

src/standard/x-styling.html
sorvell: Another good catch. I think a customStyle property should only apply if it's truthy.

kevinpschaaf:
LGTM.

Issue #1902, opened 2015-06-17
With Shady DOM, <content> doesn't get passed to another element inside dom-if


trevordixon:
See http://plnkr.co/edit/k159u1diDGJ1PNgfwAvB?p=preview. Using v1.0.3, there are JS errors. Using master, there aren't JS errors, but it still doesn't work.

It does work, in v1.0.3 and master, with native shadow DOM. See http://run.plnkr.co/plunks/k159u1diDGJ1PNgfwAvB/?dom=shadow.

@kevinpschaaf Probably related to #1816?

sorvell:
Fixed via https://github.com/Polymer/polymer/pull/1917.

PR #1917, opened 2015-06-19
Fixes: #1902.

sorvell:
Use Polymer.dom to find insertion points in document fragments added to elements.

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -261,7 +261,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             var dc$ = factory(content).getDistributedNodes();
             for (var j=0; j<dc$.length; j++) {
               hostNeedsDist = true;
-              var node = dc$[i];
+              var node = dc$[j];
               var parent = node.parentNode;
               if (parent) {
                 removeFromComposedParent(parent, node);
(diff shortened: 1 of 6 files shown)


trevordixon:
Another issue: http://plnkr.co/edit/jW0woRzBQpZ050Oih1r3?p=preview

Click the toggle button a couple times, and you'll see JS errors.

src/lib/dom-api.html
```diff
@@ -172,10 +172,11 @@
         // when the need arises.  For now, the user must call
         // distributeContent(true), which updates insertion points manually
         // and forces distribution.
-       var fragContent = (node.nodeType === Node.DOCUMENT_FRAGMENT_NODE) &&
-            node.querySelector(CONTENT);
+        var fragContent = (node.nodeType === Node.DOCUMENT_FRAGMENT_NODE) &&
+          !node._noContent && Polymer.dom(node).querySelector(CONTENT);
```
kevinpschaaf: I was picturing the flag to be opt-in to the scoped querySelector cost, vs opt-out.  i.e. "tell us when you have a content in there"  as opposed to "tell us when you don't", but I guess it's not trivial for annotator to put the flag on, because it would need to be copied from `template.content` to the `root` fragment after `importNode`, is that right?

kevinpschaaf:
Re: the bug @trevordixon pointed out, looks like [this line](https://github.com/Polymer/polymer/blob/fix-1902/src/lib/dom-api.html#L264) needs to dereference using `j` instead of `i` (appears unrelated to this change):

```js
            for (var j=0; j<dc$.length; j++) {
              ...
              var node = dc$[i];   // << should be dc$[j]
```

kevinpschaaf:
LGTM, after review.

PR #1920, opened 2015-06-19
Refer to globals explicitly using window.[property].

trevordixon:
Makes Polymer a little more friendly to Closure compilation.

### DIFF
--- a/src/lib/async.html
+++ b/src/lib/async.html
@@ -52,7 +52,7 @@ Polymer.Async = (function() {
     lastVal += len;
   }
 
-  new (window.MutationObserver || JsMutationObserver)(atEndOfMicrotask)
+  new (window.MutationObserver || window.JsMutationObserver)(atEndOfMicrotask)
     .observe(twiddle, {characterData: true})
     ;
   
--- a/src/lib/dom-module.html
+++ b/src/lib/dom-module.html
@@ -44,21 +44,21 @@
   // imports tree. (Note: this should only upgrade any imports that have been
   // loaded by this point. In addition the HTMLImports polyfill should be
   // changed to upgrade elements prior to running any scripts.)
-  var cePolyfill = window.CustomElements && !CustomElements.useNative;
+  var cePolyfill = window.CustomElements && !window.CustomElements.useNative;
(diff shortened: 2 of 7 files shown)


rictic:
Hey, thanks for the contribution!

Do you have more info on this change? I wasn't aware that Closure Compiler cared either way about referring to globals as `window.identifierName` vs `identifierName`, it's just important that it know about the existence of the globals.

Is the issue that the definitions of these variables aren't visible to the compiler?

rictic:
I'm thinking that these should be solved with externs rather than in polymer.  

There are three symbols here that aren't getting resolved and adding `window.identifier` seems to just be papering over that problem. 

* `window.addEventListener` can be assumed to exist in the web platform as of IE 9 and all other browsers: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
* `CustomElements` is [exported by webcomponents.js](https://github.com/webcomponents/webcomponentsjs/blob/c9bceeedb4077b5658d95f5755883592be92fcaf/src/CustomElements/CustomElements.js)
* `JsMutationObserver` is [exported by webcomponents.js](https://github.com/webcomponents/webcomponentsjs/blob/938051272519053867b5bafa7411e7afefdd2c01/src/MutationObserver/MutationObserver.js#L562)

My instinct is to address this as follows:

The first one I think the responsibility of the user for configuring the compiler to target modern browsers.

The second two could be addressed by giving webcomponents.js an externs file. There's an argument to be made that Polymer doesn't want to assume the existence of webcomponents.js (and indeed this code is careful to only dereference those identifiers inside of conditionals), but I think there's an argument for just adding in the webcomponents.js externs for all compiled polymer code.

@jklein24 What do you think?

trevordixon:
In my setup, the `addEventListener` listener error happens even though `window.prototype.addEventListener` is defined in externs.

rictic:
I'd love to make core more compiler friendly, but I don't think this is the route that we're going to take.

The recommended setup is to avoid compiling polymer core and to use the externs defined in https://github.com/PolymerLabs/polymer-externs/

Issue #1839, opened 2015-06-11
bindings to concrete types not propagating correctly from template to collection


sjmiles:
Given:

```html
     <template is="dom-repeat" items="{{somethings}}">
        <paper-input value="{{item}}"></paper-input>
      </template>
```

changes to `paper-input.value`appear in `template.getModelForElement(paper-input).item` but not in `template.items[n]`.

http://jsbin.com/qocuka/2/edit

pvmart:
Might be related.
Overwriting an object in array with the new one destroys binding.
``` js
         this.push('test', {value: 1});
         this.set('test.0.value',  2);
         this.set('test.0',{value: 3});
         this.set('test.0.value',  4);
```
The result is 3 but in the model is 4.
http://plnkr.co/edit/AH5SKoedUQgeyfBgsfWL?p=preview

kevinpschaaf:
@pvmart Yep, they are the same root cause.  A fix is in the works that should solve both this and https://github.com/Polymer/polymer/issues/1854.

kevinpschaaf:
This should be fixed by PR #1855.

Confirmed it fixes repro case here:
http://jsbin.com/ziseru/2/edit

kevinpschaaf:
Root cause here w.r.t. notifying based on individual `dom-repeat` instances' `item` changng was fixed in #1854.

Note related issue opened here: https://github.com/Polymer/polymer/issues/1913, regarding problems when primitive values are duplicated in the array.

PR #1940, opened 2015-06-22
Disambiguate config fromAbove and local defaults, pass to _setProperty.

kevinpschaaf:
- Fixes bug in previous fix for #1839/#1854 where notifying properties would not notify during configuration
- renames `_setPropertyQuiet` to `_setProperty` with `quiet` arg
- renames `_propertySet` to `_propertySetter` for clarity


sorvell:
LGTM

Issue #1946, opened 2015-06-24
Consider reseting property to default when binding undefined structured data


kevinpschaaf:
When binding structured data (e.g. `{{item.type}}`) to an element, if a new item is assigned that no longer has a sub-property where the previous object did (e.g. item goes from `{name: 'foo', type: 'zaz'}` to `{name: 'bar'}`, an `undefined` value will be pushed through the binding to the element.  Since `undefined` has special meaning, especially for multi-property observers which are not called until all properties are defined, this can result in unexpected state.

See use case at https://github.com/Polymer/polymer/issues/1742 and proposed fix in https://github.com/Polymer/polymer/pull/1941

kevinpschaaf:
Closing, we likely won't pursue this gambit as a solution.

PR #1947, opened 2015-06-24
Use single-property observer to deal better with if=undefined. Fixes #1742

### LINKED ISSUE #1742: [1.0] dom-if : undefined is considered falsy only once
Hey !

In a conditional template, undefined is considered falsy only on the first invocation.
As of today, the only way to handle this use case is to declare a boolean computed property that returns false if the target value is undefined.

I made the following jsbin to reproduce the bug :
http://jsbin.com/retejotege/1/edit?html,output

Hope that helps :)

kevinpschaaf:
This takes advantage of the fact that single property observers are called for each change (including undefined), whereas multi-property observers wait for all properties to be defined.  Since the side-effect of the observe is to queue an async render, there is no real benefit from this being a multi-property observer.  #1946 is open to track a more general fix for this issue.

### DIFF
--- a/src/lib/template/dom-if.html
+++ b/src/lib/template/dom-if.html
@@ -44,7 +44,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        */
       'if': {
         type: Boolean,
-        value: false
+        value: false,
+        observer: '_queueRender'
       },
 
       /**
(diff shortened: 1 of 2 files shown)


sorvell:
Needs a test.

sorvell:
LGTM

PR #1949, opened 2015-06-24
Catch errors in async queue & re-start. Fixes #1933. Fixes #1759.

### LINKED ISSUE #1759: If an async callback throws an error, it's never removed from the callback list
... and all other async functions are blocked from running.

See https://github.com/Polymer/polymer/blob/master/src/lib/async.html#L48

### LINKED ISSUE #1933: Polymer.Base.async "infinite loop" condition
Please consider the following contrived usage of `Polymer.Base.async`:

```html
<script>
  function AsyncAlpha() {
    // Queue a task:
    Polymer.Base.async(function() {
      console.log('alpha');

      // Cause a second task to be queued:
      AsyncBeta();

      // Throw an exception before the first task returns:
      throw new Error('ANY ERROR');
    });
  }

  function AsyncBeta() {
    Polymer.Base.async(function() {
      console.log('beta');
    });
  }

  AsyncAlpha();
</script>
```

The above example will cause the first task ("alpha") to be re-called in what is effectively an infinite loop. In order to reproduce this condition, a second task ("beta") _must_ be queued before the first task returns, and an exception _must_ be thrown after the second task is queued but before the first task returns.

kevinpschaaf:


sorvell:
LGTM

PR #1958, opened 2015-06-24
Make computed properties implicitly readOnly. Fixes #1925.

### LINKED ISSUE #1925: Setting readOnly AND computed on properties
Obviously, we can't set a value on a computed property, but setting `readOnly: true` causes the computed property handler to never be called. Maybe there should be some warning, or even an error if a property is configured like this. Or possibly just ignore the `readOnly` flag if `computed` is defined.

```javascript
				_optionLabels: {
					type: Array,
					value: [],
					readOnly: true,
					computed: '_computedOptionLabels(options)'
				},
```

kevinpschaaf:


src/lib/bind/accessors.html
```diff
@@ -174,7 +174,8 @@
       // ReadOnly properties have a private setter only
       // TODO(kschaaf): Per current Bind factoring, we shouldn't
       // be interrogating the prototype here
-      if (model.getPropertyInfo && model.getPropertyInfo(property).readOnly) {
+      var info = model.getPropertyInfo(property);
```
sorvell: No longer guarding against model not having `getPropertyInfo`. We still need to do this.

src/lib/bind/effects.html
```diff
@@ -77,7 +77,7 @@
       if (args) {
         var fn = this[effect.method];
         if (fn) {
-          this[effect.property] = fn.apply(this, args);
+          this._setProperty(effect.property, fn.apply(this, args));
```
sorvell: Cannot make a readOnly property named `property`. Perhaps we should change the name of `_setProperty`?

sorvell:
LGTM

PR #1969, opened 2015-06-26
Fix typos in changelog.

arthurevans:
There was an unescaped `<content>` tag in one item, and a repeated item.

Then I tweaked some other stuff, because I liked it better that way.


sorvell:
LGTM

PR #1989, opened 2015-06-30
Keep a sum-total count of native events to properly tally for `unlisten`

### LINKED ISSUE #1988: `node.unlisten` removes native event listeners too often
Native event listeners are removed when a gesture event count reaches 0, but this will break other gesture events that are listening to the same node. Instead, a sum total count should be used to remove native listeners only when all gestures are removed.

azakus:
Fixes #1988


kevinpschaaf:
LGTM

PR #1993, opened 2015-07-01
Ensure stamped DOM reflects _hideTemplateChildren state. Fixes #1751

### LINKED ISSUE #1751: `dom-repeat` in a falsy `dom-if` should hide newly stamped children
If the `dom-repeat` in a `dom-if` stamps new children while `dom-if` is false (`_hideTemplateChildren == true`), those should be hidden.

Discovered during code review of #1635.

kevinpschaaf:


src/lib/template/templatizer.html
```diff
@@ -320,9 +320,13 @@
       // each row is a document fragment which is lost when we appendChild,
       // so we have to track each child individually
       var children = [];
+      var hide = host._hideTemplateChildren;
       for (var n = this.root.firstChild; n; n=n.nextSibling) {
```
sorvell: Consider giving TemplateInstance a `showHideChildren` method? Then use it here and in repeater's `_shotHideChildren`.

sorvell:
LGTM

PR #2004, opened 2015-07-01
Move modelForElement to Templatizer, use Polymer.Base.fire in Bind lib. Fixes #2003

### LINKED ISSUE #2003: Low-level changes for `iron-list` integration (`fire` & `modelForElement`)
Need two changes for `iron-list` integration:

- `Polymer.Bind` lib has a layering violation (noted TODO) where it calls `this.fire` from Bind.  Should be changed to call `Polymer.Base.fire`, so that the model API decoupled (`iron-list` prepares the vanilla template as a bind model, which is not a `Polymer.Base` extension)
- Since `Templatizer` performs the `_templateInstance` decoration of top-level stamped children, `modelForElement` should be a `Templatizer` API since it relies on this implementation detail.

kevinpschaaf:


sorvell:
LGTM

PR #2008, opened 2015-07-01
Fixed the google-map tag in the README

abdonrd:
The google-map element has latitude and longitude properties, not lat and long.


tjsavage:
Thanks!

PR #2027, opened 2015-07-04
Do not surround empty textnodes with span tag (dom-if)

LM450N:


LM450N:
I would love to run again the tests. Because the first time only IE10 on Windows 7 failed. Can you re-run them please ?

LM450N:
PR #2058 does the job.

sorvell:
Thanks for the PR.

Since this was so simple and it needed a test, we went ahead and just did it via https://github.com/Polymer/polymer/pull/2058.

LM450N:
No problem. I do my best but you know, Google engineers will always be the best. :wink:

PR #2031, opened 2015-07-04
"Unprevent" tap events

GrahenKraken:
After touchmove, the next tap stays prevented. Set info.prevent to false when "saving" a new tap event so any calls to Gestures.prevent('tap') only prevent the "current" tap.

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -492,6 +492,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     save: function(e) {
       this.info.x = e.clientX;
       this.info.y = e.clientY;
+      this.info.prevent = false;
     },
 
     mousedown: function(e) {


GrahenKraken:
CLA signed :)

azakus:
Can you explain a bit more about the order of operations you're seeing this in?

Are you calling `Gestures.prevent` outside of an event loop, or using a gesture event's `detail.prevent` function?

GrahenKraken:
I'm not calling prevent at all. The info.prevent property on the tap gesture stays prevented from the previous track gesture. In my case, after I swiped to open a drawer or scrolled a paper-scroll-header-panel, the next tap event had its info.prevent set to true.

kevinpschaaf:
Closing, resolved by https://github.com/Polymer/polymer/pull/2124.

PR #2036, opened 2015-07-07
Fixes #1972: make setting innerHTML correctly traverse added dom.

### LINKED ISSUE #1972: DOM API innerHTML adds only first element
[JSbin + workaround](http://jsbin.com/rufuge/edit?html,output)

When trying to set a inner HTML only the first element is added to the node. 
`Polymer.dom(node).innerHTML = "<div>added</div><div>missing</div>"`

A work around is to wrap the new html inside a div (see JS bin)

I think the culprit is [this](https://github.com/Polymer/polymer/blob/master/src/lib/dom-api.html#L617) that uses `e.nextSibling`.

I think the for should loop over `d.children` or I miss something (ex. performance reason, etc...)?

sorvell:


kevinpschaaf:
LGTM

PR #2041, opened 2015-07-08
Correct custom property regexp's to catch new lines.

### LINKED ISSUE #1974: with Polymer#1.05-update, style-sheets and custom-style-elements are not parsed in my project anymore
I think, I kept following rules for mixins and css-variables but console says "Uncaught TypeError: Cannot read property 'trim' of undefined"
Everything is fine with my style-sheets with Polymer#1.04, and I don't see anything wrong in my code
https://github.com/Fairyon/webvisual/tree/develop/src/public/css

sorvell:
Fixes #1974: correct custom property regexp's so that `--foo: \n var(--bar);` is supported.

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -344,15 +344,15 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       },
 
       rx: {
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
LGTM https://regex101.com/r/eH4aH7/1

PR #2051, opened 2015-07-08
Fixes #1888

### LINKED ISSUE #1888: Expected behavior for importNode,cloneNode
I am not sure if this is by design and/or if it is a `Polymer` or rather a general `Web-Components` spec issue. 

When I use `importNode` or `cloneNode` with `deep=true` of a custom-element that contains another custom element (via Light DOM) and then insert it into the document, the resulting DOM structure is not the same as if I would create it manually. 

See the jsbin: http://jsbin.com/haxeni/edit?html,console

sorvell:
Fixes #1888: add `cloneNode`, `importNode` to `Polymer.dom`

Includes some minor patch-dom refinements


kevinpschaaf:
LGTM

Issue #1725, opened 2015-06-02
[1.0] dom-if adds <span> tag (with lots of whitespace) above and below element


lambertsteven:
the dom-if template adds a `<span>` tag above and below the element which contains a lot of extra whitespace. When placed inside a css `display: table` layout, it completely kills the layout.

```
<dom-module id="test">
  <template>
    <div class="test">

      <!-- portrait -->
      <template is="dom-if" if="true">
        <div>Hello</div>
      </template>

    </div>
  </template>
</dom-module>
```

results in 

```
<test>
    <div class="hi style-scope test">

      <!-- portrait -->
      <span>
        </span><div class="style-scope test">Hello</div><span>
      </span>
     <template is="dom-if" if="true" class="style-scope test"></template>

    </div>
</test>
```

tregoning:
@lambertsteven just FYI a temporary workaround is to simply remove all white space in the if statement as such:
```
<dom-module id="test">
  <template>
    <div class="test">

      <!-- portrait -->
      <template is="dom-if" if="true"><div>Hello</div></template>

    </div>
  </template>
</dom-module>
```

LM450N:
I have got the same trouble. Any news about this issue ? Thanks @tregoning for the workaround.

tregoning:
Another workaround (at least when dealing with tables) is to simply remove the spans from the rendering tree (e.g set them to  display: none)

This rule worked for my case : 

tr > span {display:none}

kevinwestern:
+1 to getting this merged. This bug throws off layouts depending on flex styling, such as space-between.

kevinwestern:
FWIW: I patched this locally and got more bugs:

TypeError: Cannot set property 'display' of undefined at dom-if.Polymer._showHideChildren (<my-path>/polymer/polymer.html:3889:17) at dom-if.Polymer._render (<my-path>/polymer/polymer.html:3841:6) at dom-if.Polymer.Templatizer._flushTemplates (<my-path>/polymer/polymer.html:3071:8) at Object.Polymer.Debounce.Debouncer.complete (<my-path>/polymer/polymer.html:968:15) at Object.Polymer.Async._atEndOfMicrotask

LM450N:
I have processed the `dom-if` unit tests before send this PR and all was right. I have executed all units tests just right now and they're all passed.

[...]

PR #2058, opened 2015-07-09
Fix #1925

### LINKED ISSUE #1925: Setting readOnly AND computed on properties

sorvell:
dom-if avoids wrapping text nodes containing only whitespace


kevinpschaaf:
LGTM

LM450N:
Fix also the #1725 . And my PR #2027 may be closed.

Issue #2007, opened 2015-07-01
Array observer delivers incorrect changeRecord.path index after deleting item in array


borismus:
I register an array observer:

    observers: [
        'sightChanged(sights.*)'
      ],

sightChanged is then called argument changeRecord = {path: 'sights.2.description', ...}. If I remove something from the sights array before index 2, the path 'sights.2.description' remains the same, even though it should now be decremented by one. This results in index out of bound problems:

![screen shot 2015-07-01 at 12 27 14 pm](https://cloud.githubusercontent.com/assets/277632/8463022/8c237074-1fec-11e5-82e1-34847b99f776.png)

Seems like a bug to me. For the time being, is there a workaround to observing arrays?

kevinpschaaf:
Notifications for arrays use unique "keys" for rather than index, so what you are seeing is expected, although possibly unintuitive and not yet well documented.  This is done to ensure path linkage between objects in arrays are maintained despite possible insertions or deletions in the array.

For now, you can do this to workaround, using the `Polymer.Collection` abstraction that is used internally to manage keys for arrays:
```js
var collection = Polymer.Collection.get(this.sights);
var idx = this.sights.indexOf(collection.getItem(key));
```
Where `key` is the number pulled from `changeRecord.path`.

We'll keep this issue open to consider how to improve the API.

borismus:
Thanks. It might be nice also to have a helper method that would simplify parsing keys from pathes. Perhaps you could provide the key as `changeRecord.key`, or provide a `Polymer.getModelByPath()` method?

kevinpschaaf:
One thought is to disambiguate keys from indices in paths by using a token like `#` to denote a key, so that you could pass paths to `set` and `get` (currently `set` and `get` expect array indices, not keys, so the dichotomy is obviously bad and confusing).  The path in the above example would then be `sights.#2.description`, and calling `this.get('sights.#2')` would return the correct base object and `this.set('sights.#2.name', 'foo')` would set a field on the base object.

Aside from needing to parse the base path (e.g. `sights.#2.description` -> `sights.#2`), would this addition to the API have helped solve your use case?  Or did you actually _need_ the index of the array (as opposed to the value at the index)?

tomalec:
Forgive me being paranoid, but how about escaping both `#` and even `.` used in current implementation? as `.something` and `#1` are valid object keys

PR #2063, opened 2015-07-10
Fix setting of array indexes in path

b2ornot2b:
@set wasn't choosing the indexes correctly for arrays

e.g. @set 'arr.1.k', 123 looked up index 1 incorrectly.

### DIFF
--- a/src/standard/notify-path.html
+++ b/src/standard/notify-path.html
@@ -147,9 +147,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var last = parts[parts.length-1];
         if (parts.length > 1) {
           for (var i=0; i<parts.length-1; i++) {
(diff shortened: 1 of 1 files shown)


b2ornot2b:
I signed it!

kevinpschaaf:
The `set` API accepts array _index_ for paths containing arrays.  The intention is to dereference the value by index, but notify using collection key.  Your change seems to flip it so that that the API accepts collection _key_ and dereferences using that, which would be a breaking API change.

See this related issue and [comment here](https://github.com/Polymer/polymer/issues/2007#issuecomment-119305600), which proposes an improvement to the API to allow key and index to be disambiguated, allowing you to pass a notified path (which contains collection keys) to `set`, which may address what you were attempting to do here.

PR #2079, opened 2015-07-13
Fixes #2078

### LINKED ISSUE #2078: Custom properties fail to resolve when an element is attached to an unknown shadowRoot
Given an element added to an unknown shadowRoot like this:

```
var root = unknownHostElement.createShadowRoot();
root.appendChild(document.createElement('x-foo'));
```

An exception will occur due to the fact that it's assumed that the dom host of the `x-foo` element has a known styling api which can be interrogated. In this case the dom host of `x-foo` is not a Polymer element. We'll need to make the property resolver smarter about finding a styling host to avoid this problem and ensure that elements in unknown hosts are properly styled.

sorvell:
When computing custom style properties, make sure the styling scope is valid when the element is attached to a shadowRoot whose host is not a Polymer element.

### DIFF
--- a/polymer-micro.html
+++ b/polymer-micro.html
@@ -31,8 +31,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       this._prepAttributes();
       // shared behaviors
       this._prepBehaviors();
-      // inheritance
-      this._prepExtends();
       // factory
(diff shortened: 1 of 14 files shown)


kevinpschaaf:
LGTM

PR #2080, opened 2015-07-14
Fixes #2077

### LINKED ISSUE #2077: Bindings near special characters fail in IE11 / IE10
I have a pretty simple component to validate this with a binding like so:

```html
<link rel="import" href="../pathto/polymer.html"/>

<dom-module id="my-component">

  	<template>
             <p>&copy;</p>
             <p>{{myText}}</p>
  	</template>
</dom-module>
<script>
	Polymer({
		is: "my-component",
		properties: {
			myText: {
				type: String,
				value: 'Sample'
			}
		},
	});
</script>

```

Where myText is a property with any string value.  In IE 11 this ends up getting rendered like:

image

Here's a plunker with the issue demonstration: http://plnkr.co/edit/5LDuhpPahqKgR25Z4CZG

I did work around it by making the © being a property as well and databinding it, but this seems a bit excessive for static text.

sorvell:
Workaround IE text node splitting issue that can make text bindings fail. The behavior is such that node.normalize() fails and so we address adjacent text nodes manually.

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -147,12 +147,23 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // `annote` is supplied as it is the annote.parent of added annotations
     _parseChildNodesAnnotations: function(root, annote, list, callback) {
       if (root.firstChild) {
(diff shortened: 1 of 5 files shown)


kevinpschaaf:
LGTM

PR #2082, opened 2015-07-14
Fixes #2081

### LINKED ISSUE #2081: element.getContentChildNodes generates exception under Shadow dom when a content is not found
The `element.getContentChildNodes(selector)` generates an exception when run under Shadow DOM when the `selector` argument does not correspond to an existing `<content>` element. When run under Shady DOM, this returns an empty array. We should normalize this behavior.

The root cause is that `Polymer.dom(element).getDistributedNodes` generates an exception under Shadow DOM when element does not have this api. This should be normalized as well.

sorvell:
* make `Polymer.dom(element).getDistributedNodes` and `Polymer.dom(element).getDestinationInsertionPoints()` always return at least an empty array (was generating exception under Shadow DOM); 

* make `element.getContentChildNodes` and `element.getContentChildren` always return at least an empty array when a selector is passed that does not find a `<content>`  (was generating exception under Shadow DOM)

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -688,12 +688,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
 
       DomApi.prototype.getDestinationInsertionPoints = function() {
-        var n$ = this.node.getDestinationInsertionPoints();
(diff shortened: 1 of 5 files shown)


kevinpschaaf:
LGTM

PR #2089, opened 2015-07-15
Fix reuse logic to handle multiple mutations in same turn. Fixes #2009.

### LINKED ISSUE #2009: JS error with dom-repeat and async splice array element in followed by remove last element
Demonstration: http://plnkr.co/edit/isFkxmfNUa5mrhZL1wST?p=preview

If you splice an element into an array then remove the last element (in that order), in an async handler, and the array is items for a dom-repeat, you'll trigger a javascript error "Cannot read property '_children' of undefined".

kevinpschaaf:


sorvell:
LGTM

Issue #2010, opened 2015-07-02
Template stamping doesn't work in custom element light DOM


nomego:
Considering the following code:

```js
var appendTo = this.$.myCustomElement,
  t = document.createElement('template', 'dom-bind'),
  span = document.createElement('span');
span.innerHTML = '{{hello}}';
t.content.appendChild(span);
t.hello = 'hey';
```

Appending this to the light DOM won't work:

```js
Polymer.dom(appendTo).appendChild(t);
```

This will only render the template tag with the content within and data-binding placeholders emptied.
The `dom-change` event will actually fire and informs me that the `attached()` callback stuff of the `dom-bind` element actually have been run.
But stamping fails and moving the content outside the template never happens.
Tries with `dom-template` and manual `stamp()` also failed.

It actually needs to be in the local DOM:

```js
Polymer.dom(appendTo.root).appendChild(t);
```

Is this a bug or by design?
Could not find any documentation on this.

sorvell:
I can confirm this issue. Here's a simple test and a workaround:

http://jsbin.com/vagugufowi/1/edit?html,output (see the `//workaround` comment)

We'll try to get this fixed by the next release.

sorvell:
Related to https://github.com/Polymer/polymer/issues/1818.

nomego:
Thank you, and thanks for the workaround.
It would be semantically more accurate to append to the light DOM with consideration of the content.
Also, it's good if this process works the same way for native and custom elements.

Issue #1968, opened 2015-06-26
Auto binding template not stamped when async import is used to load polymer


ebidel:
If one makes their main page import that loads polymer.html async, e.g. `<link rel="import" href="elements.html" async>`, auto binding templates on the page do not stamp their content.

ebidel:
Current workaround:

    function importLoaded(e) {
      document.querySelector('#t')._readySelf();
    }
    <link rel="import" href="..." async onload="importLoaded(event)">

sorvell:
It looks like this only occurs on Chrome when WebComponents polyfill (`webcomponents-lite.js`) is not loaded. Can you confirm?

ok: http://jsbin.com/cahuha/1/edit?html,output
not working: http://jsbin.com/wubulo/edit?html,output

ebidel:
Yep confirmed. I'm dynamically loading the polyfill based on feature detection, which means Chrome is never getting it. Confirmed your bins too.

sorvell:
Thanks. There is an easy fix that will get the `dom-bind` to stamp, but there will be some issues that are worth a bit of discussion.

The reason that `dom-bind` does not stamp immediately is that it's a requirement that all custom elements used inside the `dom-bind` are available when the template stamps. Otherwise, bindings may propagate property values to un-upgraded elements that go unseen when the element is upgraded. This is, perhaps, unlikely (a value comes out of an upgraded element and is routed into an un-upgraded one), but it can happen.

Our typical strategy is to simply wait for HTMLImports to resolve before stamping the `dom-bind`. However, this is not a generally knowable state when async imports are used.

We could simply not provide any help for this situation when async imports are used (always stamp if the dom-bind is created after document.readyState === 'complete'. This is the easy fix).

Alternatively, we can attempt to see if any currently existing imports are loaded and wait for them to load if not. Unfortunately, HTML Imports currently gives no way to distinguish that an import has already 404'd. In that case this strategy would fail.

Another option is simply to make users stamp the `dom-bind` themselves in this case. If we go with this, then a public stamp method is necessary.

ebidel:
I'd be happy with a public method for stamping. I suppose if someone is already going the async route, they're probably comfortable/interested in full control.

sorvell:
Closed via https://github.com/Polymer/polymer/pull/2090. The `dom-bind` element now has a `render` method which can be called when using async imports.

Frikki:
@sorvell So instead of @ebidel’s fix:

```
[...]

Issue #1818, opened 2015-06-10
Dom-bind not working if parent is already attached


Sleeckx:
Hello,

please consider the following test case. Clicking "**Add dom-bind directly**" for the first time, it will append the dom-bind template but not its bound children. Upon clicking it again, it adds the bound content as intended.

http://jsbin.com/weyobaqewo/6/edit?html,output

kevinpschaaf:
Root caused to distribution reentrancy during attached.  Short-term solution is for `dom-bind` to address this specifically; @sorvell to open general issue to cover the reentrancy problem.

PR #2090, opened 2015-07-15
Shady DOM rendering + Polymer.dom.flush improvements

### LINKED ISSUE #2039: Consider making shady dom distribution trigger custom elements attached callbacks
The `iron-list` element uses `Polymer.dom(this).appendChild` to attach items. It then needs to measure the item. Since `Polymer.dom` is asynchronous, it needs to either call `flush` or measure asynchronously. However, when the custom elements polyfill is used, after the item is asynchronously attached to dom, async to that any custom elements in the item fire their attached callback. This makes it difficult for `iron-list` to know when it can measure.

If `Polymer.dom`'s rendering process also called `CustomElements.takeRecords()` then elements like `iron-list` would have a clear way to know when they can measure rendered items: call `this.async` or `Polymer.dom.flush()`.

sorvell:
Fixes #2039, #2010, #1968, #1818 

* Shady DOM better handles mutations within element attached.
* Polymer.dom.flush now processes CustomElement and template rendering and can be used when dom measuring is necessary.


kevinpschaaf:
LGTM

PR #2097, opened 2015-07-16
Allow setting non-index array properties. Fixes #2096.

### LINKED ISSUE #2096: Array-like objects fail to update the template
When working with array like objects, data-binded properties do not properly update the view later when setting to those values.

http://jsbin.com/mamogo/1/edit?html,output

seems to be linked to the notifyPath reporting changes as 'complexArray..splices'

kevinpschaaf:


sorvell:
LGTM

PR #2100, opened 2015-07-17
Fix misspelling.

japacible:


kevinpschaaf:
Fixed in https://github.com/Polymer/polymer/commit/97944e42d85fb3fceff81786d38a18d6f5289a3c

PR #2109, opened 2015-07-18
Fix #2107: improve binding expression parser to match valid javascrip…

### LINKED ISSUE #2107: Binding expressions should be able to start with non-word characters like $
Here's an example: `$I(foo, baz, bar)`

The parser for expressions currently looks for non-word characters so it misses expressions that start with $. Since it's valid to write javascript functions like this, it should be supported.

sorvell:
…t property names.

### DIFF
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -168,7 +168,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     // method expressions are of the form: `name([arg1, arg2, .... argn])`
     _parseMethod: function(expression) {
-      var m = expression.match(/(\w*)\((.*)\)/);
+      // tries to match valid javascript property names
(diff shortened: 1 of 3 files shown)


src/standard/effectBuilder.html
```diff
@@ -168,7 +168,8 @@
 
     // method expressions are of the form: `name([arg1, arg2, .... argn])`
     _parseMethod: function(expression) {
-      var m = expression.match(/(\w*)\((.*)\)/);
+      // tries to match valid javascript property names
+      var m = expression.match(/([a-zA-Z_$][0-9a-zA-Z_$]*)\((.*)\)/);
```
azakus: Why not just `/([\w\d$]+)\((.*)\)/`?
That will capture more valid cases, including functions that have to be quoted in the prototype.

kevinpschaaf:
LGTM

PR #2117, opened 2015-07-19
Fixes #2065, fixes PolymerElements/paper-drawer-panel#80, duplicates @myfreeweb's commit

### LINKED ISSUE #2065: Scroll, user has to tap twice to fire tap event.
Hi,

I am developing a web app using polymer. Everything works great on desktop and almost everything works great and mobile expect one view. This view is accessible [here](https://alpha.wid.la/#!/projects).

On this specific view, the user could scroll the page and tap on a card to display a modal view. The problem is (**on mobile**):
* if user doesn't scroll and directly taps on a card. It works perfectly.
* if user scrolls and taps on a card, nothing happen but if the user taps on the same card again, it works.

I can reproduce this bug on my OnePlus One, Nexus 9 and ios simulator.

I have tried to debug it but before I explain to you what "I found", here is  a similar view structure that I use:

```html
<paper-drawer-panel> <!-- this element listen on **track** event to be able to swipe the drawer -->

  <div drawer>...</div>
  <paper-scroll-header-panel main>

   <card-list>
     <template is="dom-repeat" items="[[cards]]">
       <card model="[[item]]"></card> <!-- this element listen on **tap** event and fire a new event -->
     </template>
   </card-list>

  </paper-scroll-header-panel main>

</paper-drawer-panel>
```

So after debugging, I observed:
* when the user directly taps on a card, [info.prevent](https://github.com/Polymer/polymer/blob/815bb975dadbb494c19c875216e09fe70c643ecb/src/standard/gestures.html#L527) is `false` and everything works great, the `tap` event is fired.
* when the user scrolls, first the `<card/>` `tap` event should occurred. But as [`dy`](https://github.com/Polymer/polymer/blob/815bb975dadbb494c19c875216e09fe70c643ecb/src/standard/gestures.html#L525) is too high, the event is not fired. The gesture is also [reset](https://github.com/Polymer/polymer/blob/815bb975dadbb494c19c875216e09fe70c643ecb/src/standard/gestures.html#L535). Second the  `<paper-drawer-panel/>` `track` event should occurred and it does. So [this](https://github.com/Polymer/polymer/blob/815bb975dadbb494c19c875216e09fe70c643ecb/src/standard/gestures.html#L451) line is executed. This will prevent the next tap event. I guess all of this is normal. But then. the user will really tap on the card and now [info.prevent](https://github.com/Polymer/polymer/blob/815bb975dadbb494c19c875216e09fe70c643ecb/src/standard/gestures.html#L527) is `true` because of the previous `Gestures.prevent('tap');`. The `tap` event is not fired but the [tap gesture is reset](https://github.com/Polymer/polymer/blob/815bb975dadbb494c19c875216e09fe70c643ecb/src/standard/gestures.html#L535). The user has to tap again on the card to finally fire the `tap` event.

If I remove this [line](https://github.com/PolymerElements/paper-drawer-panel/blob/0339161b5ee1c46f587b92beab70fec13d507378/paper-drawer-panel.html#L342). Everything works perfectly (obviously except the gesture to open the drawer).

I hope I was clear enough. Does someone have an idea ?
Thanks in advance,

Thomas

m4b:
Fixes #2065, fixes PolymerElements/paper-drawer-panel#80 

Reproduces @myfreeweb's commit, but looks like he won't sign the CLA (don't blame him), so here we are.

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -504,6 +504,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     mousedown: function(e) {
+      this.info.prevent = false;
       this.save(e);
     },
     click: function(e) {
@@ -511,6 +512,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     touchstart: function(e) {
+      this.info.prevent = false;
       this.save(e.changedTouches[0]);
     },
     touchend: function(e) {


m4b:
heh; i guess cla isn't signed. what a day already

kevinpschaaf:
Resolved by https://github.com/Polymer/polymer/pull/2124

PR #2126, opened 2015-07-21
Fixes #2125: adds a register method to dom-module to support imperati…

### LINKED ISSUE #2125: Add better support for imperatively created `dom-module`
The `dom-module` element registers itself when its `createdCallback` is called. This makes it impossible to create via `createElement`. The `dom-module` element should support a `register` method to support imperative creation.

sorvell:
…ve creation.

### DIFF
--- a/src/lib/dom-module.html
+++ b/src/lib/dom-module.html
@@ -4,36 +4,76 @@
 
   var modules = {};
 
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
LGTM

PR #2134, opened 2015-07-22
Disable tracking if scrolling

azakus:


azakus:
Fixes #2135

kevinpschaaf:
LGTM

hectorguo:
Hi, I still got this issue in version 1.2.3. 

When I bind `track` event on each item of a tall list, I can't scroll up and down. 

I am trying to simulate the iOS-style list, you know, swiping left to delete. But I don't know how to disable track event when users scroll up and down.

PR #2140, opened 2015-07-24
Allow array API's accept string & negative args. Fixes #2062.

### LINKED ISSUE #2062: Array mutations can fail when arguments to splice are a string
Repro here: http://jsbin.com/faqetabada/2/edit?html,output

Should just parseInt on args before notifying splices.

kevinpschaaf:
Brings the API more in line with native splice, etc.


sorvell:
LGTM

PR #2170, opened 2015-07-30
Fixes #2154: ensure Polymer.dom always sees wrapped nodes when Shadow…

### LINKED ISSUE #2154: TypeError: 'mozMatchesSelector' called on an object that does not implement interface Element.
Polymer 1.0

Not a issue in Chrome, only Linux Firefox 31.8.0. 

Not sure if this is a Polymer issue or not, since the error is coming from `webcomonents.js` version `0.7.3`:

In the console I see `TypeError: 'mozMatchesSelector' called on an object that does not implement interface Element.` which originates from code: 

```javascript
    var MatchesInterface = {
      matches: function(selector) {
        selector = shimMatchesSelector(selector);
        return scope.originalMatches.call(unsafeUnwrap(this), selector);
      }
    };

```

This happens when I click on:

```html
        <template is="dom-bind" id="videoButtonClick">
          <div class="horizontal layout center" style="height: 100%">
            <div class="layout vertical center" style="width: 100%">
              <div id="play-button-container" class="horizontal center-center layout wrap">
                <div id="play-button-png" on-click="openVideo"></div>
                <h5>See and hear the story</h5>
              </div>
            </div>
          </div>
        </template>
```

which runs:

```javascript
var app = document.querySelector('#videoButtonClick');
app.openVideo = function () {
	videoPlayer = Polymer.dom(document.rootElement).querySelector('video-player');
	videoPlayer.open();
};
```

Also, Polymer 1.0 loads/renders extremely slowly on Firefox.

sorvell:
…DOM polyfill is in use.

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -31,6 +31,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
     };
 
+    // ensure nodes are wrapped if SD polyfill is present
+    if (window.wrap && Settings.useShadow && !Settings.useNativeShadow) {
+      DomApi = function(node) {
+        this.node = wrap(node);
+        if (this.patch) {
+          this.patch();
+        }
+      };
+    }
+
     DomApi.prototype = {
 
       flush: function() {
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
LGTM

PR #2173, opened 2015-07-30
Fix dom bind example in jsdoc

tomalec:
Rename `core-ajax` to `iron-ajax`, and add `auto` attribute as in https://github.com/Polymer/docs/pull/1276


kevinpschaaf:
Thanks!

PR #2174, opened 2015-07-30
debouncer unit tests

gronke:
- Add some unit tests for Debouncer utility
- Remove unused test assertion (tested setTimeout instead of debounce)
- cast return parameter of isDebouncerActive

### DIFF
--- a/test/unit/utils.html
+++ b/test/unit/utils.html
@@ -94,6 +94,71 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     });
 
+    test('debounce flushing', function(done) {
+
+      var called = 0;
+      var cb = function() {
+        called++;
+      };
+
+      window.el1.debounce('foo', cb);
+      window.el1.flushDebouncer('foo');
+      window.el1.debounce('foo', cb);
+
+      setTimeout(function() {
+        assert.equal(called, 2, 'debounce should be called twice');
+        done();
+      }, 100);
+
+    });
+
+    test('debounce state', function(done) {
+
+      window.el1.debounce('foo', function() {
+        assert.equal(window.el1.isDebouncerActive('foo'), false, 'debouncer is inactive after resolution');
+        done();
+      });
+
+      assert.equal(window.el1.isDebouncerActive('foo'), true, 'debouncer is active immediately after triggering');
+
+    });
+
+    test('debounce cancelling', function(done) {
+
+      var triggered = false;
+
+      window.el1.debounce('foo', function() {
+        triggered = true;
+      });
+      window.el1.cancelDebouncer('foo');
+      assert.equal(window.el1.isDebouncerActive('foo'), false, 'debouncer is inactive after cancelling');
+
+      setTimeout(function() {
+        assert.equal(triggered, false, 'debounce never fired');
+        done();
+      }, 100);
+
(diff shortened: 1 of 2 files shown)


test/unit/utils.html
```diff
@@ -91,12 +90,76 @@
 
       setTimeout(function() {
         assert.equal(called, 1, 'debounce should be called exactly once');
-        assert(Date.now() - now > 100, 'debounce should be called after at least 100ms');
```
kevinpschaaf: Is there a reason you took out this assert?

`debounce` supports two modes, one without a timeout, in which case it will run at next microtask, and one with a timeout, in which case it should run no earlier than the timeout (which is what that was verifying).

gronke: In my opinion testing of ``setTimeout`` should be done by the browser vendors and in case of Microtasks in [test/unit/async.html](https://github.com/Polymer/polymer/blob/master/test/unit/async.html), so that we can rely on them here.

@kevinpschaaf thanks for pointing out that ```wait``` and ```no-wait``` debounce behavior should be tested for different issues. Added more explicit tests for each of them and as replacement of the ``Date.now`` assertion I removed there now is a second timeout after ``50ms`` that asserts the callback not to be fired yet.

test/unit/utils.html
```diff
@@ -73,30 +73,116 @@
...
       window.el1.debounce('foo', cb);
-      window.el1.debounce('foo', cb, 100);
+      window.el1.debounce('foo', cb, 50);
       window.el1.debounce('foo', cb, 100);
 
       setTimeout(function() {
-        assert.equal(called, 1, 'debounce should be called exactly once');
-        assert(Date.now() - now > 100, 'debounce should be called after at least 100ms');
+        assert.equal(called, 0, 'callback is not called yet');
+      }, 50);
```
gronke: The lines above are replacing the the assertion that was deleted in the first commit of this PR.

gronke:
@kevinpschaaf Do you want these commits to be squashed to something more descriptive? What else can I do to get this ready to merge?

samccone:
Topically this looks good to me, @kevinpschaaf would you mind giving this a once over as well¿

gronke:
Danke @samccone :blush: - branch is rebased

gronke:
The quick fix https://github.com/Polymer/polymer/pull/2291/ patches the same :beetle:, but is less work to review - tests can be added later.

gronke:
While hunting a bug in an external element, I needed this tests again to ensure debounce is reliable. Can somebody please take a look at this tests and consider merging them? Thank you!

azakus:
LGTM2, I'll handle the rebasing.

azakus:
Merged to master as https://github.com/Polymer/polymer/commit/9b898c25cca663dec22e8368e5cfc87a08aff628, https://github.com/Polymer/polymer/commit/02068525759c6ee4cbc2178c04ceb30372f2ac40, https://github.com/Polymer/polymer/commit/391649341b6449810191b4040f7e28ab23b5420b, https://github.com/Polymer/polymer/commit/9bef4c057c6da56f534bab0beac911ba80998c67, and https://github.com/Polymer/polymer/commit/8ef7bac6dc92c44fa702c3c9606b2a5d0d1d43eb

PR #2190, opened 2015-08-03
Fixes #1673: ensure instance effects exist before marshaling attributes.

### LINKED ISSUE #1673: Can't instantiate component under CE polyfill (where hostAttributes triggers property change handler)
Polymer 1.0 appears to contain a regression from Polymer 0.9. This [gist](https://gist.github.com/JanMiksovsky/16f7f5f2fc5cca95fda1) shows a component that uses hostAttributes to set a property; setting the property triggers a property changed handler.

This element cannot be instantiated under Shady DOM. It raises an exception:

```
TypeError: undefined is not an object (evaluating 'this.__data__[property]')
```

It looks like a low-level function called _propertySet is trying to access a property bag called __data__ which has not yet been created.

Code such as this used to work up to Polymer 0.9.

This is probably a small bug somewhere, but it's currently preventing Basic Web Components like [basic-list-box](component.kitchen/components/basic-web-components/basic-list-box) and [basic-carousel](component.kitchen/components/basic-web-components/basic-carousel) from working under Polymer 1.0. We'd obviously like to be able to have those components work with the final 1.0 release.

### LINKED ISSUE #2185: Firefox fails to show custom element with properties bound with two-way binding + hostAttributes
If we 

* use Polymer 1.0.8
* have a custom element (either in a separate page or inline as below) with a property
* the property is bound using two way binding
* has a hostAttributes on the element
* we are using Firefox 41.0a2 (not tested on earlier) (it works fine on Chrome Version 42)

Then we encounter the error: 

    TypeError: this.__data__ is undefined. 

The error comes from "var old = this.__data__[property];" line 1153 from polymer.html.js according to the firefox debug tools.

If, however, we remove the hostAttributes, then it displays fine in firefox.

Below is the code that bails in Firefox. 

```
<html>
    <head>
        <script src="bower_components/webcomponentsjs/webcomponents-lite.min.js"></script>
        <link rel="import" href="bower_components/polymer/polymer.html">
    </head>

    <body>

        <dom-module id="an-element">
            <template>
                <span>{{thing}}</span>!
            </template>
            <script>
                    Polymer({
                        is: "an-element",

                        hostAttributes: {
                            thing: "yeah"
                        },

                        properties: {
                            thing: {
                                type: String,
                            }
                        }
                    });
                });
            </script>
        </dom-module>

        <template is="dom-bind" id="app"> 
                <an-element thing="{{thing}}"></an-element>
        </template>
    </body>

    <script>
        window.addEventListener('WebComponentsReady', function(e) {
            document.querySelector("#app").thing='XXX'; // XXX is only seen in chrome, not firefox.
        });
    </script>
</html>
```

sorvell:
Fixes #2185

### DIFF
--- a/polymer.html
+++ b/polymer.html
@@ -71,12 +71,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       this._popHost();
       // concretize template references
       this._marshalAnnotationReferences();
-      // install host attributes
-      this._marshalHostAttributes();
       // setup debouncers
       this._setupDebouncers();
       // concretize effects on instance
       this._marshalInstanceEffects();
+      // install host attributes
+      this._marshalHostAttributes();
       // acquire instance behaviors
       this._marshalBehaviors();
       // acquire initial instance attribute values
--- a/src/lib/base.html
+++ b/src/lib/base.html
@@ -49,8 +49,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     // reserved for canonical behavior
     attributeChangedCallback: function(name) {
-      this._setAttributeToProperty(this, name); // abstract
+      this._attributeChangedImpl(name); // abstract
       this._doBehavior('attributeChanged', arguments); // abstract
+    },
+
+    _attributeChangedImpl: function(name) {
+      this._setAttributeToProperty(this, name);
     },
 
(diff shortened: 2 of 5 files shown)


kevinpschaaf:
LGTM

Issue #1852, opened 2015-06-12
What is the replacement for template bind ref in Polymer 1.0?


miyconst:
In Polymer 0.5 we have:

``` html
<template id="common-html">
    This template has some common html code.            
</template>

<template bind ref="common-html"></template>
```

How to achieve this functionality with Polymer 1.0?

Here is the same StackOverflow question, unfortunately without answer.
http://stackoverflow.com/questions/30516697/polymer-1-0-template-bind-ref-equivalent

robrez:
+1, I've been working around it w/ a custom-element in lieu of the old approach.

kevinpschaaf:
For 1.0, we scoped the template/binding system fairly tightly to cover common CE use cases.

Adding `ref` support is a fairly large feature, but it's on our radar.  Due to how we refactored the binding system to do all the heavy lifting at registration time (rather than instance time), there are other pre-requisite features (e.g. dynamic addition of bindings at instance time) that would need to be added before this could be implemented.

We'll keep this issue open to track any future progress on the feature.

bendavis78:
@kevinpschaaf, is there an alternative/workaround in 1.0 using javascript? It's not clear looking at the existing javascript API how to take a template and "stamp" it out with a model.

kevinpschaaf:
@bendavis78 No, as mentioned there are other pre-requisite features that would need to be added.  In 1.0, templates don't "observe" their models, rather models (elements) have accessors that push data to their templates, which requires up-front cooperation from the model.

tomalec:
+1 for getting back `ref` support

gazal-k:
Is this on the roadmap for 1.x or 2.x ?

seintelligence:
+1 for getting back ~~ref~~ support

Trakkasure:
All, I created a custom tag to help with this: https://github.com/Trakkasure/dom-bindref
It relies on Object.observe since I don't have intimate knowledge of Polymer's internals nor the time to dig in at the moment.

gazal-k:
@Trakkasure thanks for sharing. Can it bind an object to the stamped template though?

For example, the following should display "value" within a span tag.

```html
<template is="dom-template" id="template-bind"><span>[[key]]</span></template>
<dom-bindref ref="template-bind" id="template-bind-ref" bind='{"key":"value"}'></dom-bindref>
```

gazal-k:
[...]

Issue #1806, opened 2015-06-09
Does version 1.0 support recursion?


masaoliou:
Version 0.5 allows us to recursively generate elements using **ref** syntax. Can this be done in version 1.0, and how?

dhubler:
Same way as in 0.5

vsimonian:
`ref` in repeating templates isn't currently supported. I'm not sure whether or not it's on the roadmap, though - that's a question worth asking, I think.

A theoretical code sample for anyone peeking into this issue:

```html
<template id="recursiveTemplate" is="dom-repeat" items="[[menuItems]]">
  <template is="dom-if" if="[[item.items]]">
    <div class="submenu">
      <header>[[item.name]]</header>
      <template is="dom-repeat" ref="recursiveTemplate"></template>
    </div>
  </template>
  <template is="dom-if" if="[[!item.items]]">
    <paper-item>[[item.name]]</paper-item>
  </template>
</template>
```

kirilledelman:
This doesn't seem to work:

            <template is="dom-repeat" items="{{items}}" id="mindmap-row">

                <div id="{{item.id}}" on-click="rowClicked"></div>

                <template is="dom-if" if="{{item.children}}">

                    <iron-collapse>
                        <template is="dom-repeat" items="{{item.children}}" ref="mindmap-row"></template>
                    </iron-collapse>

                </template>

            </template>

My collapse section is always empty - template isn't referenced.

vsimonian:
@kirilledelman If you're on 1.0, you're not doing anything wrong. There's currently no support for `ref` in templates.

kirilledelman:
Oh I see. That's too bad, as it's a very useful feature. I'll have to rewrite using recursive elements instead.

zikes:
You can include a component inside of itself: http://stackoverflow.com/questions/30468581/polymer-recursive-template-binding/30469213#30469213

masaoliou:
Including a component inside of itself as zikes pointed out works for myself.

vsimonian:
@zikes, now that's a creative solution! Never even occurred to me. :+1:

robrez:
Did anyone get word if "ref" is going to be implemented?  Recursion-aside, it's nice to be able to "stamp" stuff out in a named template.

Maybe a custom element for any such case is going to be the recommend approach.

robrez:
It looks like "ref" is officially on the radar...
See here: #1852

PR #2196, opened 2015-08-04
Recursive template reference capability

gazal-k:
Added new extension to template element; 'dom-bindref'.

Fixes:
 - https://github.com/Polymer/polymer/issues/1852
 - https://github.com/Polymer/polymer/issues/1806


gazal-k:
I signed it!

gazal-k:
I'll add the tests for "dom-bindref" soon. The failures on IE 7 don't seem to be a result of the changes here.

gazal-k:
added tests for `dom-bindref`. Seems like those are also failing in IE 10 along with the already failing cases for `dom-bind` and `dom-repeat` in IE 10. Not sure whether `dom-bindref` cases are failing due to failures in `dom-bind` and `dom-repeat`.

simonfan:
Registering the element using the `Polymer.Class` and `document.registerElement('el-ement', ElementClass)` syntax seems to be working quite nicely. 

Haven't tested thoroughly though..

```
<link rel="import" href="../bower_components/polymer/polymer.html">

<dom-module is="recursive-polymer">
    
    <template>

        <h1>{{ data.name }}</h1>

        <div id="children">

            <template is="dom-repeat" items="{{ data.children }}">
                <recursive-polymer id="{{ item.name }}" data="{{ item }}"></recursive-polymer>
            </template>

        </div>

    </template>

    <script>
        var RecursivePolymer = Polymer.Class({
            is: 'recursive-polymer',

            listeners: {
                tap: 'handleTap'
            },

            handleTap: function () {
                console.log('tapped')
                this.set('data.name', 'test')
            },

            properties: {
                data: {
                    type: Object,
                    notify: true
                }
            }
        });

        document.registerElement('recursive-polymer', RecursivePolymer);

    </script>


</dom-module>
```

zmoshansky:
@azakus  et al. Feedback to land this PR please? It would be very useful, and based on interest on my [stack overflow question](https://stackoverflow.com/questions/34645114/polymer-dom-repeat-binding-to-content?noredirect=1#comment57249182_34645114), quite in demand.

btelles:
+1 I'm wishing this feature were available as well. 
Context:
I'm creating a generic CRUD element that, when declared, one of its "content" elements is a template that is passed into a dom-repeat template . It seems quite a few people have run into this problem as well (per StackOverflow question in zmoshansky's comment above).

ebidel:
cc @kevinpschaaf @sorvell @sjmiles

sorvell:
We plan to add support for runtime binding which is a pre-requisite to fully supporting a bind ref feature like this. This implementation won't support, for example, outer scope (host) binding.

The element defined in this PR is useful for a subset of cases, and I'd recommend publishing it separately so users can take advantage of it. Sorry for the delay and thanks for the effort!

PR #2198, opened 2015-08-04
Refines how element styles are marshaled for custom elements

sorvell:
Refines how element styles are marshaled for custom elements. 

* `<custom-style>` now supports a `module` attribute that is a space separated list of `<dom-module>` `id`'s which contain style data.
* `<style>` elements in custom element `<dom-module>`'s also support a `module` attribute
* `<dom-module>`'s can marshal style data from `<style>` elements inside themselves OR inside contained `<template>`'s.

Loading remote style data should now be done by loading an import containing a dom-module and marshaling the data using the module attribute as described above.

We will soon be deprecating the  use of an HTML Import inside a `<dom-module>` to load style data. This allows a file to contain pure css text like a stylesheet file, but it causes the file to parse as HTML which has too many negative side effects: possible security concern, performance issue, and generates warning message in chrome.


kevinpschaaf:
LGTM

Issue #2176, opened 2015-07-30
Feature request: simplify properties definition with default value


nazar-pc:
Sometimes we don't have default value for property, in this case we write:
```javascript
...
  properties: {
    prop : String
  }
...
```

It is simple enough and quite readable.
But if we need to specify default property value it instantly become unnecessary more complex (no notifying or reflection to property, just default value):
```javascript
...
  properties: {
    prop : {
      type : String,
      value : "Default value"
    }
  }
...
```
I think there is no problem to infer property type from default value, so we can write just like something closer to what Polymer 0.5 had, but still do the same thing as full definition:
```javascript
...
  properties: {
    prop : "Default value"
  }
...
```
Determining property type is simple:
```javascript
var type = (function(prop){
  switch (typeof prop) {
    case 'boolean':
      return Boolean;
    case 'number':
      return Number;
    case 'string':
      return String;
    default:
      if (prop instanceof Date) {
        return Date;
      } else if (prop instanceof Array) {
        return Array;
      }
      return Object;
  }
})(this.prop);
```
Example above basically covers all types currently supported.

Such simplification might make a lot of code easier to read, small piece of magic that makes life better.

If you think it will be accepted, I can prepare PR with this feature + tests for it.

nazar-pc:
@sorvell, should this be closed since #2205 was rejected or there is still some space for discussion?

PR #2205, opened 2015-08-05
Simplified default value decaration

nazar-pc:
I was not able to wait and implemented #2176 myself.
It allows turn:
```javascript
...
  properties: {
    prop : {
      type : String,
      value : "Default value"
    }
  }
...
```
into
```javascript
...
  properties: {
    prop : "Default value"
  }
...
```
which radically improves readability of this common use case without sacrificing performance.

It works with strings, numbers, boolean values, date objects and arrays (everything supported by Polymer), generic objects are ignored because it will cause random errors otherwise.

I wasn't able to run WCT in any way I've tried, so tests were not actually tested by myself.

### DIFF
--- a/polymer.html
+++ b/polymer.html
@@ -42,6 +42,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       this._prepEffects();
       // shared behaviors
       this._prepBehaviors();
+      // convert simplified default value for properties into expanded form
+      this._prepSimplifiedProperties();
       // fast access to property info
       this._prepPropertyInfo();
       // accessors part 2
@@ -85,7 +87,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // acquire instance behaviors
       this._marshalBehaviors();
       /*
-      TODO(sorvell): It's *slightly() more efficient to marshal attributes prior 
+      TODO(sorvell): It's *slightly() more efficient to marshal attributes prior
       to installing hostAttributes, but then hostAttributes must be separately
       funneled to configure, which is cumbersome.
       Since this delta seems hard to measure we will not bother atm.
--- a/src/standard/configure.html
+++ b/src/standard/configure.html
@@ -208,6 +208,35 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // property, Chrome will leak memory across page refreshes
       // https://crbug.com/529941
       this._handlers = [];
+    },
+
+    _prepSimplifiedProperties: function() {
+      var property, type, value;
+      for (property in this.properties) {
+        value = this.properties[property];
+        type = (function() {
+          switch (typeof value) {
+            case 'boolean':
+              return Boolean;
+            case 'number':
+              return Number;
+            case 'string':
+              return String;
+            default:
+              if (value instanceof Date) {
+                return Date;
+              } else if (value instanceof Array) {
+                return Array;
+              }
+          }
+        })();
+        if (type) {
+          this.properties[property] = {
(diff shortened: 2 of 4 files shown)


nazar-pc:
Same tests failed in #2204 which fixes just single typo, so no new issues introduced.

nazar-pc:
Rebased against master, moved to Polymer micro, otherwise attributes values were not applied during `ready` callback, one more test added for mentioned issue.

tjsavage:
The feature here seems clear as per [this issue](https://github.com/Polymer/polymer/issues/2176) - @kevinpschaaf @sorvell @sjmiles thoughts on the ergonomic impact?

nazar-pc:
Again moved to Standard, should be better place for it

nazar-pc:
Eventually combined hacks for Micro and Standard together

nazar-pc:
Welcome version 4 of implementation.
Now it is pretty simple, reliable and contains even more tests.
I think it is ready to be upstreamed, give me some feedback here, please.
This PR is more than 4 months old.

TimvdLippe:
Correct me if I am wrong, but I think this currently does not support behaviors?

sorvell:
Thanks for the effort here and sorry for the delay. After review, we've elected not to merge this PR. In Polymer 0.5 type info was automated as in the PR here. There was too much confusion around whether the type of the property reflected the current type or the type of the default value, since these may be different.

If you like this functionality, I recommend implementing it via a behavior. The correct place to add the fix up is by overriding `_prepBehavior` since every behavior can provide properties.

nazar-pc:
@TimvdLippe, apparently it doesn't, but I'll add it for sure.

@sorvell, what do you mean by *current type*? I just do see the source of confusion if we reduce `x : {type : Y, value : Z}` to `x : Z`. For me it is absolutely clear where type is, this is just convenient simplification, can you elaborate on this a bit?

sorvell:
Let's say you have `properties: { foo: 'default' }` then in your code you set `this.foo = 5`. The initial value was a String and the current value is a Number. What type should the `foo` attribute deserialize to? We're hoping that the need to explicitly set foo's `type: String` will help people avoid mistakes and confusion. An element behavior that implements this is a good way to customize the 'behavior' here as you like.

nazar-pc:
Well, there is basically no difference whether you have expanded declaration or compact. It will be definitely serialized to string as declared initially. My though was just to add this to documentation. If someone have difficulties with remembering this - they are free to use either format. I just found annoying writing expanded form in some cases and saw other people here in issues and on StackOverflow that were wondering why compact form is not supported.

For instance here: https://github.com/Zecat/paper-typeahead/commit/2c014a7356ef213b43638f953bebed537413679a#diff-61fcf7ebb4ae4e94dbd85773cbc463c5R9 It is so natural that people making mistakes.

hawkett:
I agree with @nazar-pc - there is no ambiguity with the compact version - 100% clear that it should deserialize to string - foo:'default' and foo:String say the same thing wrt type.

Compacting property declarations this way would be very nice IMO.

nazar-pc:
For those curious I've reworked this PR into standalone version that might be included right after Polymer itself ([source](https://github.com/nazar-pc/CleverStyle-CMS/blob/02522d16c7124ee82ded69f7f1b06284b8a05f01/includes/js/a5.Polymer-simplified-default-value-declaration.ls), [compiled](https://github.com/nazar-pc/CleverStyle-CMS/blob/02522d16c7124ee82ded69f7f1b06284b8a05f01/includes/js/a5.Polymer-simplified-default-value-declaration.js))

PR #2214, opened 2015-08-05
Always trigger tap for synthetic click events

### LINKED ISSUE #2212: Tap gesture won't be generated by synthetic `click` under certain circumstances
When calling the `click` method on an element to generate a synthetic click event, the values for `clientX` and `clientY` on the corresponding `click` event are different across browsers and generally not useful. The code that generates `tap` gesture events relies on measuring the distance between a given `click` event and a preceding "down" pointer event.

https://github.com/Polymer/polymer/blob/098b2991ec7e5720cdbc39c6d4dfef79a31129a4/src/standard/gestures.html#L627

When generating a synthetic `click`, this conditional behavior can result in a failure to fire a corresponding `tap` event. This has negative consequences for a11y (among other things) in some of our elements.

azakus:
Fixes #2212

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -111,6 +111,25 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
   }
 
+  function isSyntheticClick(ev) {
+    if (ev.type === 'click') {
+      // ev.detail is 0 for HTMLElement.click in most browsers
+      if (ev.detail === 0) {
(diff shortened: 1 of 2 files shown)


azakus:
@cdata @sorvell @kevinpschaaf PTAL

cdata:
Me likey!

kevinpschaaf:
LGTMish

PR #2243, opened 2015-08-10
Fixes #2235. Manages logical information in shady distribution more d…

### LINKED ISSUE #2235: Invalid childNodes on previous parent is returned after calling domApi.appendChild
http://jsbin.com/litugehexu/edit?html,console,output

Also the following issue could be a different mutation of the reported issue. Both issue are due to caching the _lightChildren on the parent

After calling domApi(customElement).appendChild(domApi(customElement).nextSibling) and then flush, the nextSibling of the custom element is not reported correctly by domApi.

http://jsbin.com/hijoxawewa/edit?html,console,output

sorvell:
…irectly by capturing it explicitly when needed and not whenever distribution is run.

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -223,8 +223,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       },
 
       _updateInsertionPoints: function(host) {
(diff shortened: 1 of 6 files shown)


kevinpschaaf:
LGTM

PR #2247, opened 2015-08-11
 Some default computed bindings methods

### LINKED ISSUE #2237: Some default computed bindings methods
There are some very often cases that should be available out of the box, for instance, I have following and similar methods in multiple places:
```coffeescript
icon_class : (icon) ->
	"fa fa-#{icon}"
```

I suggest following list of default methods for computed bindings:
* `if(condition, then [, else [, prefix [, postfix]]])` - allows conditionally applying `active` CSS class alongside with other "constant" classes and the likes
* `join(array [, separator = ','])` - might be used for applying array of classes and things like that
* `concat(thing [, another [,...]])` - allows to do simple concatenations, like in example above
* `and(x, y [, z [,...]])` - since it is not allowed to use any expressions except negation, this might be useful for boolean attributes/properties
* `or(x, y [, z [,...]])` - similar to previous
* `xor(x, y [, z [,...]])` - similar to previous

This is bare minimum I think, with #2163 implemented this will cover most of trivial cases without any need to reinvent the wheel.

nazar-pc:
Added to Polymer Standard
Fixes #2237

### DIFF
--- a/polymer.html
+++ b/polymer.html
@@ -111,3 +111,4 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="src/lib/template/array-selector.html">
 <link rel="import" href="src/lib/template/dom-if.html">
 <link rel="import" href="src/lib/template/dom-bind.html">
+<link rel="import" href="src/lib/bind/default-computed-bindings.html">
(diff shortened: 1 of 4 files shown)


azakus:
Sorry for the delay in looking at this.

Given our response to the issue (https://github.com/Polymer/polymer/issues/2237#issuecomment-179532171), we've decided not to accept this PR.

We appreciate the effort you put into making Polymer better!

nazar-pc:
For those curious I've reworked this PR into standalone version that might be included right after Polymer itself ([source](https://github.com/nazar-pc/CleverStyle-CMS/blob/02522d16c7124ee82ded69f7f1b06284b8a05f01/includes/js/a5.Polymer-default-computed-bindings.ls), [compiled](https://github.com/nazar-pc/CleverStyle-CMS/blob/02522d16c7124ee82ded69f7f1b06284b8a05f01/includes/js/a5.Polymer-default-computed-bindings.js))

PR #2258, opened 2015-08-12
Fix for loosing prototype's property value during getter/setter initialization

### LINKED ISSUE #2257: Strange issue with data binding when properties are in prototype
```html
<dom-module id="my-element">
    <template>   
      <span>{{x}}</span>
      <span>{{y}}</span>
    </template>
    <script>
      Polymer({
        is: 'my-element',
        x: 'x',
        y: '',
        ready: function () {
          this.y = 'y';
        }
      });
    </script>
</dom-module>
```
In this example only `y` will be rendered, `x` property of created element is `undefined` O_o.

nazar-pc:
Basically prototype's value was overridden with accessor.
Check added - if instance value doesn't exist - use value from prototype, just like it work during regular prototype inheritance.
Fixes #2257 (partially)

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -157,10 +157,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // create accessors that implement effects
 
     _createAccessors: function(model, property, effects) {
+      var prototype_value = model[property];
       var defun = {
         get: function() {
           // TODO(sjmiles): elide delegation for performance, good ROI?
-          return this.__data__[property];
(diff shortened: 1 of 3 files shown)


tjsavage:
Adding @kevinpschaaf to match corresponding [issue](https://github.com/Polymer/polymer/issues/2257)

nazar-pc:
With master code causes errors, something changed withing Polymer:(

sorvell:
Sorry for the delay and thanks for the effort here. The supported way to define default values for properties is via the value field for a property in the `properties` object. After review, we've decided not to support 2 ways to do this. If you like this feature, it's something that could possibly be implemented via a behavior.

PR #2266, opened 2015-08-13
Small performance tweak during object cloning

nazar-pc:
If most of properties are non-undefined, suggested improvement makes cloning twice as fast.
Performance test (results in 2x improvement in stable Firefox and Chromium):
```js
var w = {};
for (var i = 0; i < 10000; ++i) {
	w[i] = Math.random();
}

for (var i = 0; i < 1000; ++i) {
	w[i] = null;
}


var result1, result2;
var time1, time2, time3;
time1 = new Date;
result1 = {};
for (i in w) {
	if (w[i] !== null) {
		result1[i] = w[i];
	}
}
time2 = new Date;
result2 = Object.create(w);
for (i in result2) {
	if (result2[i] === null) {
		delete result2[i];
	}
}
time3 = new Date;
console.log(time2 - time1);
console.log(time3 - time2);
```

Even with half of elements being `null` it is still a bit faster.

### DIFF
--- a/src/standard/configure.html
+++ b/src/standard/configure.html
@@ -45,11 +45,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     // storage for configuration
     _setupConfigure: function(initialConfig) {
-      this._config = {};
+      // If we have more non-undefined properties, it is more efficient to clone whole object
+      // and then remove undesired properties
+      this._config = Object.create(initialConfig || {});
       // don't accept undefined values in intialConfig
       for (var i in initialConfig) {
-        if (initialConfig[i] !== undefined) {
-          this._config[i] = initialConfig[i];    
+        if (initialConfig[i] === undefined) {
+          delete this._config[i];
         }
       }
       this._handlers = [];


tjsavage:
This seems like a potential win... could you add a corresponding issue to help guide the conversation? Adding @kevinpschaaf to take a look

nazar-pc:
Issue: #2507

nazar-pc:
Fails with master code on one test: `unit/dom-bind.html:initial value notifies to dom-bind`

src/standard/configure.html
```diff
@@ -45,11 +45,13 @@
 
     // storage for configuration
     _setupConfigure: function(initialConfig) {
-      this._config = {};
+      // If we have more non-undefined properties, it is more efficient to clone whole object
+      // and then remove undesired properties
+      this._config = Object.create(initialConfig || {});
```
sorvell: I think all we need here is:

`this._config = initialConfig ? Object.create(initialConfig) : {};`

doesn't seem like the delete is necessary

sorvell:
Sorry for the delay on this. Made one suggestion above and it seems that this breaks at least one test. That'll need to be sorted out prior to merging. Thanks!

src/standard/configure.html
nazar-pc: Do you mean properties deletion is not needed? Because original comment below shows that filtering from `undefined` properties was intentional.

src/standard/configure.html
```diff
@@ -45,11 +45,13 @@
...
       // don't accept undefined values in intialConfig
       for (var i in initialConfig) {
-        if (initialConfig[i] !== undefined) {
-          this._config[i] = initialConfig[i];    
+        if (initialConfig[i] === undefined) {
+          delete this._config[i];
```
azakus: This won't actually work as you expect. `delete` only deletes properties off of the given object, and does not walk the prototype chain.

```js
var a = {a: 1};
var b = Object.create(a);
delete b.a;
b.a === 1;
```

Therefore, undefined properties will still exist in `this._config`. Instead, you'd have to do something like this:
```js
this._config = initalConfig;
for (var i in initialConfig) {
  if(initialConfig[i] === undefined) {
   delete this._config[i];
 }
}
```

Also, the use of `delete` slows down property accesses in v8 by deoptimizing the object. To accept this PR, I we need a benchmark showing that using `delete` is faster than cloning the object without `undefined` values.

PR #2286, opened 2015-08-17
Support for negative numbers in computed bindings

### LINKED ISSUE #2264: Negative numbers in computed bindings
```javascript
Polymer(
...
    properties : {
        negative_num : {type: Number, value: -1},
    },
    some_method : function (num) {
        return num;
    }
....

```
Not working, empty inside:
```html
<span>[[some_method(-1)]]</span>
```
Working, `-1` text inside:
```html
<span>[[some_method('-1')]]</span>
```
Working, `-1` text inside:
```html
<span>[[some_method(negative_num)]]</span>
```

nazar-pc:
Fixes #2264

### DIFF
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -221,6 +221,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           a.literal = true;
           break;
         case '#':
+        case '-':
           a.value = Number(arg);
           a.literal = true;
           break;
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -211,6 +211,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       };
       // detect literal value (must be String or Number)
       var fc = arg[0];
+      if (fc === '-') {
+        fc = arg[1];
+      }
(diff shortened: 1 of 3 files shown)


azakus:
Can you add a test with invalid data, such as `{{computeNegativeNumber(-A)}}` ?

nazar-pc:
Do you mean test as test or as additional check? I'm not sure how it should behave in such case.

src/standard/effectBuilder.html
```diff
@@ -221,6 +221,7 @@
           a.literal = true;
           break;
         case '#':
+        case '-':
```
azakus: I think we should check this far earlier, and then check that the next character is a number literal.

Something like

```js
var fc = arg[0];
if (fc === '-') {
  fc = arg[1];
}
if (fc >= '0' && fc <= '9') {
```

azakus:
@nazar-pc That is a great question. It could either be a `NaN` or use '-A' as a literal. @sorvell, do you have a preference?

nazar-pc:
I wanted to make patch as simple as possible.
Moving check earlier as you suggested naturally results in `undefined` for `{{computeNegativeNumber(-A)}}`, which in my opinion is reasonable.

azakus:
LGTM!

PR #2291, opened 2015-08-17
make isDebouncerActive actually return a bool

jakemac53:
Previously it would return a `truthy` value, but not a boolean as the type annotation specifies.

### DIFF
--- a/test/unit/utils.html
+++ b/test/unit/utils.html
@@ -66,9 +66,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         called++;
       };
 
+      assert(window.el1.isDebouncerActive('foo') === false);
       window.el1.debounce('foo', cb);
       window.el1.debounce('foo', cb);
       window.el1.debounce('foo', cb);
+      assert(window.el1.isDebouncerActive('foo') === true);
 
       setTimeout(function() {
         assert.equal(called, 1, 'debounce should be called exactly once');


gronke:
Hey @jakemac53, have you seen #2174 ? Looks similar.
https://github.com/gronke/polymer/blob/master/src/micro/debouncer.html#L52

jakemac53:
@gronke yep that also looks like it would solve the issue. I will leave this open for now, but feel free to close this if that one gets merged.

.gitignore
```diff
@@ -1,3 +1,4 @@
 bower_components
 node_modules
 dist
+.idea
```
gronke: what's this line good for?

gronke:
I'm in favor of merging this (smaller) PR first, so that the bug is fixed sooner than later.
Will rebase #2174 afterwards, so that it stays fast-forward mergeable.

Except the comment on the `.gitignore` file :+1:

jakemac53:
The gitignore change is for people who use an intellij ide (webstorm, etc).

gronke:
@jakemac53 Ah, I see.. Sounds like you want to have this line in your global `.gitignore`, so that it does not have to be duplicated in every project you work on.

jakemac53:
Updated and squashed

JeremybellEU:
The changes in this PR already landed in master as of #2174 so this PR can be closed.

kevinpschaaf:
@jakemac53 Let's merge this, would you mind rebasing the branch to master? Thanks!

jakemac53:
It looks like the change already got submitted, although you still may want to merge the tests I added so make sure it doesn't regress. I just rebased and pushed so feel free to merge (or not).

kevinpschaaf:
@jakemac53 Ah sorry, missed the comments above. Thanks, we'll add the tests.

kevinpschaaf:
Tested based on https://github.com/Polymer/polymer/pull/3374 to confirm test failure reported here is fixed.

LTGM

PR #2293, opened 2015-08-17
Fixes #2253 and fixes #2283

### LINKED ISSUE #2253: Invalid childNodes after domApi(node).appendChild(documentfragment)
https://jsbin.com/puwaduvaqa/edit?html,console,output

### LINKED ISSUE #2283: `getDistributedNodes()` returns nodes that are already removed under shady dom
Hi, I'm not sure if this is expected behavior. Say I have an element `<x-test>` that selects and inserts content from its parent,

`<x-test>`
```html
<dom-module id="x-test">
  <template>
    <div>x-test's distributed nodes:</div><br>
    <content id="content" select="[data-test]"></content>
  </template>
</dom-module>
```

and from my parent, I call it like that,

`<parent-element>`
```html
<dom-module id="parent-element">
  <template>
  ...
    <x-test id="app">
      <template is="dom-repeat" items="{{list}}">
        <div data-test>{{item.foo}}</div>
      </template>
    </x-test>
  </template>
</dom-module>
```

where `<parent-element>`'s `list` array looks like
```js
this.push("list", { foo: "item 1" });
this.push("list", { foo: "item 2" });
this.push("list", { foo: "item 3" });
```

everything works like a champ so far. Calling `Polymer.dom(content).getDistributedNodes()` returns the 3 items.

However, if I remove an item from the list,
```js
this.splice("list", 2, 1);
```

while the template renders just fine, `getDistributedNodes()` still returns the full 3 items - when I expect 2 items. This works correctly under native shadow dom behaviour and I thought it should be normalised. 

Here is a repro: http://jsbin.com/xebelufika/edit?html,output

Try adding some items, log the nodes, remove some, log the nodes, then add again. It seems that adding an item corrects this behavior.

Thanks so much,
Jason

sorvell:
Fixes #2253: refine logical tree check and populate parents of insert…ion points with logical info only if necessary.

Fixes #2283: when a node is removed, we need to potentially distribute not only its host but also its parent.

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -64,19 +64,18 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // container to container.host.
       // 3. node is <content> (host of container needs distribution)
       appendChild: function(node) {
-        var handled;
         this._removeNodeFromHost(node, true);
-        if (this._nodeIsInLogicalTree(this.node)) {
-          // if a <content> is added, make sure it's parent has logical info.
+        // if a <content> is added, make sure it's parent has logical info.
+        if (this.getOwnerRoot()) {
           this._ensureContentLogicalInfo(node);
-          this._addLogicalInfo(node, this.node);
-          this._addNodeToHost(node);
-          handled = this._maybeDistribute(node, this.node);
-        } else {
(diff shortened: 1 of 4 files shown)


frankiefu:
lgtm :+1:

PR #2294, opened 2015-08-17
fix for loop on targets

darkslategrey:
I try to use opalrb with polymer but there is an issue: opal/opal#1065
As was mentionned in the issue, I added the array index into the for loop on targets in the src/lib/style-cache.html file


darkslategrey:
I signed it!

darkslategrey:
I signed it!

vais:
@darkslategrey it looks like `_objectsEqual` is supposed to be usable with both objects and arrays. The change you made works for arrays, but not for objects?

PR #2295, opened 2015-08-17
Fix for method parsing in computed binding

nazar-pc:
Quantifier should be greedy, otherwise it fails to parse correctly such construction:
```html
<span attr="[[computed('(', foo)]]">
```


tjsavage:
Could you create a corresponding issue for this? It seems clear in the PR, but an issue will help the conversation.

nazar-pc:
Issue: #2506

samccone:
@tjsavage any word on this one?

azakus:
LGTM

PR #2306, opened 2015-08-18
Fixes #2304: avoid trying to read style data from imports that did no…

### LINKED ISSUE #2304: Attempting to load a <link rel="import"> inside a dom-module template causes exception
The styling system attempts to read style data from this element's .import property. Instead we should ignore link elements that have no import.

sorvell:
…t load.


frankiefu:
lgtm :+1:

PR #2314, opened 2015-08-19
Use numeric sort when removing dom-repeat instances

### LINKED ISSUE #2297: Sort on dom-repeat template causes exception on splice
When I use a sort function on a `dom-repeat` template and call `splice` to remove more than 10 items from the list at once, the following exception is thrown and the splice is only partially applied:

```
Uncaught TypeError: Cannot read property 'isPlaceholder' of undefined
    Polymer._detachRow	@	polymer.html:3906
    Polymer._applySplicesUserSort	@	polymer.html:3833
    Polymer._render	@	polymer.html:3741
    Debouncer.complete	@	polymer.html:1118
    Polymer.Async._atEndOfMicrotask	@	polymer.html:1079
```

For some reason, removing 10 items or less is fine. Here is a minimal example for reproducing the problem:

```html
<dom-module id="list-test">
    <template>

        <template is="dom-repeat" items="{{ items }}" sort="sort">

            <div>{{ item }}</div>

        </template>

        <button on-tap="removeItems">Remote first 11 items</button>

    </template>

    <script>

        Polymer({
            is: "list-test",
            properties: {
                items: {
                    type: Array,
                    value: [0,1,2,3,4,5,6,7,8,9,10,11]
                }
            },
            sort: function(a, b) {
                return a - b;
            },
            removeItems: function() {
                // This causes an exception
                // Removing less than 11 items is fine for some reason
                this.splice("items", 0, 11);
            }
        });

    </script>
</dom-module>
```

azakus:
Fixes #2297

### DIFF
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -402,6 +402,10 @@ Then the `observe` property should be configured as follows:
       return this.collection.getKey(a) - this.collection.getKey(b);
     },
 
+    _numericSort: function(a, b) {
+      return a - b;
+    },
+
     // Render method 2: incremental update using splices with user sort applied
     // ----
     // Removed/added keys are deduped, all removed rows are detached and pooled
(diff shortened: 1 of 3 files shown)


sorvell:
LGTM

Issue #2310, opened 2015-08-19
IE11 error "unable to get property '_lightChildren' when component is just a <content> tag


twlamb:
If you create a component such as:

```html
<link rel="import" href="../bower_components/polymer/polymer.html" />

<dom-module id="test-component">
    <template>
        <content></content>
    </template>
</dom-module>
<script>
    Polymer({
        is: "test-component"
    });
</script>
```

IE11 will throw an error when the page is rendered.  It seems the page renders OK, but if you use "unresolved" on the body of your page, it will never resolve.

If you add an html tag around `<content>` such as a `<div>`
```html
<link rel="import" href="../bower_components/polymer/polymer.html" />

<dom-module id="test-component">
    <template>
        <div>
            <content></content>
        </div>
    </template>
</dom-module>
<script>
    Polymer({
        is: "test-component"
    });
</script>
```
Then it works just fine.  If you use a paper-material in your page and "unresolved" it will never be removed in IE11, since paper-material's template is just `<content></content>`.

sorvell:
@rictic Since we've discovered this is a` webcomponentsjs` issue, can you close this in favor of that one when it's filed? Thanks.

PR #2321, opened 2015-08-19
Fix a rare, racey IE11 bug

rictic:
In very rare cases that I can only sometimes reproduce, and only sometimes on IE11 running within certain VMs, `c` (which is a <content> node) has no parent at this point in the code.

Charitably, perhaps this setup is exercising certain unusual event timings, causing a legal but unusual order of operations.

Bisect indicates that this "bug" was introduced at https://github.com/Polymer/polymer/commit/21500fb53f4dc186ffd1099d5492c711e097c8a4

### DIFF
--- a/src/mini/shady.html
+++ b/src/mini/shady.html
@@ -64,7 +64,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         for (var i=0, c; i < i$.length; i++) {
           c = i$[i];
           saveLightChildrenIfNeeded(c);
-          saveLightChildrenIfNeeded(c.parentNode);
+          // Also check for _lightParent because of an IE 11 bug, where
+          // a child of a document fragment doesn't always know about its
+          // parent, even when the parent knows about the child.
+          // More info: https://github.com/Polymer/polymer/pull/2321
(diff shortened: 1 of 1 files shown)


rictic:
IE 11.0.9600.17126 --- has this bug
IE 11.0.9600.17959 -- does not have this bug

Can reproduce this by trying to do `document.createElement('paper-material')` or any Polymer element whose direct template child is a `<content>` node. IE gives us an element without a parent, even though the element is in its parent's list of children.

Fixes #2310

rictic:
Updated the commit to use `_lightParent` when `parent` can't be found. Also added a comment documenting why we're checking both.

sorvell:
@rictic Let's check one more thing before merging this... I found the code that's supposed to work around this IE bug: https://github.com/webcomponents/webcomponentsjs/blob/master/src/CustomElements/register.js#L353.

Let's try to determine if this code path is running or not when the problem occurs. If it is then it suggests our fix there is incomplete.

rictic:
@sorvell Nice! Yeah, the better fix for this is in webcomponentsjs. Opened PR there: https://github.com/webcomponents/webcomponentsjs/pull/385

PR #2332, opened 2015-08-20
Fixes #2263: ensure custom-style can parse variable definitions in su…

### LINKED ISSUE #2263: custom-style containing '/deep/' or '::shadow' produces an SyntaxError in Firefox
The problem occurs when the rule is passed to the Element.prototype.matches function. 
Firefox throws the following: 
```
SyntaxError: An invalid or illegal string was specified
```

The location of the error is in Polymer.StyleProperties.propertyDataFromStyles, when the matchesSelector is called on the rule. The rule with which I reproduce it is 'body /deep/ *', but any rule which contains /deep/ and ::shadow will trigger the error.

Before the custom-style element, the html contains both the script for webcomponents-lite.min.js, as well as the import for polymer.html.

sorvell:
…pported selectors (e.g. /deep/) without exception due to unknown css.

### DIFF
--- a/src/lib/custom-style.html
+++ b/src/lib/custom-style.html
@@ -135,11 +135,13 @@ Note, all features of `custom-style` are available when defining styles as part 
       if (this.include) {
         e.textContent += styleUtil.cssFromModules(this.include);
       }
+      var rules = styleUtil.rulesForStyle(e);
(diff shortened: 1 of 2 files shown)


frankiefu:
LGTM :+1:

PR #2338, opened 2015-08-21
Don't call shady dom method when using shadow dom

rictic:


rictic:
Yeah, #2341 is a more solid, principled approach to fixing this bug. Closing this PR.

PR #2340, opened 2015-08-22
Fixes #2334: when composing nodes in shady dom, check if a node is wh…

### LINKED ISSUE #2334: Invalid fragment childNodes
https://jsbin.com/yejizipile/edit?html,console,output
different path from #2311

for what it's worth, my temp fix is to always check in _updateChildNodes/shady.html if the real parentNode is the container.  See code below.

```javascript
_updateChildNodes: function(container, children) {
        var composed = getComposedChildren(container);
        var splices =
          Polymer.ArraySplice.calculateSplices(children, composed);
        // process removals
        for (var i=0, d=0, s; (i<splices.length) && (s=splices[i]); i++) {
          for (var j=0, n; (j < s.removed.length) && (n=s.removed[j]); j++) {
            if (container === n.parentNode) {
                remove(n);
            }
            composed.splice(s.index + d, 1);
          }
          d -= s.addedCount;
        }
        // process adds
        for (var i=0, s, next; (i<splices.length) && (s=splices[i]); i++) {
          next = composed[s.index];
          for (var j=s.index, n; j < s.index + s.addedCount; j++) {
            n = children[j];
            insertBefore(container, n, next);
            composed.splice(j, 0, n);
          }
        }
      },
```

sorvell:
…ere we expect it to be before removing it from its distributed position. We do this because the node may have been moved by Polymer.dom in a way that triggered distribution of its previous location. The node is already where it needs to be so removing it from its parent when it's no longer distributed is destructive.

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -238,8 +238,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _tryRemoveUndistributedNode: function(node) {
         if (this.node.shadyRoot) {
-          if (node._composedParent) {
-            nativeRemoveChild.call(node._composedParent, node);
+          var parent = getComposedParent(node);
+          if (parent) {
+            nativeRemoveChild.call(parent, node);
           }
           return true;
         }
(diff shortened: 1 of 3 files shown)


frankiefu:
LGTM :+1:

PR #2342, opened 2015-08-22
Fixes #2341: branch Polymer.dom to use native dom methods under Shado…

### LINKED ISSUE #2341: Polymer.dom can call some shady dom code when running shadow dom
For example, the dom mutation methods call shady dom code paths. Instead they should be routed to the native dom methods: `appendChild`, `insertBefore`, `removeChild`, `replaceChild`, `cloneNode`.

sorvell:
…w DOM for: appendChild, insertBefore, removeChild, replaceChild, cloneNode.


rictic:
LGTM!

Issue #2344, opened 2015-08-22
Binding to reflected property


nazar-pc:
```haml
%dom-module(id="cs-icon")
	%template
		%template(is="dom-if" if="[[multiple_icons]]")
			X
		%template(is="dom-if" if="[[!multiple_icons]]")
			Y
	%script(src="script.js")
```
```coffeescript
Polymer(
	'is'				: 'cs-icon'
	properties			:
		icon			:
			reflectToAttribute	: true
			type				: String
		multiple_icons	:
			computed	: '_multiple_icons(icon)'
			type		: Boolean
	_multiple_icons : (icon) ->
		icon.split(' ').length > 1
)
```
```html
<cs-icon icon="[[icon]]"></cs-icon>
```
Two issues here:
* eventually neither `X` nor `Y` is shown, even though computed `multiple_icons` is boolean
* while `icon` property is reflected, attribute is removed during binding to property

No errors in browser console.
When bind as to attribute, everything works as expected.

sorvell:
Closed in favor of https://github.com/Polymer/polymer/issues/2348.

PR #2349, opened 2015-08-22
Fix for 2348 and 2344

### LINKED ISSUE #2348: Getters and setters made inaccessible
```javascript
Polymer({
	'is'		: 'cs-nav-tabs',
	'extends'	: 'nav',
	properties	: {
		active	: {
			observer	: 'active_changed',
			type		: Number
		}
	},
	ready : function () {
		if (!this.querySelector('button[active]')) {
			this.active = 0;
		}
	},
	active_changed : function () {
		var buttons = this.querySelectorAll('button');
		var button;
		for (var index = 0; index < buttons.length; ++index) {
			button = buttons[index];
			button.active = index == this.active;
		}
	}
});
```
```javascript
Polymer({
	'is'		: 'cs-button',
	'extends'	: 'button',
	properties	: {
		active	: {
			reflectToAttribute	: true,
			type				: Boolean
		}
	}
});
```
```html
<nav is="cs-nav-tabs">
	<button is="cs-button" type="button">One</button>
	<button is="cs-button" type="button">Two</button>
</nav>
```

At the moment when `button.active` is assigned (at `createdCallback` of `[is=cs-nav-tabs]`, you know), there are no getters/setters yet in `button` element, which eventually makes accessors below this newly declared property in prototypes chain and basically inaccessible.

**Firefox Nightly, Shadow DOM polyfill**

nazar-pc:
Fix for getters/setters for property become inaccessible when propery set on element before it is ready
Did not succeed in creating test case this time:(
Fixes #2348

### DIFF
--- a/src/standard/configure.html
+++ b/src/standard/configure.html
@@ -114,8 +114,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _configureProperties: function(properties, config) {
       for (var i in properties) {
         var c = properties[i];
-        // don't accept undefined values
-        if (c.value !== undefined) {
+        if (this[i] !== undefined) {
+          config[i] = this[i];
+          delete this[i];
+        } else if (c.value !== undefined) {
           var value = c.value;
           if (typeof value == 'function') {
             // pass existing config values (this._config) to value function
--- a/test/unit/configure-elements.html
+++ b/test/unit/configure-elements.html
@@ -165,3 +165,27 @@
     });
   </script>
 </dom-module>
+
+<script>
+  Polymer({
+    is: 'x-configure-internal',
+    properties: {
+      shouldChange : {
+        observer: 'shouldChangeCallback',
+        type: String
+      }
+    },
+    shouldChangeCallback: function() {
+      this.innerHTML = this.shouldChange;
+    }
+  });
+</script>
+<script>
+  Polymer({
+    is: 'x-configure-external',
+    ready: function() {
+      var e = this.querySelector('x-configure-internal');
+      e.shouldChange = 'It works';
+    }
+  });
+</script>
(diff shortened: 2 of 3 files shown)


nazar-pc:
Indeed it fixes #2344 as well...

tjsavage:
added @sjmiles to take a look on this as he was active on the [corresponding issue](https://github.com/Polymer/polymer/issues/2348)

src/standard/configure.html
```diff
@@ -121,6 +121,9 @@
             value = value.call(this, this._config);
           }
           config[i] = value;
+        } else if (this.hasOwnProperty(i)) {
```
sorvell: `this[i] !== undefined` should be faster than `hasOwnProperty`.

sorvell:
Sorry for the time lag on this. The main concern with this change is incurring a performance penalty on elements that do not have early-set properties. Would you be willing to do some performance testing on this? Thank.

nazar-pc:
Modified as you suggested.
About performance: created 10 elements with 300 String properties there, 10 instances of each element on page (30k properties in total, all passing through newly added code).

Without patch `readystatechange` took 45.56ms in Chromium, with patch 79.45ms. This is 1.74 times slower, but since this step is executed only once for each element and takes so small amount of time even for 300 properties this is an affordable change.

Code I've used for testing: http://pastebin.com/ZdMnrZqN

src/standard/configure.html
```diff
@@ -121,6 +121,9 @@
             value = value.call(this, this._config);
           }
           config[i] = value;
+        } else if (this[i] !== undefined) {
```
sorvell: Seems like the check `if` and the `else if` should be switched here. If there's an existing value, we should take that and not the default value.

sorvell:
We plan to do some perf testing to see what overhead this adds to elements without early-property sets.

nazar-pc:
Modified as suggested

sorvell:
The performance of this change looks ok to me. Can you please add a test? Thanks.

nazar-pc:
@sorvell, I've managed to write test for this change, also rebased against master

PR #2359, opened 2015-08-24
Remove forceUpgraded check in dom-module.import

azakus:
Multiple elements depend on dom-module.import and may register before dom-module
is registered.

With the forceUpgraded check, only the first module to call .import will be
upgraded, typically custom-style.

This will break the subsiquent elements, such as dom-template.

Fixes Polymer/vulcanize#234


sorvell:
@azakus Looks ok, but can you please add a test?

sorvell:
LGTM

Issue #2329, opened 2015-08-20
ES6: Better support for ES6 class syntax


ebidel:
@sjmiles and I talked about this syntax. Can we land this asap before the Summit?

```
class MyElement {
    registered() { // TODO: Polymer core make this work
      this.is = 'my-element';  
      this.properties = {
        foo: {
          type: String,
          value: 'foo prop made it!!'
        }
      }; 
    }

    // everything else is business as usual
    created() {}
    ready() {}
    attached() {}
    detached() {}
    attributeChanged() {}
  }

  Polymer(MyElement); // TODO: Polymer core need to support taking `MyElement` instead of `MyElement.prototype`.
```

```
    // @sjmiles: We will put this in Polymer core
    Polymer.Base.registerCallback = function() {
      this._doBehavior('registered'); // abstract
      this._registerFeatures();  // abstract
    };
```

laco0416:
:+1:

redian:
+1

ikeagold:
:+1:

JableSoHo:
When I declare properties in my ES6 class as shown above it works fine, but when I add an observer to a property the observer function does not fire.  I have a jsbin that illustrates this:

http://jsbin.com/senulobuho/edit?html,js,output

laco0416:
@JableSoHo i think `this.properties` is a mistake. `properties` should be *static* like as `get is()`. 

http://jsbin.com/yajehiweha/edit?html,js,console,output

JableSoHo:
@laco0416 Thanks a lot, that fixed the issue!!

sjmiles:
@laco0416, @JableSoHo: beware of making a `properties` getter that returns an inline object. Polymer uses `this.properties` more than once, and the getter as written will return a new object every time.

sjmiles:
These requests should be fulfilled now via https://github.com/Polymer/polymer/pull/2365

ebidel:
Confirmed the changes :100:

JableSoHo:
@sjmiles Would it be best then to declare this.properties in the registered function and in the getter just return that?

ebidel:
@JableSoHo.  See my syntax in https://github.com/Polymer/polymer/issues/2329#issue-102177066

JableSoHo:
@ebidel I used your syntax which works.  Except now neon-animations, and dialog buttons throw an exception. This is the exception : 

[...]

PR #2365, opened 2015-08-25
Es6 friendliness

### LINKED ISSUE #2363: ES6: vulcanize + babel produces errors b/c of strict mode issue(s)
Atm, building an element in ES6 classes requires a transpiler like babel to work x-browser. That workflow looks something like this:

1. vulcanize | cripser (to get separate output.js file)
3. babel output.js

The full command is below. BabelJS adds `"use strict";` to the top of the transpiled output:

screen shot 2015-08-24 at 4 26 47 pm

Unfortunately, this means JS errors, as parts of core 'use strict' and other parts do not.

**Fix**
Easiest fix is to get ride of [this global](https://github.com/Polymer/polymer/blob/master/src/lib/settings.html#L12), make it `window.Polymer = {...`. I've confirmed changing this one line in the transpiled output works!

` vulcanize polymer.html --inline-script | crisper -h polymer.v.html -j polymer.v.b.js; babel --compact false polymer.v.b.js -o polymer.v.b.js
`

sjmiles:
Aimed to fix #2363, #2329, #1903.

### DIFF
--- a/src/lib/polymer-bootstrap.html
+++ b/src/lib/polymer-bootstrap.html
@@ -34,15 +34,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // Polymer.Base is now chained to factory.prototype, and for IE10 compat
       // this may have resulted in a new prototype being created
       prototype = factory.prototype;
-      // native Custom Elements treats 'undefined' extends property
-      // as valued, the property must not exist to be ignored
-      // TODO(sjmiles): Custom Elements no longer has the above behavior
       var options = {
-        prototype: prototype
+        prototype: prototype,
+        extends: prototype.extends
       };
-      if (prototype.extends) {
-        options.extends = prototype.extends;
-      }
       Polymer.telemetry._registrate(prototype);
       document.registerElement(prototype.is, options);
       return factory;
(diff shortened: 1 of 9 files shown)


src/lib/polymer-bootstrap.html
```diff
@@ -20,12 +20,23 @@
...
+      }
+      // desugar the prototype and return a factory object
+      var factory = desugar(prototype);
+      // Polymer.Base is now chained to factory.prototype, and for IE10 compat
       // this may have resulted in a new prototype being created
-      prototype = ctor.prototype;
+      prototype = factory.prototype;
       // native Custom Elements treats 'undefined' extends property
       // as valued, the property must not exist to be ignored
+      // TODO(sjmiles): Custom Elements no longer has the above behavior
```
sorvell: Will investigate if we can clean this up.

src/lib/polymer-bootstrap.html
```diff
@@ -20,12 +20,23 @@
     var userPolymer = window.Polymer;
 
     window.Polymer = function(prototype) {
-      var ctor = desugar(prototype);
-      // Polymer.Base is now chained to ctor.prototype, and for IE10 compat
+      // if input is a `class` (aka a function with a prototype), use the prototype
+      // remember that the `constructor` will never be called
+      if (typeof prototype === 'function') {
```
sorvell: Wondering about ES6 syntax and people wanting to use `extends` with, something. Think that is worth supporting? If so, seems like it'd go somewhere right in here.

sorvell:
LGTM

PR #2368, opened 2015-08-25
Set of various micro optimizations that may be useful

nazar-pc:


tjsavage:
Thanks for the PR - apologies for the delay on feedback. To make it a little easier to discuss the impetus and reasoning behind the change, would you be able to create an issue that goes over the reasoning and then link to it from this PR? For this one in particular, any small perf benchmarks you could create to show the performance improvement gained here would be really helpful.

nazar-pc:
Here is issue: #2502
Also I've added some more explanations for these changes

nazar-pc:
Fails with master code on 3 tests in `unit/dom-bind.html`

azakus:
Sorry for the delay in reviewing this.
We've done some micro optimizations since this PR was made, and it may no longer be relevant.
It also doesn't merge against master, so please rebase if this is still an issue.

PR #2375, opened 2015-08-26
Fixes #2356, #2357

### LINKED ISSUE #2356: custom-style should generate a better error when its `include` is not found
At the moment, this generates an arcane error. Instead, it should warn that the include name cannot be found.

### LINKED ISSUE #2357: Imported styles take precedence over local styles when declared in same tag
So, if I do:

    <style include="shared-styles">
      .foo { color: red; }
    </style>

I would expect the local `.foo` rule to override any imported one. But it _appears_ that the imported rules are added after any rules defined in the body of the tag, so they take precedence:

http://jsbin.com/jivupu/edit?html,output

I'm not sure whether a) this is the expected behavior, and b) whether we even want to document using a style tag with both `include` and local rules.

Please advise, @sorvell. Thanks.

sorvell:
Fixes #2356: issue a warning and don't throw an exception when a style include cannot be found.

Fixes #2357: include data now comes before any textContent in a style element.

### DIFF
--- a/src/lib/custom-style.html
+++ b/src/lib/custom-style.html
@@ -133,31 +133,37 @@ Note, all features of `custom-style` are available when defining styles as part 
       // used applied element from HTMLImports polyfill or this
       var e = this.__appliedElement || this;
       if (this.include) {
-        e.textContent += styleUtil.cssFromModules(this.include);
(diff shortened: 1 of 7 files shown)


frankiefu:
LGTM :+1:

PR #2379, opened 2015-08-26
move the mixing-in of behaviors so that it happens before `register`is invoked

sjmiles:
We recently changed the flow so that `doBehavior('register')` is invoked before features are registered, which had the side-effect of making the behavior methods unavailable on the prototype (because the mixing in step hadn't happened yet).

This patch promotes mixing in so that behavior methods are installed on the prototype before `doBehavior('register')`.


sorvell:
Change was introduced to help address: https://github.com/Polymer/polymer/issues/2363

sorvell:
LGTM.

PR #2395, opened 2015-09-02
Undefined configuration fixes. Fixes #2360

### LINKED ISSUE #2360: `undefined` values are not consistently prevented from entering the configuration process
Generally setting a value to `undefined` is an anti-pattern, since we use `undefined` to indicate the state of not being set.

In fact, if you try set a value explicitly to `undefined` as a value in `properties`, the configuration object summarily ignores it. 

However, if you bind to an object path that evaluates to `undefined` or that cannot be evaluated, the configuration code does not ignore this value, leading to an inconsistency (and scenarios where it's impossible to control an element value with defaulting, see @sjmiles for specfics).

Ignoring `undefined` as a configuration value when it flows from bindings is a reasonably easy change which restores consistent behavior (i.e. see https://github.com/Polymer/polymer/blob/master/src/standard/configure.html#L70).

Note that make the behavior consistent can result in two change sequences in certain scenarios which is sub-optimal, again see @sjmiles for specfics). It seems that the 'treat undefined structured values as null` gambit coincidentally resolves this problem optimally, but we plan on evaluating that change separately.

sorvell:
Never allows an undefined value to be configured.

### DIFF
--- a/test/unit/configure.html
+++ b/test/unit/configure.html
@@ -83,7 +83,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     test('binding to undefined path not propagated during configuration', function() {
       var e = document.querySelector('x-configure-host').$.child;
-      assert.isTrue(e._definedChangedCalled);
+      assert.equal(e.defined, undefined);
+      assert.equal(e._definedChangedCalled, 1);
     });
 
   });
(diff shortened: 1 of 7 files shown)


test/unit/configure.html
```diff
@@ -81,6 +81,11 @@
       assert.equal(e.readOnly, 'default');
     });
 
+    test('binding to undefined path not propagated during configuration', function() {
+      var e = document.querySelector('x-configure-host').$.child;
+      assert.isTrue(e._definedChangedCalled);
```
kevinpschaaf: also add
```
assert.isTrue(e.defined);
```
for completeness?

sorvell:
This turned out to be more troublesome than expected. There are plans to address this a different way so closing this.

Issue #1727, opened 2015-06-02
[1.0] Do not require prefixing ::content if native support of polyfill available


nazar-pc:
`.content-wrapper > ::content .special` is really ugly, would be nice do not require prefix if native Shadow DOM support available or full WebComponents.js polyfill included.
(from https://github.com/Polymer/polymer/issues/1698#issuecomment-108030075)

nazar-pc:
Actually, when using Shadow DOM this recommendation can be just ignored, so I'm fine here.
But ideally this should work the same under Shady DOM too (at least if it is planned to be supported in long-run), just to be closer to native standard behavior

nazar-pc:
I've approached this issue second time and found very simple way to fix this inconvenience.
@sorvell, take a look at #2419, please.

nazar-pc:
@sorvell, this was fixed by #2419 on 30 September, can you take a look at it?
PR is really trivial.

nazar-pc:
Fixed by merged #2419, thanks

PR #2419, opened 2015-09-05
Eliminate the need to write `:host ::content` instead of just `::content`

nazar-pc:
While keeping the same processing under the hood.
I've also removed workaround from tests and they don't fail, which is good.

### DIFF
--- a/src/lib/style-transformer.html
+++ b/src/lib/style-transformer.html
@@ -30,9 +30,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
      * :host(...) -> scopeName...
 
-     * ::content -> ' ' NOTE: requires use of scoping selector and selectors
-       cannot otherwise be scoped:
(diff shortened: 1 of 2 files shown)


tjsavage:
Thanks for the PR - apologies for the delay on feedback. To make it a little easier to discuss the impetus and reasoning behind the change, would you be able to create an issue that goes over the reasoning and then link to it from this PR?

nazar-pc:
It was already created and linked back: https://github.com/Polymer/polymer/issues/1727

nazar-pc:
Just rebased this against master to resolve conflicts and did some small improvements.
Merge this small fix soon, please. It is almost 2 months old already.

nazar-pc:
Rebased against master once again. Really simple fix, sad to see it is not merged months later :-1:

sorvell:
LGTM. Sorry for the delay and thanks for the effort!

PR #2428, opened 2015-09-10
Add note on arrayDelete with array vs. path

### LINKED ISSUE #2427: PolymerBase.arrayDelete does not notify when Array argument is passed
It looks like arrayDelete uses path.splice instead of this.splice if an array argument is passed rather than a string argument.

https://github.com/Polymer/polymer/blob/master/src/standard/utils.html#L214

This means that the array is modified, but no observers are called.

arthurevans:
Fixes #2427. Well, actually it just documents the limitation, it doesn't fix anything :).

@sorvell or @kevinpschaaf should review -- if this behavior is unintended, please correct me, but I don't think there's any way to reconstruct the path from the array.

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -199,9 +199,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     /**
      * Removes an item from an array, if it exists.
+     * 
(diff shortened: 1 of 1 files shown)


kevinpschaaf:
:100:

PR #2432, opened 2015-09-11
pass event registered node to handler

WilliamSWoodruff:
This let's you write event handlers with a reliable reference to the node the event was registered to.

```javascript
this.listen(polymerElement, 'tap', 'tapHandler');

tapHandler: function(e, detail, node) {
  console.log(node) // polymerElement
}
```
Has been tested with adding event handlers via 
```javascript
on-eventName="eventHandler"
this.listen(element, eventType, eventHandler),
listeners : {'myId.tap': 'eventHandler'}
```

### DIFF
--- a/src/standard/events.html
+++ b/src/standard/events.html
@@ -119,7 +119,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var host = this;
       var handler = function(e) {
         if (host[methodName]) {
-          host[methodName](e, e.detail);
+          host[methodName](e, e.detail, node);
         } else {
           host._warn(host._logf('_createEventHandler', 'listener method `' +
             methodName + '` not defined'));


devinivy:
This is a common issue in the Polymer world, where any event registered using `this.listen()`, for example, is handled by a function with the listening Polymer component as the bound context.  There's no way to access the original element to which the event was registered.

Would love to see this go through!

azakus:
This isn't actually necessary. In `tapHandler`, `e.currentTarget` will be `polymerElement` as the event handler was registered on that node: [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget)

[JSBin example](http://jsbin.com/mebiha/edit?html,console,output)

devinivy:
Thx for the follow-up @azakus– not sure how I missed that!

PR #2471, opened 2015-09-24
Reset handlers queue after finished replaying events

### LINKED ISSUE #2452: Memory garbage collection
While developing today i noticed that my Polymer powered page had an enormous hunger for memory after manually refreshing the tab several times.
I noticed the same behavior on https://www.polymer-project.org/summit (and all Poylmer sites i tested for that matter)
Both in vanilla Firefox and Chrome, after a few reloads, memory usage already exceeds 200Mb with open end, depending of the reload count. I pushed it over 500Mb and after 15 minutes idle tab, it didn't show any sign of reduction.
Same test with GitHub frontpage: While memory usage increases after a few moments the GC kicked in and reduced a lot.

Is there a special reason why this happens?

azakus:
Holding on to an array of CustomEvents can leak memory across page refreshes in
Chrome: https://crbug.com/529941

Fixes #2452

### DIFF
--- a/src/standard/configure.html
+++ b/src/standard/configure.html
@@ -49,7 +49,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // don't accept undefined values in intialConfig
       for (var i in initialConfig) {
         if (initialConfig[i] !== undefined) {
-          this._config[i] = initialConfig[i];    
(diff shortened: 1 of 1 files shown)


kevinpschaaf:
LGTM

PR #2484, opened 2015-09-26
Add .gitattributes to solve line endings cross-OS (merge after other PRs)

### LINKED ISSUE #2592: Cloning and editing Polymer on Linux machines causes diff clutter
When developing Polymer on a Linux machine, diff clutter is caused due to line-ending changes. In order to globally configure line-endings and prevent this clutter from happening, a .gitattributes file should be added.

Steps to reproduce (on a Linux machine):
1. Clone/Fork Polymer
2. Open the project in your prefered editor
3. Open a file with `\r\n` line-endings (currently on master: `src/lib/css-parse.html`)
4. Insert a character somewhere in the file
5. Save the file
6. Delete the just inserted character (e.g. no textual changes should remain)
7. Save the file
8. Run `git diff` and observe the output below:

```javascript
diff --git a/src/lib/css-parse.html b/src/lib/css-parse.html
index a176d7c..30c9bee 100644
--- a/src/lib/css-parse.html
+++ b/src/lib/css-parse.html
@@ -43,7 +43,7 @@ Polymer.CssParse = (function() {
             n = {start: i+1, parent: p, previous: previous};
             p.rules.push(n);
             break;
-          case this.CLOSE_BRACE: 
+          case this.CLOSE_BRACE:
             //console.groupEnd(n.start);
             n.end = i+1;
             n = n.parent || root;
@@ -84,9 +84,9 @@ Polymer.CssParse = (function() {
       if (r$) {
         for (var i=0, l=r$.length, r; (i<l) && (r=r$[i]); i++) {
           this._parseCss(r, text);
-        }  
+        }
       }
-      return node;  
+      return node;
     },
 
     // stringify parsed css.
@@ -99,10 +99,10 @@ Polymer.CssParse = (function() {
         if (r$ && (preserveProperties || !this._hasMixinRules(r$))) {
           for (var i=0, l=r$.length, r; (i<l) && (r=r$[i]); i++) {
             cssText = this.stringify(r, preserveProperties, cssText);
-          }  
+          }
         } else {
-          cssText = preserveProperties ? node.cssText : 
-            this.removeCustomProps(node.cssText);  
+          cssText = preserveProperties ? node.cssText :
+            this.removeCustomProps(node.cssText);
           cssText = cssText.trim();
           if (cssText) {
             cssText = '  ' + cssText + '\n';
@@ -138,7 +138,7 @@ Polymer.CssParse = (function() {
     },
 
     removeCustomPropApply: function(cssText) {
-      return cssText            
+      return cssText
         .replace(this._rx.mixinApply, '')
         .replace(this._rx.varApply, '');
     },
@@ -171,7 +171,7 @@ Polymer.CssParse = (function() {
   };
 
 
-  // exports 
+  // exports
   return api;
 
 })();
```
The actual changes are `^M` characters indicating the line-ending changed.
image

Currently if you want to get rid of these changes:
1. Add a `.gitattributes` in the top-level folder
2. In the file insert `* binary`
3. `git add .`
4. `git reset --hard`

However this makes you lose ALL changes you made.
Therefore to prevent diff clutter and being able to contribute on Linux machines, the `.gitattributes` file should be added with `* text=auto`

JeremybellEU:
When checking out Polymer on a Linux machine, several files will be completely added due to non-CRLF endings. Therefore if you edit just 1 line, the whole file will be triggered as changed.

This change prevents the git diff from being cluttered, uniforming all line-endings across the project.
Fixes #2592

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto


samccone:
I am :+1: to this thanks @JeremybellEU :)

JeremybellEU:
Thanks for your support @samccone . For future pull-requests in the affected files (total of 58 files), will generate full on diffs. If you want to prevent this, all files have to be fixed before that. However, in order to prevent merge issues, this should be done when merging. Because every single change in one of those 58 files (which is very very likely) will cause merge conflicts which are very hard to solve. Maybe the full diffs would be more suitable, to also keep `git blame` sane.

samccone:
yeah @JeremybellEU I think it would be wise to go ahead and correct the line endings across the project, but I am going to defer to the maintainers

JeremybellEU:
:+1:

JeremybellEU:
Finally found time to create the issue: https://github.com/Polymer/polymer/issues/2592

ebidel:
@JeremybellEU Thanks for this. We need it :) I think we'll need to wait to merge it until the other PRs are addressed. We're going to go through those on a weekly basis now. cc @tjsavage

azakus:
LGTM, but I'll hold off merging until after the next release.

azakus:
I'm reverting this patch for now. Too awkward to handle the file changes piecemeal. If we decide to merge this again, it will be with every file normalized at once.

JeremybellEU:
Yes merging this causes quite a headache if there are a lot of open PRs. I would suggest to only merge this if there is some sort of code-freeze/very few PRs. Then the merges are managable, right now with 40 open PRs it is probably a nightmare.

For now I would advise all developers to set `git config --global core.autocrlf input` as described in https://help.github.com/articles/dealing-with-line-endings/#global-settings-for-line-endings . Therefore whenever the developer commits, just these files will have the correct line endings. This creates incremental fixing of the line endings and makes merging a lot easier.

Off Topic: This whole line endings is just a pain if not done pre-emptively and git can't really deal with them nicely.

PR #2485, opened 2015-09-26
Add expose method to be able to delegate Polymer object functions.

JeremybellEU:
When passing callbacks to external APIs, it is common to fetch
the `this` reference from the callback. This `this` references the
element that invoked the callback.

However, in order to be able to use Polymer functions, the `this`
that originally referenced the Polymer object, is lost.

Therefore in order to maintain both references, expose links `this`
to the Polymer object and the first argument references the element
that called the callback.

I personally came accross this problem while working on a `fullCalendar` Polymer element. The conflicting callback was `drop`. [In their docs](http://fullcalendar.io/docs/dropping/drop/) it is stated:
`this holds the DOM element that has been dropped.`

I do need the element, because it contains some data/properties I need process. Therefore I would lose my this reference to my Polymer element, making me unable to use any of all the functions defined by Polymer (or my custom functions/properties).

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -150,7 +150,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      *   payload.
      * @param {Object=} options Object specifying options.  These may include:
      *  `bubbles` (boolean, defaults to `true`),
-     *  `cancelable` (boolean, defaults to false), and 
+     *  `cancelable` (boolean, defaults to false), and
      *  `node` on which to fire the event (HTMLElement, defaults to `this`).
      * @return {CustomEvent} The new event that was fired.
      */
@@ -199,11 +199,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     /**
      * Removes an item from an array, if it exists.
-     * 
-     * If the array is specified by path, a change notification is 
+     *
+     * If the array is specified by path, a change notification is
      * generated, so that observers, data bindings and computed
-     * properties watching that path can update. 
-     * 
+     * properties watching that path can update.
+     *
      * If the array is passed directly, **no change
      * notification is generated**.
      *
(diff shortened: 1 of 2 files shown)


samccone:
correct me if I am wrong @JeremybellEU but would using `.bind` also solve this problem?

JeremybellEU:
@samccone That was my initial thought too, but that means that you lose the `this` of the element being dropped. Meaning you are unable to access any of the data/properties of the element that called the callback. An example usage can be found here:
[The callback function in the configuration object](https://github.com/Mentoraat/plannings-tool-client/pull/4/files#diff-e9fbc97ddec464a79752b0d174f22bb8R80)
[The usage with this and event](https://github.com/Mentoraat/plannings-tool-client/pull/4/files#diff-e9fbc97ddec464a79752b0d174f22bb8R100)

JeremybellEU:
Sidenote: I would like to be proven wrong that this method is unnecessary. Together with a fellow student colleague we have tried for several days to get it to work with `.bind`. However, as we (the receiver) don't have the reference to the calling element, we only know the reference upon invoking our function. At which point the `this` reference gives us the pointer.

Once again, if there is a solution with `.bind` we are highly interested in it, because we spend quite some hours this week trying to solve the issue. We are currently convinced it is not possible unless you use an extra function like expose.

JWGmeligMeyling:
You are looking for a principle known as currying. Currying is included in several libs such as [lodash](https://lodash.com/docs#curry), and the principle is explained in for example [this](https://www.youtube.com/watch?v=iZLP4qOwY8I) video.

For the sake of not having to use lodash, I will give an example using the following curry function, taken from [this](https://medium.com/@kbrainwave/currying-in-javascript-ce6da2d324fe) blogpost.
```js
var currier = function(fn) {
   var args = Array.prototype.slice.call(arguments, 1);
   return function() {
      return fn.apply(this, args.concat(Array.prototype.slice.call(arguments, 0)));
   }
}
```

Then, you can do the following:
```js
var a = { value: 'a' }
var b = { value: 'b' }
var c = { value: 'c' }

b.drag = currier(function(a, b) { 
   console.log(this, a, b);
}, b)

b.drag.call(a, c) // will print a, b and c
```

Or applied to your own usage:
```js
    //...
    drag: curry(function drag(calendar, event) {
      // interact with this (the element), calendar and event
    })(this)
   //---
```

JeremybellEU:
@samccone Hey, I just saw commit https://github.com/Polymer/polymer/commit/6ed836f9aceb25c012786cac76c2be267710699b Should I open an issue for this PR retroactively?

arthurevans:
@JeremybellEU Yes, please. It's helpful if the bug includes the simplest possible repro case for the issue you're seeing (so we understand the problem and so we can test the patch).

JeremybellEU:
While thinking about the repro case, I mostly could come up with the FullCalendar example. Searching on the internet shows similar problems, but the impact is minor. Therefore I think it is better to add this feature to my local Polymer instead of for everyone.

PR #2486, opened 2015-09-26
Allow newlines in computed binding argument list

mgiuffrida:
This allows breaking in the middle of computed bindings, e.g.:

```
<div hidden$="[[_someVeryLongFunctionName(
                  someVeryLongArg,
                  anotherCrazyLongArg]]">
```

Note that `[\s\S]*` [doesn't seem to have worse performance]( http://jsperf.com/javascript-multiline-regexp-workarounds/5) than `.*`.


samccone:
Hey @mgiuffrida awesome spot!!

Would you mind adding a new unit test here?
https://github.com/Polymer/polymer/blob/36b9fe034f5e67820a8aaca9020120415397cd15/test/unit/bind-elements.html

Thanks a million!!! :sparkles:

mgiuffrida:
@samccone Done and they pass locally in Cr and FF. PTAL, thanks!

Chrome still limits HTML to 80 chars in WebUI so being able to break up long lines is helpful.

samccone:
awesome work @mgiuffrida this LGTM :+1: 

ping @ebidel for :eyes:

nazar-pc:
Nice addition

mgiuffrida:
@danbeam for FYI

david-saslawsky:
I can merge this fix into #2580 if that helps. 

Would be nice to get rid of the parsing oddities in computed bindings since they're frequently used in Polymer 1.0.

````
[[format(
    '[[D\'oh!]], "{0}" items left', 
    items.count
)]]
````

nazar-pc:
There is even one more fix for this area: #2295

david-saslawsky:
BTW, it looks like the intention in the original regular expression was to have a `\s*` before the `(` like this:

<pre>var m = expression.match(/([^\s]+)<b>\s*</b>\(([\s\S]*)\)/);</pre>

otherwise why use `[^\s]+` instead of `[^\(]+` ?

Maybe the expression should be:

````
var m = expression.match(/([^\s]+?)\s*\(([\s\S]*)\)/);
````
or
````
var m = expression.match(/([^\s\(]+)\s*\(([\s\S]*)\)/);
````

TimvdLippe:
If I am not mistaken, this was added in #3017 with [`\\s*`](https://github.com/Polymer/polymer/pull/3017/files#diff-870cfe064a36c23b69e8da891fdff6beR98)

mgiuffrida:
No, this still seems to be an issue: http://jsfiddle.net/mgiuffrida/x921yb0L/1/

As the expression has changed repeatedly since September, could someone from Polymer own this issue?

kevinpschaaf:
We'll take a look at merging this; just need to make sure it's compatible with the first-level regex change in https://github.com/Polymer/polymer/pull/3017/files#diff-870cfe064a36c23b69e8da891fdff6beR92.

mgiuffrida:
I'll give it a shot tonight.

mgiuffrida:
I've opened #3370 and rebased there, my original repo was deleted.

mgiuffrida:
Okay, I think this is right now? I always screw up GitHub pull requests...

mgiuffrida:
triggering new build

kevinpschaaf:
LGTM

kevinpschaaf:
Thanks @mgiuffrida

PR #2503, opened 2015-09-30
Added note about including a clear repro case.

arthurevans:
Thought this note was worth adding.

Do we want to add a note on unit tests? I don't know what the policy is, but maybe something like, "We **strongly** prefer PRs that include unit tests for any new or changed functionality." 

Feels like we would want to require unit tests for new & changed functionality at the minimum.


samccone:
:+1:

ebidel:
LGTM

PR #2508, opened 2015-09-30
Parent property values should come from template. Fixes #2504.

### LINKED ISSUE #2504: Templatizer: Variables in the parent scope are not passed during initialization
When the templatized template is not the templatizer client, variables in the parent scope are not passed during initialization.

Originally filed as: https://github.com/PolymerElements/iron-list/issues/107

Root caused to https://github.com/Polymer/polymer/blob/master/src/lib/template/templatizer.html#L385-L389; parent props need to be pulled from the <template> rather than the <iron-list> element.

kevinpschaaf:


samccone:
is there a test we can put in place here¿

PR #2512, opened 2015-09-30
Fixes: 2511: adds optional `strip-whitespace` attribute to templates …

sorvell:
…which causes Polymer to eliminate extraneous whitespace in templates.

### DIFF
--- a/src/mini/template.html
+++ b/src/mini/template.html
@@ -34,10 +34,30 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           'Move inside simple <template>.'));
       }
       // bootstrap the template if it has not already been
-      if (this._template && !this._template.content && HTMLTemplateElement.bootstrap) {
-        HTMLTemplateElement.decorate(this._template);
-        // recurse if necessary
-        HTMLTemplateElement.bootstrap(this._template.content);
+      if (this._template) {
+        if (!this._template.content && HTMLTemplateElement.bootstrap) {
+          HTMLTemplateElement.decorate(this._template);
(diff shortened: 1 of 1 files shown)


src/mini/template.html
```diff
@@ -34,10 +34,30 @@
...
-        // recurse if necessary
-        HTMLTemplateElement.bootstrap(this._template.content);
+      if (this._template) {
+        if (!this._template.content && HTMLTemplateElement.bootstrap) {
+          HTMLTemplateElement.decorate(this._template);
+          // recurse if necessary
+          HTMLTemplateElement.bootstrap(this._template.content);  
+        }
+        if (this._template.hasAttribute('strip-whitespace')) {
+          this._stripWhiteSpaceNodes(this._template.content);
```
kevinpschaaf: Consider moving to annotator? May not be worth it?

src/mini/template.html
```diff
@@ -34,10 +34,30 @@
           'Move inside simple <template>.'));
       }
       // bootstrap the template if it has not already been
-      if (this._template && !this._template.content && HTMLTemplateElement.bootstrap) {
-        HTMLTemplateElement.decorate(this._template);
-        // recurse if necessary
-        HTMLTemplateElement.bootstrap(this._template.content);
+      if (this._template) {
```
kevinpschaaf: While here, would be nice to document that this is "when element templates are in the main document (e.g. vulcanized) they are processed before the template polyfill runs"

azakus:
Dupe of https://github.com/Polymer/polymer/commit/35a1b94

Issue #2509, opened 2015-09-30
Improve notifyPath/notifySplices API for public usage


kevinpschaaf:
The current path notification API has a couple of warts making it problematic for use by data adapter layers providing alternate forms of data observation:

* `notifyPath` accepts only collection keys for arrays (as opposed to array indicies as `set` does).  With the changes to resolve https://github.com/Polymer/polymer/issues/2007, `set` will accept either keys or indicies, and the public version of `notifyPath` should do so as well; this will likely involve refactoring internal usage of `notifyPath` to avoid the overhead of key/index disambiguation when it is known that the path is key-based
* `_notifySplices` is currently private and undocumented; for a fully operable data observation layer, this needs to be exposed publicly

kevinpschaaf:
PR here: https://github.com/Polymer/polymer/pull/2513

devinivy:
@kevinpschaaf I think the docs may need some adjustment now, given that `set` no longer leverages `notifyPath` for sugar.  Thinking of [this](https://github.com/Polymer/polymer/blob/master/PRIMER.md#set-path) quote,
> Since in the majority of cases, `notifyPath` will be called directly after an assignment, a convenience function `set` is provided that performs both the assignment and notify actions.

I realize that it doesn't explicitly say that `set` uses `notifyPath` directly, but it seems to imply that this is the case.  Now `set` uses the private interface `_notifyPath`.

Out of curiosity, is there any way `set`, `splice`, `push`, etc. could continue to use the `notifyPath` public interface?  We were patching `notifyPath` to add functionality to `set`, `splice`, `push`, etc.  It's a nice assumption to be able to make, that `notifyPath` is used by all those methods.

kevinpschaaf:
That the other API's now use `_notifyPath` rather than the public `notifyPath` is an implementations detail.  We do this for performance reasons: since the public `notifyPath` has to do the index->key normalization, we are careful to only do that once in the other public API's, and call `_notifyPath` internally, which accepts only pre-normalized key-based paths.  So I don't think we'd consider changing that back as it's a performance consideration.

devinivy:
Cool– that makes sense.  Thanks for getting back!

PR #2513, opened 2015-10-01
Improvements to path API related to arrays. Fixes #2007 and #2509

### LINKED ISSUE #2007: Array observer delivers incorrect changeRecord.path index after deleting item in array

kevinpschaaf:
Previously, `get` & `set` expected paths with array _indicies_ (only), whereas `notifyPath` expected paths with collection _keys_ (only).  Furthermore, paths in wildcard `.*` observer info objects always collection _keys_ (rather than indices for arrays), which could not be passed directly to `get` or `set` since they expected indices.

Included in this PR:

* Denote keys with `#` to disambiguate from index
  * `items.3.name` evaluates to `items[3].name`
  * `items.#3.name` evaluates to `Polymer.Collection.get(this.items).getItem('#3').name`
* `get`, `set`, `notifyPath`, and all other public API's that take paths now accept paths with either keys or indices
  * Involves separating public `notifyPath` (which does index->key conversion) from private `_notifyPath` (which expects key-based paths only), and changing all internal use of `notifyPath` to `_notifyPath`
* Added public `notifySplices` API for integration with alternate data observation systems


kevinpschaaf:
Topics to review:
* Continue storing collection key as `store` index and converting to/from `#`-prefixed string? Or switch to dictionary + running count and store as `#`-prefixed string and eliminate conversion step?  May need to measure.
* `info` arg added to `get` is kinda weird, but saves 2x walk over path parts.  Alternatives?
* `_notifyPathUp` naming alternatives?

src/standard/notify-path.html
```diff
@@ -146,29 +157,56 @@
...
+              // Replace item associated with key in collection
+              coll.setItem(key, value);
+            }
           }
+          // Set value to object at end of path
           prop[last] = value;
+          // Notify observers of path change
           if (!root) {
-            this.notifyPath(parts.join('.'), value);
+            this._notifyPath(parts.join('.'), value);
```
sorvell: Why only if not `root`?

src/lib/collection.html
```diff
@@ -46,10 +46,11 @@
       } else {
         this.pmap[item] = key;
       }
-      return key;
+      return '#' + key;
```
sorvell: Storing these including the `#` seems simpler.

src/standard/notify-path.html
```diff
@@ -146,29 +157,56 @@
...
+          // item in collection associated with key for that item
+          if (array) {
             var coll = Polymer.Collection.get(array);
-            var old = prop[last];
-            var key = coll.getKey(old);
-            parts[i] = key;
-            coll.setItem(key, value);
+            if (parseInt(last) == last) {
+              // Dereference index & lookup collection key
+              var old = prop[last];
```
sorvell: Couldn't last be a key? If so, this lookup will fail.

src/standard/notify-path.html
```diff
@@ -186,23 +224,42 @@
...
+       *   indices, the index may be used as a dotted part directly
        *   (e.g. `users.12.name` or `['users', 12, 'name']`).
        * @param {Object=} root Root object from which the path is evaluated.
+       * @param {Object=} info Info object; if supplied, a `path` property
+       *   will be added containing path with array indices converted to keys
        * @return {*} Value at the path, or `undefined` if any part of the path
        *   is undefined.
        */
-      get: function(path, root) {
+      get: function(path, root, info) {
```
sorvell: Can we factor `info` into another function for internal use?

src/standard/notify-path.html
```diff
@@ -146,29 +157,56 @@
...
-            coll.setItem(key, value);
+            if (parseInt(last) == last) {
+              // Dereference index & lookup collection key
+              var old = prop[last];
+              var key = coll.getKey(old);
+              // Translate array indices to collection keys for path notificaiton
+              parts[i] = key;
+              // Replace item associated with key in collection
+              coll.setItem(key, value);
+            } else if (last[0] == '#') {
```
sorvell: Flip the condition since this is the more common case.

sorvell: Flip the condition.

src/standard/notify-path.html
```diff
@@ -330,27 +387,67 @@
...
+       *   ordered and must be normalized/merged before notifying).
+      */
+      notifySplices: function(path, splices) {
+        var info = {};
+        var array = this.get(path, this, info);
+        // Notify change to key-based path
+        this._notifySplices(array, info.path, splices);
+      },
+
+      _notifySplices: function(array, path, splices) {
```
sorvell: Add note here that path must contain keys not indices.

src/standard/notify-path.html
```diff
@@ -146,29 +157,56 @@
...
-            prop = prop[part];
-            if (array && (parseInt(part) == part)) {
-              parts[i] = Polymer.Collection.get(array).getKey(prop);
+            if (array && part[0] == '#') {
+              // Part was key; lookup item in collection
+              prop = Polymer.Collection.get(array).getItem(part);
+            } else {
+              // Get item from simple property dereference
+              prop = prop[part];
+              if (array && (parseInt(part) == part)) {
```
samccone: We might want to specify the radix here, to eliminate an entire class of odd behavior 

screen shot 2015-10-08 at 5 47 35 pm

sorvell:
LGTM!

PR #2517, opened 2015-10-02
add util method for shadow children

notwaldorf:
Sometimes it's useful to tell whether a node is in your light or shadow DOM*, so this PR adds a convenience method for that. It's the `shadowDomChild` version, since the similar `isLightDom` would also return true for sibling elements which is a bit of a lie.

* Mostly around event handling; light dom children should usually handle their own events (see: https://github.com/PolymerElements/paper-item/issues/28, https://github.com/PolymerElements/paper-menu/issues/44)

Let me know if the space trimming changes drive you insane -- Chromium used to disallow trailing spaces, so my editor auto-nukes them on sight.

/cc @cdata @blasten

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -150,7 +150,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      *   payload.
      * @param {Object=} options Object specifying options.  These may include:
      *  `bubbles` (boolean, defaults to `true`),
-     *  `cancelable` (boolean, defaults to false), and 
+     *  `cancelable` (boolean, defaults to false), and
      *  `node` on which to fire the event (HTMLElement, defaults to `this`).
      * @return {CustomEvent} The new event that was fired.
      */
@@ -199,11 +199,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     /**
      * Removes an item from an array, if it exists.
-     * 
-     * If the array is specified by path, a change notification is 
+     *
+     * If the array is specified by path, a change notification is
      * generated, so that observers, data bindings and computed
-     * properties watching that path can update. 
-     * 
+     * properties watching that path can update.
+     *
      * If the array is passed directly, **no change
      * notification is generated**.
      *
@@ -306,6 +306,29 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         }
       }
       return elt;
+    },
+
+    /**
+     * Checks whether an element is in this element's light DOM tree.
+     *
+     * @method isLightDescendant
+     * @param {HTMLElement=} node The element to be checked.
+     * @return {Boolean} true if node is in this element's light DOM tree.
+     */
+    isLightDescendant: function(node) {
+      return this.contains(node) &&
+          Polymer.dom(this).getOwnerRoot() === Polymer.dom(node).getOwnerRoot();
+    },
+
+    /**
+     * Checks whether an element is in this element's local DOM tree.
+     *
+     * @method isLocalDescendant
+     * @param {HTMLElement=} node The element to be checked.
+     * @return {Boolean} true if node is in this element's local DOM tree.
+     */
+    isLocalDescendant: function(node) {
+      return this.root === Polymer.dom(node).getOwnerRoot();
     }
 
   });
--- a/test/unit/utils-content.html
+++ b/test/unit/utils-content.html
@@ -26,7 +26,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <div></div>
     <div></div>
   </x-content>
-  
+
   <x-content-multi id="elt3">
     <span></span>
     <div></div>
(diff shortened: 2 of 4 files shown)


src/standard/utils.html
```diff
@@ -306,6 +306,18 @@
...
+    /**
+     * Convenience method for checking whether an element is in this element's
+     * shadow DOM tree;
+     *
+     * @method isShadowDomChild
+     * @param {HTMLElement=} node The element to be checked.
+     * @return {Boolean} true if node is in this element's shadow DOM tree.
+     */
+    isShadowDomChild: function(node) {
+      return (Polymer.dom(this).getOwnerRoot() !== Polymer.dom(node).getOwnerRoot());
```
cdata: :gem:

cdata: (I think the wrapping parens are not necessary in this case).

samccone: super minor nit: why the wrapping `( )`?

samccone: we could guard clause this method to see if node is even defined as well

notwaldorf: because i am paranoid that's why. nuking.

blasten: 🍭

samccone: :jack_o_lantern:

cdata: :ghost:

samccone: :-1: on the guard clause? too fancy? waste o LOC¿ :skull:

notwaldorf: @samccone Don't need to guard: Polymer.dom(null) is well defined and returns `document` :)

test/unit/utils-content.html
```diff
@@ -37,41 +37,46 @@
...
-    
+
     suite('content utils', function() {
-      
+
       var elt1 = document.querySelector('#elt1');
       var elt2 = document.querySelector('#elt2');
       var elt3 = document.querySelector('#elt3');
-      
+
```
cdata: Many trailing whitespaces died to bring us this PR...

notwaldorf: image

src/standard/utils.html
samccone: :goat: 

oh yeah the idea was simply to bypass the method and just return false maybe slightly faster :speedboat: 

but maybe not worth it

test/unit/utils-content.html
```diff
@@ -88,10 +93,30 @@
...
     });
-  
+
+    suite('isShadowDomChild utils', function() {
+      var elt1 = document.querySelector('#elt1');
+      var elt4 = document.querySelector('#elt4');
+
+      test('isShadowDomChild is false for light children', function() {
+        var span = elt4.querySelector('span');
```
cdata: Technically, should we be using `Polymer.dom` to search for these element references?

notwaldorf:
@sorvell i don't think the failing tests are my fault -- they're in `dom-repeat`

src/standard/utils.html
```diff
@@ -306,6 +306,18 @@
         }
       }
       return elt;
+    },
+
+    /**
+     * Convenience method for checking whether an element is in this element's
```
arthurevans: For consistency, we should start with a verb and end with a period, like:

    Checks whether an element is in this element's shadow DOM tree.

Or:

    Checks whether _node_ is in this element's shadow DOM tree.

blasten: +1

src/standard/utils.html
```diff
@@ -306,6 +306,17 @@
...
+    },
+
+    /**
+     * Checks whether an element is in this element's shadow DOM tree.
+     *
+     * @method isShadowDomChild
+     * @param {HTMLElement=} node The element to be checked.
+     * @return {Boolean} true if node is in this element's shadow DOM tree.
+     */
+    isShadowDomChild: function(node) {
```
sorvell: The name should be `isLocalDomChild`. Shadow and Shady are both implementations Polymer uses to provide local DOM.

src/standard/utils.html
```diff
@@ -306,6 +306,17 @@
...
+      return Polymer.dom(this).getOwnerRoot() !== Polymer.dom(node).getOwnerRoot();
```
sorvell: You also need to check `contains`. Simply being in the same ownerRoot does not guarantee that the element is inside the element's subtree.

cdata: @sorvell From an event dispatch perspective, does this matter? Put another way: in ShadowDOM, when I measure `event.target`, isn't the target always someone in my scope?

sorvell: Yes, that's true. Why is that relevant? This method has nothing to do with events afaict.

notwaldorf: I thiiiink this will return true for somebody else's local child. So if I have a `paper-item`
```
<paper-item>
  <span></span>
  <paper-checkbox></paper-checkbox>
</paper-item>
```
I believe with the current method, `paperItem.isShadowDomChild(paperCheckbox.someShadowChild)` will return true, and shouldn't.

Hence `this.contains(node)` or something like that.

sorvell: Sorry, I misread this...

it should be `this.root === Polymer.dom(node).getOwnerRoot();`

src/standard/utils.html
sorvell: Correction, let's not confuse the term `child` which has specific DOM meaning. Let's go with `isLocalDescendant`.

src/standard/utils.html
notwaldorf: Talked to @sorvell, the negation of this isn't "light descendant", so going back to `isLightDescendant", which is the one we really care about.

src/standard/utils.html
sorvell: I think it's reasonable also to add `isLightDescendant` too.

src/standard/utils.html
notwaldorf: :musical_note: gonna add both, gonna be fun times :notes:

cdata: I'm a little worried that the use case is getting lost in the shuffle here. Can someone clarify what the current direction is, so that I can give a meaningful response?

cdata: @sorvell @notwaldorf

notwaldorf:
@cdata: i've added correct implementations for `isLocalDescendant` and `isLightDescendant` (where descendant = child but i can't use that word). We still want `isLocalDescendant`

test/unit/utils-content.html
```diff
@@ -88,10 +97,55 @@
         var nodes = elt3.getContentChildren('[dne]');
         assert.equal(nodes.length, 0, 'should find 0 nodes');
       });
-      
+
+    });
+
+    suite('isShadowDomChild utils', function() {
```
sorvell: let's update this suite name

test/unit/utils-content.html
```diff
@@ -88,10 +97,55 @@
...
+      var elt6 = document.querySelector('#elt6');
+
+      test('isLightDescendant is true for light children', function() {
+        var span = elt4.querySelector('span');
+        var customElement = elt4.querySelector('x-content');
+        assert.isTrue(elt4.isLightDescendant(span));
+        assert.isTrue(elt4.isLightDescendant(customElement));
+      });
+
+      test('isLocalDescendant is false for local children', function() {
```
sorvell: light?

sorvell:
couple little test nits but otherwise LGTM.

notwaldorf:
@sorvell Done! Who merges in Polymer land?

sorvell:
Thx...

PR #2535, opened 2015-10-06
Only one real listener per `listen` call

### LINKED ISSUE #2534: Polymer events creates a new bound handler for every `listen` call
As discussed in person, this could be made slightly more optimal by looking to see if a listener has been created already.

Fixing this would allow `listen` to be called n times with the same arguments, while ensuring that only one listener is ever actually added.

azakus:
Actually use event listener cache correctly

Fixes #2534 Polymer Gestures creates a new bound handler for every `listen` call

### DIFF
--- a/src/standard/events.html
+++ b/src/standard/events.html
@@ -80,8 +80,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @param {string} methodName Name of handler method on `this` to call.
      */
     listen: function(node, eventName, methodName) {
-      this._listen(node, eventName,
-        this._createEventHandler(node, eventName, methodName));
+      var handler = this._recallEventHandler(this, eventName, node, methodName);
+      // reuse cache'd handler
+      if (!handler) {
+        handler = this._createEventHandler(node, eventName, methodName);
+      }
+      this._listen(node, eventName, handler);
     },
 
     _boundListenerKey: function(eventName, methodName) {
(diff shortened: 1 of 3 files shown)


samccone:
Is there a test we can put in place for this?

azakus:
@samccone yep

cdata:
A good test might be..
1. Add a listener twice
2. Verify that the event is only handled once
3. Remove the listener
4. Verify that the event is no longer handled

src/standard/events.html
```diff
@@ -80,8 +80,12 @@
      * @param {string} methodName Name of handler method on `this` to call.
      */
     listen: function(node, eventName, methodName) {
-      this._listen(node, eventName,
-        this._createEventHandler(node, eventName, methodName));
+      var handler = this._recallEventHandler(this, eventName, node, methodName);
+      // reuse cache'd handler
```
cdata: :beers:

cdata:
Relevant for https://github.com/PolymerElements/iron-selector/pull/64

azakus:
@sorvell or @kevinpschaaf PTAL

test/unit/events.html
```diff
@@ -141,6 +141,41 @@
           assert.deepEqual(el._removed[3], {target: 'div', event: 'bar'});
         });
       });
+
+      suite('Event Listener Cache', function() {
```
samccone: woot

sorvell:
LGTM

PR #2545, opened 2015-10-08
Missing var keyword

trevordixon:


samccone:
this LGTM unless there is some crazy reason for this to be as is :)

sorvell:
LGTM, thanks!

PR #2548, opened 2015-10-08
Effective children

kevinpschaaf:


src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+      _updateContentElements: function(info) {
+        this._observeContentElements(info.addedNodes);
+        this._unobserveContentElements(info.removedNodes);
+      },
+
+      _observeContentElements: function(elements) {
+        for (var i=0, h, n; (i < elements.length) && (n=elements[i]); i++) {
+          if (this._isContent(n)) {
+            n.__observeNodesMap = n.__observeNodesMap || new WeakMap();
+            if (n.__observeNodesMap.get(this) == null) {
```
kevinpschaaf: `n.__observeNodesMap.has(this)`

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+      _observeContent: function(content) {
+        return Polymer.dom(content).observeNodes(this._notify.bind(this));
+      },
+
+      _unobserveContentElements: function(elements) {
+        for (var i=0, n, h; (i < elements.length) && (n=elements[i]); i++) {
+          if (this._isContent(n)) {
+            h = n.__observeNodesMap.get(this);
+            if (h) {
+              n.__observeNodesMap.set(this, null);
```
kevinpschaaf: `n.__observeNodesMap.delete(this)`

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+      },
+
+      _ensureCleanup: function() {}, // abstract
+
+      _hasListeners: function() {
+        return Boolean(this._listeners.length);
+      },
+
+      addNode: function(node) {
+        if (this._hasListeners()) {
```
kevinpschaaf: Break out doc fragment passed as `nodes` into individual `addedNodes`?

src/lib/dom-api.html
```diff
@@ -508,45 +524,27 @@
...
-        if (!this._classList) {
-          this._classList = new DomApi.ClassList(this);
+      observeNodes: function(callback, options) {
+        if (!this.observer) {
+          this.observer = this.node.localName === CONTENT ?
+            new DomApi.MutationContent(this) :
+            new DomApi.Mutation(this);
         }
-        return this._classList;
+        return this.observer.addListener.apply(this.observer, arguments);
```
sorvell: Remove apply here.

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+        },
+
+        _unobserveAttributes: function(elements) {
+          for (var i=0, h, n; (i < elements.length) && (n=elements[i]); i++) {
+            if (n.nodeType === Node.ELEMENT_NODE) {
+              h = this._attrObservers.get(n);
+              if (h) {
+                h.disconnect(n);
+              }
+              this._attrObservers.set(n, null);
```
sorvell: delete

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+          for (var i=0, h, n; (i < elements.length) && (n=elements[i]); i++) {
+            if (n.nodeType === Node.ELEMENT_NODE) {
+              this._ensureElementAttrObserver(n);
+            }
+          }
+        },
+
+        _ensureElementAttrObserver: function(element) {
+          this._attrObservers = this._attrObservers || new WeakMap();
+          if (!this._attrObservers.get(element)) {
```
sorvell: has

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+
+      Polymer.Base.extend(DomApi.Mutation.prototype, {
+        
+        _ensureSetup: function(options) {
+          this._trackAttributes = this._trackAttributes ||
+            options.attributes;
+          if (!this._observer) {
+            this._observer = new MutationObserver(this._notify.bind(this));
+            this._preflush = this._flush.bind(this);
+            Polymer.dom.addPreflush(this._preflush);
```
kevinpschaaf: Minor: Rather than adding bound `_flush` functions to `_preFlushList`, why not just add `this` and have `flush` loop over `_preFlushList` and call `mutaiton._flush()` and avoid binding all the functions?  Want to keep the preflush concept abstract?

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+
+        _notify: function(mxns) {
+          var info = {
+            target: this.node,
+            addedNodes: [],
+            removedNodes: []
+          };
+          // collapse multiple mutations into one view
+          if (Array.isArray(mxns)) {
+            Array.prototype.forEach.call(mxns, function(m) {
```
kevinpschaaf: Avoid `forEach` going forward?

src/lib/dom-api.html
```diff
@@ -435,28 +451,28 @@
...
-          if ((d.nodeType === Node.ELEMENT_NODE) &&
-            (d.localName !== CONTENT) &&
-            matchesSelector.call(d, selector)) {
-            results.push(d);
+      getEffectiveChildNodes: function() {
+        var list = [];
+        var c$ = this.childNodes;
+        for (var i=0, l=c$.length, c; (i<l) && (c=c$[i]); i++) {
+          if (c.localName === CONTENT) {
+            list = list.concat(factory(c).getDistributedNodes().slice());
```
kevinpschaaf: What's the `slice()` for?

src/lib/dom-api.html
```diff
@@ -752,7 +750,35 @@
         return n$ ? Array.prototype.slice.call(n$) : [];
       };
 
-      DomApi.prototype._distributeParent = function() {}
+      DomApi.prototype._distributeParent = function() {};
+
+      DomApi.prototype.notifyObservers = function() {
+        if (this.node.shadowRoot) {
```
kevinpschaaf: Need help understanding
* Why notify only shadow root observers?
* Where is notifyObservers for shady?
* Notify only schedules notify... not a flush?

src/lib/dom-api-flush.html
```diff
@@ -0,0 +1,88 @@
...
+    _FLUSH_MAX: 100,
+    _needsTakeRecords: !Polymer.Settings.useNativeCustomElements,
+    _debouncers: [],
+    _preFlushList: [],
+    _finishDebouncer: null,
+
+    // flush and debounce exposed as statics on Polymer.dom
+    flush: function() {
+      for (var i=0; i < this._preFlushList.length; i++) {
+        this._preFlushList[i]();
```
kevinpschaaf: Move `_preFlush` into `_flush` so it loops?

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+      addListener: function(callback, options) {
+        options = options || {};
+        this._ensureSetup(options);
+        var listener = {fn: callback, options: options};
+        this._notifyInitial(listener);
+        return this._listeners.push(listener);
+      },
+
+      removeListener: function(handle) {
+        this._listeners.splice(handle - 1, 1);
```
kevinpschaaf: This invalidates previously held handles.

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+        }
+      },
+
+      _ensureCleanup: function() {}, // abstract
+
+      _hasListeners: function() {
+        return Boolean(this._listeners.length);
+      },
+
+      addNode: function(node) {
```
kevinpschaaf: Note which are shady-only?

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+      },
+
+      removeNode: function(node) {
+        if (this._hasListeners()) {
+          this._removedNodes.push(node);
+          this._scheduleNotify();
+        }
+      },
+
+      addAllNodes: function(force) {
```
kevinpschaaf: Decide if force is required

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+          removedNodes: this._removedNodes
+        }
+        this._updateContentElements(info);
+        this._callListeners(info);
+        this._addedNodes = [];
+        this._removedNodes = [];
+      },
+
+      _notifyInitial: function(listener) {
+        var info = {
```
kevinpschaaf: Only go if there are `childNodes`?  Otherwise will get an empty notification followed by the one from shady's call to `addAllNodes` if observed before distribution?

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+      },
+
+      addAllNodes: function(force) {
+        if (this._hasListeners() || force) {
+          var c$ = this.domApi.childNodes;
+          if (c$.length) {
+            for (var i=0, c; (i < c$.length) && (c=c$[i]); i++) {
+              this._addedNodes.push(c);
+            }
+            this._scheduleNotify();
```
kevinpschaaf: Is there a chance for double-notification in case of:

* observe no. 1
* initial distribution (schedules notify because there is a listener)
* observe no. 2 (notifies initial & receives scheduled notification also?)

`_notifyInitial` should be gated on whether initial distribution notification has been sent yet

src/lib/dom-api-mutation.html
kevinpschaaf: Is the check necessary? If everything is working correctly, should only ever observe one content once?

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+      },
+
+      _notify: function(mxns) {
+        var info = mxns || {
+          target: this.node,
+          addedNodes: this._addedNodes,
+          removedNodes: this._removedNodes
+        }
+        this._updateContentElements(info);
+        this._callListeners(info);
```
kevinpschaaf: Consider filtering out <content>'s between updating content elements and calling listeners?

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+            Polymer.dom.addPreflush(this._preflush);
+            // note: doing this > 1x is a no-op
+            this._observer.observe(this.node, {childList: true});
+          }
+          ensureSetup.call(this, options);
+        },
+
+        _ensureCleanup: function() {
+          this._observer.disconnect();
+          Polymer.dom.removePreflush(this._preflush);
```
kevinpschaaf: null this._observer so MO is re-opened in _ensureSetup?

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+    };
+
+    if (Settings.useShadow) {
+
+      var ensureSetup = DomApi.Mutation.prototype._ensureSetup;
+
+      Polymer.Base.extend(DomApi.Mutation.prototype, {
+        
+        _ensureSetup: function(options) {
+          this._trackAttributes = this._trackAttributes ||
```
kevinpschaaf: Impractical in shadow? Remove?  Otherwise, note that _trackAttributes tracking is coarse (ideally flag is cleared when last callback with attributes:true is removed.

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+          }
+        },
+
+        _ensureElementAttrObserver: function(element) {
+          this._attrObservers = this._attrObservers || new WeakMap();
+          if (!this._attrObservers.get(element)) {
+            var self = this;
+            this._attrObservers.set(element, new MutationObserver(
+              function(mxns) {
+                self._callListeners(mxns, function(options) {
```
kevinpschaaf: These are attribute mutations (not add/remove mutations)... Ideally needs to be converted to addedNodes/removedNodes.

src/lib/dom-api-mutation-content.html
```diff
@@ -0,0 +1,111 @@
...
+    };
+
+    DomApi.MutationContent.prototype = Object.create(DomApi.Mutation.prototype);
+
+    Polymer.Base.extend(DomApi.MutationContent.prototype, {
+
+      addListener: function(callback, options) {
+        var h = DomApi.Mutation.prototype.addListener.call(this, callback, 
+          options);
+        this._scheduleNotify();
```
kevinpschaaf: Super does notifyInitial (no-op since childNodes is empty).

Dirty check approach seems broken for initial?  Consider simplifying initial by constraining when you can observe?

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,280 @@
...
+
+      removeListener: function(handle) {
+        this._listeners.splice(handle - 1, 1);
+        if (!this._hasListeners()) {
+          this._ensureCleanup();
+        }
+      },
+
+      _ensureSetup: function(options) {
+        if (!this._isSetup) {
```
kevinpschaaf: options unused?

src/lib/dom-api-mutation-content.html
```diff
@@ -0,0 +1,111 @@
...
+        _ensureSetup: function(options) {
+          this._trackAttributes = this._trackAttributes ||
+            options.attributes;
+          if (!this._observer) {
+            var root = this.domApi.getOwnerRoot();
+            var host = root && root.host;
+            if (host) {
+              this._observer = Polymer.dom(host).observeNodes(
+                this.notify.bind(this), {attributes: this._trackAttributes});
+            }
```
kevinpschaaf: Will not recurse back up with attributes:true if was called previously with attributes:false, since ensureSetup is 1-shot

src/lib/dom-api-mutation.html
```diff
@@ -0,0 +1,219 @@
...
+            Polymer.dom.addPreflush(this._preflush);
+            // NOTE: subtree true is way too aggressive, but it easily catches
+            // attribute changes on children. These changes otherwise require
+            // attribute observers on every child. Testing has shown this
+            // approach to be more efficient. 
+            // TODO(sorvell): do we still need to include an option to defeat
+            // attribute tracking?
+            this._observer.observe(this.node, {
+              childList: true,
+              attributes: true,
```
kevinpschaaf: Sucks that there's no `childList` + `attributes` combo that actually means "any attribute changes on my childList".

So any change (add, remove, attribute change) to any of its light children descendants will incur the cost of an `_unobserveContentElements` iteration (once) and a levenshtein (per listener).

Seems potentially expensive enough to keep the attributes option, since it allows you to remove _both_ `attributes` + `subtree`.

Is there an option to auto-enable attributes if you encounter a `select` on a `<content>` while observing?  Or is that too complicated?

src/lib/dom-api-observe-distributed-nodes.html
```diff
@@ -0,0 +1,91 @@
...
+        _setup: function() {
+          if (!this._observer) {
+            var root = this.domApi.getOwnerRoot();
+            var host = root && root.host;
+            if (host) {
+              this._observer = Polymer.dom(host).observeNodes(
+                this._scheduleNotify.bind(this));
+              if (this._hasAttrSelect()) {
+                Polymer.dom(host).observer.enableShadowAttributeTracking();
+                this._observer._alwaysCallListener = true;
```
kevinpschaaf: `_alwaysCallListener` needs to get passed up via `enableShadowTracking`?

src/lib/dom-api-observe-nodes.html
```diff
@@ -0,0 +1,251 @@
...
+    var hasDomApi = Polymer.DomApi.hasDomApi;
+
+    /**
+     * DomApi.ObserveNodes tracks changes to an element's effective child nodes,
+     * the same list returned from `Polymer.dom(node).getEffectiveChildNodes()`.
+     * It is not meant to be used directly; it is used by
+     * `Polymer.dom(node).observeNodes(callback)` to observe changes.
+     */
+    DomApi.ObserveNodes = function(domApi) {
+      this.domApi = domApi;
```
kevinpschaaf: NodeObserver ?

src/lib/dom-api.html
```diff
@@ -266,6 +277,17 @@
         return parent && parent.shadyRoot && hasInsertionPoint(parent.shadyRoot);
       },
 
+      _removeNodeFromParent: function(node) {
+        var parent = node._lightParent;
+        if (parent && hasDomApi(parent)) {
+          var d =  factory(parent);
+          if (d._observer) {
+            d._observer.removeNode(node);
```
kevinpschaaf: Remove cruft.

src/mini/shady.html
```diff
@@ -463,6 +476,24 @@
...
+            dc.observer.notify();
+          }
+        }
+      }
+    }
+
+    function notifyInitialDistribution(host) {
+      if (hasDomApi(host) && Polymer.dom(host).observer) {
+        Polymer.dom(host).observer.notify();
+      }
```
kevinpschaaf: cache Polymer.dom(host)

kevinpschaaf:
LGTM!  :+1: :trophy: children: [ :boy: :girl: :baby: ] changed: [ :children_crossing: ]

samccone:
@sorvell / @kevinpschaaf any design docs around this feature / why it is good / how/when to use it?

sorvell:
Will be forthcoming after discussion with @arthurevans ...

arthurevans:
What I have learned so far is:

- use cases for the effective children APIs may be a trifle abstruse. But perhaps not so much as I think.
- observeNodes is nifty. 

For example, consider a tab panel element that we use like this:

    <simple-tab-panel>
      <simple-tab></simple-tab>
      <simple-tab></simple-tab>
      <div class="tab-page"></div>
      <div class="tab-page"></div>
    </simple-tab-panel>

The contract is simply: clicking the first tab shows the first tab-page, and so on. But since I'm a good citizen, I want to generate unique IDs for each of these elements and set up ARIA attributes to link them: `aria-controls="tab-page-someuniqueid"`.

It's easy to do this once in `ready`, but what if the user adds tabs later? I used to have to add a mutation observer, which still wouldn't catch the case where one of my light DOM children is a `<content>` element that may have children distributed into it.

Now I can do something like:

    Polymer.dom(this.$.tabContainer).observeNodes(tabsChanged);

Which is pretty nifty, right?

arthurevans:
See also: Chris's PR, above for example niftiness.

samccone:
Wooot for observeNodes makes sense thx @arthurevans , is there a way to unregister an observation?

cdata:
@sorvell @arthurevans The asynchrony involved in using a `MutationObserver` to measure added and removed items suggests that `IronSelectableBehavior` might be redesigned to handle items that "don't exist yet" a little more gracefully. This is because an item "not existing" could mean A) the item doesn't actually exist as a child of the selectable, or B) the item exists but the `MutationObserver` hasn't told us about it yet. Right now, `IronSelectableBehavior` is designed around the assumption that the set of items is whatever you get when you `querySelectorAll` for items, and this is a fundamentally different approach that does not take full advantage of the change notifications you get from `observeNodes`.

cdata:
@samccone Yup, see https://github.com/PolymerElements/iron-selector/blob/observe-nodes/iron-selectable.html#L144

samccone:
Q: should `unobserveNodes` automatically be called when an element is detached?

cdata:
@samccone I don't think that you necessarily want that always, although I admit it is usually dubious for a user to want to respond to these types of changes while detached.

samccone:
I only wonder because of the potential for users to shoot themselves in the foot memory leak wise with this non private API.

cdata:
In support of what you are saying, `listeners` automatically "unlisten" when you detach (I think).

cdata:
Okay, I take that back. `listeners` _does not_ automatically "unlisten" upon detach.

samccone:
yeah so, imho unregistering automatically is a huge win for users from a memory and leak standpoint.. with the ability to conditionally opt out of this behavior

Can anyone come up with a case why this should not act this way?

ebidel:
Can you open an issue to discuss? This PR probably isn't the right place to
suggest new features :)

PR #2559, opened 2015-10-12
Fix URL to component.kitchen

Gpx:
## Problem

The guide links to [`www.component.kitchen`](http://www.component.kitchen) which does not exist.

## Solution

Change the link to [`component.kitchen`](http://component.kitchen).


samccone:
heh :+1: lgtm

Gpx:
How come the tests are failing on IE10-11?

PR #2568, opened 2015-10-15
Add travis config

azakus:
Set up travis to test chrome and linux always, and all of our supported browsers
for first-party PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,30 @@
+language: node_js
+sudo: false
+cache:
+  directories:
+  - node_modules
+matrix:
+  include:
+  - node_js: stable
+    script: xvfb-run -a wct --simpleOutput -l firefox -l chrome
+    addons:
+      firefox: latest
+      apt:
+        sources:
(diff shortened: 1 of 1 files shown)


azakus:
After this pull, I will disable our homegrown CI runner for polymer.

samccone:
Woot!

azakus:
As you can see, Travis and our homegrown CI server agree that IE 10 is crazy. My current best guess here is that this is related to the polyfilled MutationObserver.

Disturbingly, IE 10 works just like the rest if you run wct and sauce manually behind the $GOOG corporate proxy, so that leads me to thing this weird quirk is also networking related.

samccone:
@azakus if you add something like this
https://github.com/tastejs/todomvc/commit/abee8bdedb4913a6f70b4ad5167fad16ad333def

It should make CI super super easy to debug, it was a huge win for us over in todomvc land

kevinpschaaf:
LGTM

PR #2572, opened 2015-10-15
contributing copy fixup

samccone:
Issue describing the intend of the change: https://github.com/Polymer/polymer/issues/2573


ebidel:
LGTM

Issue #2579, opened 2015-10-16
Handle quoted strings when parsing annotations and method calls


david-saslawsky:
The Polymer annotation and function parsers should be aware of quoted string literals 
For example, The following annotation fails without a proper error message:
````
[[localize("Welcome back, (Mr {0})")]]
````

This would fix #1784, #2295 and any other tricky case where a string contains chars like `}`, `]`, `)`

TimvdLippe:
I think this issue was fixed as of #3017

kevinpschaaf:
This is fixed per #3017 assuming commas are escaped as described in the docs: https://www.polymer-project.org/1.0/docs/devguide/data-binding.html#literals

azakus:
JSBin working with escaped commas: http://jsbin.com/jimave/edit?html,output

PR #2580, opened 2015-10-16
#2579 Handle quoted string escaping when parsing annotations and methods

### LINKED ISSUE #1784: Computed binding can't handle comma in literal string
See http://plnkr.co/edit/G8hEnW0qwlB88ljO5V4S?p=preview, whose `dom-element.html` is below. If you add a comma in one of the arguments to `concat`, the rendering breaks.

```html
<link rel="import"
      href="http://www.polymer-project.org/1.0/samples/components/polymer/polymer.html">

<dom-module id="dom-element">
  <template>
    <p>{{concat('I am a DOM element and this is my ', field, ' DOM!')}}</p>
  </template>
</dom-module>

<script>
  Polymer({
    is: "dom-element",
    
    properties: {
      field: {
        type: String,
        value: 'local',
      },
    },
    
    concat: function() {
      return Array.prototype.join.call(arguments, '');
    },
  });
</script>
```

Amusingly, if you use `{{concat('I am a DOM element, ' and this is my ', field, ' DOM!')}}` as in http://plnkr.co/edit/TCFCC98q8utBvz7hg7oo?p=preview, it removes the character before the `,`, and recovers from there.

david-saslawsky:
Allows expressions like this one without escaping
````
[[localize("Welcome back, (Mr {0})")]]
````
Fixes  #1784, #2295

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -88,12 +88,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           this._parseElementAnnotations(node, list);
     },
 
-    _bindingRegex: /([^{[]*)({{|\[\[)([^}\]]*)(?:]]|}})/g,
+    _bindingRegex: (function() {
+        var dblQuotStr = '"(?:\\\\.|[^"\\\\])*"';
+        var quotStr = "'(?:\\\\.|[^'\\\\])*'";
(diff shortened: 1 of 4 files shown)


david-saslawsky:
@googlebot CLA signed.

david-saslawsky:
Fixed by #3017

PR #2584, opened 2015-10-16
Fix IE10 regressions. Fixes #2582

### LINKED ISSUE #2582: IE test regressions in dom-repeat, dom-bind, and dom-module-inline
IE tests have been failing due to an arcane timing bug in webcomponents-js, which have masked a couple of other bugs that have snuck in.

kevinpschaaf:
* Copy attribute list before modifying it
* Fall back to document for current document if no currentScript

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: node_js
 sudo: false
-cache:
-  directories:
-  - node_modules
 matrix:
   include:
   - node_js: stable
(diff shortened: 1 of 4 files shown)


sorvell:
LGTM

azakus:
Yeah, npm installs concurrently I believe, and wct also depends on bower. If bower is installing while wct is installing, npm assumes all the dependencies have loaded.

We've had bad lifecycle postinstalls like this in the past with selenium-standalone.

samccone:
>  If bower is installing while wct is installing, npm assumes all the dependencies have loaded.

woah, sounds like a great bug to file over at https://github.com/npm/npm/

azakus:
Here's a similar one: https://github.com/npm/npm/issues/7688

PR #2585, opened 2015-10-16
Do not configure compound property/attribute binding if literal if em…

### LINKED ISSUE #2583: Unset compound bindings with no literals should not override default target property value
To match the semantics of single-property bindings, unset compound bindings with no literals should not override default target property value.

kevinpschaaf:
…pty. Fixes #2583.


sorvell:
LGTM

kevinpschaaf:
ALL GREEEEEEEN!!!!!

PR #2588, opened 2015-10-17
Fix #2587: When Polymer.dom(el).appendChild(node) is called, cleanup …

### LINKED ISSUE #2587: `Polymer.dom(...).observeNodes` does not notify for some DOM removal
Please see the contrived test case here: http://jsbin.com/dajife/2

Removing nodes from the example element (notably, an element without a shadow root), fails to cause an expected notification from `observeNodes`.

sorvell:
…work must be performed on the existing parent of node. This change fixes a missing case in this cleanup work: if the existing parent has a observer via `Polymer.dom(parent).observeNodes`, it needs to be notified that node is being removed even if the node does not have specific logical info. For example, if an observed node has no Shady DOM and has a child that is removed. A test for this case was added.

### DIFF
--- a/src/lib/dom-api.html
+++ b/src/lib/dom-api.html
@@ -274,7 +274,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       },
 
       _removeNodeFromParent: function(node) {
-        var parent = node._lightParent;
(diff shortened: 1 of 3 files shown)


kevinpschaaf:
LGTM

PR #2600, opened 2015-10-20
[wip] Lint the javascript code with eslint

fredj:
Display errors with ```gulp lint```

work in progress

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,14 @@
+{
+  "extends": "eslint:recommended",
+  "env": {
+    "browser": true
+  },
+  "plugins": [
+    "html"
+  ],
+  "globals": {
+    "CustomElements": true,
+    "HTMLImports": true,
+    "Polymer": true
+  }
+}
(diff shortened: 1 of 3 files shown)


BLamy:
Should probably be called `gulp eslint` as I would expect `lint` would be reserved for polylint once they fit that into the toolchain.

fredj:
http://jscs.info/ should probably used instead as it's already the linter used in https://github.com/PolymerElements/polymer-starter-kit

LM450N:
:+1: for following the validation rules of the PSK. If you prefer ESLint maybe you could propose it for the PSK then add the `eslint` task to Polymer.

azakus:
Thanks for the PR!

I'm marking this as needs investigation and/or docs to make sure we evaluate linters for the best use case.

Is the community collecting around eslint the best, or is jshint, xo, jscs, etc. the best?

azakus:
Alright, after reviewing other linters in #3368, I think eslint is the way to go!

I have a few tweaks I want to make on this PR, so I'm going to merge it into a different branch and close this one. I'll reference it when it's ready.

Thanks again!

Issue #2520, opened 2015-10-02
ES6: support class extends


gutenye:
Given

```
class Base {
   say() { .. }
}

class User extends Base { }
Polymer(User)
```

It's not working. tested on `v1.1.4`

possible related to https://github.com/Polymer/polymer/pull/2365#discussion_r37827984

ebidel:
Confirmed this doesn't work. Repo: http://jsbin.com/wahepakeri/edit?html,output

Note, vanilla custom elements can extend from a base class that's extending `HTMLElement`: 
http://jsbin.com/neyijafoqo/edit?html,output

sjmiles:
Polymer doesn't currently support chained-prototypes as arguments to `Polymer()` in any context. This is a Roadmap Feature.

ebidel:
If we get element inheritance back, does it unlock this feature?

sjmiles:
It's more complicated than that.

laco0416:
@sjmiles  I tried to solve this issue.
https://github.com/Polymer/polymer/pull/2604
Please review this!

azakus:
We're currently working on inheritance support, and we will have a better opportunity to support this use case, #3345.

There are 4 cases we will support for inheritance

1. `MyPolymerElement extends OtherPolymerElement`
1. `MyPolymerElement extends HTMLElement`
1. `MyPolymerElement extends HTMLOtherNativeElement`
1. `MyPolymerElement extends RandomJSClass`

For case 1, the answer is simple: do nothing.
For case 2 and 3, we will need to replace the native html element prototype with one that includes `Polymer.base`.
For case 4, we will find the lowest prototype, Object, and replace that with `Polymer.Base` on `HTMLElement`.

laco0416:
@azakus multi-staged inheritance will be supported?

```
class MyBaseElement extends SomeClass

class MyElement extends MyBaseElement
```

azakus:
@laco0416 That should work, but my original statement in https://github.com/Polymer/polymer/issues/2520#issuecomment-176497918 appears to be too hard.

With a few helper functions we can support these use cases:

1. `class MyElement`
2. `class MyElement extends Polymer.Element` where `Polymer.Element` is `HTMLElement` plus the Polymer base class
3. `class MyElement extends Polymer.Extend('input')` where `Polymer.Extend` creates `HTMLInputElement` chained with Polymer base class
[...]

PR #2604, opened 2015-10-21
#2520 Support es6 class extends

laco0416:
Ref. https://github.com/Polymer/polymer/issues/2520

To support class extends, the prototype inheritances must be flattened to single prototype.
Sample: http://jsbin.com/wetiwi/edit?html,console

Please review my PR! Thanks.

### DIFF
--- a/src/lib/polymer-bootstrap.html
+++ b/src/lib/polymer-bootstrap.html
@@ -23,6 +23,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // if input is a `class` (aka a function with a prototype), use the prototype
       // remember that the `constructor` will never be called
       if (typeof prototype === 'function') {
+        // flatten the prototype if it has super class
+        flatten(prototype);
         prototype = prototype.prototype;
       }
       // if there is no prototype, use a default empty object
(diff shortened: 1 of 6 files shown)


laco0416:
Currently, Firefox (latest in travis) does not support es6 class but nightly build does.
To test, I used babel and compiled es6 to es5.

BLamy:
:+1:

laco0416:
@ebidel May I ask for reviewing my PR to someone?

ebidel:
Core team's job :) @sjmiles @sorvell @kevinpschaaf

src/lib/polymer-bootstrap.html
```diff
@@ -47,6 +49,22 @@
       return factory;
     };
 
+    var flatten = function(child) {
```
azakus: This feels super weird. If we want to attach the Polymer.Base prototype to the right spot (just above HTMLElement), then the proper procedure would be to write a function that traverses the given prototype chain and attaches Polymer.Base in the correct spot.

azakus:
We're working on Polymer element inheritance that will make for a better staring point in adding this feature.

Issue #2610, opened 2015-10-22
node.isLightDescendant(node) returns true and shouldn't.


notwaldorf:
Sigh. Never let me contribute to core again.

PR #2611, opened 2015-10-22
isLightDescendant should return false for self

notwaldorf:
Fixes https://github.com/Polymer/polymer/issues/2610


sorvell:
LGTM, thx :+1: :100:

PR #2626, opened 2015-10-26
deepActiveElement added to dom API

MartinMoizard:


MartinMoizard:
`According to the current spec,
http://w3c.github.io/webcomponents/spec/shadow/#active-element
Document.activeElement returns "adjusted" element, i.e.
when the real focused element is within a shadow tree, it returns
the top-most shadow host in the Document treescope.`
(w3c/webcomponents#104)

azakus:
Can you rebase this against master, I fixed a bunch of whitespace lines in `dom-api.hml` and it will be easier to look at :)

From what I can see, if you change https://github.com/MartinMoizard/polymer/blob/deepActiveElement/src/lib/dom-api.html#L899 to be `wrap(document).activeElement`, then you can make https://github.com/MartinMoizard/polymer/blob/deepActiveElement/src/lib/dom-api.html#L905 just be `return node`.

azakus:
@sorvell for style review

MartinMoizard:
@azakus I rebased, that's now easy to read :). I also made the changes you suggested but a unit test is failing on Firefox 41. I will check on Monday what's happening.

MartinMoizard:
@azakus I fixed the unit tests. I had to return wrap(node) instead of node like you suggested. Tell me if I'm missing something. Thanks.

hypnoce:
:+1:

MartinMoizard:
@sorvell did you have time to check this pull request? Thanks!

sorvell:
Sorry for the delay. Is there an issue filed that this is intending to solve? We'll need to discuss why this type of API is necessary before accepting this PR. Thanks.

MartinMoizard:
@sorvell There's no issue filed regarding this fix (for now).

Here is a use case that needs this new API: we are building a pivot widget, and we need to store the focused element (in our case this is a cell of the pivot) in order to restore the focus later in a specific scenario.

Let me know if you need more details. Thanks

sorvell:
We're trying to enforce a policy that every PR has an issue, so that'd be best.

We're disinclined to add this feature to Polymer. Strictly speaking this API is breaking encapsulation since it exposes a node available in a scoped tree. While the platform explicitly allows users to look inside scoped trees (via `this.shadowRoot`), it's probably not a pattern we want to codify within Polymer itself. 

This is definitely something your pivot widget can, itself, do; alternatively, you could expose this information more directly within the elements you're using. Of course, that is likely only feasible if the set of elements that can be inside the pivot widget is known and under your control.

In general, focus treatment in Shadow DOM is an area that we think needs more work so it would be helpful to know more about the details of your use case.

MartinMoizard:
@sorvell Thank you for your insights, we managed to use js events to capture the currently focused dom element.

Issue #2621, opened 2015-10-24
The custom properties shim alters the cascade order incorrectly


philipwalton:
For example, I have the following CSS in my `<flex-line>` element definition:

```css
:host {
  --flex-line-gutter: 1em;
}
:host ::content > * {
  margin: var(--flex-line-gutter) 0 0 var(--flex-line-gutter);
}
:host ::content > flex-line {
  margin: 0;
}
```

It get translated into this:

```html
<style scope="flex-line">
  flex-line > flex-line {
    margin: 0;
  }
</style>
<style scope="flex-line-0">
  flex-line.flex-line-0 > * {
    margin: 1em 0 0 1em;
  }
</style>
```

As you can see, the order of the rules has been reversed and the specificity of the first content rule has increased, completely changing the cascade order.

If you're going to change the specificity of a rule (which I understand you have to in some cases) you should normalize it so all rules are increased by the same relative specificity.

TimvdLippe:
This issues was fixed in #2642 but not closed.

ebidel:
Thanks Tim.

Phil holler if you see anything contrary.

Issue #2531, opened 2015-10-06
CSS over specificity when using vars 


samccone:
```html
<dom-module id="x-foo">
  <template>
     :host {
        background: var(--foo, red);
     }
    :host(.zap) {
        background: orange;
     }
  </template>
  <content></content>
</dom-module>
```

```html
<x-foo>i am red</x-foo>
<x-foo class='zap'>i am red also?!</x-foo>
```

------

Discovered here:
https://github.com/PolymerElements/paper-toolbar/pull/40/files

samccone:
xref: https://github.com/Polymer/polymer/issues/2439#issuecomment-145927714

ebidel:
Here's a working bin: http://jsbin.com/suhaqezoqa/edit?html,output

samccone:
thanks @ebidel +100

abdonrd:
Your jsbin, @ebidel, is not working for me. :/

ebidel:
It shouldn't work, right? It shows reproducing the bug.

abdonrd:
Aah, okey! Sorry @ebidel! :sweat_smile:

samccone:
xref #1873

clintwood:
@ebidel, @samccone, I needed to solve this now, so based off of your jsbin above I have a hacky workaround here: http://jsbin.com/xemuxe/edit?html,output that solves the specificity issue for now.

Essentially:
```html
<template>
<style>
:host {
  background: var(--foo, red);
  /* specificity - hack */
  --zap-styles: {
    background: orange;
  }
}
:host(.zap) {
  /* specificity - hack */
  @apply(--zap-styles);
}
</style>
<content></content>
</template>
```

azogheb:
:+1: on resolving this issue. Much like @clintwood we have used a workaround by using dummy / no-op var() on the rule that should have more specificity. 

I guess the question is why does Polymer split the CSS into two <style> blocks, one for rules with vars in the first one and one for rules with vars in the second?

nazar-pc:
I'm working on this issue. Got it kind of working, but some tests are failing, so I'll make PR when fix them and add new tests on top.

clintwood:
@nazar-pc, awesome - much needed!

nazar-pc:
Done in #2642, give it a try:)
@ebidel, no one is assigned to this issue, whom should we ask for PR review?

TimvdLippe:
This issues was fixed in #2642 but not closed.

[...]

Issue #1873, opened 2015-06-15
Overwriting mixin properties 


Pupix:
Hello! I am trying to overwrite a property coming from a CSS mixing, but the way Polymer handles the variables makes it impossible to do so (without using "!important").

```html
<style>
    .foo {
        @apply(--paper-font-caption)
        white-space: normal;
    }
</style>
```

The expected result would be to have "white-space: normal" applied on top of the one provided by paper-styles, however that's not the case.

Do I have to rewrite the entire mixing just to change one property or is there an alternative way?

Cheers.

sorvell:
This is due to the way that Polymer separates static styles from styles which contain custom properties. This use case is reasonable and our approach needs to be modified slightly to support this better.

We will probably not separate custom properties css from static css within rules that contain properties.

Pupix:
Thanks for the response. Will wait for an update then.

sorvell:
This issue also leads to specificity issues when combining static and custom property rules.

robrez:
:+1:

jmcgdz:
:+1:

nazar-pc:
@Pupix, can you provide complete test case? I guess this might be either already fixed by #2642 or I can try to extend PR to cover this case as well and add corresponding test case.

robrez:
@nazar-pc Here's a jsbin that, i think, demonstrates the issue:

https://jsbin.com/kefuzi/edit?html,output

nazar-pc:
OK, here is easier to understand example: https://jsbin.com/xupufudivu/1/edit

nazar-pc:
Seems to be different area than covered in #2642, but I'll take a look on this as well.

nazar-pc:
#2642 now fixes this issue as well.

PR #2642, opened 2015-10-29
Fix for incorrect CSS selectors specificity (fixes at least #2531, #1873 and #2621)

nazar-pc:
As reported in #2531 (while I expect many other similar might be fixed as well).
I've also fixed 2 existing tests where order of styles was hardcoded (it changed a bit now, but preserves priority even better and no other tests affected).
Tested manually under Shady DOM, Shadow DOM and Shadow DOM polyfill.
More tests specific to these improvements will be added soon (now it is time to sleep:)).

### DIFF
--- a/src/lib/css-parse.html
+++ b/src/lib/css-parse.html
@@ -11,11 +11,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 /*
   Extremely simple css parser. Intended to be not more than what we need
-  and definitely not necessarly correct =).
+  and definitely not necessarily correct =).
 */
 Polymer.CssParse = (function() {
 
-  var api = {
+  return {
     // given a string of css, return a simple rule tree
     parse: function(text) {
       text = this._clean(text);
@@ -31,7 +31,7 @@ Polymer.CssParse = (function() {
     _lex: function(text) {
       var root = {start: 0, end: text.length};
       var n = root;
-      for (var i=0, s=0, l=text.length; i < l; i++) {
+      for (var i=0, l=text.length; i < l; i++) {
         switch (text[i]) {
           case this.OPEN_BRACE:
             //console.group(i);
@@ -123,7 +123,7 @@ Polymer.CssParse = (function() {
           }
         }
       }
-      // emit rule iff there is cssText
+      // emit rule if there is cssText
       if (cssText) {
         if (node.selector) {
           text += node.selector + ' ' + this.OPEN_BRACE + '\n';
@@ -184,10 +184,6 @@ Polymer.CssParse = (function() {
     AT_START: '@'
 
   };
-
-
-  // exports
-  return api;
 
 })();
 
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -87,9 +87,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var parts = cssText.split(';');
         for (var i=0, p; i<parts.length; i++) {
           p = parts[i];
-          if (p.match(this.rx.MIXIN_MATCH) || p.match(this.rx.VAR_MATCH)) {
-            customCssText += p + ';\n';
-          }
+          customCssText += p + ';\n';
         }
         return customCssText;
       },
@@ -181,7 +179,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         rule.cssText = output;
       },
 
-      // Test if the rules in these styles matche the given `element` and if so,
+      // Test if the rules in these styles matches the given `element` and if so,
       // collect any custom properties into `props`.
       propertyDataFromStyles: function(styles, element) {
         var props = {}, self = this;
@@ -208,7 +206,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         return {properties: props, key: o};
       },
 
-      // Test if a rule matches scope crteria (* or :root) and if so,
+      // Test if a rule matches scope criteria (* or :root) and if so,
       // collect any custom properties into `props`.
       scopePropertiesFromStyles: function(styles) {
         if (!styles._scopeStyleProperties) {
@@ -218,7 +216,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         return styles._scopeStyleProperties;
       },
 
-      // Test if a rule matches host crteria (:host) and if so,
+      // Test if a rule matches host criteria (:host) and if so,
       // collect any custom properties into `props`.
       //
       // TODO(sorvell): this should change to collecting properties from any
@@ -319,7 +317,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           // otherwise, if we have css to apply, do so
           } else if (cssText) {
             // apply css after the scope style of the element to help with
-            // style predence rules.
+            // style precedence rules.
             style = styleUtil.applyCss(cssText, selector,
               nativeShadow ? element.root : null, element._scopeStyle);
           }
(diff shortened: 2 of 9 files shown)


nazar-pc:
Tests are also added (pretty basic, but anyway).

nazar-pc:
Added fix for #1873 since it touches similar places, so it would be much easier to merge, tests added for this case as well.
One caveat here: `@apply` location doesn't have any difference, namely:
```css
.green-2 {
  border-top: 2px solid green;
  @apply(--x-overriding);
}
```
works in the same way as:
```css
.green {
  @apply(--x-overriding);
  border-top: 2px solid green;
}
```
So, `@apply` rule kind of pops-up.
I'm not sure if it is according to specs, but implementing it in such way that takes order into account would be MUCH more difficult.
Also thinking logically, if you have local property - it should override one from mixin, otherwise it doesn't really make much sense.

nazar-pc:
@sorvell, it seems  you're the one whom I should ask for review here.

mfrost66:
@nazar-pc 
>>> So, @apply rule kind of pops-up.
I'm not sure if it is according to specs, but implementing it in such way that takes order into account would be MUCH more difficult.
Also thinking logically, if you have local property - it should override one from mixin, otherwise it doesn't really make much sense.

I would love to see order taken into account
1. it is consistent with current CSS behavior where order drives specificity.  Besides being "correct" it would meet the  [principal of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)
2. it allows order to be used to set a default style in case the @apply doesn't.  Users of libraries don't always know (or want to know) the details of the library - especially with complex libraries or novice/casual users.
3. If there is to be pain (it's hard), let's keep the pain point as small as possible - in your code in this case :)

Using your example, this allows me to set a default border-top.  The @apply may/may not have border styles
```css
.green-2 {
  border-top: 2px solid green;
  @apply(--x-overriding);
}
```

nazar-pc:
1. Agree, but I'll need to read through specs draft first, I'll do that a bit later
2. Makes sense
3. It would not be too easy, but I have some thoughts how to make it order-aware, I just need some time for this, stay tuned:)

JeremybellEU:
Is it possible to include a fix for #2621 too? (if this PR doesn't already fix the problem described) Seems they are closely related.

nazar-pc:
@JeremybellEU, this PR should fix #2621 already.

JeremybellEU:
:+1: The PR description should probably add "Fixes #2621" to get it in the changelog. Great work @nazar-pc !

nazar-pc:
Asked about `@apply` position awareness here: https://github.com/tabatkins/specs/issues/45
No response yet, but since it is much more logical to actually take position into account - I've added that as well in latest commit.

Now this PR seems to be complete and ready to radically improve experience of Polymer users when dealing with CSS mixins/variables :rocket::rocket::rocket:

@mfrost66, give it a try.

P.S. @sorvell, I hope you'll find some time for review in nearest future. Some of my non-merged PRs are waiting from 5th of August till today :cry:

azogheb:
+1
Will be nice to see these styling issues resolved!!

samccone:
R: @tjsavage for slotting

azakus:
This PR is pretty big, so I'm reviewing it last. I'll do the merge conflict work unless it gets too nasty.

nazar-pc:
I can rebase it if necessary

azakus:
that would be great :)

nazar-pc:
Rebased against master

BLamy:
:+1: I can finally stop spamming `!important`.

azakus:
@sorvell for review

nazar-pc:
WOW, just simplified it A LOT!
Also added more tests.
Turns out just one `if` caused all kinds of issues here:)

sorvell:
Thanks for doing this PR. We were considering solving this issue as in the PR, but we're a little concerned about the performance burden of having excess rules in the custom properties css. I'm going to try to dive into this and if the cost looks low, we'll go ahead and merge this.

Our longer term plan is to explore integrating custom property css together with the rest of the element css, making updates to custom properties via the CSSOM and using a single stylesheet for all of an element's styles. This is fairly involved and will mean more significantly forking the code for Shadow v. Shady DOM.

nazar-pc:
Rebased against master to resolve merging conflicts, squashed everything into single commit, since history was not useful.

ebidel:
https://github.com/Polymer/polymer/pull/2642 is another specificity issue likely fixed by this.

nazar-pc:
I guess it was a typo, #2642 is this PR's id.

ebidel:
Yes #2731 is the issue

mfrost66:
@ebidel Can you provide a status update?

I'm discouraged by the level of changes since the 1.0 release in regards to styling.  Not only have you significantly changed the feature set of 1.x with deprecation of /deep/ and css imports, but the proposed solution (mixins,vars) is buggy and of apparent low priority as evidenced by the age of this PR.  The level of effort required on our (users) part to stay up with an every-changing API is non-trivial.  More so when the features are either buggy or not well thought out.

>>>Our longer term plan is to explore integrating custom property css together with the rest of the element css, making updates to custom properties via the CSSOM and using a single stylesheet for all of an element's styles. This is fairly involved and will mean more significantly forking the code for Shadow v. Shady DOM

Your statement above implies that there is more significant change planned for the future.  My initial enthusiasm for web components is slowly turned into trepidation.  A clear statement on the status of styles and roadmap/timeline for a "final" solution would be greatly appreciated.  A timely fix for the current bugs surrounding styles would also be wonderful.  A caveat emptor on the polymer home page wouldn't be out of order until the team's able to sort this mess out.

nazar-pc:
@mfrost66, if you don't mind you can use often up to date version of Polymer with bunch of extremely useful patches in JS form (not HTML imports form to speed up things in browsers other than Chromium) here: https://github.com/nazar-pc/CleverStyle-CMS/tree/master/includes/js
There is file `a4.polymer-1.2.2+patches.min.js` there currently, you can see file history, there is complete list of patches included in commit message.

And yes, I'm very sad with pace of PRs merging too, since most of major issues are in fact already fixed for a long time now, just not upstreamed.

mfrost66:
@nazar-pc Thanks, which of the libraries would I need - just the a4.polymer*.js file?

I actually tried to apply your patches locally but ran into the sad fact that polymer [doesn't build on windows](https://github.com/Polymer/polymer/issues/2679)  and I was too flabbergasted to keep trying.

nazar-pc:
@mfrost66, yes, just that file - it contains Micro, Mini and Standard - full Polymer version that you'd get by including `polymer.html`.

mfrost66:
@nazar-pc 
I had expected the styles would penetrate into module content.  Is this a bug or is there an alternate way of doing this? 

In the example below, "inner content" should be colored red.
Example

_test.html_
```html
<!DOCTYPE html>
<html>
<head>
  <script charset="utf-8" src="bower_components/webcomponentsjs/webcomponents.js"></script>
  <link rel="import" href="bower_components/polymer/polymer.html">
  <link rel="import" href="my-modules.html">
</head>
<body>
<my-paragraph>
  <my-text>inner content</my-text>
</my-paragraph>
</body>
<script>
</script>
</html>
```

_my-modules.html_
```html
```

nazar-pc:
It should not work, you're applying color to `my-text` host, not its contents. If you have any questions - ask on stackoverflow or somewhere else, please, since it doesn't really relate to this issue directly.

azakus:
This change looks fine to me. @nazar-pc can you rebase this PR?

mfrost66:
Hmm, I believe it should in fact work.  From polymer styles doc's:
>>>
These custom properties can be defined similarly to other standard CSS properties and will **inherit from the point of definition down the composed DOM tree**, similar to the effect of color and font-family

Perhaps its not related to this PR?  I'll post it as a new bug.

azakus:
@nazar-pc scratch that, I'll rebase it since the conflict is with something I wrote :P

azakus:
Merged!

Sorry for the delay, the team has been heads down on some performance work, and I'm helping get everything back on track.

PR #2657, opened 2015-10-30
Produce nicer error on malformed observer

### LINKED ISSUE #2555: malformed observer registration yields a difficult to debug error
```js
...
observers: ['foo(zap, tap']
...
```

screen shot 2015-10-09 at 1 52 52 pm

Seems like something we might want to catch and surface a bit better

JeremybellEU:
It will now check if the regex in `_parseMethod` was unsuccesful (`sig == null`) and throw a more convenient exception for the user.

Also deduplicated lots of code in that test suite, as it became quite cumbersome to scroll and look at it.

Fixes #2555


samccone:
oh wow this is so nice @JeremybellEU 

cc @ebidel

JeremybellEU:
Rebased to solve merge conflict.

src/standard/effectBuilder.html
```diff
@@ -104,16 +104,28 @@
...
-    _addComplexObserverEffect: function(observer) {
-      var sig = this._parseMethod(observer);
-      for (var i=0, arg; (i<sig.args.length) && (arg=sig.args[i]); i++) {
-        this._addPropertyEffect(arg.model, 'complexObserver', {
-          method: sig.method,
-          args: sig.args,
-          trigger: arg
-        });
+    _addComplexObserverEffect: (function() {
+      function MalformedObserverException(observer) {
```
azakus: Just throw a regular `Error` with this message.
There's no need to make a custom error type.

azakus:
LGTM with nit.

JeremybellEU:
Thanks for the review @azakus . I have changed the code accordingly.

I was under the impression that this PR was on hold, since there is an ongoing discussion about error handling in #2736 .

azakus:
@JeremybellEU No, that discussion is complicated by the fact of the try/catch in registration to try and recover from broken element registration.

Since this PR is just making a better error message, it is totally fine.

JeremybellEU:
Glad to hear!

samccone:
wooot :dancer:

PR #2659, opened 2015-10-31
Adds incremental rendering to dom-repeat. Fixes #2537.

### LINKED ISSUE #2537: Performance: dom-repeat does not perform on large datasets
When using dom-repeat with large datasets the browser becomes unresponsive for a period of time while the render task is being completed. The `<iron-list>` component I believe was designed to solve this. However there are use cases that `<iron-list>` does not support (a grid view until that is implemented, or any other UI where repeated item position is calculated/randomized, etc).

I have thrown together a demo on github (was going to do a fiddle but the dummy api I am calling causes the request to throw a `NET::ERR_CERT_COMMON_NAME_INVALID` error)

http://jshcrowthe.github.io/dom-repeat-perf/non-chunked.html

In the following timeline you can see that there is an long, event-loop blocking, render task that takes place immediately after the request completes

screen shot 2015-10-06 at 5 11 02 pm

_NOTE: My machine is a 2013 MacBook Pro 2.3 GHz i7 with 16 GB RAM_

If you repeatedly click on the counter button while the request is being made you will notice that, up until the XHR request completes, the UI is responsive. However once the render task for the `dom-repeat` begins the subsequent click events are queued up and are unable to process until the full render has completed.

I hacked around with chunking the rendering task (utilizing `setTimeout(function() {...}, 0)`)  and was able to create a functional proof of concept that illustrates a potential solution. A demo can be found at the following URL:

http://jshcrowthe.github.io/dom-repeat-perf/chunked.html

Before I go too much further with this I wanted to get feedback to see if this was a good potential solution or if I am altogether going about this wrong. You can see the actual change I made here:

https://github.com/jshcrowthe/polymer/commit/a5d5edd930db3dba873056261e0234acb08f9ce9

_NOTE: I have only modified the `_applyFullRefresh` render method for the sake of demo. If this is a good solution all 3 of the render types would probably want to have this type of logic. In addition to that a way to abort an "in progress" render would need to be implemented to ensure that multiple calls to `_applyFullRefresh` don't result in duplication. You could also implement this type of chunking with DOM removals if it proves to be similarly slow._

If you compare the timeline for the chunked render you can see a drastic change in the timeline. The UI is not rendered all at one time however the render task never takes so long (at least in this simple instance) where it blocks user interaction.

screen shot 2015-10-06 at 6 45 57 pm

kevinpschaaf:
Adds optional incremental "chunked" rendering to `dom-repeat`.

New API:
* `limit` - a static value that can be used to limit the number of instances rendered
* `initialCount` - enables incremental rendering and sets initial render count (when used, limit will be automatically controlled and should not be set by the user)
* `chunkCount` - sets incremental count to be rendered per rAF; default is `'auto'`, which adjusts the chunk size using simple throttling scheme to target a set frame rate
* `targetFramerate` -  when `'auto'` chunk size is used, determines the target frame budget

If `initialCount` is set, after setting (or re-setting) `items`, the initial count will be rendered pre-paint, and all remaining items will be incrementally rendered at `requestAnimationFrame` timing in increments of `chunkCount`.

### DIFF
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -194,7 +194,7 @@ Then the `observe` property should be configured as follows:
        * Defines an initial count of template instances to render after setting
        * the `items` array, before the next paint, and puts the `dom-repeat`
        * into "chunking mode".  The remaining items will be created and rendered
-       * incrementally at each animation frame therof until all instnaces have
+       * incrementally at each animation frame therof until all instances have
        * been rendered.
        */
       initialCount: {
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -491,7 +491,7 @@ Then the `observe` property should be configured as follows:
       }
       // Remove any extra instances from previous state
       for (var j=this._instances.length-1; j>=i; j--) {
-        this._removeInstance(j);
+        this._detachAndRemoveInstance(j);
       }
     },
 
@@ -544,7 +544,7 @@ Then the `observe` property should be configured as follows:
           var idx = removedIdxs[i];
           // Removed idx may be undefined if item was previously filtered out
           if (idx !== undefined) {
-            this._removeInstance(idx);
+            this._detachAndRemoveInstance(idx);
           }
         }
       }
@@ -606,7 +606,7 @@ Then the `observe` property should be configured as follows:
       splices.forEach(function(s) {
         // Detach & pool removed instances
         for (var i=0; i<s.removed.length; i++) {
-          this._removeInstance(s.index);
+          this._detachAndRemoveInstance(s.index);
         }
         for (var i=0; i<s.addedKeys.length; i++) {
           this._insertPlaceholder(s.index+i, s.addedKeys[i]);
@@ -632,7 +632,7 @@ Then the `observe` property should be configured as follows:
       }
     },
 
-    _removeInstance: function(idx) {
+    _detachAndRemoveInstance: function(idx) {
       var inst = this._detachInstance(idx);
       if (inst) {
         this._pool.push(inst);
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -241,18 +241,6 @@ Then the `observe` property should be configured as follows:
         value: 20
       },
 
-      /**
-       * Maximum number of removed instances to pool for reuse when rows are
-       * added in a future turn.  By default, pooling is enabled.
-       *
-       * Set to 0 to disable pooling, which will allow all removed instances to
-       * be garbage collected.
-       */
-      poolSize: {
-        type: Number,
-        value: 1000
-      },
-
       _targetFrameTime: {
         computed: '_computeFrameTime(targetFramerate)'
       }
@@ -338,10 +326,10 @@ Then the `observe` property should be configured as follows:
       return Math.ceil(1000/rate);
     },
 
-    _initializeChunkCount: function(initialCount, chunkCount) {
-      if (initialCount) {
-        this.limit = initialCount;
-        this._chunkCount = parseInt(chunkCount, 10) || initialCount;
+    _initializeChunkCount: function() {
+      if (this.initialCount) {
+        this.limit = this.initialCount;
+        this._chunkCount = parseInt(this.chunkCount, 10) || this.initialCount;
       }
     },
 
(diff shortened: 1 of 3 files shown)


src/lib/template/dom-repeat.html
```diff
@@ -196,23 +252,32 @@
...
     observers: [
-      '_itemsChanged(items.*)'
+      '_itemsChanged(items.*)',
+      '_initializeChunkCount(initialCount, chunkCount)'
     ],
 
     created: function() {
       this._instances = [];
+      this._pool = [];
+      this._boundRenderChunk = this._renderChunk.bind(this);
```
sorvell: Probably worth avoiding Function.bind here.

src/lib/template/dom-repeat.html
```diff
@@ -253,6 +316,50 @@
...
+      if (this.initialCount) {
+        this.limit = this.initialCount;
+        this._chunkCount = parseInt(this.chunkCount, 10) || this.initialCount;
+      }
+    },
+
+    _tryRenderChunk: function() {
+      if (this._chunkCount &&
+          Math.min(this.limit, this._instances.length) < this.items.length) {
+        this.debounce('renderChunk', this._requestRenderChunk);
```
kevinpschaaf: comment: debounce required to ensure chunked rendering & data mutation rendering don't cause duplicate rAF's

src/lib/template/dom-repeat.html
```diff
@@ -253,6 +316,50 @@
...
+    _initializeChunkCount: function() {
+      if (this.initialCount) {
+        this.limit = this.initialCount;
+        this._chunkCount = parseInt(this.chunkCount, 10) || this.initialCount;
+      }
+    },
+
+    _tryRenderChunk: function() {
+      if (this._chunkCount &&
+          Math.min(this.limit, this._instances.length) < this.items.length) {
```
kevinpschaaf: condition should just be `(this.limit < this.items.length)`

src/lib/template/dom-repeat.html
```diff
@@ -253,6 +316,50 @@
...
+    },
+
+    _computeFrameTime: function(rate) {
+      return Math.ceil(1000/rate);
+    },
+
+    _initializeChunkCount: function() {
+      if (this.initialCount) {
+        this.limit = this.initialCount;
+        this._chunkCount = parseInt(this.chunkCount, 10) || this.initialCount;
```
kevinpschaaf: Don't init chunk count here (unnecessary); measure initial render time & use for first chunk

src/lib/template/dom-repeat.html
```diff
@@ -253,6 +316,50 @@
...
+      if (this.items) {
+        this._debounceTemplate(this._render);
+      }
+    },
+
+    _computeFrameTime: function(rate) {
+      return Math.ceil(1000/rate);
+    },
+
+    _initializeChunkCount: function() {
```
kevinpschaaf: Remove observer; only needs to be checked in itemsChanged

src/lib/template/dom-repeat.html
```diff
@@ -324,7 +434,7 @@
       if (this._needFullRefresh) {
         this._applyFullRefresh();
         this._needFullRefresh = false;
-      } else {
+      } else if (this._keySplices.length) {
```
kevinpschaaf: Comment re: 3 cases (items/sort/filter changed, splices, limit)

src/lib/template/dom-repeat.html
```diff
@@ -385,17 +513,18 @@
         var key = keys[i];
         var inst = this._instances[i];
         if (inst) {
-          inst.__setProperty('__key__', key, true);
-          inst.__setProperty(this.as, c.getItem(key), true);
+          inst.__key__ = key;
+          if (!inst.isPlaceholder && i < this.limit) {
```
kevinpschaaf: Is this possible?

src/lib/template/dom-repeat.html
```diff
@@ -497,7 +594,7 @@
         idx = end + 1;
       }
       // Insert instance at insertion point
-      this._instances.splice(idx, 0, this._insertRow(idx, key, pool));
+      this._insertPlaceholder(idx, key);
```
sorvell: Consider if we want to make this a real instance if we can here.

sorvell:
LGTM.

JeremybellEU:
The tests in this PR currently cause master builds too fail. The test suite chunked rendering failed 3 tests. The failing build is https://travis-ci.org/Polymer/polymer/builds/89713538

PR #2662, opened 2015-10-31
Fix parsing of custom properties with 'var' in value

### LINKED ISSUE #2660: Custom properties with 'var' in the value aren't applied
ex: `--my-font: 'Varela Round'` will not be applied because of the word `Var` in `Varela`. @sorvell :cry:

JeremybellEU:
(Note: the newline 'changes' are solved after #2484)

image

The screenshot shows the new behaviour. It only consumes lines with
`<css-property-name>:<anything not a ;>var(<anything not a ;>)<optional ; and \n>`

* It leaves the first line which was the original issue.
* It consumes the second line, which is a var assignment
* It consumes the third line which is preceded with a var assignment
* It leaves the fourth line, a property between assignments
* It consumes the whole fifth line, which does not have a new line between the two variable assignments
* It consumes the second part of the sixth line, which does not have a new line between the custom property and the variable assignment
* It consumes the first part of the seventh line, which does not have a new line between the variable assignment and the custom property

Fixes #2660


nazar-pc:
LGTM, :+1:

samccone:
:bow: this is amazing work @JeremybellEU thanks a million! 

R: @tjsavage

tjsavage:
This LGTM. Adding @sorvell to take a look.

azakus:
LGTM

PR #2663, opened 2015-11-01
Fix for BEM-like CSS selectors under media queries, fixes #2639

### LINKED ISSUE #2639: CSS class names with BEM modifiers not processed correctly in custom-style by Polymer when inside a media query
As a simple test case the following CSS inside shared-styles.css is imported as a custom-style:

```
.foo {
  width: 100px;
  height: 100px;
  background: #0f0;
}

@media screen and (min-width: 48rem) {
  .foo {
    background: #f00; 
  } 
}

.foo--bar {
  background: #00f;
}

@media screen and (min-width: 48rem) {
  .foo--bar {
    background: #000; 
  } 
}
```

After this is processed by Polymer (using version 1.2.0) it is output in the head like so:

```
.foo:not([style-scope]):not(.style-scope) {
  width: 100px;
  height: 100px;
  background: #0f0;
}

@media screen and (min-width: 48rem) {
  .foo:not([style-scope]):not(.style-scope) {
    background: #f00;
  }
}

.foo--bar:not([style-scope]):not(.style-scope) {
  background: #00f;
}

@media screen and (min-width: 48rem) {
  .foo--bar {
    background: #000; 
  }
}
```

Notice the `:not([style-scope]):not(.style-scope)` part is missing from `.foo--bar` within the media query, which means the rule with never be applied as it's specificity is too low to override the rule outside the media query.

nazar-pc:
Simple fix for #2639 + corresponding test case

### DIFF
--- a/src/lib/css-parse.html
+++ b/src/lib/css-parse.html
@@ -123,7 +123,7 @@ Polymer.CssParse = (function() {
     },
 
     _hasMixinRules: function(rules) {
-      return (rules[0].selector.indexOf(this.VAR_START) >= 0);
+      return rules[0].selector.indexOf(this.VAR_START) === 0;
     },
 
     removeCustomProps: function(cssText) {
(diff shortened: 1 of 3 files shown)


azakus:
LGTM, manually merged

samccone:
@nazar-pc coming in big :hand:

PR #2664, opened 2015-11-01
Added support for short unicode escape sequences, fixes #2650

### LINKED ISSUE #2650: unicode escape sequences in css not handled correctly
consider the following Polymer element:
```html
<dom-module id="model-viewer">
  <style>
    /* represents ':host 3d-model' */
    :host .\33 d-model
    {
      border: solid 1px #000;
      width: 100px;
      height: 100px;
    }
  </style>
  <template>
    <canvas id="3d-model"></canvas>
  </template>
  <script>
    var Viewer = Polymer
    ({
      is: 'model-viewer',
      factoryImpl: function(name)
      {
      },
      ready: function(){},
      attached: function(){},
      detached: function(){}
    });
  </script>
</dom-module>
```

When Polymer parses the element declaration, specifically the `style`, it incorrectly treats the space in this sequence `\33 d-model` as a real space, when in fact it terminates the unicode escape sequence `\33`. As a result it transforms the selector to: `model-viewer .\33.model-viewer d-model.model-viewer` instead of `model-viewer .\33 d-model.model-viewer` as expected.

nazar-pc:
Fixes #2650 + corresponding test case

### DIFF
--- a/src/lib/css-parse.html
+++ b/src/lib/css-parse.html
@@ -60,6 +60,7 @@ Polymer.CssParse = (function() {
       if (node.parent) {
         var ss = node.previous ? node.previous.end : node.parent.start;
         t = text.substring(ss, node.start-1);
+        t = this._expandUnicodeEscapes(t);
         // TODO(sorvell): ad hoc; make selector include only after last ;
         // helps with mixin syntax
         t = t.substring(t.lastIndexOf(';')+1);
@@ -87,6 +88,18 @@ Polymer.CssParse = (function() {
         }
       }
       return node;
+    },
+
+    // conversion of sort unicode escapes with spaces like `\33 ` (and longer) into
+    // expanded form that doesn't require trailing space `\000033`
+    _expandUnicodeEscapes : function(s) {
+      return s.replace(/\\([0-9a-f]{1,6})\s/gi, function() {
+        var code = arguments[1], repeat = 6 - code.length;
+        while (repeat--) {
+          code = '0' + code;
+        }
+        return '\\' + code;
+      });
     },
 
     // stringify parsed css.
--- a/test/unit/css-parse.html
+++ b/test/unit/css-parse.html
@@ -60,6 +60,27 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
     /* comment */
   </style>
+
+  <style id="short-escape-sequence">
+    .\33 d-model {
+      border-top: 3px solid red;
+    }
+    .\a33 d-model {
+      border-top: 3px solid red;
+    }
+    .\b333 d-model {
+      border-top: 3px solid red;
+    }
+    .\c3333 d-model {
+      border-top: 3px solid red;
+    }
+    .\d33333 d-model {
+      border-top: 3px solid red;
+    }
+    .\e33333d-model {
+      border-top: 3px solid red;
+    }
+  </style>
 <script>
 
   function sanitizeCss(text) {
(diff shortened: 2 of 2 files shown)


src/lib/css-parse.html
```diff
@@ -89,6 +90,12 @@
       return node;
     },
 
+    // conversion of sort unicode escapes with spaces `\33 ` into
+    // expanded form that doesn't require space `\000033`
+    _expandUnicodeEscapes : function (s) {
+      return s.replace(/\\([0-9]+)\s/g, '\\0000$1');
```
arthurevans: Thid is not correct for the general case. 

-  Should be `[0-9a-fA-F]` -- escapes are hexadecimal.
-  Hex number needs to be limited to six hex characters, otherwise you could match a valid, non-space-delimited escape, followed by letters and a meaningful space, like `\0000E9da `.
-  Last and most challenging issue: you can't assume the unicode escape is exactly two digits. 3,4,5 and 6 digit escapes followed by a space are valid (for 6-digit escapes, the space is optional, but we'd need to consume it here).

More information here:

http://www.w3.org/International/questions/qa-escapes#cssescapes

nazar-pc: @arthurevans, I wanted to avoid using functions initially.
Now it should be fine.

src/lib/css-parse.html
```diff
@@ -89,6 +90,21 @@
       return node;
     },
 
+    // conversion of sort unicode escapes with spaces like `\33 ` (and longer) into
```
samccone: I am not sure I follow what is going on here.. mind explaining?

nazar-pc: The problem is that CSS allows unicode escapes like `.\33 d-model` (because `.3d-model` seems to be invalid), which requires space after escape sequence in short form (2 hex digits in this example).
It is all good, but current CSS parser in Polymer is simple regexp based one, so space after escape sequence causes problems for it, see https://github.com/Polymer/polymer/issues/2650 for background. Make parser that covers all cases would make everything unnecessary more complex, so I've chosen another way.

This function converts short escape sequence into long (namely `\33 ` into `\000033`, 6 hex digits), which doesn't actually require space after it and thus doesn't cause any problems for Polymer parser, while functioning in exactly the same way (see tests added).

src/lib/css-parse.html
```diff
@@ -89,6 +90,21 @@
       return node;
     },
 
+    // conversion of sort unicode escapes with spaces like `\33 ` (and longer) into
+    // expanded form that doesn't require space `\000033`
+    _expandUnicodeEscapes : function(s) {
+      return s.replace(/\\([0-9a-f]{1,6})(\s?)/gi, function(source, code, whitespace) {
```
arthurevans: I don't think we need to modify escape sequences _unless_ they have a trailing space.

nazar-pc: You are right, I'll modify it.

nazar-pc: @arthurevans, done

src/lib/css-parse.html
```diff
@@ -89,6 +90,18 @@
       return node;
     },
 
+    // conversion of sort unicode escapes with spaces like `\33 ` (and longer) into
+    // expanded form that doesn't require trailing space `\000033`
+    _expandUnicodeEscapes : function(s) {
+      return s.replace(/\\([0-9a-f]{1,6})\s/gi, function() {
```
JeremybellEU: `\s` matches more than a trailing space (tabs, spaces and new-lines). Is that intentional? If you only want to match a trailling space, it should be `/\\([0-9a-f]{1,6}) /gi`

nazar-pc: Actually yes: http://www.w3.org/TR/CSS2/syndata.html#escaped-characters
Together with #2671 it should be solid fix for mentioned issue.

azakus:
LGTM

PR #2671, opened 2015-11-03
Fix for multiple consequent spaces present in CSS selectors, fixes #2670

### LINKED ISSUE #2670: Multiple spaces between selector classes
Currently it seems like there is an issue where multiple spaces between classes will cause that selector to not be applied.
```
.parent .child-0 {}
.parent  .child-1 {} // double space between classes in selector will not be applied to element
```
Once fixed of course things are fine but was quite painful to track down.

nazar-pc:
Fixes #2670 + corresponding test case

### DIFF
--- a/src/lib/css-parse.html
+++ b/src/lib/css-parse.html
@@ -60,6 +60,7 @@ Polymer.CssParse = (function() {
       if (node.parent) {
         var ss = node.previous ? node.previous.end : node.parent.start;
         t = text.substring(ss, node.start-1);
+        t = t.replace(this._rx.multipleSpaces, ' ');
         // TODO(sorvell): ad hoc; make selector include only after last ;
         // helps with mixin syntax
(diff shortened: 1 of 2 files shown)


src/lib/css-parse.html
```diff
@@ -162,6 +163,7 @@
       mixinApply: /@apply[\s]*\([^)]*?\)[\s]*(?:[;\n]|$)?/gim,
       varApply: /[^;:]*?:[^;]*var[^;]*(?:[;\n]|$)?/gim,
       keyframesRule: /^@[^\s]*keyframes/,
+      multipleSpaces: /\s+/g
```
JeremybellEU: A bit nitpicky, but as pointed out in the other PR, `\s` matches whitespace. In my eyes it would be more logical to name this regex `whitespace`.

Same for the id name and description in the testcase.

azakus:
LGTM, manually fixed conflict.

PR #2673, opened 2015-11-03
Fix using value$ on input element

### LINKED ISSUE #2666: Failed to execute 'setAttribute' on 'Element': 'value$' is not a valid attribute name.
```
Uncaught InvalidCharacterError: Failed to execute 'setAttribute' on 'Element': 'value$' is not a valid attribute name.
```

Code:
https://github.com/liuwenchao/aha-table/blob/master/src/aha-table.html#L280

The error will disappear if I remove L280, L286. L292, L298. Any idea?

Refer to https://github.com/liuwenchao/aha-table/issues/25

JeremybellEU:
The error was thrown because of a specific edge case for `<input value$="">`. This check was introduced in #1589 for #1578 and latest modified on https://github.com/Polymer/polymer/commit/a300862f5d3d34b01f09f7290eb8540f501c424d#diff-870cfe064a36c23b69e8da891fdff6beR299

For both `value$` and `value` `name === 'value'`. However, if it was originally `value$` it should not be executed. Therefore the check should be changed to `origName === 'value'`.

Fixes #2666

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -313,7 +313,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // `value` attribute if it previously had a value (can't
         // unconditionally set '' before removing since attributes with `$`
         // can't be set using setAttribute)
-        if (node.localName == 'input' && name == 'value') {
+        if (node.localName === 'input' && origName === 'value') {
           node.setAttribute(origName, '');
         }
         // Remove annotation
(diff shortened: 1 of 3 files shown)


jmcgdz:
:+1:

justjico:
:+1:

liuwenchao:
:+1:

varun-aggarwal:
:thumbsup:

sky-dev:
:+1:

kevinpschaaf:
LGTM.  :+1: 

Thanks @JeremybellEU!

PR #2676, opened 2015-11-03
[API Proposal] Data Validations

BLamy:
Many web-components expose an external API. Configuring your element's attributes incorrectly typically results in your web component not working (usually a white screen of death). This can be relatively hard to debug if documentation is lacking. This PR proposes an API enhancement for supporting data validations in polymer. The goal is to be able to catch errors sooner and make for easier to debug applications. This is done by adding your data validations directly in the properties info, which gives you one centralized place to look (other then docs) to figure out the required structure for a property.

This solution adds only about 50-75 lines to polymer and maintains backwards compatibility given that it is just progressive enhancement for your properties. 

Data validations are added via `rules` and applied directly inside of the property's definition.
```
Polymer({
      is:'x-foo',
      rules: {
        maxLength: function(val, arg) {
          return (val.length <= arg);
        },
        minLength: function(val, arg) {
          return (val.length >= arg);
        }
      },
      properties: {
        name: {
          type: String,
          maxLength: 5,
          minLength: 3,
          notify: true
        }
      }
});
```

I'm no @robdodson but I attempted to create some videos explaining basic use cases. 

# Basic string rules 
[Basic string rules](http://www.youtube.com/watch?v=9hEoSML4lPA)

# Using behaviors and external libraries
[Using behaviors and external libraries](http://www.youtube.com/watch?v=hKKwdzEREBA)

# Enum use case
[Enum use case](http://www.youtube.com/watch?v=dBjcpvLZjj8)

# property-name-invalid Notification 
[property-name-invalid Notification](http://www.youtube.com/watch?v=GgahM8FEKrY)


BLamy:
I signed it!

PR #2677, opened 2015-11-03
[API Proposal] Property Validations

BLamy:
Many web-components expose an external API. Configuring your element's attributes incorrectly typically results in your web component not working (usually a white screen of death). This can be relatively hard to debug if documentation is lacking. This PR proposes an API enhancement for supporting data validations in polymer. The goal is to be able to catch errors sooner and make for easier to debug applications  (Can provide strong static analysis of when using attributes improperly in HTML). This is done by adding your data validations directly in the properties info, which gives you one centralized place to look (other then docs) to figure out the required structure for a property.

This solution adds only about 50-75 lines to polymer and maintains backwards compatibility given that it is just progressive enhancement for your properties. 

Data validations are added via `rules` and applied directly inside of the property's definition.
```
Polymer({
      is:'x-foo',
      rules: {
        maxLength: function(val, arg) {
          return (val.length <= arg);
        },
        minLength: function(val, arg) {
          return (val.length >= arg);
        }
      },
      properties: {
        name: {
          type: String,
          maxLength: 5,
          minLength: 3,
          notify: true
        }
      }
});
```

I'm no @robdodson but I attempted to create some videos explaining basic use cases. 

# Basic string rules 
[Basic string rules](http://www.youtube.com/watch?v=9hEoSML4lPA)

# Using behaviors and external libraries
[Using behaviors and external libraries](http://www.youtube.com/watch?v=hKKwdzEREBA)

# Enum use case
[Enum use case](http://www.youtube.com/watch?v=dBjcpvLZjj8)

# property-name-invalid Notification 
[property-name-invalid Notification](http://www.youtube.com/watch?v=GgahM8FEKrY)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 bower_components
 node_modules
 dist
+.DS_Store
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -163,7 +163,27 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           return this.__data__[property];
         }
       };
+
+
       var setter = function(value) {
+        // TODO: Am I allowed to call to micro from the core lib like this?
+        // Doing a soft check for now
+        if (Polymer.Base.validateRules && this.getPropertyInfo) {
+          if (Polymer.Base.validateRules(this, property, value)) {
+            this._propertySetter(property, value, effects);
+            // If a property fails validation.
+            // Then a user tries to reassign the last valid value, Fire a change event
+            var old = this.__data__[property];
+            if (effects && old && old === value && this.properties[property] && this.properties[property].isInvalid) {
+              this._notifyChange(property);
+            }
+          } else {
+            if (this.properties[property]){
+              this.properties[property].isInvalid = true;
+            }
+          }
+          return;
+        }
         this._propertySetter(property, value, effects);
       };
       // ReadOnly properties have a private setter only
--- a/test/unit/rules-elements.html
+++ b/test/unit/rules-elements.html
@@ -1,0 +1,192 @@
+<script src="https://cdnjs.cloudflare.com/ajax/libs/validator/4.0.5/validator.js"></script>
+<script>
+var Validator = Validator || {};
+Validator.js = Validator.js || {rules: validator};
+</script>
+
+<dom-module id="external-library">
+  <template>
+    <style>
+      :host {
+        display: block;
+      }
+    </style>
+    <h1>Card Number:</h1>
+    <h3 id="ccLabel">{{cardNumber}}</h3>
+    <input id="ccInput" value="{{cardNumber::input}}"></input>
+  </template>
+  <script>
+    Polymer({
+      is: 'external-library',
+
+      properties: {
+        cardNumber: {
+          type: String,
+          notify: true,
+          isCreditCard: true
+        }
+      },
+
+      behaviors: [
+        Validator.js
+      ],
+
+      listeners: {
+        'card-number-changed': '_cardNumberChanged',
+        'card-number-invalid': '_cardNumberFailedValidation'
+      },
+
+      _cardNumberChanged: function () {
+        this.$.ccInput.style['border-color'] = 'green';
+      },
+
+      _cardNumberFailedValidation: function (event) {
+        this.$.ccLabel.innerHTML = 'Failed: ' + event.detail.ruleName;
+        this.$.ccInput.style['border-color'] = 'red';
+      }
+    });
+  </script>
+</dom-module>
+
+
+<dom-module id="simple-min-max-rules">
+  <template>
+    <style>
+      :host {
+        display: block;
+      }
+    </style>
+    <h1>Name:</h1>
+    <h3 id="nameLabel">{{name}}</h3>
+    <input id="nameInput" value="{{name::input}}"></input>
+  </template>
+  <script>
+    Polymer({
+      is: 'simple-min-max-rules',
+
+      rules: {
+        maxLength: function(val, arg) {
+          return (val.length <= arg);
+        },
+        minLength: function(val, arg) {
+          return (val.length >= arg);
+        }
+      },
+
+      properties: {
+        name: {
+          type: String,
+          maxLength: 5,
+          minLength: 3,
+          notify: true
+        }
+      },
+
+      listeners: {
+        'name-changed': '_nameChanged',
+        'name-invalid': '_nameFailedValidation'
+      },
+
+      _nameChanged: function () {
+        this.$.nameInput.style['border-color'] = 'green';
+      },
+
+      _nameFailedValidation: function (event) {
+        this.$.nameLabel.innerHTML = 'Failed: ' + event.detail.ruleName;
+        this.$.nameInput.style['border-color'] = 'red';
+      }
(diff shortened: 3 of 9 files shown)


BLamy:
I signed it!

BLamy:
I signed it!

laco0416:
:+1:

azogheb:
:+1: on this. Much needed enhancement!

zoechi:
Great work, love it!

src/lib/bind/accessors.html
```diff
@@ -163,7 +163,27 @@
           return this.__data__[property];
         }
       };
+
+
       var setter = function(value) {
+        // TODO: Am I allowed to call to micro from the core lib like this?
```
BLamy: Am I allowed to call to to micro from the core lib like this? I noticed that properties is also in micro so is it safe to assume that micro is always imported? Furthermore, is this a feature that even makes sense being in micro? It doesn't seem like 'bare minimum custom element sugaring' it seems more like it should be a part of polymer standard but then i'm calling to standard all the way from the core lib.

JeremybellEU:
In your proposal, the `rules` object is a collection of functions. Wouldn't it be easier to instead of searching for rules in a seperate object, to directly calls on `this`? That way your validation can also be reused for pre-emptively checking for example.

I can imagine a form element where validation is required. You then add a validation rule onto a property which sends a request. But instead of waiting it to fail, you can pre-emptively link the validation to the input on-insert (for example)

BLamy:
I personally think validation rules should be kept separate from `this`. You can still access them all through `this.rules`. Rules are typically highly reusable and therefore, I feel like most people would be implementing them using behaviors. Because of this, I think it's best to put rules in their own namespace to avoid naming collisions. 

If something like this where to be implemented, Google would probably have to publish their own element line full of behaviors for common validation use cases. (`Chlorine behaviors`?). That way people have a common toolchain to use for data validations instead of pulling code from a million different sources. 

I've been using sails on the backend and I find that about 80-85% of my use cases have been handled by this list. 
http://sailsjs.org/documentation/concepts/models-and-orm/validations

Doing this will allow us to do some pretty elaborate static analysis on our HTML.

Imgur

.gitignore
```diff
@@ -1,3 +1,4 @@
 bower_components
 node_modules
 dist
+.DS_Store
```
nazar-pc: This is your local environment, ignore it locally, do not add to `.gitignore`

test/unit/rules-elements.html
```diff
@@ -0,0 +1,192 @@
+<script src="https://cdnjs.cloudflare.com/ajax/libs/validator/4.0.5/validator.js"></script>
```
nazar-pc: Bad idea to include it from CDN, there should be possibility to test everything completely offline.
Also what is the need to include something from CDN at all?
Unit tests should be simple, short and self-contained.

test/unit/rules.html
```diff
@@ -0,0 +1,158 @@
...
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
+
+
+
+  <script src="../../../webcomponentsjs/webcomponents-lite.js"></script>
+  <script src="../../../web-component-tester/browser.js"></script>
+  <link rel="import" href="../../polymer.html">
+  <script src="/bower_components/test-fixture/test-fixture-mocha.js"></script>
+  <link rel="import" href="/bower_components/test-fixture/test-fixture.html">
+  <link rel="import" href="/bower_components/iron-test-helpers/iron-test-helpers.html">
```
nazar-pc: Why do you need those 3 things?
Again, this is unit test, it should be simple, short and self-contained.
WCT already includes everything you'll likely NEED to write unit tests.

test/unit/rules-elements.html
BLamy: Since this was more of a proposal then a serious PR. I was shooting to make the test more like code samples then test. Realistically, If I where writing a test for this, I wouldn't need to use a lib as a normal behavior would have been fine.. However, the goal was to demo how we could utilize existing validation frameworks to beef up polymer.

test/unit/rules.html
BLamy: Is using text-fixture not generally considered good practice? I didn't use it because I needed it; I used it because I thought it was recommended. I did notice that all the other test weren't using it but I just figured the test where written pre-dating test-fixture.

.gitignore
BLamy: :+1:

MeTaNoV:
Ping! :+1:

bedeoverend:
:+1:

robdodson:
We think this is a really cool idea however we're also very concerned about adding additional file size to our elements and even more API to maintain. We don't currently plan to include this PR at this time because it doesn't align with the current priorities of making components as small and fast as possible.

Having said that, this is obviously a popular idea with some folks so it may be worth exploring if it can be turned in to an optional behavior that you house under your own GitHub repo. An ecosystem of behaviors and add-ons would be awesome.

I'd suggest in the future, to save yourself time, please open an issue discussing your ideas before putting together a PR. We really don't want to disappoint folks and I know it's frustrating to spend hours on something which the team doesn't merge.

zoechi:
This would be fine. Just ensure Polymer provides the necessary hooks to make it easy to implement this in behaviors.

BLamy:
@robdodson This was actually a direct result of your data binding video. This was just my project to understand data binding better. I thought it was kind of cool and thought I would share. Not frustrated at all. In fact, I would actually prefer if google focused on things such as the carbon elements, designer and tooling for the current api (to name a few). 

As far as keeping my own branch goes I don't know If I have time to maintain a branch of something as fast moving as polymer and provide enough behaviors/add-ons to make it a practical progressively enhanced alternative. So unless someone here feels like taking on the challenge.

BLamy:
@zoechi I think he's suggesting that someone maintain their own branch of polymer which supports this. Sort of like the 'typescript of polymer'. 
I don't think google is planning on changing any hooks.

zoechi:
@blamy I don't see how to interpret it this way. He mentioned behavior and I assume he means something like iron-behaviors

robdodson:
No not suggesting you maintain your own branch of Polymer. I'm curious if
we could make this something that a behavior could hook into

JeremybellEU:
Maybe a new issue can be created to discuss a proposal for hooks for `properties`, based on this one? I have some other ideas that might could benefit from such extension points.

Something along the lines of `pre`, `transform`, `post` hooking of the setters? My 2 cents.

zoechi:
Might also be useful to delay propagation of values until animations are done when value changes are supposed to invoke animations.

BLamy:
@zoechi , I guess i'm just a little confused because I don't know how that differs from what I have here. All this is doing is proposing an API for that hook. I don't package any validations or anything. You need to mixin your own validations using behaviors.

zoechi:
@BLamy I didn't look to close at what you implemented. I guess you should rather address this comment at @robdodson

robdodson:
It looked to me like the changes were specifically looking for the existence of a validateRules subproperty. I think any sort of hooking would have to be more generic than that (kind of like what @JeremybellEU was getting at, general pre/post hooks). Apologies if I've misread the PR, that was just my take on it.

PR #2684, opened 2015-11-04
Fix Gestures when using SD polyfill

### LINKED ISSUE #2641: Gestures.handleNative sometimes throws "Unable to get property mousedown of undefined or null reference" in IE11
When using `<paper-dropdown-menu>` and selecting an item in IE11 I get the error "Unable to get property mousedown of undefined or null reference". The error originates from `Gestures.handleNative`. The statement `node[GESTURE_KEY]` is undefined. If i add the following line 
```javascript
    if (!gobj) gobj = unwrap(node)[GESTURE_KEY];
```
everything works perfect. So the problem seems to be that the node is not wrapped when passed to `Gestures.add` but is when the event is received.

azakus:
Test with SD polyfill and native ShadowDOM
Requires webcomponents/webcomponentsjs#434

Fixes #2641

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -199,8 +199,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     handleNative: function(ev) {
       var handled;
       var type = ev.type;
-      var node = ev.currentTarget;
+      var node = wrap(ev.currentTarget);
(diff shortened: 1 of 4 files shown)


azakus:
@sorvell or @kevinpschaaf PTAL

sorvell:
LGTM :+1:

PR #2686, opened 2015-11-05
Micro-optimizations to Polymer's core performance

sorvell:
Lots of little things that add up to ~20% savings to element construction.


kevinpschaaf:
LGTM

Issue #1389, opened 2015-04-09
Shady DOM doesn't correctly parse custom property rules.


philipwalton:
Consider the following element:

```html
<dom-module id="my-element">
  <style>
    :host {
      x--color: red;
    }
    p {
      color: var(--color);
    }
  </style>
  <template>
    <div>
      <p>Some paragraph text...</p>
    </div>
  </template>
</dom-module>

<script>
  Polymer({
    is: "my-element",

    // The custom properties shim is currently an opt-in feature
    enableCustomStyleProperties: true
  });
</script>

<my-element></my-element>
```

The text should not be red because `x--color: red` is not a valid property/value combination, but in Chrome it is red.

PR #2689, opened 2015-11-05
Fix parsing of minimized css output

### LINKED ISSUE #2483: CSS parser: incorrectly handing polybuild output
When Polybuild crushes element CSS and removes whitespace, custom property rules end up running into each other. Unfortunately, this causes our parser's regexing to start ignore rules. The issue can be seen when you have two `--var` in a row (one won't do it).

    top: 400px;--paper-fab-background: var(--paper-green-500);--paper-fab-keyboard-focus-background: var(--paper-green-700);

putting as space in between properties works:

    top: 400px; --paper-fab-background: var(--paper-green-500); --paper-fab-keyboard-focus-background: var(--paper-green-700);

http://jsbin.com/wayoholaxi/edit?html,output - you'll see `top: 400px` is ignored

cc @azakus

JeremybellEU:
The problem was the expression `[\s;]` in the first non-capturing group. This incorrectly consumed the `;` from a previous css rule. 

image

This check was introduced at https://github.com/Polymer/polymer/commit/6433f626c02c1037dbe30f755b67f018cddf4ef6 for #1389.

Fixes #2483

### DIFF
--- a/src/lib/css-parse.html
+++ b/src/lib/css-parse.html
@@ -171,7 +171,7 @@ Polymer.CssParse = (function() {
     _rx: {
       comments: /\/\*[^*]*\*+([^/*][^*]*\*+)*\//gim,
       port: /@import[^;]*;/gim,
-      customProp: /(?:^|[\s;])--[^;{]*?:[^{};]*?(?:[;\n]|$)/gim,
+      customProp: /(?:^[^;\-\s}]+)?--[^;{]*?:[^{};]*?(?:[;\n]|$)/gim,
       mixinProp: /(?:^|[\s;])?--[^;{]*?:[^{;]*?{[^}]*?}(?:[;\n]|$)?/gim,
       mixinApply: /@apply[\s]*\([^)]*?\)[\s]*(?:[;\n]|$)?/gim,
       varApply: /[^;:]*?:[^;]*?var\([^;]*\)(?:[;\n]|$)?/gim,
--- a/test/unit/custom-style.html
+++ b/test/unit/custom-style.html
@@ -235,6 +235,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           --variable-property-after-property: 3px;
           --variable-property-after-assignment: 4px;
           --variable-property-before-assignment: 5px;
+          --variable-into-first-variable: 9px;
+          --variable-into-second-variable: 10px;
+          --variable-into-third-variable: 11px;
         }
       </style>
       <child-variable-with-var id="child"></child-variable-with-var>
@@ -251,7 +254,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           --variable-between-properties: 6px;
           background-color: var(--variable-property-before-property); padding-top: var(--variable-property-after-property);
           --variable-assignment-before-property: 7px; padding-bottom: var(--variable-property-after-assignment);
-          padding-left: var(--variable-property-before-assignment);--variable-assignment-after-property: 8px
+          padding-left: var(--variable-property-before-assignment);--variable-assignment-after-property: 8px;
+          top: 12px;--variable-from-other-variable: var(--variable-into-first-variable);--variable-from-another-variable: var(--variable-into-second-variable); --variable-from-last-variable: var(--variable-into-third-variable);
         }
       </style>
       <child-of-child-with-var id="child"></child-of-child-with-var>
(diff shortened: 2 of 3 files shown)


samccone:
R: @tjsavage for slotting

tjsavage:
Adding @azakus who has been taking a look

nazar-pc:
There was somewhat similar fix in 883aa5c, just did PR to this PR to improve regexp a bit: https://github.com/JeremybellEU/polymer/pull/1

JeremybellEU:
Yes I have tried that solution, but that fails on detecting incorrect variable names like `name--incorrect`. It will leave up the `name`, causing weird CSS to happen. Therefore I decided to explicitly capture more characters in front, in order to circumvent the 'failures'.

You can try this yourself by adding the incorrect variable and inspecting using chrome dev tools when running the unit test. The element should have a CSS rule with warning, indicating invalid CSS is applied to the element.

However, if this CSS weirdness is more desired than capturing everything, your patch seems fine. I will leave that for the core team to decide.

nazar-pc:
Well, I've tried that and unfortunately both our regexps are not good, they start capturing incorrect CSS property - in your case completely (doesn't harm, incorrect property), in my case partially (might override some custom properties?).
I'll try to find something that will cover all cases.

JeremybellEU:
What I tried (and did work together with my regex) was throwing a SyntaxError upon catching incorrect CSS variable names. However, this is against the CSS spec if I am not mistaken, but at least clearly notifies the user what is going on. It therefore failed [this testcase](https://github.com/Polymer/polymer/blob/master/test/unit/styling-cross-scope-var.html#L523).

To achieve this I changed https://github.com/Polymer/polymer/blob/master/src/lib/css-parse.html#L150 to

```js
.replace(this._rx.customProp, function(match) {
  if (match.trim().indexOf(this.VAR_START) !== 0) {
    throw new SyntaxError("Invalid CSS variable defined '" + match + "'");
  }
  return '';
})
```

nazar-pc:
I think throwing exception is a good idea, never seen `--` in the middle of any css property.

JeremybellEU:
I will squash the commits tomorrow to also fix the cla

nazar-pc:
@googlebot, what is wrong?) I'm OK with that, confirm or whatever work you're expecting

azakus:
googlebot gets confused if the commiter email does not match the author email

JeremybellEU:
Commits have been squashed and the PR is again available for review @azakus . If/when this PR is merged, I shall create an issue to discuss the addition of the `SyntaxError`.

azakus:
LGTM, @sorvell for review

ebidel:
Ping

rubenstolk:
It's like that one month before `Spectre`, can we see `polybuild` soon yes or no :)

rubenstolk:
Any update on this @sorvell?

JeremybellEU:
PR rebased onto current master.

JeremybellEU:
@azakus for some reason the PR did not fail, but when merged into master Travis complained on Safari: https://travis-ci.org/Polymer/polymer/builds/94970278#L1630

I don't have safari myself, but could you confirm/deny the tests break in that browser? For some reason the tests in this PR were not run in safari, as seen at https://travis-ci.org/Polymer/polymer/builds/94621570#L1557

azakus:
Reverting for now, moved into branch `JeremybellEU-fix-min-css`

kevinpschaaf:
@JeremybellEU FYI, since we moved our CI runner to Travis, 3rd-party PR's are only tested against browsers that we can run on the Travis VM (Firefox & Chrome); the full browser matrix is only run on Sauce for pushes to non-forked branches, due to Travis's lack of support for a secured test configuration needed to keep Sauce keys private (something we're hoping to work with them to fix).

nazar-pc:
@kevinpschaaf, in fact Travis have support for private environmental variables, configs in `.travis.yml` and more. I have such key for decrypting certificate to create digital signatures on releases in one of my projects.
Here are docs:
* https://docs.travis-ci.com/user/environment-variables/#Encrypted-Variables
* https://docs.travis-ci.com/user/encrypting-files/
* https://docs.travis-ci.com/user/encryption-keys/

azakus:
@nazar-pc Yep, we use encrypted variables for the sauce configuration.https://github.com/Polymer/polymer/blob/master/.travis.yml#L18-L21

kevinpschaaf:
@nazar-pc Travis will only decrypt such keys for first-party branches, not for forks, since a malicious forker could echo the decrypted keys to the log by modifying the `.travis.yml` file to do just that.

From the page you linked: 
> Please note that encrypted environment variables are not available for pull requests from forks.

When I say "secured test configuration", I mean we'd like to be able to specify the `.travis.yml` file in a secured location (e.g. not in the repo, which the contributor can change in their fork, but on the travis website, for example).

PR #2692, opened 2015-11-06
Fix for custom styles parsing.

### LINKED ISSUE #2381: Big CSS causes styling issues when elements combined together
As usual, I've spent many long hours in debugger, but yet don't know why it happens.
Where are 2 cases:
* few elements imported from separate files, styling works properly
* put everything in the same order into one file - styling becomes partially broken

It was not possible to put into jsfiddle/jsbin, so I've reduced my real case and put it into repository so that you can play with it: https://github.com/nazar-pc/Polymer-style-bug

Live demo:
* https://nazar-pc.github.io/Polymer-style-bug/good.html - separate files, input styled properly
* https://nazar-pc.github.io/Polymer-style-bug/bad.html - single import, input styled incorrectly

The problem disappears when you remove big chunk of CSS that comes from Font Awesome, it seems to cause an issue.

Problem appears in Firefox (stable & nightly) and doesn't in Chromium 44.

Please, find time to investigate this.

nazar-pc:
When custom style is second element in HTML import after script with `Polymer()` call.
Fixes #2381

### DIFF
--- a/src/lib/custom-style.html
+++ b/src/lib/custom-style.html
@@ -11,7 +11,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="style-transformer.html">
 <link rel="import" href="style-defaults.html">
 
-<!-- 
+<!--
 
 The `custom-style` extension of the native `<style>` element allows defining styles
 in the main document that can take advantage of several special features of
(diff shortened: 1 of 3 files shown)


sorvell:
Thanks for doing so much work on this. The fix looks simple but I'd like to understand the nature of the breakage before merging. 

Do you know if the issue is isolated to using the Shadow DOM polyfill?

nazar-pc:
No, it happens with WebComponenst lite and Shady DOM as well (in fact I've debugged it in such configuration to simplify stacktrace). It is actually because of how HTML Imports polyfill works. I guess HTML Imports polyfill will require to parse imports into set of root elements and insert them one by one in proper order manually instead of all together in order to mimic native behavior entirely, but it seems to overcomplicate polyfill's code base a lot and doesn't worth it.

PR #2696, opened 2015-11-06
Ensure parent node exists when stamping. Fixes #2685.

### LINKED ISSUE #2685: dom-if throws if detached before instance is stamped
Repro: http://jsbin.com/xomime/edit?html,console,output

When dom-if is attached, it queues an async task to append its template content to its parent.  If before the quited task runs it then gets detached, it has no parentNode and attempting to append throws.

kevinpschaaf:


sorvell:
LGTM :+1:

PR #2705, opened 2015-11-09
Fix compound bindings with braces in literals

### LINKED ISSUE #2704: Curly braces within Polymer bindings stopped working with Polymer 1.2
Our translation system uses placeholders like `{0}` for interpolation, which stopped working now with Polymer 1.2, where the binding string is not evaluated.
Consider the following:

```html
<template>
  <span>{{ _("Hello {0}", userName) }}</span>
</template>
```

In Polymer 1.0, this results in "Hello user", but in Polymer 1.2, the whole expression is printed verbatim, i.e. `{{ _("Hello {0}", userName) }}`.

JeremybellEU:
The problem in this regex was the `[^}\]]*` which also triggers on a single `}`. Therefore we have to check for a group of two `}` or `]`. This is now changed to a negative look-ahead `?!` which checks for `\}\}` and `\]\]`. After that the rest of the matches are non-greedy, to prevent double consuming of `{{binding1}}{{binding2}}` to be translated into a binding with variable name `binding1}}{{binding2`.

I also escaped all `}` and `]` to force matching on literal characters.

image

Fixes #2704

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -88,7 +88,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           this._parseElementAnnotations(node, list);
     },
 
-    _bindingRegex: /([^{[]*)({{|\[\[)([^}\]]*)(?:]]|}})/g,
+    _bindingRegex: /([^{[]*)(\{\{|\[\[)(?!\}\}|\]\])(.+?)(?:\]\]|\}\})/g,
 
     // TODO(kschaaf): We could modify this to allow an escape mechanism by
     // looking for the escape sequence in each of the matches and converting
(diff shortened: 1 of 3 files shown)


samccone:
wow.. you are too fast @JeremybellEU :)

samccone:
cc @kevinpschaaf @ebidel for :eyes:

kevinpschaaf:
LGTM :+1:

david-saslawsky:
@JeremybellEU many thanks for the fix.


Shouldn't the same solution be applied to the `([^{[]*)` at the beginning ? Not very likely I know, but I'm thinking of cases like this one:
````
<span>[YES]/[NO] {{computeCompound('world', 'username ', 'Hello {0} ')}}</span>
````

Also, how does the negative lookahead works ? I might be missing something but it seems to me that the regexp works fine without it because of the non greedy match on `(.+?)`:
````
/([^{[]*)(\{\{|\[\[)(.+?)(?:\]\]|\}\})/g
````

Anyway, thanks again. I can remove ugly workarounds in my application now :+1:

PR #2707, opened 2015-11-09
Fix templatizer under the template polyfill. forceDocumentUpgrade

garlicnation:
was interacting problematically with un-prepped HTMLTemplateElements.
By forcing a bootstrap before the document upgrade, we sidestep this issue.

### DIFF
--- a/src/lib/dom-module.html
+++ b/src/lib/dom-module.html
@@ -98,12 +98,16 @@
   document.registerElement('dom-module', DomModule);
 
   function forceDocumentUpgrade() {
+    var script = document._currentScript || document.currentScript;
+    var doc = script && script.ownerDocument || document;
+    if (!doc) {
+      return;
+    }
+    if (HTMLTemplateElement && HTMLTemplateElement.bootstrap) {
+      HTMLTemplateElement.bootstrap(doc);
+    }
     if (cePolyfill) {
-      var script = document._currentScript || document.currentScript;
-      var doc = script && script.ownerDocument || document;
-      if (doc) {
-        CustomElements.upgradeAll(doc);
-      }
+      CustomElements.upgradeAll(doc);
     }
   }
 
--- a/src/mini/template.html
+++ b/src/mini/template.html
@@ -33,12 +33,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           'must not be a type-extension, found', this._template,
           'Move inside simple <template>.'));
       }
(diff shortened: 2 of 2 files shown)


azakus:
We should be able to remove this code then: https://github.com/Polymer/polymer/blob/master/src/mini/template.html#L36-L42

azakus:
LGTM with nit to remove the leftover template bootstrap code in `_prepTemplate`

azakus:
Notes for reasoning:

The `HTMLTemplateElement` polyfill will not run until `DOMContentLoaded`.
The `HTMLImports` Polyfill works around this by upgrading templates in each document before asking the `CustomElements` polyfill to observe the document and allow for element upgrades.

However, when using vulcanize with all the inlining flags, all elements will upgrade before the template polyfill runs, because of the call to `forceDocumentUpgrade` in the element upgrade flow. This can break `dom-repeat` and a few other elements that expect their template to have a `content` document.

This patch ensures that all templates in the document will be upgraded before all of the elements upgrade, just as in the standard use with HTMLImports.

kevinpschaaf:
Reviewed along with @sorvell.

Making the fix in `dom-module`'s `forceDocumentUpgrade` isn't the best place, since ideally elements that are known to not have `dom-module`'s (like `dom-repeat`) wouldn't try to look for a `dom-module` in the first place, fail, and call that function.

I'm inclined to instead add specific code to bootstrap template extension elements in Polymer's startup flow, ala _add_ this to `_prepTemplate` (leaving the existing code as-is):

```js
if (this.localName == 'template') {
  HTMLTemplateElement.decorate(this);		
  HTMLTemplateElement.bootstrap(this.content);		
}
```

This issue highlights a potentially serious performance issue with the way `forceDocumentUpgrade` does an expensive tree walk each time an element that intentionally has no `dom-module` is registered; will open separate issue to address.

kevinpschaaf:
Per discussion, consider moving decorate & bootstrap code to CustomElements polyfill: https://github.com/webcomponents/webcomponentsjs/blob/master/src/CustomElements/upgrade.js#L60

azakus:
I'm not sure I follow. Do you want the Template polyfill to actually use the CustomElement machinery?

azakus:
Just to be 100% clear, this PR should be closed and we will proceed with webcomponents/webcomponentsjs#439 right?

kevinpschaaf:
Yes, that's the plan.

Issue #1734, opened 2015-06-03
Consider annotator differentiating between `[[3, 4]]` and `[[binding]]`


sjmiles:
Currently `[[` characters trigger binding logic immediately, even if the expression is actually JSON data.

atotic:
Why is there no warning when binding logic encounters: '[["Jan", 31],["Feb", 28],["Mar", 31]]'. It is not parseable as a binding, the code must swallow an error quietly somewhere.

zikes:
I think I found the offending check at https://github.com/Polymer/polymer/blob/master/src/lib/annotations/annotations.html#L93

I think the fix would be to check that the third character is either an underscore or a letter, as those are the only valid starting characters for bindable properties or computable functions, right?

SmokyBob:
@zikes I think checking the 3rd character is not enough, for example 
> '[[true, "Permission1"],[false, "Permission2"]]'

Will still fail to work properly.
Perhaps, checking the presence of a comma in the remaining text might be enough?
Trying to `JSON.parse` the remaining string I think would be the safes, but I think it's too detrimental for performance

zikes:
@SmokyBob I forgot about booleans, that throws a real wrench into that idea.  Unfortunately that also rules out the comma check, as `[[true]]` is still valid JSON.

I think a `JSON.parse` may actually be necessary, but could be a last-resort check after anything else that might be more certain and performant.

zikes:
I just remembered that computed annotations can have multiple parameters, which would also contain commas.

wq9:
You can solve this by adding a space between `[[`. For example, 

```attribute='[ ["Jan", 31],["Feb", 28],["Mar", 31] ]'```

Now, attribute contains an array instead of data binding.

SmokyBob:
@zikes you are right, forgot about that computed properties can have multiple values (or an array as the parameter).

I like @wq9 idea of using {[ for the one way binding, but it will not solve the issue; I can pass an object as parameter where the first item is an array.

zikes:
`{[` normally not valid JSON notation as the property name would have to come first.

Unfortunately `{[` is a very breaking change.  Even if there is a significant performance hit, better to `JSON.parse` first and try to optimize from there.

SmokyBob:
@zikes got me :) answering before coffee was not a good idea.

[...]

PR #2711, opened 2015-11-10
Fix compound binding regex when a binding is preceded with text with { or ]

JeremybellEU:
As outlined by @david-saslawsky in [his comment](https://github.com/Polymer/polymer/pull/2705#issuecomment-155527368) another edge-case for the regex is
`[YES]/[NO] {{computeCompound('world', 'username ', 'Hello {0} ')}}`

Therefore the regex is once again (sorry for the multiple PR's) changed to also cover this edge-case. Updated the original test-case to reflect this change.

image

@kevinpschaaf PTAL

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -88,7 +88,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           this._parseElementAnnotations(node, list);
     },
 
-    _bindingRegex: /([^{[]*)(\{\{|\[\[)(?!\}\}|\]\])(.+?)(?:\]\]|\}\})/g,
+    _bindingRegex: /((?!\{\{|\[\[).+?)?(\{\{|\[\[)(.+?)(?:\]\]|\}\})/g,
 
     // TODO(kschaaf): We could modify this to allow an escape mechanism by
     // looking for the escape sequence in each of the matches and converting
--- a/test/unit/bind-elements.html
+++ b/test/unit/bind-elements.html
@@ -26,7 +26,7 @@
       no-computed="{{foobared(noInlineComputed)}}"
       compoundAttr1$="{{cpnd1}}{{ cpnd2 }}{{cpnd3.prop}}{{ computeCompound(cpnd4, cpnd5, 'literal')}}"
       compoundAttr2$="literal1 {{cpnd1}} literal2 {{cpnd2}}{{cpnd3.prop}} literal3 {{computeCompound(cpnd4, cpnd5, 'literal')}} literal4"
-      compoundAttr3$="{{computeCompound('world', 'username ', 'Hello {0} ')}}"
+      compoundAttr3$="[yes/no]: {{cpnd1}}, {{computeCompound('world', 'username ', 'Hello {0} ')}}"
       >
       Test
       </div>
--- a/test/unit/bind.html
+++ b/test/unit/bind.html
@@ -767,10 +767,12 @@ suite('compound binding / string interpolation', function() {
     assert.equal(el.$.boundChild.getAttribute('compoundAttr2'), 'literal1  literal2  literal3 literal literal4');
   });
 
-  test('compound property attribute with {} in text', function() {
+  test('compound property attribute with {} and [] in text', function() {
     var el = document.createElement('x-basic');
 
-    assert.equal(el.$.boundChild.getAttribute('compoundAttr3'), 'Hello {0} username world')
+    el.cpnd1 = 'cpnd1';
+
+    assert.equal(el.$.boundChild.getAttribute('compoundAttr3'), '[yes/no]: cpnd1, Hello {0} username world')
   })
 
   test('compound adjacent textNode bindings', function() {


JeremybellEU:
There has been a problem reported in https://github.com/GoogleWebComponents/google-chart/issues/73#issuecomment-156607495 where the regex updated in #2705 matched the `JSONArray`, which it did not before. I have derived two possible actions:

1. Do not do anything. While this sounds weird, the `JSONArray` does match the description in [the docs](https://www.polymer-project.org/1.0/docs/devguide/data-binding.html#property-binding): `A binding annotation consists of a property name or subproperty name enclosed in curly brackets ({{}}) or square brackets ([[]]).`. The "name" is a list of `JSONObjects`. However, in order to circumvent this issue, a new regex can be even more strict, as shown in option 2:
* Further restrict the allowed names to bind on. This solution is again more intelligent, but with every update it also lowers the maintainability and might cause collateral damage (like what happened with the JSON). It would allow all tests to pass, but creates more edge-cases. The regex I came up with:
```javascript
// Regex with comments, compiles to:
// ((?!\{\{|\[\[).+?)?(\{\{|\[\[)((?:!)?(?:(?:[\w\s\-\_\.\:\$]+?)|(?:[\w\s\-\_\.\:\$]+\(.*?\))))(?:\]\]|\}\})
_bindingRegex: (function() {
  // Opening Binding options
  var ob = '\\{\\{|\\[\\[';
  // All allowed characters in a name of property or computed function
  var name = '[\\w\\s\\-\\_\\.\\:\\$]+';

  return new RegExp('((?!' + ob + ').+?)?'  // Everything that is not a {{ or [[
    + '(' + ob + ')'                        // The mode (two- or one-way)
    + '('                                   // Start of contents of the binding
      + '(?:\!)?'                           // Is the binding a negation
      + '(?:'                               // Can either be name or computed function
        + '(?:' + name + '?)'               // A normal name for a property
      + '|'                                 // Or
        + '(?:' + name                      // The name of the computed function
        + '\\(.*?\\)'                       // The arguments of the function within ()
        + ')'                               // End of computed function match
      + ')'                                 // End of name of property or computed function match
    + ')'                                   // End of the whole match returned as m[2]
    + '(?:\\]\\]|\\}\\})'                   // The binding closing

    ,'g')
  })(),
```
image

Both options have pro's and con's. Option 1 requires users to put JSON as return value of a computed function or property, option 2 is significantly more difficult to maintain and understand. Personally, I don't know what the way to go is. I only wanted to show that both options are possible.

I hope this information is useful, please let me know what option would be desirable.

cc @ebidel @kevinpschaaf @azakus @sorvell

ebidel:
1 would be a breaking change, so I'm not sure we have much of an option other than to get this working again as it was.

arthurevans:
For option #1, couldn't you also put a space between the brackets? I believe "[ [1, 2, 3], [3, 4, 5] ]" would passed through unscathed.

JeremybellEU:
@arthurevans Yes that is correct. This would not match on the regex, but parse as JSON.

jongeho1:
Future complexity may be tamed by decoupling the one-way data binding anchors from valid JSON (choosing new anchors)

Here are some more cases but the intent here may be ambiguous between the use of existing features:

* one-way data binding w/ literal 
* JSON string literals in attributes with intent of Array or Object deserialization:

```
[[0]]
[[0.01]]
[[true]]
````

JeremybellEU:
I have inserted another option to counterfeit JSON edge-cases as pointed out by @jongeho1

david-saslawsky:
I think the problem with the option #3 is not that is is too complex but that it not tight enough to prevent all false positive matches.

The current parser is not that simple to maintain, it does not recognizes things it should:
````
[[localize(',')]]

[[alert('[[caution]]')]]

[[
foo (
   bar, 123
)
]] 
````

It also recognizes things it shouldn't - which creates a risk of collision with other syntax and makes typos harder to find:
````
[[compound(foo]]     // silently removed
````

So it seems to me that there are also maintenance and complexity costs with option #1

I mean, polymer is great as it is. You sure don't run into these problems everyday and there are workarounds in pretty much all cases - but still, trying to match an airtight syntax does not seem that expensive to me: 
````
var NUMBER = '(:?' + '[0-9]+' + ')'; 
var STRING = '(:?' + '\'[^\']*\'' + ')';    // todo: string escaping and double quotes
var IDENTIFIER = '(:?' + '[\\w\\s\\-\\_\\.\\:\\$]+' + ')';

var ARGUMENT = '(:?' + IDENTIFIER + '|' + NUMBER + '|' +  STRING + ')';
var ARGUMENT_LIST = '(:?' + ARGUMENT + ',?' + ')*';
var COMPUTED_FUNCTION = '(:?' + IDENTIFIER + '\\(' + ARGUMENT_LIST + '\\)' + ')';
var BINDING = '(:?' + COMPUTED_FUNCTION + '|' + IDENTIFIER + ')';
var EXPRESSION = '\\[\\[' + BINDING + '\\]\\]' + '|' + '\\{\\{' + BINDING + '\\}\\}';

var bindingRegex = new RegExp(EXPRESSION, "g")
````
Improving and testing each individual sub-expression is still easy.




If this "one regular expression to rule them all" approach doesn't work (it runs but I haven't tested the speed or the compatibility), another option is to use the RE as a poor man's lexer that split the input string into a stream of tokens - as soon as something that looks like a binding is detected.

The trick would be to bail out and leave the HTML untouched in case of parser error. Ambiguities between json and polymer syntax are rare and wouldn't take long to resolve anyway.

kevinpschaaf:
@ebidel Just as a point of order, I want to point out that not disambiguating nested JSON array literals from one-way bindings has been a long-standing issue (https://github.com/Polymer/polymer/issues/1734) which only temporarily worked for two releases (`v1.2.0` and `v1.2.1`) as an unintentional side-effect of changes that came with compound binding, so I'm inclined to go a little more slowly w.r.t. reacting to that particular "breakage".

That said, I'm reviewing @david-saslawsky's expression for performance impact and if we can make it work with the existing bespoke string escaping.

JeremybellEU:
I really like the syntax of @david-saslawsky , but keep in mind this is a breaking change regarding naming of variables. (I think this limitation is the way to go to prevent updating this regex every week) The consequence would be that update must become a MAJOR update according to semver.

david-saslawsky:
Sorry if this is not the right place to discuss this.

I've made more a complete version of the regex:

````
var NUMBER = '(:?' + '[-+]?[0-9]*\.?[0-9]+(:?[eE][-+]?[0-9]+)?' + ')'; 

var SQ_STRING = "\\'(?:\\\\.|[^'\\\\])*\\'";
var DQ_STRING = '\\"(?:\\\\.|[^"\\\\])*\\"';
var STRING = '(:?' + SQ_STRING + '|' + DQ_STRING + ')';

var IDENTIFIER = '(:?' + '[\\w\\s\\-\\_\\.\\:\\$]+' + ')';

var ARGUMENT = '(:?' + IDENTIFIER + '|' + NUMBER + '|' +  STRING + ')';
var ARGUMENT_LIST = '\\(' + '(:?' + ARGUMENT + '(:?,|\\))' + ')*';
var COMPUTED_FUNCTION = '(:?' + IDENTIFIER + ARGUMENT_LIST + ')';
var BINDING = '(:?' + COMPUTED_FUNCTION + '|' + IDENTIFIER + ')';
var EXPRESSION = '(\\[\\[|\\{\\{)' + '(' + BINDING + ')' + '(:?\\]\\]|\\}\\})';

var bindingRegex = new RegExp(EXPRESSION, "g")
````

Floats and string escaping work. Spaces and newlines are not supported but this shouldn't have much impact on the performance (I would expect that most of the time is spent looking for the `[[` or `{{`).

This is almost airtight but I can't find a way to handle comma correctly in the argument list.
````
[[foo(123,'d\'oh','[[x]]')]]    // this is recognized
[[foo(123,'d\'oh','[[x]]',]]    // ... but so is this
````

In the long term, the "Regexp as a lexer" approach seems even more robust and flexible. Maybe something like:
````
var SYMBOL = "[,\\(\\)]";
var re = new RegExp('(' + NUMBER + ')|(' + STRING + ')|(' + SYMBOL + ')' + ...);
...
re.lastIndex = (index of the '[[') + 2;
while((match=re.exec(str)) !== null) {
    // parser state machine
    switch(parser.state) {
        ...
        case STATE_ARGUMENT_LIST:
            if(match[1])) { /* accept number */ }
            else if(match[2])) { /* accept string */ }
            else { /* error */ }
    }
}
````

This can parse things impossible to parse with a regexp:
````
foo(bar(),[['baz']])    // nested functions and JSON constants
````


Please let me know if I can be of any help.

JeremybellEU:
@david-saslawsky It might be good to create a PR with your fix so it can be reviewed more easily. I think I have the solution for the argument comma problem.

kevinpschaaf:
@david-saslawsky I modified your regex to pass the tests when integrated with Polymer and then ran some rudimentary perf tests against it.  While much more correct than the previous, it highlighted a potential [catastropic backtracking](http://www.rexegg.com/regex-explosive-quantifiers.html) situation due to the repeating quantifiers for `IDENTIFIER` characters nested inside repeating `ARGUMENT` expressions when failing a partial match.  I finally resolved the perf issue using an approximation for atomic groups using lookahead+immediate back-referencing, which made it a little more complex but still an overall improvement.

In my tests, it now parses faster than the regex on master and pretty close to the parsing speed in v1.1.5 before we added compound bindings, so I'm inclined to go with this solution, despite being slightly more strict about identifier characters allowed (I fail to see this as any sort of practical problem).  I will put up a PR for this shortly.

It also has the happy side-effect of once-again excluding most nested JSON array literals (due to the stricter expression matching) and makes the google-chart repro work again.

JeremybellEU:
Closing in favor of #3017

PR #2717, opened 2015-11-11
Adds support for `activeElement` to `Polymer.dom`.

bicknellr:
For document and shadow root nodes, if the active element is a (possibly deep) descendant of the given node, `Polymer.dom(node).activeElement` will return the deepest element within the light tree of the node which is or deeply contains the active element. For all other nodes, the getter and setter will forward to the `activeElement` property on the node.

### DIFF
--- a/src/lib/dom-api-shadow.html
+++ b/src/lib/dom-api-shadow.html
@@ -60,11 +60,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     activeElement: {
       get: function() {
-        return Polymer.DomApi.wrap(this.node).activeElement;
+        var node = Polymer.DomApi.wrap(this.node);
+        var activeElement = node.activeElement;
+        return node.contains(activeElement) ? activeElement : null;
       },
       configurable: true
     },
-    
+
     childNodes: {
       get: function() {
         return TreeApi.arrayCopyChildNodes(this.node);
--- a/src/lib/dom-api-shadow.html
+++ b/src/lib/dom-api-shadow.html
@@ -58,6 +58,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   Object.defineProperties(DomApi.prototype, {
 
+    activeElement: {
+      get: function() {
+        return Polymer.DomApi.wrap(this.node).activeElement;
+      },
+      configurable: true
+    },
+    
     childNodes: {
       get: function() {
         return TreeApi.arrayCopyChildNodes(this.node);
--- a/src/lib/dom-api-shadow.html
+++ b/src/lib/dom-api-shadow.html
@@ -60,8 +60,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     activeElement: {
       get: function() {
-        var node = Polymer.DomApi.wrap(this.node);
+        var node = DomApi.wrap(this.node);
         var activeElement = node.activeElement;
+        // Prevents `activeElement` from returning elements outside of the
+        // ShadowRoot, even if they would become descendants of the ShadowRoot
+        // in the composed tree. See w3c/webcomponents#358.
         return node.contains(activeElement) ? activeElement : null;
       },
       configurable: true
--- a/src/lib/dom-api-shady.html
+++ b/src/lib/dom-api-shady.html
@@ -430,35 +430,35 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         }
         var isShadyRoot = !!this.node._isShadyRoot;
         if (this.node !== document) {
-          // If this node isn't a document or shady root, then it doesn't
-          // have an active element.
+          // If this node isn't a document or shady root, then it doesn't have
+          // an active element.
           if (!isShadyRoot) {
             return null;
           }
           // If this shady root's host is the active element or the active
-          // element is not a descendant of the host, then it doesn't have
-          // an active element.
+          // element is not a descendant of the host (in the composed tree),
+          // then it doesn't have an active element.
           if (this.node.host === active ||
               !this.node.host.contains(active)) {
             return null;
           }
-          // If the active element is a light descendant of the shady root's
-          // host, return the active element.
-          if (this.node.host !== active &&
-              this._contains(this.node.host, active)) {
-            return active;
-          }
         }
-        // This node is either the document or a shady root of which the
-        // active element is a descendant of its host; iterate upwards to
-        // find the active element's most shallow host.
+        // This node is either the document or a shady root of which the active
+        // element is a (composed) descendant of its host; iterate upwards to
+        // find the active element's most shallow host within it.
         var activeRoot = Polymer.dom(active).getOwnerRoot();
-        while (activeRoot && activeRoot !== this.node &&
-            !(isShadyRoot && this._contains(this.node.host, active))) {
+        while (activeRoot && activeRoot !== this.node) {
           active = activeRoot.host;
           activeRoot = Polymer.dom(active).getOwnerRoot();
         }
-        return active;
+        if (this.node === document) {
+          // This node is the document, so activeRoot should be null.
+          return activeRoot ? null : active;
+        } else {
+          // This node is a non-document shady root, and it should be
+          // activeRoot.
+          return activeRoot === this.node ? active : null;
+        }
       },
       configurable: true
     },
--- a/src/lib/dom-api-shady.html
+++ b/src/lib/dom-api-shady.html
@@ -422,6 +422,47 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   Object.defineProperties(DomApi.prototype, {
 
+    activeElement: {
+      get: function() {
+        var active = document.activeElement;
+        if (!active) {
+          return null;
+        }
+        var isShadyRoot = !!this.node._isShadyRoot;
+        if (this.node !== document) {
+          // If this node isn't a document or shady root, then it doesn't
+          // have an active element.
+          if (!isShadyRoot) {
+            return null;
+          }
+          // If this shady root's host is the active element or the active
+          // element is not a descendant of the host, then it doesn't have
+          // an active element.
+          if (this.node.host === active ||
+              !this.node.host.contains(active)) {
+            return null;
+          }
+          // If the active element is a light descendant of the shady root's
+          // host, return the active element.
+          if (this.node.host !== active &&
+              this._contains(this.node.host, active)) {
+            return active;
+          }
+        }
+        // This node is either the document or a shady root of which the
+        // active element is a descendant of its host; iterate upwards to
+        // find the active element's most shallow host.
+        var activeRoot = Polymer.dom(active).getOwnerRoot();
+        while (activeRoot && activeRoot !== this.node &&
+            !(isShadyRoot && this._contains(this.node.host, active))) {
+          active = activeRoot.host;
+          activeRoot = Polymer.dom(active).getOwnerRoot();
+        }
+        return active;
+      },
+      configurable: true
+    },
+
     childNodes: {
       get: function() {
         var c$ = TreeApi.Logical.getChildNodes(this.node);
--- a/src/lib/dom-api-shady.html
+++ b/src/lib/dom-api-shady.html
@@ -446,10 +446,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // This node is either the document or a shady root of which the active
         // element is a (composed) descendant of its host; iterate upwards to
         // find the active element's most shallow host within it.
-        var activeRoot = Polymer.dom(active).getOwnerRoot();
+        var activeRoot = dom(active).getOwnerRoot();
         while (activeRoot && activeRoot !== this.node) {
           active = activeRoot.host;
-          activeRoot = Polymer.dom(active).getOwnerRoot();
+          activeRoot = dom(active).getOwnerRoot();
         }
         if (this.node === document) {
           // This node is the document, so activeRoot should be null.
--- a/test/unit/polymer-dom-elements.html
+++ b/test/unit/polymer-dom-elements.html
@@ -357,3 +357,245 @@
   });
   </script>
 </dom-module>
+
+
+<dom-module id="x-shadow-host-root">
+  <template>
+    <content></content>
+    <div>
+      <div>
+        <x-shadow-host-root-0>
+          <x-shadow-host-root-0-light></x-shadow-host-root-0-light>
+        </x-shadow-host-root-0>
+      </div>
+    </div>
+    <div>
+      <x-shadow-host-root-1>
+        <x-shadow-host-root-1-light></x-shadow-host-root-1-light>
+      </x-shadow-host-root-1>
+    </div>
+  </template>
+  <script>
+    Polymer({
+      is: 'x-shadow-host-root',
+      hostAttributes: {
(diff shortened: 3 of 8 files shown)


bicknellr:
@sorvell ping

src/lib/dom-api.html
```diff
@@ -593,6 +593,55 @@
 
     };
 
+    Object.defineProperties(DomApi.prototype, {
+      activeElement: {
```
sorvell: Let's separate the implementations of Shadow/Shady. Should make this more understandable.

src/lib/dom-api.html
```diff
@@ -593,6 +593,55 @@
...
+
+          // Not ShadowRoot-like?
+          if (!this.node.host && ownerDocument) {
+            return this.node.activeElement;
+          }
+
+          var active = trueDocument.activeElement;
+          if (!active) return null;
+
+          if (window.ShadowDOMPolyfill) {
```
sorvell: Does ShadowDOMPolyfill not have support for activeElement according to the ShadowDOM spec? If not, support should be added there.

src/lib/dom-api.html
```diff
@@ -593,6 +593,55 @@
...
+            active = ShadowDOMPolyfill.wrap(active);
+          }
+
+          var activeRoot = Polymer.dom(active).getOwnerRoot();
+          while (activeRoot && activeRoot !== this.node) {
```
sorvell: I suggest making the contract that you get the active element in the same scope as this node. I believe that would make this:

```
var ownerRoot = this.getOwnerRoot();
while (activeRoot && activeRoot !== ownerRoot) {...}
```

src/lib/dom-api.html
```diff
@@ -593,6 +593,55 @@
...
+            active = activeRoot.host;
+            activeRoot = Polymer.dom(active).getOwnerRoot();
+          }
+
+          return (!ownerDocument ? !activeRoot : activeRoot === this.node) ? active : null;
+        },
+        set: function(value) {
```
sorvell: Why is this settable? The native version appears to be read-only (https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement).

bicknellr: For elements that don't have a special activeElement property, this just lets it work as a normal property accessor (the default behavior). It definitely feels like an unlikely use case though, cleanliness is probably more important than correctness here. Think I should remove it?

src/lib/dom-api.html
bicknellr: [Doesn't look like it.](https://github.com/webcomponents/webcomponentsjs/search?q=activeElement)

src/lib/dom-api.html
sorvell: Yeah, I'd say remove.

src/lib/dom-api.html
```diff
@@ -593,6 +593,55 @@
...
+            active = ShadowDOMPolyfill.wrap(active);
+          }
+
+          var activeRoot = Polymer.dom(active).getOwnerRoot();
+          while (activeRoot && activeRoot !== this.node) {
+            active = activeRoot.host;
+            activeRoot = Polymer.dom(active).getOwnerRoot();
+          }
+
+          return (!ownerDocument ? !activeRoot : activeRoot === this.node) ? active : null;
```
bicknellr: This is confusing - I'll break it apart when updating.

src/lib/dom-api.html
bicknellr: Wouldn't this cause only siblings of the ShadowRoot's host / the host itself to be returned?

bicknellr: (from offline) Leaving this as is, but it needs a comment explaining that `this.node` will always be a shady root or document at this point.

src/lib/dom-api.html
bicknellr: Working on it here: webcomponents/webcomponentsjs#457.

bicknellr: Just merged by @azakus.

src/lib/dom-api.html
bicknellr: I think this might be a little better now.

src/lib/dom-api.html
bicknellr: Removed.

bicknellr:
@sorvell This is ready for review again.

src/lib/dom-api.html
```diff
@@ -593,6 +593,61 @@
...
     };
 
+    Object.defineProperties(DomApi.prototype, {
+      activeElement: {
+        configurable: true,
+        get: function() {
+          var ownerDocument = this.node.ownerDocument;
+
+          // Shadow DOM
+          if (Polymer.Settings.useShadow) {
```
sorvell: This section should be its own getter in the `useShadow` section of this module.

src/lib/dom-api.html
```diff
@@ -593,6 +593,61 @@
...
+            }
+          }
+
+          // Shady DOM
+
+          var trueDocument = ownerDocument || this.node;
+
+          // Not a document or shady root.
+          if (ownerDocument && !this.node._isShadyRoot) {
+            return this.node.activeElement;
```
sorvell: This only makes sense if this.node == document doesn't it? If so, let's make that explicit `if (this.node === document)`...

src/lib/dom-api.html
```diff
@@ -593,6 +593,61 @@
...
+          }
+
+          var active = trueDocument.activeElement;
+          if (!active) return null;
+
+          // `this.node` is either a document or shady root; iterate upwards
+          // to find the active element's most shallow host (relative to this
+          // node or its containing document).
+          var activeRoot = Polymer.dom(active).getOwnerRoot();
```
sorvell: Consider using `contains` to bail out of doing this work. (https://developer.mozilla.org/en-US/docs/Web/API/Node/contains)

src/lib/dom-api.html
```diff
@@ -593,6 +593,61 @@
...
+          while (activeRoot && activeRoot !== this.node) {
+            active = activeRoot.host;
+            activeRoot = Polymer.dom(active).getOwnerRoot();
+          }
+
+          // This node is a document and we’ve reached it; we should return the
+          // active element's most shallow host.
+          if (!ownerDocument && !activeRoot) {
```
sorvell: Why do we need to check `ownerDocument` here?

src/lib/dom-api.html
bicknellr: This is for specifically when it isn't a document or a shady root, it just pulls the property off the node by name. This was the getter portion of the setter I had in earlier that you wanted me to remove, maybe this would be better returning null? I suppose without the setter it's not really helpful.

src/lib/dom-api.html
bicknellr: `activeRoot` can be null if `this.node` is a shady root and not an ancestor of the active element, but we only want to return `active` when `activeRoot` is null if `this.node` is also the document.

src/lib/dom-api.html
bicknellr: How would contains make this faster? Do you mean skip `getOwnerRoot` when the active element is actually in this shady root (rather than nested deeper)?

bicknellr:
@sorvell ping again

src/lib/dom-api.html
sorvell: Yeah, let's remove.

src/lib/dom-api.html
sorvell: The native `contains` should be significantly faster than the work below. Using it allow this to fail fast.

src/lib/dom-api.html
sorvell: If we handle `this.node === document` explicitly and do the `contains` test then we only get here if we should return the activeRoot, isn't that the case?

src/lib/dom-api.html
bicknellr: (from offline discussion) I forgot that in this case we know for sure that we're in shady DOM rather than true Shadow DOM, so native `contains` will effectively be deep - I'll fix this with the next push.

src/lib/dom-api.html
bicknellr: (from offline discussion) `getOwnerRoot` returns null if the element is within the light DOM of the document, similar to how `ownerDocument` is null for documents. Also, I'll remove the use of `ownerDocument` and just compare with `document` directly for the sake of simplicity.

src/lib/dom-api.html
bicknellr: Moved.

src/lib/dom-api.html
bicknellr: Removed.

bicknellr:
So, since Polymer runs the tests on Chrome's native version of Shadow DOM, it looks like I didn't correctly interpret the retargeting algorithm for `activeElement` on ShadowRoots. Specifically, it seems like if the active element is within the light DOM of a ShadowRoot's host, `activeElement` of that ShadowRoot should be the active element rather than null. I need to fix this in the web components polyfills and then come back here. :/

src/lib/dom-api.html
```diff
@@ -593,11 +593,73 @@
...
+    Object.defineProperties(DomApi.prototype, {
+      activeElement: {
+        configurable: true,
+        get: function() {
+          var ownerDocument = this.node.ownerDocument;
+
+          if (ownerDocument || !window.ShadowDOMPolyfill) {
+            return this.node.activeElement;
+          } else {
+            return window.ShadowDOMPolyfill.wrapIfNeeded(this.node).activeElement;
```
sorvell: This file has a wrap implementation so I think this can just be:

`return DomApi.wrap(this.node).activeElement;`

src/lib/dom-api.html
```diff
@@ -593,11 +593,73 @@
...
+    });
+
     // changes and accessors...
     if (!Settings.useShadow) {
 
       Object.defineProperties(DomApi.prototype, {
 
+        activeElement: {
+          get: function() {
+            var active = document.activeElement;
```
sorvell: Generally we don't space-ify within functions and always include `{}` around conditionals.

src/lib/dom-api.html
```diff
@@ -593,11 +593,73 @@
...
+            }
+
+            // This node is a shady root and we've found the active element's
+            // most shallow host (within the shady root).
+            if (!nodeIsDocument && activeRoot === this.node) {
+              return active;
+            }
+
+            // The active element is within the light DOM of this shady root.
+            return null;
```
sorvell: As you noted, the expectation here is to return active. So it seems like the 2 conditionals above are not necessary and we can always just return `active` if we get to here.

bicknellr:
([PR for `activeElement` fix in web components polyfill.](https://github.com/webcomponents/webcomponentsjs/pull/463))

src/lib/dom-api.html
bicknellr: Fixed.

src/lib/dom-api.html
bicknellr: Fixed.

src/lib/dom-api.html
bicknellr: Fixed.

bicknellr:
Updated with the expectation that the web components polyfill PR will make it in, tests will fail until then.

src/lib/dom-api.html
```diff
@@ -593,11 +593,57 @@
 
     };
 
+    Object.defineProperties(DomApi.prototype, {
+      activeElement: {
+        configurable: true,
+        get: function() {
+          return wrap(this.node).activeElement;
```
sorvell: Should be `DomApi.wrap`

sorvell:
Only 1 little nit left, otherwise LGTM after we get the webcomponents change merge. Thanks!

src/lib/dom-api.html
bicknellr: Do you mean `Polymer.DomApi.wrap`? The `DomApi` constructor defined in this file doesn't have a `wrap` property - the object returned / assigned to `Polymer.DomApi` does though. However, the other function calling `wrap` calls it directly since it's defined in this scope. Thoughts?

bicknellr:
I had to add additional tests and make an update to the host-finding loop for this new case - might want to double check that.

src/lib/dom-api.html
```diff
@@ -593,11 +593,57 @@
...
+              // host, return the active element.
+              if (this.node.host !== active && this._contains(this.node.host, active)) {
+                return active;
+              }
+            }
+            // This node is either the document or a shady root of which the
+            // active element is a descendant of its host; iterate upwards to
+            // find the active element's most shallow host.
+            var activeRoot = Polymer.dom(active).getOwnerRoot();
+            while (activeRoot && activeRoot !== this.node && !(this.node._isShadyRoot && this._contains(this.node.host, active))) {
```
bicknellr: Here's the new change, is there a preferred way to format really long conditionals like this?

sorvell: Yeah, 80 character line limit + indenting.

valdrinkoshi:
@sorvell ping for this one, needed by https://github.com/PolymerElements/iron-overlay-behavior/pull/68

bicknellr:
Depends on webcomponents/webcomponentsjs#475.

src/lib/dom-api-shadow.html
```diff
@@ -58,6 +58,15 @@
 
   Object.defineProperties(DomApi.prototype, {
 
+    activeElement: {
+      get: function() {
+        var node = Polymer.DomApi.wrap(this.node);
+        var activeElement = node.activeElement;
+        return node.contains(activeElement) ? activeElement : null;
```
sorvell: Let's add a comment here that explains the `contains` check. I believe this is so that we ignore distributed nodes. We should also link to the spec bug where the native behavior is changing.

src/lib/dom-api-shadow.html
```diff
@@ -58,6 +58,15 @@
 
   Object.defineProperties(DomApi.prototype, {
 
+    activeElement: {
+      get: function() {
+        var node = Polymer.DomApi.wrap(this.node);
```
sorvell: DomApi is available locally so remove `Polymer.`

src/lib/dom-api-shady.html
```diff
@@ -422,6 +422,47 @@
...
+            return null;
+          }
+        }
+        // This node is either the document or a shady root of which the active
+        // element is a (composed) descendant of its host; iterate upwards to
+        // find the active element's most shallow host within it.
+        var activeRoot = Polymer.dom(active).getOwnerRoot();
+        while (activeRoot && activeRoot !== this.node) {
+          active = activeRoot.host;
+          activeRoot = Polymer.dom(active).getOwnerRoot();
```
sorvell: The convention in this module is to use `dom` which is locally mapped to `Polymer.dom`.

test/unit/polymer-dom-elements.html
```diff
@@ -357,3 +357,245 @@
   });
   </script>
 </dom-module>
+
+
+<dom-module id="x-shadow-host-root">
```
sorvell: Make sure to always register dependent elements before an element that uses them (e.g. `x-shadow-host-root-0`)

src/lib/dom-api-shady.html
bicknellr: Fixed.

bicknellr:
@sorvell: fixed stuff from your comments

sorvell:
LGTM

PR #2718, opened 2015-11-11
Add Polymer.DomApi.wrap

azakus:
- Make sure nodes are wrapped under ShadowDOMPolyfill
- Forwards `window.wrap` when defined
- Remove weird forked `DomApi` definition
- Remove calls to `wrap` from gestures


sorvell:
LGTM

arthurevans:
Does this need API docs? Would user code ever need to call this explicitly?

PR #2736, opened 2015-11-13
Better handling and error msgs for registration oopsies

### LINKED ISSUE #1163: improve error message on invalid element names
I just got this error message (because I didn't have a "-" in my element name)

Uncaught SyntaxError: Failed to execute 'registerElement' on 'Document': Registration failed for type 'sentence'. The type name is invalid.
and a JS stack trace.

There was no reference to the file:line that was doing something bad.

Also the use of "type" in the exception but "element" everywhere else is unhelpful.

### LINKED ISSUE #2730: Check if an element has been registered
hi, in my current project, I have errors in the console about elements that have been registered. This errors are related with some partials vulcanized elements. 

I have a base file with some `core` elements (this file is vulcanized) and then, each element in different pages will have their own components. 

Based on an idea of @jorgecasar, I suggest this code inside `polymer-bootstrap.html` to avoid these console errors:

```
      if (!Polymer.telemetry.isRegistered(prototype.is)) {
        Polymer.telemetry._registrate(prototype);
        document.registerElement(prototype.is, options);
      }
```

The idea is simple, add a method to `telemetry` to check if the new component has been registered. If is not, we register it. 

I've used my current project to test and check the result with the new code. I will try to make a test inside Polymer to send the pull request with the modifications.

### LINKED ISSUE #2735: Registering invalid element name or re-using name should not crash app
Currently, trying to register an invalid custom element name or re-registering an existing naw causes JS errors and a white screen of death. This is because `document.registerElement()` throws and further JS doesn't run.

This can be very frustrating for developers. A bad name or re-use of an existing name should "crash" an app! Polymer mini can smooth this over.

Examples:

```
Polymer({is: 'x'}); // Attempt to register an invalid CE name bombs out app.
Polymer({is: 'x-foo'});
Polymer({is: 'x-foo'}); // Attempt to re-register an existing name bombs out app!
console.log('I wont run');
```

ebidel:
- Fixes #2730
- Fixes #1163
- Fixes #2735 

Instead of throwing and preventing apps from loading, this changes warns to the console for:
- attempt to register an element with a name that's already been registered.
- invalid element name (e.g. "x" instead of "x-foo").

R: @kevinpschaaf @sorvell @azakus

### DIFF
--- a/src/lib/polymer-bootstrap.html
+++ b/src/lib/polymer-bootstrap.html
@@ -37,13 +37,28 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var options = {
         prototype: prototype
       };
-      // NOTE: we're specifically supporting older Chrome versions here 
+      // NOTE: we're specifically supporting older Chrome versions here
       // (specifically Chrome 39) that throw when options.extends is undefined.
       if (prototype.extends) {
         options.extends = prototype.extends;
       }
-      Polymer.telemetry._registrate(prototype);
-      document.registerElement(prototype.is, options);
+
+      try {
+        Polymer.telemetry._registrate(prototype);
+        document.registerElement(prototype.is, options);
+      } catch(e) {
+        if (prototype.is.indexOf('-') === -1) {
+          Polymer.Base._warn([prototype.is, '::',
+              'Attempted to register invalid custom element name "' +
+              prototype.is + '". Names must include a "-". See https://goo.gl/RQIJXZ.'
+          ]);
+        } else if (e.name === 'NotSupportedError') {
+           Polymer.Base._warn([
+              prototype.is, '::', 'Element name is already registered.']);
+        } else {
+          throw e; // Throw all other errors we encounter.
+        }
+      }
       return factory;
     };
 
--- a/test/unit/micro.html
+++ b/test/unit/micro.html
@@ -139,6 +139,22 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var a = document.createElement('falsey-polymer-arg');
       assert.equal(a.is, 'falsey-polymer-arg');
     });
+
+    test('register invalid element names', function() {
+      var func = function() {
+        Polymer({is: 'invalid_element_name'});
+      };
+      assert.doesNotThrow(func, 'SyntaxError: Failed to execute',
+                          'registering invalid element name does not throw');
+    });
+
+    test('re-register element names', function() {
+      var func = function() {
+        Polymer({is: 'x-trivial'});
+      };
+      assert.doesNotThrow(func, 'name is already registered',
+                    're-registration does not throw');
+    });
   });
 
 </script>


src/lib/polymer-bootstrap.html
```diff
@@ -37,13 +37,28 @@
...
-      // NOTE: we're specifically supporting older Chrome versions here 
+      // NOTE: we're specifically supporting older Chrome versions here
       // (specifically Chrome 39) that throw when options.extends is undefined.
       if (prototype.extends) {
         options.extends = prototype.extends;
       }
-      Polymer.telemetry._registrate(prototype);
-      document.registerElement(prototype.is, options);
+
+      try {
```
samccone: I wonder if there is a way to check this without dropping into a try catch deopt

azakus: The only other option is to try creating the element and see if it has a prototype that isn't `HTMLElement`

```js
var e = prototype.extends, i = prototype.is;
var el = e ? document.createElement(e, i) : document.createElement(e);
if (Object.getPrototype(el) !== HTMLElement.prototype) {
  // ...
}
```

ebidel: Keep in mind that every try/catch doesn't necessarily go through deopt. This piece of code would also need to be optimized in the first place for it to deopt. Registering elements only happens at boot and very few times. It's like this would never get optimized in the first place.

ebidel: I just heard the v8 PM stay himself, "it used to be the case that try/catch was always de-optimized. That is no longer the case!"

ebidel:
Ping

jcorrall:
ping +1

src/lib/polymer-bootstrap.html
addyosmani: There's two more bugs left to nail down for try/catch to no longer cause deopts in V8. Getting super close!

ebidel: Yep. Until then, my point was more that this piece of code gets executed very few times. We have other hot paths in polymer core (`endOfMicroTask`) that are deptop'd due to try/catch that should be looked at first.

sorvell:
We're disinclined to accept this code in the core library. Instead, the polymer linter should be able to deal with these cases. Our general rule is to push detailed and user friendly error messages to the linter rather than taxing the code in the core library.

ebidel:
We should reconsider for this case. This fixes 3 common issues as well as being helpful to users.

The current browser/Polymer messages are pretty vague and it's a pretty common issue for users. Errors like this shouldn't prevent the entire app from loading.

How is the code taxed?

sorvell:
Cases like this are precisely why the polymer linter was created. It's a great tool and is at a point now that it should be useful to solve problems like this. Custom detailed error messages inflate the core library's code size. We could very easily accept 100 individually reasonable error messages. These things add up.

samccone:
I would argue that having detailed error messages inside of the dev version of the code is really valuable, both angular 1.x and react have these features which are incredibility useful to developers (this feature in these libs has saved me many many hours over the past few years).

As far as overall codesize, a naive compiler phase on release to build a production version that strips these messages would remove this concern.

Pushing this responsibility off to another tool assumes that people are comfortable setting up a build / lint / dev flow, where in reality a lot of people use a much more bare bones approach in development or are simply not comfortable setting these things up.

imho helpful error messages are super valuable to devs and the byte cost of the helper code can be negated with a production version of the lib.

JeremybellEU:
My 2 cents:

I agree with @samccone on this one. Creating a special development version with useful error messages and using a seperate production ready polymer version would greatly improve my development workflow. Installing a seperate linter for just more detailed error messages seems weird for me.

azakus:
@samccone I think making a `dev` version of polymer has a few side-effects we need to be careful about. For instance, what happens when someone writes a component that has a dependency on `polymer-dev.html` and then releases that to the public?
How does angular/react/whatever handle that now?

azakus:
@JeremybellEU The linter will warn about more than this case, such as misspelled observers, or broken binding syntax.

samccone:
xref https://github.com/Polymer/polymer/issues/2555#ref-commit-5c46983

ebidel:
@sorvell Our users clearly want this. I don't think it's reasonable to offload developer feedback like this to a linter. Very few people will ever install and use a linting tool.

A linter is not a great tool for runtime errors. Polylint is good about analyzing a single file, but when it comes to an entire app it falls short. There are also other factors in the mix when you start to developing a real app: vulcanize, crisper, etc. 

Re. codesize: This is a 218 byte small change that would have a large impact for users. Agreed, we shouldn't accept 100 PRs for error checking. But we should accept a few :)

azakus:
@sorvell I think the actual issue here is worth fixing: Screwing up an element registration shouldn't leave you with a broken, blank page.

Yes, it is a band-aid, as your app won't work as expected.
Yes, we might get more bug reports about stuff not working.
Yes, it is more work for Polymer to do than to not do.
Yes, it is a slippery slope.

But in my mind, the cost of having users get frustrated and stop trying is worse than any performance hit we could incur.

We should evaluate the perf costs of putting a try/catch into element registration, throw some semblance of a usable error message, and mention `polylint` more prominently in the README for when things go weird.

@ebidel People should learn to use linters. There are whole classes of errors that can be found statically that a linter can reduce. Polylint isn't a panacea, but it worth training developers on tools that will help them avoid mistakes.

No application environment is as forgiving to developer mistakes as the web is, but in more cases than not, the right behavior is to stop and fail, rather than arriving at a broken state.

If a developer is building an application, they should be linting, they should be testing, they should be evaluating their runtime performance and payload sizes. This is why the tools exist, to make this easier.

However, I do see the value of an easier learning curve, especially at the early stage of prototyping.

I think we might be able to slice this the right way with a file a user only includes for development, something that uses `window.onerror` to parse Error events at runtime and offers up suggestions for tools to help diagnose the issues statically.

WDYT @ebidel, @sorvell ?

MeTaNoV:
1000% agree on linting, I have raised 2 issues ([78](https://github.com/PolymerElements/seed-element/issues/78) and [79](https://github.com/PolymerElements/seed-element/issues/79)) on mentioning it on the `<seed-element>` to be used far before web-cpmponent-testing...

ebidel:
> People should learn to use linters.

No arguments from me. Any tool we recommend to devs, we should also be dogfooding on our own elements :)

> the cost of having users get frustrated and stop trying is worse than any performance hit we could incur.

YES. This! If we want to 10x Polymer adoption, stuff like this is low hanging fruit.

@azakus that's an interesting idea if we can make it work. An easy drop in solution would make everyone happy. Not sure it will work for all cases though without changes to Polymer. For example: http://jsbin.com/quyagumawu/edit?html,output. After the 2nd `doc.registerElement` throws, no further scripts run :(

sorvell:
Note the behavior we get here is governed by Polymer's use of custom elements. If the behavior of custom elements is not desirable, bugs should be filed against the spec and Chrome's implementation. In the meantime, Polymer can smooth over the experience, but ideally, the platform does what we want.

There are 2 separate issues here: (1) content of the error messages, (2) behavior of re-registering an existing element. Let's take each of these issues in turn.

First, the error messages. Here's an example of native (when run in Chrome) error messages: http://jsbin.com/lihoxukuda/edit?html,console,output. The re-registration error seems pretty good, but the error for an invalid name does not provide info about the `-`'. To streamline, perhaps we can pass through the native error message and simply append the link to the Polymer docs.

Now re-registration behavior. It's not obvious to me if always squelching this exception is good. In some cases, this is a minor issue and the page would otherwise be mostly functional. In other cases, it's a major problem and making information about the error less obvious would make it harder to fix.

ebidel:
> perhaps we can pass through the native error message and simply append the link to the Polymer docs.

If we have to catch the error anyways (to link to the docs), might as well provide a more helpful message inline. We call out needing a "-" in many places throughout the docs, but people still miss it even when we point them to it. I'm also worried about deep links changing with our new site.

I'll file a crbug.com for improved error messages in spec/Blink. That stuff takes time though. This is where libraries like Polymer can help.

For reference, here's what x-tags does for element registration error handling:
https://github.com/x-tag/core/blob/30f25ec97a724a7dbb95765c1bbee28bcf2df142/dist/x-tag-core.js#L2081-L2092

> Now re-registration behavior. It's not obvious to me if always squelching this exception is good. In some cases, this is a minor issue and the page would otherwise be mostly functional. In other cases, it's a major problem and making information about the error less obvious would make it harder to fix.

I'm looking at this from the other side. One bad actor (component) shouldn't be able to take down an entire app. Squelching is a good idea for that reason. A couple of other examples this mitigates:

1. for big apps, vulcanized bundles are really tough to get right today. web-components-shard _may_ be a solution, but squelching this exception would help users naively use vulcanized bundles, today.
2. (more hypothetical) In a world of http2 and CDN components, some 3rd party refactored x-header to use a new x-subheader element. cnn.com was already using x-subheader on their site and now it's loading as a blank page :(

The proposed changes in this PR help solve the problem of a bad actor and improved error messaging. It doesn't suppress information about errors or make them less obvious on how to fix. It makes them more obvious :)

ebidel:
Ping.

ebidel:
@sorvell is there anything to use here or should we close entirely?

sorvell:
These types of warnings are blocked on [this issue](https://github.com/Polymer/polymer/issues/3367). We'll leave this PR open so it can be considered after it is unblocked.

ebidel:
This PR has been open for 7 months. Can we move forward on https://github.com/Polymer/polymer/issues/3367 (5 mos old) or close with a reason.

@tjsavage @kevinpschaaf @mattsmcnulty

joh1234:
Update README
https://goo.gl/JBljdR

sorvell:
Closing this since we're not adding new features to Polymer 1.x. In Polymer 2.x the user calls `customElements.define` and just gets the native error message if the element is already defined.

PR #2737, opened 2015-11-14
Fixes custom-style ordering issue and perf concern related to dom-module

### LINKED ISSUE #2708: `dom-module` forces `CustomElements.upgradeAll` each time an element without a dom-module is registered
Code in `dom-module` added to ensure non-upgraded dom-modules can be found by registering elements has an undesired side-effect of doing potentially expensive tree walks for each element registered that intentionally has no `dom-module`.  Issue is to consider avoiding these expensive tree-walks somehow.

kevinpschaaf:
Fixes #2692.
Fixes #2708.


kevinpschaaf:
LGTM

kevinpschaaf:
The failing test is due to https://github.com/webcomponents/webcomponentsjs/issues/443, which will be resolved shortly.

PR #3005, opened 2015-11-18
Simpler travis config

azakus:
Use a simpler, serial test structure:
- one build for a push, and one for a pr
  -  rather than 2 for each
- Use nodejs 4 LTS for faster installation
- Hardcode firefox 42.0 to work around travis-ci/travis-ci#5082


kevinpschaaf:
LGTM

Issue #3014, opened 2015-11-18
new Event breaks IE11


davidlwatsonjr:
The new code to cache events is using `new Event`:

https://github.com/Polymer/polymer/blob/cb68ce54ebdcfb2d6f7342e801574881fb158076/src/standard/utils.html#L245

This breaks IE11:
`Object doesn't support this action`

Should this be a CustomEvent instead of an Event?

davidlwatsonjr:
https://github.com/Polymer/polymer/pull/3015

azakus:
Do you have webcomponents.js v0.7.18 installed? We provide a polyfill for the `Event` constructor.

jshcrowthe:
I was having this issue as well. Updating to `v0.7.18` solved it for me.

azakus:
@jshcrowthe Polymer 1.2.3+ depends on webcomponents `v0.7.18`.
Are you saying that bower didn't update them together for you?

jshcrowthe:
I wasn't consuming webcomponentsjs via bower due to some weirdness in my asset pipeline. The polymer update did request `v0.7.18` I just had to manually bring it over.

davidlwatsonjr:
Closing as 0.7.18 fixes the issue.

PR #3015, opened 2015-11-18
Use CustomEvent instead of Event

davidlwatsonjr:
Per the issue described here, https://github.com/Polymer/polymer/issues/3014, new Event breaks functionality in IE11. This is easily resolved by using new CustomEvent as was previously. This also makes more sense as we are, indeed, creating a custom event.


azakus:
Base issue is closed, so I'm closing this PR.

PR #3019, opened 2015-11-18
Avoid throwing with invalid keys/paths. Fixes #3018.

### LINKED ISSUE #3018: Stricter key & path tests may cause breakage with non-defensive code
Recent changes caused what used to gracefully fail on unexpected input now inadvertently throw, which could cause unexpected breakage:

* `Polymer.Collection.getItem(key)` with an `undefined` key now throws where it didn't previously
* `Polymer.Base.notifyPath(path, value)` where the path cannot be dereferenced to the leaf now throws where it didn't previously

kevinpschaaf:


sorvell:
LGTM

PR #3027, opened 2015-11-19
Suppress warnings and expected errors in test suite

### LINKED ISSUE #3026: Clean up expected errors in test console output
When the test suite is run, a couple of errors/warnings pop up in the console. While the initial reaction is "I broke something", all errors/warnings are actually expected. Therefore the thrown errors/messages have to be caught and verified to not confuse developers when running the full test suite.

Behaviour is null correctly generates the following warning:
image

Undefined observers correctly generate the following warnings:
image

Resolve url correctly resolves, but generates a 404 on images:
image

Scoped styling generates 404s:
image

Custom style generates a 404:
image

Left-over console statements:
image
image

JeremybellEU:
This suppresses several warnings and console output to not confuse developers when running the test suite.

Fixes #3026

### DIFF
--- a/test/unit/behaviors.html
+++ b/test/unit/behaviors.html
@@ -127,6 +127,27 @@ suite('multi-behaviors element', function() {
     assert.equal(el.attributes.user.value, 'user', 'Behavior hostAttribute overrode user attribute');
   });
 
+  test('behaviour is null generates warning', function() {
+    var warned = false, oldWarn = Polymer.Base._warn;
+    Polymer.Base._warn = function (message) {
+      assert.match(message, /behavior is null/);
+      warned = true;
+    };
+
+    Polymer({
+
+      behaviors: [
+        null
+      ],
+
+      is: 'behavior-null'
+
+    });
+
+    assert.equal(warned, true, 'Null behaviour should generate warning');
+    Polymer.Base._warn = oldWarn;
+  });
+
 });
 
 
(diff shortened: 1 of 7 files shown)


azakus:
Seems ok to me, but why did the zero size images change :confused:

test/unit/behaviors.html
```diff
@@ -127,6 +127,27 @@
     assert.equal(el.attributes.user.value, 'user', 'Behavior hostAttribute overrode user attribute');
   });
 
+  test('behaviour is null generates warning', function() {
+    var warned = false, oldWarn = Polymer.Base._warn;
+    Polymer.Base._warn = function (message) {
```
azakus: It looks like `this._warn` is used for this warning, so let's just override the `_warn` of this element, if possible.

JeremybellEU: I indeed tried to do this, but sadly it was not possible. The warning is generated on call-time of `Polymer({})` and therefore there is no reference to `this` yet to override the function. Therefore I had to override the prototype, call `Polymer({})` which generated the warning, and then verify that upon creating this warning was generated.

JeremybellEU:
My original reported issue was lackluster, I apologize for that. The 'empty' images were added to prevent 404's from happening. I included all screenshots of warnings/errors in #3026 and they are fixed in this PR

test/unit/behaviors.html
azakus: Ah that makes sense.

azakus:
LGTM

PR #3066, opened 2015-11-20
Fixes #3065: Add dom-repeat.renderedItemCount property

### LINKED ISSUE #3065: Enhancement Request:  Expose a 'filtered count' property on dom-repeat
I need to apply specific logic based on the number of results displayed in a dom-repeat after applying a filter.  It would be great to listen for changes to a simple property that exposes the count as it changes.

arthurevans:
I'm not sure whether this is desirable or not, but it seems like the `dom-repeat` has this information readily available, and extracting it yourself relies on either accessing internal properties (`_instances`) or using side effects (`domRepeat.indexForElement(domRepeat.previousSibling)+1` for example).

Seems like it could either be exposed as a property with `notify`, or simply as a property on the `dom-change` event. The event property might be the most efficient, but an element property seemed more Polymeric.

Feel free to close if this isn't the right API or I'm going about it the wrong way.

### DIFF
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -199,7 +199,8 @@ Then the `observe` property should be configured as follows:
        */
       renderedItemCount: {
         type: Number,
-        notify: true
+        notify: true,
+        readOnly: true
       },
 
       /**
@@ -450,7 +451,7 @@ Then the `observe` property should be configured as follows:
       // the same item.
       this._pool.length = 0;
       // Set rendered item count 
-      this.renderedItemCount = this._instances.length;
+      this._setRenderedItemCount(this._instances.length);
       // Notify users
       this.fire('dom-change');
       // Check to see if we need to render more items
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -192,6 +192,17 @@ Then the `observe` property should be configured as follows:
       delay: Number,
 
       /**
+       * Count of currently rendered items after `filter` (if any) has been applied.
+       * If "chunking mode" is enabled, `renderedItemCount` is updated each time a 
+       * set of template instances is rendered.
+       * 
+       */
+      renderedItemCount: {
+        type: Number,
+        notify: true
+      },
+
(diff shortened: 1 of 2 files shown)


src/lib/template/dom-repeat.html
```diff
@@ -192,6 +192,17 @@
...
 
       /**
+       * Count of currently rendered items after `filter` (if any) has been applied.
+       * If "chunking mode" is enabled, `renderedItemCount` is updated each time a 
+       * set of template instances is rendered.
+       * 
+       */
+      renderedItemCount: {
+        type: Number,
+        notify: true
```
azakus: Let's add `readOnly: true`, because setting it doesn't make any sense.

src/lib/template/dom-repeat.html
```diff
@@ -438,6 +449,8 @@
       // `item` may not be sufficient if the pooled instance happens to be
       // the same item.
       this._pool.length = 0;
+      // Set rendered item count 
+      this.renderedItemCount = this._instances.length;
```
azakus: With readOnly true, this line should be
```js
this._setRenderedItemCount(this._instances.length);
```

azakus:
Can you add a test case for reacting to `renderedItemCount` changing?

src/lib/template/dom-repeat.html
arthurevans: Good point!

arthurevans:
I'm not sure I how to test that using WCT. Is there a similar test case somewhere I can look at?

Specifically, I don't see any tests in the dom-repeat tests that are catching an event. 

Could it be as simple as something like:

- set items
- render()
- check renderedItemCount
- remove some items
- render()
- check renderedItemCount

Or do I need to do some kind of async thing to catch the event?

arthurevans:
I don't know if this is what you're looking for, but I added a test showing the `renderedItemCount` changing when the repeat is filtered (which was the original use case for this property).

arthurevans:
Ping. Let me know if this is what you intended.

azakus:
LGTM!

PR #3067, opened 2015-11-21
Fixes #3023: Add style support for css selectors type [foo~=bar].

### LINKED ISSUE #3023: [attribute~="value"] selector not working on :host or :host-context
Hello,

the following test case illustrates the problem: http://jsbin.com/rexehifigo/edit?html,output

```html
<dom-module id="my-element">
  <template>
    <style>
      :host {
        display: flex;
      }
      
      :host .detail {
        flex: 1;
      }
      
      :host[layout="master-detail"] .master {
        width: var(--master-width, 40%);
      }
      
      :host[layout~="master-detail"] .master {
        width: var(--master-width, 40%);
      }
     
      :host-context([layout~="master-detail"]) .master {
        width: var(--master-width, 40%);
      }
    </style>
    <div class="master">MASTER</div>
    <div class="detail">DETAIL</div>
  </template>
  <script>
    Polymer({
      is: 'my-element'
    });
  </script>
</dom-module>
```

Polymer generates the following custom style for the element:
```html
<style scope="my-element-0">
    my-element.my-element-0[layout="master-detail"] .master.my-element {
      width:40%;
    }

    my-element.my-element-0[layout~="master-detail"].my-element .master.my-element {
      width:40%;
    }

    [layout~="master-detail"] my-element.my-element-0.my-element .master.my-element,
    my-element.my-element-0[layout~="master-detail"].my-element .master.my-element {
      width:40%;
    }
</style>
```

The first one is generated correctly, the second and third add ```.my-element``` after the attribute selector.

Edit: Included ```:host-context``` test case.

adelarosab:
It was a problem with css separators. It was just looking for ~ as separator but there is a possibility to use ~ as part of an attribute css selector.

```css
element[myattribute~=partial] {
    // some visuals
}
```

Edit: http://www.w3.org/TR/css3-selectors/

### DIFF
--- a/src/lib/style-transformer.html
+++ b/src/lib/style-transformer.html
@@ -250,7 +250,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     var SCOPE_DOC_SELECTOR = ':not([' + SCOPE_NAME + '])' +
       ':not(.' + SCOPE_NAME + ')';
     var COMPLEX_SELECTOR_SEP = ',';
-    var SIMPLE_SELECTOR_SEP = /(^|[\s>+~]+)([^\s>+~]+)/g;
+    var SIMPLE_SELECTOR_SEP = /(^|[\s>+~]+)([^\s>+~=]+)/g;
     var HOST = ':host';
     var ROOT = ':root';
     // NOTE: this supports 1 nested () pair for things like
--- a/test/unit/css-parse.html
+++ b/test/unit/css-parse.html
@@ -49,6 +49,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         margin-left: 0 !important;
       }
     }
+
+    .foo[bar~=baz] {
+      background: black;
+    }
   </style>
 
   <style id="test-ignore">
@@ -123,6 +127,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       assert.equal(tree.rules[0].selector, ':host', 'unexpected selector');
       assert.equal(tree.rules[2].selector, '@-webkit-keyframes fill-unfill-rotate', 'unexpected selector in keyframes');
       assert.equal(tree.rules[3].selector, '@media (max-width: 400px)', 'unexpected selector in @media');
+      assert.equal(tree.rules[4].selector, '.foo[bar~=baz]', 'unexpected selector in attributes');
     });
 
     test('rule cssText parse', function() {
--- a/test/unit/css-parse.html
+++ b/test/unit/css-parse.html
@@ -117,7 +117,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     });
 
     test('css rules parse', function() {
-      assert.equal(tree.rules.length, 4, 'unexpected number of rules');
+      assert.equal(tree.rules.length, 5, 'unexpected number of rules');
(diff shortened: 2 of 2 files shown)


adelarosab:
I signed it!

samccone:
@adelarosab would you mind adding a test for this functionality?

@nazar-pc would you mind reviewing this? 

Thanks!!!! :sparkles:

adelarosab:
Yeah, you caught me adding test for this.

adelarosab:
I'm having a problem with my nodejs version. I cannot test in my local machine. 

Sorry.

nazar-pc:
LGTM :+1:

JeremybellEU:
Correct me if I am wrong, but I think this breaks the following cases:

```css
.foo[bar=baz] {
}
.foo[bar=baz] > * {
}
```
image

JeremybellEU:
This regex seems to catch this edge-case:

```
(^|[\s>+~]+)([^\s>+~=\[]*(?:\[.+?\])?)
```

E.g. A CSS seperator followed by anything that is not a seperator followed by (optionally) a selector between squared brackets.

image

adelarosab:
Hi @JeremybellEU,

Nope, it doesn't break anything. You can test exactly same case you did with the jsbin from #3023.

Regards

JeremybellEU:
Hm I checked your PR locally and I am indeed unable to reproduce what I expected.

When I output the value of `m` at https://github.com/adelarosab/polymer/blob/master/src/lib/style-transformer.html#L165 for the test-case
```
:host[attribute="true"] {
  background: white;
}
```
the console output is `:host[attribute` as expected by the updated regex. However, this does not seem to effect the output of the CSS.

After some more checking, it replaces `:host[attribute` with `my-element[attribute` and then substitutes this back into the selector. Therefore the case is fixed, but the regex does not parse everything. If the regex is not supposed to parse everything, this is fine. If it however should parse everything, then the update breaks the existing selectors.

adelarosab:
Hi @JeremybellEU,

I haven't made this code. I'm not 100% sure if the purpose of regexp was match everything. The only one who can answer that question is @sorvell. 

I suppose @samccone or @nazar-pc, could bring light to this.

I dare to guess you are right. The code still works because they analyse selector just for host replacement purposes.

azakus:
This change seems ok to me, but @sorvell should really take a look

sorvell:
I don't think this regex change will be sufficient. The problem is that `~` has 2 meanings in css: it's a [sibling combinator](https://www.w3.org/TR/selectors/#general-sibling-combinators) and part of an attribute selector `[attr~=value]`. The regex needs to match the first case and not the second. 

We also don't have a sibling combinator test or this change would have made it fail! So we need to add a test for this as well.

Sorry for the delay. If you're still willing to work on this that'd be great. Otherwise we'll close this PR and put #3023 back in our queue.

adelarosab:
I'll do it ;)

Also i'm worry about performance because the regex we are looking for increase complexity a lot `(^|[\s>+~]+)((?:\[.+?\]|[^\s>+~=\[])+)`, I think.

nazar-pc:
Well, talking about performance it is better to write some simple benchmark to measure what difference we are talking about. And how it affects component's performance in general.

sorvell:
Fixed via #3413. Thanks!

PR #3068, opened 2015-11-21
ES5 strict doesn't like function declarations inside inner blocks.

rictic:
(review changes with [github's whitespace-ignoring view](?w=1) to reduce the solid-green diff line-ending fight)

### DIFF
--- a/src/lib/custom-style.html
+++ b/src/lib/custom-style.html
@@ -11,7 +11,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <link rel="import" href="style-transformer.html">
 <link rel="import" href="style-defaults.html">
 
-<!-- 
+<!--
 
 The `custom-style` extension of the native `<style>` element allows defining styles
 in the main document that can take advantage of several special features of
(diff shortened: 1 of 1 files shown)


rictic:
This came up with some code that compiled everything down to one 'use strict' file, and jscompiler becomes quite unhappy in that case.

samccone:
to be clear the error here is that because there was not var, fn was being defined in the global scope.

I wonder if integrating a tool like jshint or jscs or eslint or (whatever tool here) could prevent these kind of issues in the future.

rictic:
The function declaration wasn't pulled up to global scope, but in non-strict code it's hoisted to the top of the containing function, whereas in strict code it's only defined in its containing block.

Closure compiler probably throws the error here because this behavior can be confusing. The original code here isn't wrong, in strict mode or in lax mode.

samccone:
screenshot from 2015-11-23 12 50 07

hmm interesting! there must be another fn defined somewhere I guess

rictic:
Try:

function lax() {if (true) { function g() {return 'yes';}} return g()}
function strict() {'use strict'; if (true) { function g() {return 'yes';}}
return g()}

lax(); // "yes"
strict(); // throws ReferenceError

azakus:
LGTM

PR #3075, opened 2015-11-23
Put beforeRegister in the behaviorProperties.

### LINKED ISSUE #3076: `beforeRegister` called twice when added from a behavior
beforeRegister is being copied into the element prototype, causing doBehavior_ to call beforeRegister twice.

zhaoz:
Fix a problem where beforeRegister was being copied into the element prototype, causing doBehavior_ to call beforeRegister twice.

Fixes #3076.


kevinpschaaf:
LGTM

PR #3079, opened 2015-11-24
loosen isLightDescendant's @param type to Node

danbeam:
`isLightDescendant` doesn't use any `HTMLElement`'s API and can validly be simply `Node` when processing events on `document`.

also, I've removed the trailing `{...=}` because `Polymer.dom()` requires `?Node|?Event` as far as I can tell (from the externs) and `node` probably shouldn't be optional.  if somewhere really does use `.isLightDescendant()`, closure should let us know.

note: this doesn't address the `EventTarget` use-case (i.e. events on `window`).

/cc @notwaldorf @sorvell

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -402,7 +402,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * Checks whether an element is in this element's light DOM tree.
      *
      * @method isLightDescendant
-     * @param {Node} node The element to be checked.
+     * @param {?Node} node The element to be checked.
      * @return {Boolean} true if node is in this element's light DOM tree.
      */
     isLightDescendant: function(node) {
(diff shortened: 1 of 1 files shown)


src/standard/utils.html
```diff
@@ -402,7 +402,7 @@
      * Checks whether an element is in this element's light DOM tree.
      *
      * @method isLightDescendant
-     * @param {HTMLElement=} node The element to be checked.
+     * @param {Node} node The element to be checked.
```
jklein24: Same comment here about making this `?Node` or `!Node`

danbeam: Done. (went with ?Node because Polymer.dom() accepts it)

jklein24:
LGTM. Note this won't have an effect because this code isn't actually compiled yet. Good for documentation nonetheless.

danbeam:
the externs are compiled and they should both match

jklein24:
I definitely agree although unfortunately, that's not the case in many other areas of polymer core where the externs are more accurate or thorough than the annotations in the code itself. I agree we should be bridging that gap and will need to do so if we want to compile this repo.

sorvell:
LGTM

PR #3082, opened 2015-11-24
Revert fromAbove in applyEffectValue. Add test. Fixes #3077.

### LINKED ISSUE #3077: Listeners on data-change events may not be called if the property is bound
We recently changed property sets at bind data propagation so that they avoid firing data change events. This was done as an optimization: these events are only necessary for 2-way binding and if the value is being sent from the bindor, then no notification is necessary. The problem is that users who have added event listeners for data change events (e.g. `foo-change`) will not be notified when the property changes due to binding. We may need to consider backing this optimization out.

kevinpschaaf:


sorvell:
LGTM

PR #3083, opened 2015-11-24
Fix caching the original target causing memory leaks.

beaudesigns:
Caching the original target in `t` causes a reference to be held for
that target in detached DOM nodes.

I am binding to the `tap` gesture for a “close” action. The element is removed on the close action.

screen shot 2015-11-24 at 11 29 20 am

screen shot 2015-11-24 at 11 29 55 am

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -516,7 +516,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (!hasLeftMouseButton(e)) {
         return;
       }
-      var t = Gestures.findOriginalTarget(e);
       var self = this;
       var movefn = function movefn(e) {
         var x = e.clientX, y = e.clientY;
@@ -529,7 +528,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             self.info.state = 'end';
             untrackDocument(self.info);
           }
-          self.fire(t, e);
+          self.fire(Gestures.findOriginalTarget(e), e);
           self.info.started = true;
         }
       };
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -428,22 +428,21 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (!hasLeftMouseButton(e)) {
         return;
       }
-      var t = Gestures.findOriginalTarget(e);
       var self = this;
       var movefn = function movefn(e) {
         if (!hasLeftMouseButton(e)) {
-          self.fire('up', t, e);
+          self.fire('up', Gestures.findOriginalTarget(e), e);
           untrackDocument(self.info);
         }
       };
       var upfn = function upfn(e) {
         if (hasLeftMouseButton(e)) {
-          self.fire('up', t, e);
+          self.fire('up', Gestures.findOriginalTarget(e), e);
         }
         untrackDocument(self.info);
       };
       trackDocument(this.info, movefn, upfn);
-      this.fire('down', t, e);
+      this.fire('down', Gestures.findOriginalTarget(e), e);
     },
     touchstart: function(e) {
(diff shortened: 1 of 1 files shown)


samccone:
Hi @beaudesigns, would you mind creating or describing your test case for this? I would love to reproduce locally, and perhaps add a test back into polymer around this leaky behavior

beaudesigns:
@samccone We are investigating now. I created a JS Bin test to try to replicate it and am seeing different behavior. It's possible another feature is messing with it. Will report back soon.

samccone:
sounds good, jsbin is probably a bad env for testing memory leaks :) Just a simple html doc should do just fine... let me know if you need any help around this.

beaudesigns:
Sorry for the wait:

Here is a working example: https://gist.github.com/beaudesigns/f3398c59ed7542461598

What we found is that it's actually the `down` gesture that creates the leak, not the `tap` gesture. In light of that information, there may be a better fix than this pull request. I'm anxious to see what you brilliant people think. :)

If the example needs further explanation, let me know. (run it with the console open).

samccone:
cc @ebidel ... is https://gist.github.com/beaudesigns/f3398c59ed7542461598#file-polymer-memory-leak-html-L32 the "blessed" way to remove an upgraded element from the DOM?

beaudesigns:
One potential issue with my pull request is that `e` might be different if we're inside the `movefn` and `upfn` closures.

beaudesigns:
@samccone I'm curious to know if it's the "blessed" way as well. Though, it has nothing to do with the leak in this case. Even if the element is not removed, the leak occurs, the example removes the element just to demonstrate a more complete use case.

samccone:
>  Even if the element is not removed

I am confused by this ^

So long as the element is in the DOM tree I would expect there to be a js reference to the DOM node.... Once the element is removed from the DOM it should ideally unbind and free the reference, thus no leaks.


I will take a more in depth look in a bit

samccone:
Alright great spot @beaudesigns this totally is a real leak (if removing the DOM in this manner is the correct way of doing things)

screen shot 2015-11-24 at 10 45 04 am

It looks like we are retaining the entire the entire element... very interesting... I will defer to those that know the code a bit better to explain what is going on here, but I feel pretty confident in confirming this issue.

shepbox:
@samccone You are correct. @beaudesigns point was there is a reference to the DOM node via the path `window.Polymer.Gestures.gestures.down.info.movefn<function scope>closure.t` as well as being apart of the DOM Tree. So when removed from the DOM tree it would still have a reference.

I also noticed that this leak might happen with the `track` gesture. https://github.com/beaudesigns/polymer/blob/master/src/standard/gestures.html#L519

azakus:
I expect only one element reference to be held in this way at a time.
The next down will set a new function, which should remove the reference to the closure holding the target.

samccone:
After what @abarth said I have updated my test case https://gist.github.com/samccone/8ca56643c8e1aa4f930a

Which does confirm what @azakus is saying

screen shot 2015-11-24 at 2 02 07 pm

I still am a bit curious as to why the node is retained at all after tear down even with a force GC... but more investigation needed

azakus:
I'm not surprised the node hangs around, the function holding the reference is only dropped by the next mousedown handled.

It's probably the easiest change to null out the functions after mouseup.

samccone:
:+1: for setting to null

samccone:
also, after thinking about my test case more... because I was programatically tripping this method, i was not going through the gesture code path :) let me run this again

samccone:
Alright I have updated my test case to manually "click" so I go through the gesture path

screen shot 2015-11-25 at 9 57 38 am

^ yay we can now see the leak

screen shot 2015-11-25 at 10 00 44 am

and as @azakus said we are just holding onto the last one :)

beaudesigns:
:+1: 

Thanks so much for your help on this @samccone and @azakus.

PR #3085, opened 2015-11-24
set class attribute instead of using classname

### LINKED ISSUE #2407: SVG elements inconsistently get scoped
It appears that there are 2 code paths happening for style scoping... one here: https://github.com/Polymer/polymer/blob/master/src/standard/styling.html#L122 which does not treat SVG consistently due to the possibility that SVG elements can sometimes have .className == an SVGAnimatedString object... this will cause scopeSubtree to "skip" them since className = "new string" will just fail silently.

It seems the fix for #1490 should be expanded to be more generic to encompass all cases where styling my cross path with SVG elements.

Huge thanks to @arthurevans for helping track down why *some* SVG elements worked and others did not and still some did only under certain conditions.

Tagging @sorvell since the related fixes were primarily his work.

azakus:
Fixes #2407

### DIFF
--- a/src/standard/styling.html
+++ b/src/standard/styling.html
@@ -68,7 +68,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           }
         }
         cssText += styleUtil.cssFromModule(this.is);
-        // check if we have a disconnected template and add styles from that 
+        // check if we have a disconnected template and add styles from that
         // if so; if our template has no parent or is not in our dom-module...
         var p = this._template && this._template.parentNode;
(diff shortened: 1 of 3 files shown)


kevinpschaaf:
Can you add a test for the case of doing `appendChild` (not using `Polymer.dom`) of an SVG tree and then calling `scopeSubtree` on it and verifying the SVG gets correctly scoped?

azakus:
@kevinpschaaf Added!

kevinpschaaf:
Thanks!  LGTM.

PR #3110, opened 2015-11-30
Patch rAF to setTimeout to reduce flakiness on CI.

kevinpschaaf:


test/unit/dom-repeat.html
```diff
@@ -3648,6 +3648,16 @@
 
     suite('chunked rendering', function() {
 
+      // Patch requestAnimationFrame to setTimeout to reduce flakiness on CI
```
samccone: hmm this is interesting, I would think the same issue that would cause requestAnimationFrame to act flakey would also impact setTimeout due to how backgrounded setIntervals work in chrome

sorvell:
LGTM

PR #3111, opened 2015-11-30
Fixes #3108. Moves `debounce` functionality from polymer-micro to pol…

### LINKED ISSUE #3108: The `debounce` method is non-functional in polymer-micro
Currently polymer-micro puts debounce methods on the element's prototype but does not load the Polymer.Debounce helper library to make these methods functional. It doesn't make sense to include `debounce` at the micro layer and since this has never worked, we should consider moving it to polymer-mini instead.

sorvell:
…ymer-mini. The functionality belongs at the mini tier and was never actually functional in micro.


kevinpschaaf:
LGTM

PR #3114, opened 2015-11-30
importHref loads imports, async. Fixes #3113

### LINKED ISSUE #3113: importHref should load imports asynchronously
`Polymer.Base.importHref` doesn't load imports asynchronously. In apps, this can cause layout flashes as the import is added to the DOM and loads. I've witnessed this on a couple of apps were we're using `importHref` to dynamically load bundles, asynchronously.

Instead, loading it with the `async` attr take it out of rendering and solves the issue.

ebidel:
This changes the default behavior of `importHref` to load imports async, taking them out of the rendering path. Witnessing this is a a couple of apps we're working on, not doing so causes layout flashes as the import is added to the DOM and loads.

Adds an optional `async` arg to `importHref`. Defaults to true for perf and similar to xhr.

R: @kevinpschaaf @sorvell @azakus

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -164,7 +164,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     /**
      * Returns a list of nodes distributed to this element's `<content>`.
      *
-     * If this element contans more than one `<content>` in its local DOM,
+     * If this element contains more than one `<content>` in its local DOM,
      * an optional selector may be passed to choose the desired content.
      *
      * @method getContentChildNodes
@@ -181,7 +181,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * Returns a list of element children distributed to this element's
      * `<content>`.
      *
-     * If this element contans more than one `<content>` in its
+     * If this element contains more than one `<content>` in its
      * local DOM, an optional selector may be passed to choose the desired
      * content.  This method differs from `getContentChildNodes` in that only
      * elements are returned.
@@ -198,7 +198,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       });
     },
 
-    
+
     /**
      * Dispatches a custom event with an optional detail value.
      *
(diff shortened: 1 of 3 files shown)


ebidel:
Should we use this or make the default `false`? Async by default would be preferrable though!

JeremybellEU:
:+1:

src/standard/utils.html
```diff
@@ -358,12 +358,20 @@
      *   loaded.
      * @param {Function} onerror Callback to notify when an import
      *   unsuccessfully loaded.
+     * @param {boolean} optAsync True if the import should be loaded `async`.
+     *   Defaults to `true`.
      * @return {HTMLLinkElement} The link element for the URL to be loaded.
      */
-    importHref: function(href, onload, onerror) {
+    importHref: function(href, onload, onerror, optAsync) {
```
azakus: I'd say we make this default to `false` for back-compat.

azakus:
LGTM with nit

src/standard/utils.html
ebidel: Done

ebidel:
PTAL

src/standard/utils.html
```diff
@@ -358,12 +358,20 @@
...
+     *   Defaults to `false`.
      * @return {HTMLLinkElement} The link element for the URL to be loaded.
      */
-    importHref: function(href, onload, onerror) {
+    importHref: function(href, onload, onerror, optAsync) {
       var l = document.createElement('link');
       l.rel = 'import';
       l.href = href;
+
+      optAsync = typeof optAsync !== 'undefined' ? optAsync : false;
```
azakus: This can just be `optAsync = Boolean(optAsync)` since the default is `false`.

azakus:
Last nit, then I can merge :)

src/standard/utils.html
ebidel: Done

ebidel:
PTAL

PR #3160, opened 2015-12-04
Ensure dom-if in host does not restamp when host detaches. Fixes #3125.

### LINKED ISSUE #3125: When element is detached and reattached, its dom-ifs are torn down and restamped
This invalidates references to the dom inside the dom-if.

e.g. http://jsbin.com/vicowoxohi/1/edit?html,console,output -- the timer changes updates on restamp

kevinpschaaf:


test/unit/dom-if.html
```diff
@@ -497,19 +504,132 @@
...
+          assert.equal(clients.length, 0);
+          outerContainer.appendChild(domif);
+          setTimeout(function() {
+            var clients = outerContainer.querySelectorAll('x-client');
+            assert.equal(clients[0].uid, 0);
+            assert.equal(clients[1].uid, 1);
+            assert.equal(clients[2].uid, 2);
+            assert.equal(clients[1].previousElementSibling, clients[0]);
+            assert.equal(clients[2].previousElementSibling, clients[1]);
+            assert.equal(domif.previousElementSibling, clients[2]);
```
JeremybellEU: These 6 equals are used a number of times, with just different values. It might be possible to extract a function for this, with passing `clients` and the start number for the first 3 equals (e.g. `0` in this case and `6` for the test `move host with domif (clients persist)`). This would greatly reduce the amount of duplicate code and increase maintainability of the tests.

src/lib/template/dom-if.html
```diff
@@ -74,7 +74,9 @@
 
     detached: function() {
       // TODO(kschaaf): add logic to re-stamp in attached?
```
sorvell: This comment can be deleted now, right?

test/unit/dom-if.html
sorvell: We tend not to favor making indirections like this specifically in tests. The idea is to  try to make the asserts as straightforward as possible.

sorvell:
LGTM

PR #3161, opened 2015-12-05
Custom setProperty for bindings to hidden textNodes. Fixes #3157.

### LINKED ISSUE #3157: Text node children with bindings cause dom-if templates to update incorrectly.
(Relaying a bug @kevinpschaaf reduced offline.) If a text node containing bindings is a child of a dom-if template, text updates don't always work as expected.

[JSBin](http://jsbin.com/xipetopesa/1/edit?html,output)

1. Type into field.
2. Check the checkbox; the text shown is correct.
3. Uncheck the checkbox.
4. Type into field.
5. Text incorrectly appears in dom-if even though the checkbox is not checked. (Text would be correct if visible.)
6. Click checkbox to show text; text is shown with previous value (as entered in step 1).

(The binding within the span works as expected throughout.)

kevinpschaaf:


src/lib/template/templatizer.html
```diff
@@ -168,6 +170,13 @@
       }
     },
 
+    __setPropertyImpl: function(property, value, fromAbove, node) {
+      if (node && node.__hideTemplateChildren__ && property == 'textContent') {
+        property = '__polymerTextContent__';
+      }
+      this.__setPropertyOrig(property, value, fromAbove, node);
+    },
+
```
kevinpschaaf: Not thrilled with patching this in the hot-path here; `applyEffectValue` would be slightly better but its not as easy to patch...

sorvell:
Is it true that if we made bindings not propagate to nodes controlled by `dom-if` when its `if` is false that this issue would not occur? If so, let's factor this into our thinking about whether we want to adjust `dom-if` to have this behavior.

sorvell:
LGTM

PR #3163, opened 2015-12-05
Non-destructive `@keyframes` rule transformation.

### LINKED ISSUE #2399: Custom properties cannot be set in @keyframe rules
Polymer's property system currently does not allow properties to be used within @keyframe rules. These rules are ignored right now because they do not require selector modification. The custom property system should be made to account for this and allow properties to be consumed in @keyframe rules.

Simple repro: http://jsbin.com/mopuwe/edit?html,output

cdata:
Previously, the transformer did not disambiguate selectors in `@media`
blocks and keyframes in `@keyframes` blocks. Now, the transformer can
safely transform `@keyframes` blocks. Before a selector is transformed,
if the selector has a parent, it is checked. If the checked parent is a
`@keyframes` rule, the selector transformation is skipped.

Element-specific `@keyframes` are suffixed with the scoped element name.
For example, `@keyframes foo` in an element scoped with `x-el-0` will by
transformed to `@keyframes foo-x-el-0`. References to that animation in
the element's local styles will be updated as well.

Added tests for the new keyframes transformation.

~~This change also fixes the accidental introduction of a vestigial
semicolon at the end of some generated styles.~~

 - [x] Feedback on design
 - [x] Unit / regression tests
 - [x] Test `@apply` in keyframes
 - [x] [Ensure that `scopeId` doesn't get stale](https://github.com/Polymer/polymer/pull/3163#discussion_r49239099)

Fixes #2399

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -25,11 +25,15 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // decorates styles with rule info and returns an array of used style
       // property names
       decorateStyles: function(styles) {
-        var self = this, props = {};
+        var self = this, props = {}, keyframes = [];
         styleUtil.forRulesInStyles(styles, function(rule) {
           self.decorateRule(rule);
           self.collectPropertiesInCssText(rule.propertyInfo.cssText, props);
+        }, function onKeyframesRule(rule) {
+          keyframes.push(rule);
         });
+        // Cache all found keyframes rules for later reference:
+        styles._keyframes = keyframes;
         // return this list of property names *consumes* in these styles.
         var names = [];
         for (var i in props) {
@@ -87,7 +91,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var parts = cssText.split(';');
         for (var i=0, p; i<parts.length; i++) {
           p = parts[i];
-          if (p.match(this.rx.MIXIN_MATCH) || p.match(this.rx.VAR_MATCH)) {
+          if (p.match(this.rx.MIXIN_MATCH) ||
+              p.match(this.rx.VAR_MATCH) ||
+              this.rx.ANIMATION_MATCH.test(p) ||
+              styleUtil.isKeyframesSelector(rule)) {
             customCssText += p + ';\n';
           }
         }
@@ -181,6 +188,46 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         rule.cssText = output;
       },
 
+      // Apply keyframe transformations to the cssText of a given rule. The
+      // keyframeTransforms object is a map of keyframe names to transformer
+      // functions which take in cssText and spit out transformed cssText.
+      applyKeyframeTransforms: function(rule, keyframeTransforms) {
+        var input = rule.cssText;
+        var output = rule.cssText;
+        if (rule.hasAnimations == null) {
+          // Cache whether or not the rule has any animations to begin with:
+          rule.hasAnimations = this.rx.ANIMATION_MATCH.test(input);
+        }
+        // If there are no animations referenced, we can skip transforms:
+        if (rule.hasAnimations) {
+          var transform;
+          // If we haven't transformed this rule before, we iterate over all
+          // transforms:
+          if (rule.keyframeNamesToTransform == null) {
+            rule.keyframeNamesToTransform = [];
+            for (var keyframe in keyframeTransforms) {
+              transform = keyframeTransforms[keyframe];
+              output = transform(input);
+              // If the transform actually changed the CSS text, we cache the
+              // transform name for future use:
+              if (input !== output) {
+                input = output;
+                rule.keyframeNamesToTransform.push(keyframe);
+              }
+            }
+          } else {
+            // If we already have a list of keyframe names that apply to this
+            // rule, we apply only those keyframe name transforms:
+            for (var i = 0; i < rule.keyframeNamesToTransform.length; ++i) {
+              transform = keyframeTransforms[rule.keyframeNamesToTransform[i]];
+              input = transform(input);
+            }
+            output = input;
+          }
+        }
+        rule.cssText = output;
+      },
+
       // Test if the rules in these styles matches the given `element` and if so,
       // collect any custom properties into `props`.
       propertyDataFromStyles: function(styles, element) {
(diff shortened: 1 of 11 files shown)


cdata:
@sorvell I have a lingering uncertainty about the semantics of custom property value propagation. Since the "selector" within a `@keyframes` block isn't actually a selector, it seems that it may be necessary to inform the style system that the effective scope is not what it typically expects. For an example of what I mean, consider the following arrangement:

```html
<style is="custom-style">
  :root {
    --left: 0px;
  }

  /* A selector: */
  from {
    --left: 10px;
  }
</style>
<style is="custom-style">
  @keyframes foo {
    /* Not a selector: */
    from {
      left: var(--left);
    }
  }
</style>
<!-- A pathologically named element: -->
<from></from>
```

Ideally, the property would be transformed to `left: 0px`. However, it may be the case that the value will end up being `10px` due to selector matching.

WDYT?

cdata:
I uncommented some ShadowDOM tests since the relevant Chrome bug has been fixed.

Note: I did a lot of cleanup whitespace changes (moving `<style>` tags inside of templates), so viewing the diff with `?w=1` in the query params is recommended.

cdata:
I think there is some missing prefixed stuff that is causing Safari to fail in CI. Checking it out.

test/runner.html
```diff
@@ -56,10 +56,8 @@
       'unit/styling-remote.html',
       'unit/styling-cross-scope-var.html',
       'unit/styling-cross-scope-apply.html',
-      // TODO(sorvell): disable until this Chrome bug is addressed:
-      // https://code.google.com/p/chromium/issues/detail?id=525280
-      // 'unit/styling-cross-scope-var.html?dom=shadow',
-      // 'unit/styling-cross-scope-apply.html?dom=shadow',
+      'unit/styling-cross-scope-var.html?dom=shadow',
+      'unit/styling-cross-scope-apply.html?dom=shadow',
```
cdata: The related Chrome bug was apparently fixed in M46, so I have reactivated these.

src/lib/style-properties.html
```diff
@@ -78,18 +78,13 @@
...
-        var customCssText = '';
         var cssText = rule.parsedCssText;
         // NOTE: we support consumption inside mixin assignment
         // but not production, so strip out {...}
         cssText = cssText.replace(this.rx.BRACKETED, '')
           .replace(this.rx.VAR_ASSIGN, '');
         var parts = cssText.split(';');
-        for (var i=0, p; i<parts.length; i++) {
-          p = parts[i];
-          customCssText += p + ';\n';
```
cdata: This ad-hoc join of the parts was causing a vestigial trailing semi-colon to be introduced in cases where the last item in the `parts` list is an empty string.

src/lib/style-properties.html
```diff
@@ -254,15 +267,46 @@
...
         });
       },
 
+      // Transforms `@keyframes` names to be unique for the current host.
+      // Example: @keyframes foo-anim -> @keyframes foo-anim-x-foo-0
+      _scopeKeyframes: function(rule, scopeId) {
+        rule.keyframesNameRx = new RegExp(rule.keyframesName, 'g');
+        rule.transformedKeyframesName = rule.keyframesName + '-' + scopeId;
+        rule.transformedSelector = rule.transformedSelector || rule.selector;
+        rule.selector = rule.transformedSelector.replace(
```
sorvell: They `scopeId` for a given element may change over time. In this case won't the selector fail to update?

src/lib/style-properties.html
```diff
@@ -254,15 +267,46 @@
...
+          // For non-ShadowDOM, we transform all known keyframes rules in
+          // advance for the current scope. This allows us to catch keyframes
+          // rules that appear anywhere in the stylesheet:
+          for (var i = 0, keyframesRule = keyframesRules[i];
+               i < keyframesRules.length;
+               keyframesRule = keyframesRules[++i]) {
+            this._scopeKeyframes(keyframesRule, scopeSelector);
+            keyframeTransforms[keyframesRule.keyframesName] = function(rule, cssText) {
+              return cssText.replace(rule.keyframesNameRx, rule.transformedKeyframesName);
+            }.bind(this, keyframesRule);
```
sorvell: Prefer using a factory function instead of Function.bind or perf reasons.

src/lib/style-properties.html
```diff
@@ -179,6 +178,20 @@
         rule.cssText = output;
       },
 
+      // Apply keyframe transformations to the cssText of a given rule. The
+      // keyframeTransforms object is a map of keyframe names to transformer
+      // functions which take in cssText and spit out transformed cssText.
+      applyKeyframeTransforms: function(rule, keyframeTransforms) {
+        var output = rule.cssText;
+        if (this.rx.ANIMATION_MATCH.test(output)) {
```
sorvell: Let's memo'ize this on the rule.

sorvell: In fact since we don't expect the of animation names to change, can we memo-ize them and then only run the relevant transform functions?

src/lib/style-transformer.html
```diff
@@ -148,8 +148,12 @@
       // transforms a css rule to a scoped rule.
       _transformRule: function(rule, transformer, scope, hostScope) {
         var p$ = rule.selector.split(COMPLEX_SELECTOR_SEP);
-        for (var i=0, l=p$.length, p; (i<l) && (p=p$[i]); i++) {
-          p$[i] = transformer.call(this, p, scope, hostScope);
+        // we want to skip transformation of rules that appear in keyframes,
+        // because they are keyframe selectors, not element selectors.
+        if (!Polymer.StyleUtil.isKeyframesSelector(rule)) {
```
sorvell: `Polymer.StyleUtil` should be available in this module as `styleUtil`.

src/lib/style-util.html
```diff
@@ -44,21 +47,31 @@
         return style.__cssRules;
       },
 
-      forEachStyleRule: function(node, callback) {
+      // Tests if a rule is a keyframes selector, which looks almost exactly
+      // like a normal selector but is not (it has nothing to do with scoping
+      // for example).
+      isKeyframesSelector: function(rule) {
+        return rule.parent &&
+            rule.parent.type === Polymer.StyleUtil.ruleTypes.KEYFRAMES_RULE;
```
sorvell: scope should be `Polymer.StyleUtil`, so just `this` here?

sorvell:
Let's add a test to cover using `@apply` in keyframes. wdyt?

src/lib/style-properties.html
cdata: Will fix.

src/lib/style-transformer.html
cdata: Will fix.

src/lib/style-util.html
cdata: Will fix.

src/lib/style-properties.html
cdata: Will fix.

src/lib/style-properties.html
cdata: Will fix.

cdata:
Extra tests added for `@mixin` and style-scope changing. PTAL!

src/lib/style-properties.html
```diff
@@ -181,6 +188,50 @@
         rule.cssText = output;
       },
 
+      // Apply keyframe transformations to the cssText of a given rule. The
+      // keyframeTransforms object is a map of keyframe names to transformer
+      // functions which take in cssText and spit out transformed cssText.
+      applyKeyframeTransforms: function(rule, keyframeTransforms) {
+        var input = rule.cssText;
+        var output = rule.cssText;
+
```
sorvell: Let's remove these empty lines. We generally don't space out function contents like this and prefer factoring instead.

src/lib/style-properties.html
```diff
@@ -256,15 +307,55 @@
           hostSelector;
         var hostRx = new RegExp(this.rx.HOST_PREFIX + rxHostSelector +
           this.rx.HOST_SUFFIX);
+
+        var keyframesRules = element._styles._keyframes;
```
sorvell: Seems like this keyframes specific handling should be factored into a separate function

src/lib/style-properties.html
```diff
@@ -256,15 +307,55 @@
...
+               keyframesRule = keyframesRules[++i]) {
+            this._scopeKeyframes(keyframesRule, scopeSelector);
+            keyframeTransforms[keyframesRule.keyframesName] =
+                this._keyframesRuleTransformer(keyframesRule);
+          }
+        }
         return styleTransformer.elementStyles(element, function(rule) {
           self.applyProperties(rule, properties);
-          if (rule.cssText && !nativeShadow) {
+          self.applyKeyframeTransforms(rule, keyframeTransforms);
```
sorvell: Aren't these transforms unnecessary under native Shadow DOM? Can we avoid this in that case, if so?

src/lib/style-properties.html
cdata: Will fix.

src/lib/style-properties.html
cdata: Will fix.

src/lib/style-properties.html
cdata: Will fix.

src/lib/style-properties.html
```diff
@@ -256,15 +303,60 @@
...
+        var keyframeTransforms =
+          this._elementKeyframeTransforms(element, scopeSelector);
         return styleTransformer.elementStyles(element, function(rule) {
           self.applyProperties(rule, properties);
-          if (rule.cssText && !nativeShadow) {
+          if (!nativeShadow &&
+              !Polymer.StyleUtil.isKeyframesSelector(rule) &&
+              rule.cssText) {
+            // NOTE: keyframe transforms only scope munge animation names, so it
+            // is not necessary to apply them in ShadowDOM.
```
cdata: I left a reminder to hopefully lessen confusion in the future.

sorvell:
LGTM. Thanks for the contribution!

PR #3171, opened 2015-12-07
Fixed parsers interpreting regex as comment

Linksku:
rails-polymer and emcee were both interpreting this as an actual comment, causing them to compile incorrectly.


Linksku:
I signed it!

kevinpschaaf:
@Linksku Thanks for fixing this, but it's a little unclear what the parser was tripping up on.  Could you add a test that shows what this fixed?  Thanks.

PR #3184, opened 2015-12-10
querySelector now stops querying on first match

joineral32:
I noticed that querySelector delegates to querySelectorAll and then returns the first element in the list. Rather than finding all matches in the tree, I adjusted querySelector to only seek to find the first match.


joineral32:
Updated pull request to resolve conflicts. Based on some rough measurements with a few different test cases, this querySelector is on average about 20% faster. If needed, I can perform and upload some more rigorous test cases and results.

jimsimon-cengage:
+1

sorvell:
Sorry for the delay here and thanks for the effort. This issue was fixed in [Polymer 1.2.4](https://github.com/Polymer/polymer/releases/tag/v1.2.4).

Issue #3183, opened 2015-12-09
Add convenience method for local DOM querySelectorAll, much like $$ for local DOM querySelector


joineral32-cengage:
Polymer has a utility method $$ that provides a shorthand for querying individual elements on an element's local dom. This is very helpful and my colleagues and myself enjoy using this utility method. So much so, in fact, that we would like something similar for querySelectorAll against the local DOM.

I have opened up a pull request that adds a $$$ utility method which performs querySelectorAll against an element's local DOM, in a very similar fashion to $$: https://github.com/Polymer/polymer/pull/3186

dsschnau:
+1, just had the same idea. beats having to write `Polymer.dom(this.root).querySelectorAll(...)`

jimsimon-cengage:
+1

mkimitch:
+1

TimvdLippe:
`$$` has been removed, as the native API is a lot more intuitive: `this.shadowRoot.querySelector` Moreover, all other native DOM APIs are available on `this.shadowRoot`, including `querySelectorAll`. For more information on the discussion, see #3186. Therefore, I am closing this issue.

PR #3186, opened 2015-12-10
Added 40382$ utility for local dom querySelectorAll

joineral32:
Added a $$$ convenience method for querySelectorAll against an element's local dom, similar to $$. See issue: https://github.com/Polymer/polymer/issues/3183

### DIFF
--- a/test/unit/utils.html
+++ b/test/unit/utils.html
@@ -17,6 +17,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 </head>
 <body>
 
+<dom-module id="my-element">
+  <template>
+    <span hidden></span>
+    <span hidden class="class-selectable"></span>
+    <span hidden id="id-selectable"></span>
+    <p hidden class="class-selectable"></p>
+    <content></content>
+  </template>
+</dom-module>
+
 <script>
 
   HTMLImports.whenReady(function() {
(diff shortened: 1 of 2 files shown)


samccone:
Thanks for the PR @joineral32, someone on the team will take a look at this.

dsschnau:
+1, i was just about to open an issue requesting this sugar

test/unit/utils.html
```diff
@@ -36,6 +46,50 @@
...
+
+      var byId = window.el1.$$('#id-selectable');
+      assert(byId, 'querying local dom by id selector returns element with given id');
+      assert(!window.el1.$$('#id-nosuchthing'), 'querying local dom by id selector returns no element if given id does not exist');
+    });
+
+    test('$$$', function () {
+      var byTag = window.el1.$$$('span');
+      assert(byTag.length === 3
+              && byTag[0].tagName === 'SPAN'
```
TimvdLippe: It is better to split these asserts into seperate asserts. When one of the conditions is false, the error message is more intuitive and easier to debug.

joineral32-cengage: Sure thing! I'm hoping to get around to making that change this weekend.

sorvell:
Sorry for the delay and thanks for the effort here. We think it was probably a mistake to provide public sugaring for querying and element's local dom via `$$`. It's also a poorly named function. Because of these points, we don't want to expand the problem here.

joineral32-cengage:
Certainly we don't want to add this feature if $$ was deemed a poor choice. May I ask what are the factors that have lead you all to deem public $$ to be hazardous? (and maybe add these to the official documentation?) It would be valuable in future design decisions to understand these points.

sorvell:
`$$` queries the element's local dom which is its private implementation. There's not a good way to make this explicitly private and indeed under Shadow DOM one can do `element.shadowRoot.querySelector`. We chose a slightly obscure name to slightly obscure the function. Our current consensus is that it was just a little too unclear.

ktiedt:
@sorvell if the intent was for `$$` to be private or not used... why was it so clearly documented as a Utility function? That seems to imply it was intended to be used... https://www.polymer-project.org/1.0/docs/devguide/utility-functions.html#utility-functions

jay8t6:
are you guys gonna merge this soon or are we still directed to use querySelectorAll? It would be nice if we can use $$$.

ktiedt:
Considering this is marked as closed, and has had no further follow ups since March, I would say its pretty much decided it is not landing.

jay8t6:
Gotcha. So it's not recommended to use this.$$?

ktiedt:
I'm pushing to get more concrete answers on those statements.

tjsavage:
`this.$$` is strictly shorthand for `Polymer.dom(this.root).querySelector()`. The original thought was writing `Polymer.dom(this.root).querySelector()` was too verbose, so it'd be nice to build a relatively obscure API that's shorter but does the same thing. `$$` was chosen to be obscure so that it doesn't seem like "public" API on every element, as the DOM it queries over should be considered private.

With the benefit of hindsight, we've come to the position that using the platform API here makes it easier to figure out what is going on. While `Polymer.dom(this.root).querySelector()` is verbose, it's much more understandable than having to go to the documentation to figure out what `this.$$` means. It tells you exactly what it's doing - using the Shady DOM shim to query select over the element's shadow/shady root.

So in short: we think using `Polymer.dom(this.root).querySelector()` actually makes for better code. Of course, `this.$$` will continue to be supported, as there's plenty of code out there that uses it. In future versions of Polymer, this sort of obscuring API will likely be dropped. We'll look to update the docs to reflect this position.

jay8t6:
Makes sense. Thank you for the explanation.

PR #3200, opened 2015-12-14
Add more globals to whitelist for safari

### LINKED ISSUE #3199: Safari complains about weird globals
Safari often complains about `alert`, `prompt`, and `confirm` globals, which should be standard methods on `window`.
I think these are leftovers from the selenium plugin, so I will add them to the whitelist.

azakus:
Fixes #3199


kevinpschaaf:
LGTM

PR #3202, opened 2015-12-15
use destination insertion points when calculating the path

### LINKED ISSUE #3201: event path calculation is wrong
`Polymer.dom().path` "skips" some elements when the target is contained in a `<content>` tag, e.g.
http://jsbin.com/puxolo/edit?html,console,output

Must implement [Event Paths](http://w3c.github.io/webcomponents/spec/shadow/#event-paths)

valdrinkoshi:
Fixes #3201 by using destination insertion points for computing the path (following this algorithm https://github.com/w3c/webcomponents/blob/91f9a1fa681c39cf8ee9dff898db78fc7797354c/spec/shadow/index.html#L536). Specifically:
If a node has destination insertion points:
- append them all (except the last one) to `path`
- set `current` to be the last destination insertion point

### DIFF
--- a/src/lib/dom-api-event.html
+++ b/src/lib/dom-api-event.html
@@ -15,13 +15,13 @@ Polymer.EventApi = (function() {
   var DomApi = Polymer.DomApi.ctor;
   var Settings = Polymer.Settings;
 
-  
+
   /**
-   * DomApi.Event allows maniuplation of events compatible with 
-   * the scoping concepts in Shadow DOM and compatible with both Shady DOM 
-   * and Shadow DOM. The general usage is 
-   * `Polymer.dom(event).property`. The `path` property returns `event.path` 
-   * matching Shadow DOM. The `rootTarget` property returns the first node 
+   * DomApi.Event allows maniuplation of events compatible with
+   * the scoping concepts in Shadow DOM and compatible with both Shady DOM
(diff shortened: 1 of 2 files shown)


valdrinkoshi:
@sorvell FYI, kindly review. What's the best place to add some tests? I noticed there are already some smoke tests on `path` in `test/smoke/patch-dom.html`

src/lib/dom-api-event.html
```diff
@@ -71,10 +71,18 @@
...
+          while (current) {
+            path.push(current);
+            var insertionPoints = Polymer.dom(current).getDestinationInsertionPoints();
+            if (insertionPoints.length) {
+              for (var i = 0; i < insertionPoints.length - 1; i++) {
+                path.push(insertionPoints[i]);
+              }
+              current = insertionPoints[insertionPoints.length - 1];
+            } else {
+              current = Polymer.dom(current).parentNode || current.host;
```
valdrinkoshi: @sorvell what if we update `parentNode` to return the last destination insertion point or the deep parent?

sorvell: Don't think that makes sense. Destination insertion points and parentNode are not related.

sorvell:
LGTM. Thx!

valdrinkoshi:
\o/

PR #3208, opened 2015-12-16
Faster conversion functions and fixed double case words.

birkir:
Tested performance for these two functions and @sjmiles concerns about rejection being faster.

http://jsperf.com/kebabcase-polymer/2


TimvdLippe:
Oops I missed this PR while creating #3213 and they will therefore conflict.

kevinpschaaf:
This looks good; we need to see why the tests fail and @azakus will double-check the performance.

azakus:
#3279 looked more maintainable, so I went with that instead.
Still, thanks for making this PR!

PR #3215, opened 2015-12-18
Fix for `Polymer.dom(...)._query()` method doesn't exist which causes `Polymer.updateStyles()` to fail

### LINKED ISSUE #3214: Missing Polymer.dom()._query
After commit 306cc8130e4cab12d6e05eec6097e1ad85220c50 from #3112 Shadow DOM support is broken.
Calling `Polymer.updateStyles()` results in:
> TypeError: Polymer.dom(...)._query is not a function at HTMLElement.Polymer.Base._addFeature._updateRootStyles (...) at Function.Polymer.updateStyles (...) at HTMLDocument.ready_callback (...) at ...

Namely, Shadow DOM version of `Polymer.dom(...)` does not have `_query()` method used here: https://github.com/Polymer/polymer/blob/e3753dbaeb84e415311f478eea4dbc7c733bad6c/src/standard/x-styling.html#L262

This should be fixed before next release.
@sorvell @kevinpschaaf

nazar-pc:
Fixes #3214

### DIFF
--- a/src/lib/dom-api-shadow.html
+++ b/src/lib/dom-api-shadow.html
@@ -26,6 +26,28 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       return TreeApi.arrayCopy(this.node.querySelectorAll(selector));
     },
 
+    _query: function(matcher, node) {
+      node = node || this.node;
+      var list = [];
+      this._queryElements(node.childNodes, matcher, list);
+      return list;
+    },
+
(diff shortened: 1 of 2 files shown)


nazar-pc:
There is only one test failed, while locally tests are passing fine (though, randomly some tests may fail, but on repeated run all green)

sorvell:
Thanks for the fix! LGTM. I'm going to merge then and then massage the code slightly.

PR #3222, opened 2015-12-20
Fix for scoping when class is not specified on element

nazar-pc:
If class on element is not present then 
```javascript
(selector = element.getAttribute('class')) === null
```
And as result:
```javascript
(selector + 'style-scope') === 'nullstyle-scope'
```
While just `style-scope` was expected.


TimvdLippe:
Could you add a little test to prevent regressions? Otherwise LGTM :+1:

azakus:

nazar-pc:
Test is here

sorvell:
LGTM, thx!

PR #3234, opened 2015-12-24
Add pre-hook for properties

### LINKED ISSUE #3099: Add pre-hook to properties
This issue originates from the discussion at #2677 .

**Current post-hook**
Polymer currently allows developers to be notified whenever a property is changed. This is achieved by assigning the name of the function to be called to the `observer` field at the property in the `properties`. This hook is very useful to adding logic whenever we need to update according to the new value.

**Definition pre-hook**
In the PR linked above, it becomes apparent that developers want to insert logic to validate, transform or perform actions before the value is updated.

Therefore the new pre-hook should:
- Be called before the value is updated
- When the pre-hook returns a value (`!== undefined`), the value should be the value to update

**Example code**
Element `x-foo` declares a property `bar`:

``` js
Polymer({
  is: 'x-foo',
  properties: {
    bar: {
      type: Number,
      set: 'checkBar'
    }
  },
  checkBar: function(value) {
    if (value < 0) {
      return 0;
    }

    return value;
  }
})
```

`bar.set` references the name of the function to be called. The function has 1 argument: the value to be set. The function returns the overridden value. Therefore if bar is updated to `-1`, this value is changed to `0`. If bar is updated to `5`, the value becomes `5`.

**Use-cases**
The pre-hook allows developers to extend the properties to achieve:
1. **Validate values**<br>
   The original PR added logic to validate values to be set on to properties. This callback allows developers to insert such logic on their own, even if they are put into behaviours.
2. **Transform values**<br>
   Another use-case is transforming data. Based on the input given by the user, the element can determine if it should alter the value in order to satisfy an invariant.

**Implementation**
The added code should be fairly minimal. It requires a lookup for the `set` field in the property, a call to the function and check the return value. All other implementations of developers can extend from this hook and put into seperate behaviours to minimize the file size of the core library.

TimvdLippe:
Add pre-hooks for properties to validate and transform properties.
Adds a new field to a property called `set` that references the name of a function. This function takes a `value` and returns the new value.

Fixes #3099

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -182,16 +182,22 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       } else {
         if (info.set) {
           var setFn = model[info.set];
-          var processing = false;
 
-          setter = function(value) {
-            // When setting the value, effects are triggered. In order to
-            // prevent stackoverflows, do not re-process the updated value
-            if (!processing) {
-              processing = true;
-              this._propertySetter(property, setFn(value), effects);
-              processing = false;
+          if (setFn) {
+            var processing = false;
+
+            setter = function(value) {
+              // When setting the value, effects are triggered. In order to
+              // prevent stackoverflows, do not re-process the updated value
+              if (!processing) {
+                processing = true;
+                this._propertySetter(property, setFn(value), effects);
+                processing = false;
+              }
             }
+          } else {
+            Polymer.Base._warn(Polymer.Base._logf('_createAccessors', 'set method `' +
+              info.set + '` not defined for property ' + property));
           }
         }
 
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -23,7 +23,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       _notifyChange: function(source, event, value) {
         value = value === undefined ? this[source] : value;
         event = event || Polymer.CaseMap.camelToDashCase(source) + '-changed';
-        this.fire(event, {value: value}, 
+        this.fire(event, {value: value},
           {bubbles: false, cancelable: false, _useCache: true});
       },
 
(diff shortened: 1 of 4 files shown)


TimvdLippe:
Fixed a stack-overflow issue when a child notified an binding update.

BLamy:
:+1: This is a lot cleaner then my original proposal.

src/lib/bind/accessors.html
```diff
@@ -172,14 +172,35 @@
...
-      if (info && info.readOnly) {
+      var info = model.getPropertyInfo && model.getPropertyInfo(property) || {};
+      if (info.readOnly) {
         // Computed properties are read-only (no property setter), but also don't
         // need a private setter since they should not be called by the user
         if (!info.computed) {
           model['_set' + this.upper(property)] = setter;
         }
       } else {
+        if (info.set) {
```
BLamy: Why not `} else if(info.set) {

src/lib/bind/accessors.html
```diff
@@ -172,14 +172,35 @@
...
+          var setFn = model[info.set];
+
+          if (setFn) {
+            var processing = false;
+
+            setter = function(value) {
+              // When setting the value, effects are triggered. In order to
+              // prevent stackoverflows, do not re-process the updated value
+              if (!processing) {
+                processing = true;
```
kaste: Don't you actually need this lock/flag only b/c your setFn's in the tests are actually not idempotent?

kaste: Shouldn't this be wrapped by a  `try () finally ( ...releaseLock... )`?

TimvdLippe: The stack-overflow error can be triggered quite easily. If the parent has defined a property with a setter and binds this to a property of a child, when the child property is updated the stack-overflow error is created.

Actually I was thinking of leaving this change out (so only pre-hooks, no transformation magic) as the cons might outweigh the pros.

sorvell:
This PR may fit this category: https://github.com/Polymer/polymer/issues/3369.

kevinpschaaf:
With a feature like this, there will be a high temptation for users to depend on other element properties (e.g. `this.min`, `this.max`, etc.) in their validation logic, but this provides no way to declare those dependencies to ensure that a.) the values have been configured or b.) the validation is re-run when the dependencies change.

As-is, I don't think this would solve enough use cases to warrant an addition to the API surface right now.  Computed properties solve much of this problem space, both the dependency issue and the efficiency issue related to taking action on the unvalidated value, by differing the value on the API from the one the element actually uses internally.  As has been well stated, this comes at the cost of extra property definitions which we understand can feel burdensome.  It's a hard problem so we'll continue to consider ways to make this pattern easier.

PR #3238, opened 2015-12-27
Implement method for dirty-checking of properties

### LINKED ISSUE #2719: Dirty-checking not occuring on native Date values
Polymer's dirty-checking seems to overlook date values:

``` javascript
Polymer({
  /* ...snip... */
  _dateChanged: function(value, oldValue) {
    console.log(value === oldValue); // false
    console.log(value.getTime() === oldValue.getTime()); // true
  }
});
```

If a property is defined using the native `Date` type, polymer should convert to timestamps before dirty-checking.

TimvdLippe:
To solve dirty-checking issues with (native) Javascript Objects, this PR extracts a method to check if a property is dirty. The implementation allows custom types to be used by Polymer. Custom types should only implement the function `isNotEqualTo`. Additionally, the previous NaN checks are now only performed on Numbers as opposed to during all updates.

I did not observe any performance problems, the test suite runs on my computer in almost equal time (roughly 50 seconds).

The reason for using a function called `isNotEqualTo` is to prevent collisions with other Javascript libraries/frameworks that could possibly implement/depend on an equally named function. (for example `equals`)

Fixes #2719

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -15,7 +15,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     // for prototypes (usually)
     prepareModel: function(model) {
+      // TODO(timvdlippe): temporarily store this method as it is explicitly
+      // overridable by the user. Will be fixed in the refactoring for 2.0
+      var dirty = model._isPropertyDirty;
       Polymer.Base.mixin(model, this._modelApi);
+      if (dirty) {
+        model._isPropertyDirty = dirty;
+      }
     },
 
     _modelApi: {
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -44,10 +44,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // in the closure for the accessor for efficiency
       _propertySetter: function(property, value, effects, fromAbove) {
         var old = this.__data__[property];
-        // NaN is always not equal to itself,
-        // if old and value are both NaN we treat them as equal
-        // x === x is 10x faster, and equivalent to !isNaN(x)
-        if (old !== value && (old === old || value === value)) {
+
+        if (this._isPropertyDirty(old, value)) {
           this.__data__[property] = value;
           if (typeof value == 'object') {
             this._clearPath(property);
@@ -60,6 +58,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           }
         }
         return old;
+      },
+
+      /**
+       * Check if a property has changed.
+       * @param  {*} old   The old value of the property
+       * @param  {*} value The new value of the property
+       * @return {Boolean}       Whether the property has changed.
+       */
+      _isPropertyDirty: function(old, value) {
+        // NaN is always not equal to itself,
+        // if old and value are both NaN we treat them as equal
+        // x == x is 10x faster and equivalent to !isNaN(x)
+        return (old !== value && (old === old || value === value));
       },
 
       // Called during _applyConfig (well-known downward data-flow hot path)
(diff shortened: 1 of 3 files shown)


kevinpschaaf:
@TimvdLippe Thanks for this.  I think the value here is useful, but we're definitely opposed to modifying built-in prototypes like this.  I think the best solution would be to move the existing dirty check expression to a function, ala:

```js
isPropertyDirty: function(old, n) {
}

TimvdLippe:
Updated the PR with the extra function.

kevinpschaaf:
Thanks, Tim.  We don't want to change the actual implementation of the default dirty check at all right this moment; it is a very hot path and esp. the extra NaN implementation was tested for speed (not sure how yours compares, but I'd just rather avoid having to re-test).  And I'd prefer the non-identity Date check and `isNotEqual` call to be done in an overridden version of `_isPropertyDirty` that you can provide to your elements via a behavior, rather than in the core.  So we want the new implementation of `_isPropertyDirty` to be the current `(old !== value && (old === old || value === value)`.

TimvdLippe:
Apologies for misunderstanding your earlier comment. I have tried to implement it, however the `Polymer.Base.mixin` call overrides the custom function with the base function. I am not sure how to fix this, other than modifying the mixin function.

One remark though: If the function is overridden, only that element will properly dirty-check the property. That means that every element that uses the data-type must import the behavior. This would fall in the category of #3366 where inserting all behaviors on all elements is very inconvenient for the user.

azakus:
Hey Tim, we discussed this for a bit, and I think some of your concerns about cooperation between elements will go away when we remove the identity check for Array and Object types. When that lands overriding `_isPropertyDirty` for non-native data types would only be a performance improvement, and would not change behavior.

We'd still like the `_isPropertyDirty` name with the underscore prefix.

> the Polymer.Base.mixin call overrides the custom function with the base function.
Can you expand on this a bit? Are you seeing an element's `isPropertyDirty` being overridden by `Polymer.Base`?

TimvdLippe:
@azakus Sorry for dragging this PR so long. The current version is so different than my first proposal that I got really confused during the process.

I have added the underscore prefix again, however the test still fails. In the test, I create a Polymer object that defines the `_isPropertyDirty`. This is because all of the [methods are mixed in](https://github.com/Polymer/polymer/blob/master/src/lib/bind/accessors.html#L18) after the model is created. So if I understand the method correctly, model contains all of the properties/methods defined by the user. Then the methods in `this._modelApi` are mixed in which means they overwrite any existing methods on the model. Therefore the previously user-defined `_isPropertyDirty` gets overriden by the `_isPropertyDirty` of `this._modelApi`.

kaste:
Sorry for being late here, but can you pass the property name down into the func.

kaste:
Hm, `_isPropertyDirty` seems like an inappropriate name. This is (looks like?) a guard before setting and propagating the value. So the Polymer default is of course just a _dirty_ check, but overriding this I could just reject values from the user I don't want as well. 

This seems like a huge feature for two lines of code. Not?

kevinpschaaf:
Thanks, @TimvdLippe.  The layering in this part of the code is a bit messy.  We're doing a bunch of work right now towards a minimally-breaking 2.0 release centered around decoupling and turning all of the Polymer features into imperatively-callable libraries (that the backwards-compatible `Polymer()` call would still use implement the declarative sugar), so I'm inclined to just do what we need to do here to make `_isPropertyDirty` overridable on the element prototype/behavior since this will be cleaned up in the refactoring work that we're doing... Maybe caching it on the proto before the mixin, and restoring it afterwards?

TimvdLippe:
@kevinpschaaf Great to hear about a refactoring to improve the code architecture! Caching is I think the best solution, so I applied it and succesfully passed all tests.

TimvdLippe:
Any update on this PR @kevinpschaaf ?

TimvdLippe:
@sorvell @kevinpschaaf I assume this PR can be closed given the new approach in 2.0?

joh1234:
Update README
https://goo.gl/JBljdR

TimvdLippe:
Closing this PR as there has not been an official response on the status of this PR and I personally think this issue is resolved in 2.0.

Issue #3166, opened 2015-12-05
Fix min CSS branch fails on Safari


TimvdLippe:
#2689 introduced fixes for parsing of minimized CSS. However, after merging, the tests failed on Safari. I am not able to run the tests on Safari because it can't be installed on Linux. Therefore the following steps might be needed to investigate the issue:

The changes of the PR are available in [this branch](https://github.com/Polymer/polymer/tree/JeremybellEU-fix-min-css).

There are two possible problems:
1. The regex does not match on Safari due to a different regex engine.  This can be tested by entering the following in the console:
```js
"top: 12px;--variable-from-other-variable: var(--variable-into-first-variable);--variable-from-another-variable: var(--variable-into-second-variable); --variable-from-last-variable: var(--variable-into-third-variable);".match(/(?:^[^;\-\s}]+)?--[^;{]*?:[^{};]*?(?:[;\n]|$)/gim)
```
If the output is false, then the regex does not work on Safari. Else the output should be:
```js
["--variable-from-other-variable: var(--variable-into-first-variable);", "--variable-from-another-variable: var(--variable-into-second-variable);", "--variable-from-last-variable: var(--variable-into-third-variable);"]
```
2. Secondly, it could be a styling issue on Safari. To test this, run `polyserve` and open `http://localhost:8080/components/polymer/test/unit/custom-style.html`. Verify that `child-of-child-with-var` textContent has:
```css
bottom: 9px;
left: 10px;
right: 11px;
```
If this is not the case, styling of Safari is different than the other browsers and a possible fix is to change the CSS-properties defined at https://github.com/Polymer/polymer/pull/2689/files#diff-5ccfdb2810bc155fcc15b5a27ae88be4R273 to other properties like `border-bottom-width`.

JeremybellEU:
@JWGmeligMeyling tested the regex and confirmed it worked. So the failing test is probably related to the `bottom` css attribute.

rubenstolk:
As far as I can see, the element in the test suite is given `bottom`, `left` and `right` properties, however the element doesn't have any value for `position` (default `static`). The difference between Chrome and Safari for `window.getComputedStyle(document.querySelector('child-of-child-with-var')).bottom` is:

- Safari: `"auto"`
- Chrome: `"9px"`

Submitted a PR: https://github.com/Polymer/polymer/pull/3240

TimvdLippe:
Landed in #3241, woohoo! Thanks a ton @rubenstolk

[...]

PR #3240, opened 2015-12-28
Make tests succeed in Safari

rubenstolk:
This PR solves issue https://github.com/Polymer/polymer/issues/3166

As far as I can see, the element in the test suite is given `bottom`, `left` and `right` properties, however the element doesn't have any value for `position` (default `static`). The difference between Chrome and Safari for `window.getComputedStyle(document.querySelector('child-of-child-with-var')).bottom` is:

- Safari: `"auto"`
- Chrome: `"9px"`

Applying `position: relative` solves this.

### DIFF
--- a/test/unit/custom-style.html
+++ b/test/unit/custom-style.html
@@ -248,6 +248,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <template>
       <style>
         child-of-child-with-var {
+          position: relative;
           --variable-own-line: "Varela font";
           margin-top: var(--variable-property-own-line);
           margin-bottom: var(--variable-property-preceded-property);


TimvdLippe:
The build errored on FireFox due to a Selenium error of not being able to connect with 45000 ms. I am not sure if that is related to this issue.

However, if the CSS parsing of Safari and the other browsers are equal for `position`, I think the issue can be solved by:
Line 238
```css
--variable-into-first-variable: relative;
```
Line 274
```css
position: var(--variable-from-other-variable);
```

The `position: relative;` on line 251 can be removed again.

Another thing is that the original PR was reverted. I am not sure if this change would be applied, if the merge to master will contain the original fixes, or just these 2 line-changes. CC: @azakus

rubenstolk:
@TimvdLippe I don't see a reason of having `relative` in a variable or not, it doesn't really matter as long as relative is defined for the particular element.

I sent this PR to your branch, if you merge this PR and then send a new PR to master from your branch, I think @azakus can take it from there.

I don't know why tests fail on travis but that is not related to these fixes...

TimvdLippe:
Hm okay, it is probably me misunderstanding the CSS parsing of Safari ;p Thanks for the fix anyways and hopefully a rebuild will be green!

rubenstolk:
Re-opened as a clean PR to master in https://github.com/Polymer/polymer/pull/3241

PR #3241, opened 2015-12-28
Fix parsing of minimized css output (2)

rubenstolk:
https://github.com/Polymer/polymer/pull/2689 but with fixed tests

### DIFF
--- a/test/unit/custom-style.html
+++ b/test/unit/custom-style.html
@@ -248,6 +248,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <template>
       <style>
         child-of-child-with-var {
+          /* in certain browsers (e.g. Safari) `top`, `bottom`, `left`, `right` don't compute
+             when no explicit position is defined (`relative` / `absolute` / `fixed`) */
+          position: relative;
           --variable-own-line: "Varela font";
           margin-top: var(--variable-property-own-line);
           margin-bottom: var(--variable-property-preceded-property);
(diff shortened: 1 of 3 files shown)


rubenstolk:
@azakus @TimvdLippe I think this one is good to be merged now

rubenstolk:
@googlebot have a look at https://github.com/Polymer/polymer/pull/2689 and add `cla: yes` :)

rubenstolk:
@TimvdLippe do you think we should make the same change to `mixinProp`?

TimvdLippe:
Hm I vaguely remember I also applied the fix to `mixinProp` but I think it broke some tests. It seems logical imo and worth another shot.

rubenstolk:
I've applied the same fix to `mixinProp` as well now in this PR. @TimvdLippe would you mind cross-checking the code and then assigning it to @azakus to merge it?

test/unit/custom-style.html
```diff
@@ -245,13 +248,15 @@
     <template>
       <style>
         child-of-child-with-var {
+          position: relative;
```
TimvdLippe: We might want to add a CSS comment to clarify the usage of this rule.

TimvdLippe:
One small comment, otherwise LGTM! Thanks for your investigation and fix!

rubenstolk:
Added a comment! @TimvdLippe please assign it to @azakus so we can get this merged!

TimvdLippe:
I don't have write access to this repository, so I can not assign anyone.

rubenstolk:
@TimvdLippe just noticed you're in Delft too. Time for a Polymer@Delft meetup!?

TimvdLippe:
@rubenstolk contacted you on the Polymer Slack

azakus:
LGTM!

rubenstolk:
Yay @azakus, now when https://github.com/PolymerLabs/polyclean/pull/9 gets merged we finally have a real solution for minifying Polymer!!!

PR #3260, opened 2016-01-01
Add sugar for child-delegation for on-* methods

### LINKED ISSUE #3259: Event delegation sugar for the mediator pattern
When using the mediator pattern, often the parent is listening to an event of a child and perform operations on a different child. For example implementing a `refresh-button` with `paper-button` and `iron-ajax`:

``` html
<template>
  <iron-ajax
    id="ajax"
    ....
    ></iron-ajax>
  <paper-button on-tap="refreshClick">Refresh</paper-button>
</template>
<script>
Polymer({
  is: 'refresh-button',
  refreshClick: function() {
    this.$.ajax.generateRequest();
  }
});
</script>
```

This pattern requires additional methods in the parent to just infer a method on a child. It would be nice to have some sugar that allows direct references to a child method. E.g.:

``` html
<template>
  <iron-ajax
    id="ajax"
    ....
    ></iron-ajax>
  <paper-button on-tap="ajax.generateRequest">Refresh</paper-button>
</template>
<script>
Polymer({
  is: 'refresh-button'
});
</script>
```

This sugar allows the interaction between children via the mediator pattern, while preventing countless of methods to be defined in the parent.

TimvdLippe:
The `on-*` events are checked for containing a `.`. If this is the case, reference the corresponding child with name before the `.` and the `methodName` following the `.`.

Fixes #3259

### DIFF
--- a/src/standard/events.html
+++ b/src/standard/events.html
@@ -129,14 +129,28 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     _createEventHandler: function(node, eventName, methodName) {
       var host = this;
-      var handler = function(e) {
-        if (host[methodName]) {
-          host[methodName](e, e.detail);
+      var handler = (function() {
+        var processingNode, processingMethod;
+
+        if (methodName.indexOf('.') > 0) {
+          var split = methodName.split('.');
+          processingNode = host.$[split[0]];
+          processingMethod = split[1];
         } else {
-          host._warn(host._logf('_createEventHandler', 'listener method `' +
-            methodName + '` not defined'));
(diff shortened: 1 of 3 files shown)


sorvell:
We're not convinced that this feature fits a wide enough number of use cases to be included directly. Rather than merging this PR, we want to add better extension points so that our awesome users can innovate. See https://github.com/Polymer/polymer/issues/3369.

TimvdLippe:
I am sad to not see this merged, as I find myself in this situation more often than not. That said, I do understand the reasoning about the use case and will wait for the result of #3369 to act accordingly.

PR #3267, opened 2016-01-05
Move attribute deserialization to after template stamping. Fixes #3266

### LINKED ISSUE #3266: Move attribute deserialization to after template stamping
It would be beneficial to move attribute deserialization to after template stamping, so that custom `deserialize` methods can deserialize strings to elements.  This has proven useful for element API's that require referencing other elements, such as scroll targets, activate targets, etc. as a means to provide declarative association.

kevinpschaaf:


src/micro/attributes.html
```diff
@@ -83,7 +83,7 @@
         // NOTE: never allow 'class' to be set in hostAttributes
         // since shimming classes would make it work
         // inconsisently under native SD
-        if (!this.hasAttribute(n) && (n !== 'class')) {
+        if (!this.hasAttribute(n) && (n !== 'class') && this._config[n] === undefined) {
```
kevinpschaaf: @sorvell @cdata 

Ok, spent a little time to recall what where we ended with this.  These changes are what was in my stash from when we looked at this last.  Everything is chill except for this one change here.

Best I can (re-)work out, [this test](https://github.com/Polymer/polymer/blob/master/test/unit/attributes.html#L286) fails without the `&& this._config[n] === undefined` here.  That test tests that setting a `hostAttribute` on a host, which is then deserialized and bound down into an attribute of a child which itself also defines a `hostAttribute` with a different value, only results in the child property observer only being called once with the host value.  It fails not because the ending value is wrong, but because the callback is called more than once.  This is because of the queueing nature of `attributeChangedCallback` when `setAttribute` is called from the context of a `createdCallback` (on Chrome only); the host's setAttribute of the overriding value is queued, and runs after the child readies itself with its own attribute value. (Tangentially, it also fails in an identical way on polyfill where `setAttribute` runs the callback synchronously for different reasons, but the change here resolves the issue for both)

Next steps are:
* Do we even want to fix this?  Best I can tell the setup is totally contrived and worst case it arrives at the same values, just with an extra observer call.  We could just "fix the test".
* If we do want to fix this, we have a layering violation here between configuration (standard) & attributes (micro), so we'd need to deal with that.
* Last, I just noted that the `n` here is an attribute name, so it looks wrong that `n` is used as a key to `_config` which is property-named.  Probably requires adding a case-map call here, which is slightly worse for perf.

Seems like that makes me lean towards just fixing the test and removing the change on this line.  Let's discuss tomorrow.

sorvell:
After thinking about this more, this seems like a reasonable set of changes. Attributes are inputs that should be marshaled with other inputs (i.e. default values) at configuration time.

The necessary check to `_config` that @kevinpschaaf added indicates that we are not currently configuring hostAttributes. Even though [the test](https://github.com/Polymer/polymer/blob/master/test/unit/attributes.html#L286) passes before these changes, it still results in an extra, unnecessary attribute set.

src/micro/attributes.html
sorvell: What if we add an overridable `_canApplyAttribute` which we can implement in the configure module. This will allow us to avoid an extra setAttribute for this case. 

As for the case-map question, it seems that we aren't currently doing this mapping for hostAttributes at all so perhaps we can just ignore that? wdyt?

src/mini/ready.html
```diff
@@ -99,6 +99,11 @@
     },
 
     _ready: function() {
+      // install host attributes
+      this._marshalHostAttributes();
```
sorvell: Suggest we move this into `_configure` since this is the common spot for marshaling element inputs.

cdata:
I have tried this branch out on a prototype version of the Shared Element References behavior, and discussed with @sorvell a bit.

It seems to me that the design as outlined for Shared Element References is missing semantics for some conditions, and that Polymer's `properties` block will probably never be suitable to support a well-rounded implementation.

The gist of my rationale is that element references have no meaning when an element is detached from a host, because it is not possible to know which host to search for the element (at least, not without some cooperation at instantiation time). The semantics for what should happen on attach / detach are vaguely defined for an element implementing the behavior, but I think that the most useful thing would be for an element reference to be updated automatically as the element expecting the reference is detached / attached to the DOM (possibly in different contexts over time).

It may be best for element reference serialization / deserialization to happen outside of the property serialization / deserialization flow, since the semantics are probably different. WDYT? Back to the drawing board?

sorvell:
An alternate approach is being explored by @cdata so it looks like we don't need this change.

PR #3274, opened 2016-01-05
Remove call to `wrap` in deepContains

### LINKED ISSUE #3273: `deepContains` will fail if webcomponents polyfills not loaded
`wrap` is only defined by webcomponents.js (and lite).

azakus:
Test Polymer.dom with native shadowroot (chrome only)

Fixes #3273


azakus:
@sorvell PTAL

sorvell:
LGTM

PR #3276, opened 2016-01-07
Small performance tweak (primarily for start time)

nazar-pc:
Small performance tweak (primarily for start time) to avoid duplicated styles injections (once skipping CSS mixins and variables and second time with them)

Besides it allows to skip some CPU cycles (observably, RegExps take a bit more time, so it worth it), it also doesn't duplicate styles in 2 stylesheets inserted at different stages, especially good for Non-Chromium browsers, since allows to emulate less in case of Shadow DOM polyfill.

I do not really care about Shady DOM, so since I've encountered 2 tests failing with this patch under Shady DOM changes only affect Shadow DOM (either native or polyfilled, both are fine).

P.S. I'm experiencing significant performance penalty with multiple CSS mixins/vars in elements that have dozens of instances on page (especially under non-Chromium, this is one of results of exploration how to optimize something in styling system without crashing everything.

### DIFF
--- a/src/standard/styling.html
+++ b/src/standard/styling.html
@@ -20,10 +20,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     var prepElement = Polymer.Base._prepElement;
     var nativeShadow = Polymer.Settings.useNativeShadow;
+    var useShadow = Polymer.Settings.useShadow;
 
     var styleUtil = Polymer.StyleUtil;
     var styleTransformer = Polymer.StyleTransformer;
     var styleExtends = Polymer.StyleExtends;
+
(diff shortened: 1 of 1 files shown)


kevinpschaaf:
@sorvell PTAL

samccone:
@nazar-pc do you have any perf metrics / test repo steps to back this change up?

nazar-pc:
I've tested performance using my codebase with many elements on page through Dev Tools Timeline. Unfortunately it is not easy to execute that code standalone.
It produces the same results, just works faster, thus no new tests, sorry.
This change have especially positive impact when using full polyfill.

sorvell:
Fixed via #3413. Thanks!

Issue #3206, opened 2015-12-15
Camel case declared properties with single letter in the middle are being ignored


alexclemensua:
Camel case declared properties with single letter in the middle are being ignored. This behaviour has been introduced somewhere around 1.2.3, because in 1.0.0, 1.1.0 and 1.1.3 it works as expected. If this is not considered a bug then information about the new behaviour should be included in the [documentation](https://www.polymer-project.org/1.0/docs/devguide/properties.html#attribute-deserialization).

For example: a boolean variable named alignYAxis will be ignored when added to the element code:

    <chart-nasty align-y-axis>a very informative chart goes here</chart-nasty>

    Polymer({
      is : 'chart-nasty',
      properties: {
        alignYAxis: Boolean
      }
    });

but alignAxis works as expected:

    <chart-fine align-axis>an even more informative chart goes here</chart-fine>

    Polymer({
      is : 'chart-fine',
      properties: {
        alignAxis: Boolean
      }
    });

arthurevans:
Looks like this change was introduced in 1.2.2 as a result of the mirco-optimizations, I think specifically of this change here: https://github.com/Polymer/polymer/commit/acdd242616f77616f1f3f90e0389c4c98161bcb4

We used to do dash-case to camel-case when processing the attributes, but now we build up a `_propertyInfo` object using the `camelCaseToDashCase` function, which doesn't handle consecutive capital letters.

I'd definitely argue it's a bug, since it's a) a change in behavior, and b) the `camelCaseToDashCase` function doesn't correctly reverse the `dashCaseToCamelCase` function in `case-map.html`.

TimvdLippe:
@arthurevans Looking at the git history, it appears that the `camelCaseToDashCase` function never worked with 2 uppercase characters in sequence. Is that correct?

[Renaming of file](https://github.com/Polymer/polymer/commit/06f78ee87e8db48bfb5816550960fab6af1b2e88)
[Caching of conversions](https://github.com/Polymer/polymer/commit/ee3e0fda4b735726978168dca8939252f7af3811)
[Moving the function again](https://github.com/Polymer/polymer/commit/6c9e396a1bc9f14fa3dbd3dd7cb06a05737d8832)
[First version of the function](https://github.com/Polymer/polymer/commit/a9d714a3294414b49e978634db6ad7577428c134)

jimsimon:
This doesn't seem like a bug per the docs:
> When mapping attribute names to property names:
>
[...]

PR #3213, opened 2015-12-18
Add unit test suite for CaseMap

TimvdLippe:
Added a testsuite to test the case-map functionality. Includes a testcase for #3206. Therefore on current master the CI fails, but after a fix has been submitted (in #3279) the CI will succeed.

PR #3279, opened 2016-01-07
Improved performance of Polymer.CaseMap.* functions

nazar-pc:
Doubled `Polymer.CaseMap.dashToCamelCase` performance with simplified and once compiled RegExp.
5 times faster `Polymer.CaseMap.camelToDashCase` using simplified replace part, simplified and once compiled RegExp.

Code is more condenced, but still readable and doesn't use temporary variables.

### DIFF
--- a/src/lib/case-map.html
+++ b/src/lib/case-map.html
@@ -12,32 +12,24 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   Polymer.CaseMap = {
 
     _caseMap: {},
+    _rx: {
+      dashToCamel: /-[a-z]/g,
+      camelToDash: /([A-Z])/g
+    },
 
     dashToCamelCase: function(dash) {
-      var mapped = Polymer.CaseMap._caseMap[dash];
-      if (mapped) {
-        return mapped;
-      }
-      // TODO(sjmiles): is rejection test actually helping perf?
-      if (dash.indexOf('-') < 0) {
-        return Polymer.CaseMap._caseMap[dash] = dash;
-      }
-      return Polymer.CaseMap._caseMap[dash] = dash.replace(/-([a-z])/g, 
-        function(m) {
-          return m[1].toUpperCase(); 
-        }
+      return this._caseMap[dash] || (
+        this._caseMap[dash] = dash.indexOf('-') < 0 ? dash : dash.replace(this._rx.dashToCamel,
+          function(m) {
+            return m[1].toUpperCase();
+          }
+        )
       );
     },
 
     camelToDashCase: function(camel) {
-      var mapped = Polymer.CaseMap._caseMap[camel];
-      if (mapped) {
-        return mapped;
-      }
-      return Polymer.CaseMap._caseMap[camel] = camel.replace(/([a-z][A-Z])/g, 
-        function (g) { 
-          return g[0] + '-' + g[1].toLowerCase() 
-        }
+      return this._caseMap[camel] || (
+        this._caseMap[camel] = camel.replace(this._rx.camelToDash, '-$1').toLowerCase()
       );
     }
 


arthurevans:
I think this should also fix https://github.com/Polymer/polymer/issues/3206, which appears to be caused by the `camelToDashCase` behavior not matching that of `dashToCamelCase`.

nazar-pc:
Definitely. Should I add test case for that?

TimvdLippe:
@nazar-pc You can use the test suite I developed for #3213 : https://github.com/Polymer/polymer/pull/3213/files#diff-00b54e7068e156ebcad1fcf0eb6190b6R22

Also the behavior of the function is changed in this PR. It will convert `CamelCase` to `-camel-case`, which is confusing for the user. The previous version resulted in `Camel-case`.

nazar-pc:
Even though this behavior might be changed, but I'm not sure how many people are actually using that syntax. Even if you want `CamelCase` to be mapped to `camel-case` property, it will be converted back as `camelCase` anyway, so if we support starting with capital letter - there will be some confusion for users and from implementation perspective.

TimvdLippe:
Yes I completely agree, just wanted to point it out. Might be a good thing to put in the docs to prevent confusion/mistakes.

nazar-pc:
If this one is selected, I think @TimvdLippe can just modify his PR to contain tests only and @googlebot will be happy with CLA:)

samccone:
@nazar-pc do you have a test case and performance metrics to back this change up?

Thanks for the PR! :)

nazar-pc:
I've just took attribute with few sections and executed it something like 100k times and measured how long it took.
You can try something like this in console:
```javascript
var dash = 'relatively-long-tag-sample';
var camel = 'relativelyLongTagSample';
  var before = {
    _caseMap: {},
    dashToCamelCase: function(dash) {
      var mapped = before._caseMap[dash];
      if (mapped) {
        return mapped;
      }
      // TODO(sjmiles): is rejection test actually helping perf?
      if (dash.indexOf('-') < 0) {
        return before._caseMap[dash] = dash;
      }
      return before._caseMap[dash] = dash.replace(/-([a-z])/g, 
        function(m) {
          return m[1].toUpperCase(); 
        }
      );
    },
    camelToDashCase: function(camel) {
      var mapped = before._caseMap[camel];
      if (mapped) {
        return mapped;
      }
      return before._caseMap[camel] = camel.replace(/([a-z][A-Z])/g, 
        function (g) { 
          return g[0] + '-' + g[1].toLowerCase() 
        }
      );
    }
  };
  var after = {
    _caseMap: {},
    _rx: {
      dashToCamel: /-[a-z]/g,
      camelToDash: /([A-Z])/g
    },
    dashToCamelCase: function(dash) {
      return this._caseMap[dash] || (
        this._caseMap[dash] = dash.indexOf('-') < 0 ? dash : dash.replace(this._rx.dashToCamel,
          function(m) {
            return m[1].toUpperCase();
          }
        )
      );
    },
    camelToDashCase: function(camel) {
      return this._caseMap[camel] || (
        this._caseMap[camel] = camel.replace(this._rx.camelToDash, '-$1').toLowerCase()
      );
    }
  };
console.time('before-1');
for (var i = 0; i < 100000; ++i) {
  before.dashToCamelCase(dash + i);
}
console.timeEnd('before-1');
console.time('before-2');
for (var i = 0; i < 100000; ++i) {
  before.camelToDashCase(camel + i);
}
console.timeEnd('before-2');

console.time('after-1');
for (var i = 0; i < 100000; ++i) {
  after.dashToCamelCase(dash + i);
}
console.timeEnd('after-1');
console.time('after-2');
for (var i = 0; i < 100000; ++i) {
  after.camelToDashCase(camel + i);
}
console.timeEnd('after-2');
```
Output will be like following:
```
before-1: 1406.400ms
before-2: 1391.004ms
after-1: 342.465ms
after-2: 344.330ms
```

You can play with attribute length, `i` is added to avoid caching.

nazar-pc:
Test case for the bug that was fixed accidentally can be taken from #3213 as is.

azakus:
LGTM

PR #3284, opened 2016-01-11
Add support for multiple handlers in on-* annotations

### LINKED ISSUE #1593: Allow for assigning multiple handlers for the same event on the same element, declaratively in HTML
Perhaps `<div on-click="handler1 handler2"></div>`?

TimvdLippe:
Added support for declaring multiple handlers between brackets.
E.g. `on-multiple="[oneMethod anotherMethod]"`. Using brackets, the performance impact on existing single method annotations is minimal. The only check extra introduced is for checking the existence of `[` as first character of the annotation.

Fixes #1593


sorvell:
We're not convinced that this feature fits a wide enough number of use cases to be included directly. Rather than merging this PR, we want to add better extension points so that our awesome users can innovate. See #3369.

PR #3296, opened 2016-01-14
Fixes #3295. Only cache a false-y result for an element's owner shady…

### LINKED ISSUE #3295: getOwnerRoot incorrect when queried after an element is added from within a documentFragment
This can happen when, for example, and element queries its owner root via `Polymer.dom(this).getOwnerRoot()` while it is inside a documentFragment like a dom-if. The owner root should be null in this case, but if the element is subsequently added to a root, the owner root should be updated. It does not, in this case because when the node is added, for performance reasons Polymer.dom avoids resetting owner roots inside this tree specifically if the tree is a documentFragment. 

Alternatively, we can ensure that owner roots are never cached when an element is not in the document. This should be a better approach and maintain good performance.

sorvell:
… root iff the element is currently in the document.

### DIFF
--- a/src/lib/dom-api-shady.html
+++ b/src/lib/dom-api-shady.html
@@ -170,8 +170,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (!node) {
         return;
       }
-      if (node._ownerShadyRoot === undefined) {
-        var root;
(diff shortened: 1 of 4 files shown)


kevinpschaaf:
LGTM

PR #3302, opened 2016-01-16
Computed annotations dependencies

### LINKED ISSUE #3299: Some computed methods need non-contextual dependencies, it's not DRY to always list them
Consider a translation scheme:

```
{{translate(translator, 'Hello World')}}

translate: function(translator, text, params) {
  return translator(text, params);
}
```

The design is that the 'translator` engine is injected by some means (simple property set, e.g.), therefore we must depend on 'translator' to trigger the relevant binding updates. Having to include this static name in every translate binding is a hassle.

One possibility is to add a way to register static dependencies for a computation.

```
{{translate('Hello World')}}

depends: {
  translate: 'translator'
},
translate: function(text, params) {
  return this.translator(text, params);
}
```

Something like that.

TimvdLippe:
Add support for dependencies for computed annotations.

Fixes #3299

### DIFF
--- a/polymer.html
+++ b/polymer.html
@@ -85,7 +85,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // acquire instance behaviors
       this._marshalBehaviors();
       /*
-      TODO(sorvell): It's *slightly() more efficient to marshal attributes prior 
+      TODO(sorvell): It's *slightly() more efficient to marshal attributes prior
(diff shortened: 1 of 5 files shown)


kevinpschaaf:
Per https://github.com/Polymer/polymer/pull/3358#issuecomment-179604592, we're re-considering the approach in https://github.com/Polymer/polymer/pull/3358 as potentially a more robust approach.

azakus:
We decided to go with #3358 instead.

PR #3305, opened 2016-01-16
Remove reference to _composedChildren

rictic:
AFAICT this is the only reference to `_composedChildren`, and this code will throw if `composed` is true and `_composedChildren` doesn't exist.

`_composedChildren` doesn't seem to be set in either Polymer or WebComponents.js:
* https://github.com/Polymer/polymer/search?utf8=%E2%9C%93&q=_composedChildren
* https://github.com/webcomponents/webcomponentsjs/search?utf8=%E2%9C%93&q=_composedChildren


kevinpschaaf:
LGTM

PR #3309, opened 2016-01-19
Fixes #3308. Use an explicit undefined check to test if logical tree …

### LINKED ISSUE #3308: Polymer.dom element accessors can be incorrect when an element has no logical nodes
Consider for example this case: http://jsbin.com/jitiqu/edit?html,output.

Polymer is incorrectly falling back to the element's native firstChild when the logical firstChild is false-y. We must instead do a more explicit undefined check.

sorvell:
…information exists.


kevinpschaaf:
LGTM

Issue #3285, opened 2016-01-13
Nested svgs inside iron-icon get a "nullstyle-scope"


notwaldorf:
Repro jsbin: http://jsbin.com/zuxayik/edit?html,output

The nested `svg` inside the `iron-icon` has a thing called `nullstyle-scope` which I don't think is on purpose:
<img width="590" alt="screen shot 2016-01-11 at 3 18 37 pm" src="https://cloud.githubusercontent.com/assets/1369170/12282814/0110bad8-b956-11e5-8df2-2496cc68742f.png">

This is causing rendering issues on IE11 (in the jsbin, the icon is black on IE11, but red in Chrome)

PR #3310, opened 2016-01-19
Fix svg scope

azakus:
Fix a few spots where `null` can creep into `className` calculation for scoping.


azakus:
Fixes #3285

kevinpschaaf:
LGTM

Issue #3291, opened 2016-01-14
Cross-platform helper for setting an element's CSS transition property.


MeTaNoV:
Similar to `transform`, provide a portable way of using `transition`, should be something like:
```
    transition: function(transition, node) {
      node = node || this;
      node.style.webkitTransition = transition;
      node.style.transition = transition;
    },
```

PR #3318, opened 2016-01-22
Cross-platform helper for setting an element's CSS transition propert…

MeTaNoV:
…y. Fixes Issue #3291.

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -345,6 +345,20 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     /**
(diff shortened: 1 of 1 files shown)


kevinpschaaf:
Thanks, @MeTaNoV.  At this point we rather regret adding this type of sugar in the core Polymer library and probably wouldn't do it again.  Since prefixed `transition` isn't needed for anything > Safari 7.1, we're opting to not include this in the core library.  Behaviors exist as a good way to add this type of thing to your elements when needed.

MeTaNoV:
oki doki

PR #3322, opened 2016-01-23
Fixes 3321

sorvell:


src/lib/dom-api-shady.html
```diff
@@ -81,6 +82,11 @@
         } else {
           TreeApi.Composed.appendChild(container, node);
         }
+      // if a fragment provoked distribution, clean up it's actual dom.
+      } else if (isFragment) {
```
azakus: Shouldn't this just be an `else`?

sorvell: No, this is special handling specifically for fragments. When fragments are added to dom, all of their children are added.

azakus: Ah my bad, the multiple `if`s next to each other was confusing, and I assumed the `else if (isFragment)` was part of the `if (!isFragment)` flow.

test/unit/polymer-dom-content.html
```diff
@@ -1396,8 +1430,33 @@
...
+      frag.appendChild(document.createElement('div'));
+      Polymer.dom(x).appendChild(frag);
+      Polymer.dom.flush();
+      assert.equal(frag.childNodes.length, 0);
+    });
+
   });
 
+  
+
```
azakus: Leftover extra lines?

azakus:
LGTM

kevinpschaaf:
LGTM

PR #3323, opened 2016-01-23
Revert style properties change

rictic:
From fd5778470551f677c2aa5827398681abb1994a88

This broke a number of tests on an internal project.

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -87,7 +87,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var parts = cssText.split(';');
         for (var i=0, p; i<parts.length; i++) {
           p = parts[i];
-          customCssText += p + ';\n';
+          if (p.match(this.rx.MIXIN_MATCH) || p.match(this.rx.VAR_MATCH)) {
+            customCssText += p + ';\n';
+          }
         }
         return customCssText;
       },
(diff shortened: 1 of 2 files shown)


sorvell:
LGTM.

Our plan is to investigate the breaking tests and try to get this changed merged back asap.

nazar-pc:
This is a huge breaking change for many people. As I understand this was not fixed back before release and now implementation does not correspond to specification, am I right?

Your internal projects are awesome, but many people rely on things that were fixed. I suspect correcting of styles order should be enough in mentioned projects. It should be done anyway.

samccone:
@rictic is there anyway you can post test cases?

sorvell:
Unfortunately, there is a flaw in the css shimming for custom properties.
Consider:

<x-foo class="bar"></x-foo>

We shim the `:host` styles of the element via an `x-foo` selector. This
allows a user's style made with a `.bar` selector to properly override the
`:host` styling. However, when custom properties are shimmed, we make a
selector like `x-foo.x-foo-0` for the custom property styling. This
selector *improperly* overrides the user's `.bar` selector.

Placing more styling inside the stylesheet for custom properties exposes
more styling to this flaw. This is what the PR that we reverted did.

We're considering if there is an alternate method of shimming that avoids
this problem. We're also considering adding an option to flip the shimming
behavior between before and after this PR to give users the option to
choose what works best for them.

nazar-pc:
Is there any new test that I can use to improve implementation? Currently I see that existing test was removed, but nothing added instead.

sorvell:
The flaw is in the design of the custom property shimming. The PR simply
put more pressure on this by exposing more styling to this flaw. This bin
shows the basic problem:

http://jsbin.com/jisuxo/edit?html,output

nazar-pc:
I'll try to find some relatively easy solution for this to cover both of mentioned cases.

sorvell:
Excellent. Thanks for your help.

sorvell:
See also: https://github.com/Polymer/polymer/issues/3326

PR #3325, opened 2016-01-23
Ensure dom-if moved into doc fragment is torn down. Fixes #3324

### LINKED ISSUE #3324: dom-if appended to document fragment does not result in its children removed
1. attach a truthy dom-if so that its children are stamped into the dom
2. create a document fragment
2. append the dom-if to the document fragment

The stamped children will not be removed from their original location.

kevinpschaaf:


sorvell:
LGTM

PR #3329, opened 2016-01-24
Allow listeners to be defined by function (not just by name) 

eeid26:
Not be able to pass functions in the element prototype.listeners and the used behaviors is making using behaviors a little tedious than it should be, at least for me. I have to make sure the names are different between each of the behaviors and the element prototype. 

The pull request allows defining the exact function signature on the prototype and the behaviors and each function will be called with the Element prototype first and then the behaviors in descending order to their definition in the behaviors list.

This allows the element, if it needs to, to call the stop stopImmediatePropagation to stop the event from propagating to other behaviors. Also behaviors with a higher index can call stopImmediatePropagation to prevent behaviors from receiving the events.

Like this:

``` javascript
    Polymer.BehaviorBase1 = {

      properties: {
        disabled: {
          type: Boolean,
          notify: true
        }
      },

      listeners: {},

      _disabledChanged: function(event, newValue) {
        console.log("BehaviorBase1 _disabledChanged called.");
      }
    };

    Polymer.Base.mixin(Polymer.BehaviorBase1.listeners, {
      'disabled-changed': Polymer.BehaviorBase1._disabledChanged
    });

    Polymer.BehaviorBase2 = {

      properties: {},

      listeners: {},

      _disabledChanged: function(event, newValue) {
        console.log("BehaviorBase2 _disabledChanged called.");
      }

    };

    Polymer.Base.mixin(Polymer.BehaviorBase2.listeners, {
      'disabled-changed': Polymer.BehaviorBase2._disabledChanged
    });

    var proto = {

      behaviors: [
        Polymer.BehaviorBase1,
        Polymer.BehaviorBase2
      ],

      listeners: {
        mousedown: '_onMouseDown'
      },

      _disabledChanged: function(event, newValue) {
        console.log("Element _disabledChanged called.");
      }

    };

    Polymer.Base.mixin(proto.listeners, {
      'disabled-changed': proto._disabledChanged
    });

    Polymer(proto);
```

See the new unit tests for details.

Thank you

### DIFF
--- a/src/micro/behaviors.html
+++ b/src/micro/behaviors.html
@@ -133,10 +133,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     _marshalBehaviors: function() {
-      for (var i=0; i < this.behaviors.length; i++) {
+      // Add the element listeners first. This way, the element will be able to
+      // call stopImmediatePropagation to prevent other behaviors from getting the event.
+      this._marshalBehavior(this);
+
+      // add the behaviours listeners in reverse order what they are in the list.
+      // This will allow behaviours that have a higher index in the behaviours
+      // list to call stopImmediatePropagation to stop event propagating to other
+      // behaviours in the list with lower indices.
+      for (var i = this.behaviors.length - 1; i >= 0; i--) {
         this._marshalBehavior(this.behaviors[i]);
       }
-      this._marshalBehavior(this);
     }
 
   });
--- a/src/standard/events.html
+++ b/src/standard/events.html
@@ -129,8 +129,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     _createEventHandler: function(node, eventName, methodName) {
       var host = this;
+      var byFunction = typeof methodName === 'function';
       var handler = function(e) {
-        if (host[methodName]) {
+        if (byFunction) {
+          methodName.call(this, e, e.detail);
+        } else if (host[methodName]) {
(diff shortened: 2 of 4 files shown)


kevinpschaaf:
@eeid26 We can't accept this PR as-is with the change to the behavior order, as that would be a breaking change.  We'd recommend that the element prototype override the event _handler_ on the behavior to address the concern you raised in the comment.

The other approach that has proven useful is to just namespace behavior methods with their behavior name which is just a simpler approach.  That said, since the change is so minor, we're inclined to merge the rest of this, however if we make the change for `listeners` we think it should probably be done for `observer` in the properties block also for consistency.  Could you update the PR to include that?

Note we can't easily apply the same fix for `computed` or `observers` since those require strings to declare dependencies in addition to names.

sorvell:
Ping, @eeid26 can you address the feedback? Thanks.

eeid26:
Sorry for the late response.
I understand how changing the order of how the behaviors are registered in the listeners object might break some older code.

I think maybe I should wait and see how the new polymer inheritance is implemented and see if my change would be needed.

@kevinpschaaf about your comment  
"The other approach that has proven useful is to just namespace behavior methods with their behavior name which is just a simpler approach"

That is exact;y what I was trying to avoid. Is to have to name functions differently in different behaviors.

TimvdLippe:
As listeners are now in the legacy implementation, I am closing this PR. You can use the native methods to add event listeners with functions to the DOM.

PR #3338, opened 2016-01-27
Fixes #3337. When a doc fragment is added, only update the invalidati…

### LINKED ISSUE #3337: Incorrect distribution after adding a fragment that does not include an insertion point after adding a fragment that does
Document fragments added to shadyRoots are scanned for insertion points. If one is found then the shadyRoot's list of insertion points is invalidated and on next distribution, the list is updated. However, if a fragment that does not include an insertion point is added after one that does but before distribution, the insertion point list invalidation improperly gets unset.

Problem spot: https://github.com/Polymer/polymer/blob/master/src/lib/dom-api-shady.html#L94

sorvell:
…on state of the insertion point list of the shadyRoot IFF it is not already invalid. This fixes an issue that was detected when an a doc fragment that did not include an insertion point was added after one that did but before distribution.

### DIFF
--- a/src/lib/dom-api-shady.html
+++ b/src/lib/dom-api-shady.html
@@ -91,8 +91,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _addNode: function(node, ref_node) {
       var root = this.getOwnerRoot();
       if (root) {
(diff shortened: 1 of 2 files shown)


azakus:
LGTM

PR #3344, opened 2016-01-28
Re-introduction of improved CSS selectors specificity fix

### LINKED ISSUE #3326: Re-fix custom properties css specificity mixin issue
https://github.com/Polymer/polymer/pull/3323 reverted a desirable change from fd57784 that we'd like to have back. Need to track down the tests that broke, find the root cause, and re-integrate the change with a fix.

nazar-pc:
Re-introduction of improved CSS selectors specificity fix reverted in 0a0b580981bcd7ba6af33161993355bed99056cd

Now allows `.super-class` to override `:host`'s styling

Fixes #3326

@azakus, @sorvell, @rictic for review


sorvell:
Fixed via #3413. Thanks!

PR #3350, opened 2016-01-30
Fix bindings with special characters

### LINKED ISSUE #3349: Data binding variable with dash
As of Polymer 1.2.4 I can no longer data bind to a variable name with a dash in it.

`selected-family="{{ styles.h1.font-family }}"`

When I debug the literal value of selectedFamily is "{{styles.h1.font-family}}". This was working in 1.2.3.

TimvdLippe:
The special regex characters must be escaped to match correctly the literal meaning. 

Fixes #3349

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -90,7 +90,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     _bindingRegex: (function() {
-      var IDENT  = '(?:' + '[a-zA-Z_$][\\w.:$-*]*' + ')';
+      var IDENT  = '(?:' + '[a-zA-Z_$][\\w.:$\\-*]*' + ')';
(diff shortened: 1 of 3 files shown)


src/lib/annotations/annotations.html
```diff
@@ -90,7 +90,7 @@
     },
 
     _bindingRegex: (function() {
-      var IDENT  = '(?:' + '[a-zA-Z_$][\\w.:$-*]*' + ')';
+      var IDENT  = '(?:' + '[a-zA-Z_\\$][\\w\\.:\\$\\-\\*]*' + ')';
```
azakus: The only character that should need an escape here is `-`

TimvdLippe:
Rebased to current master and fixed comment.

sorvell:
Can you rebase one more time? Otherwise LGTM... Thanks!

TimvdLippe:
Rebased on current master.

Issue #2162, opened 2015-07-28
Two-way data binding and negation expression


tomalec:
[Docs](https://www.polymer-project.org/1.0/docs/devguide/data-binding.html#expressions-in-binding-annotations) says that Polymer supports negation expressions:
```hmtl
<template>
  <div hidden="{{!enabled}}"></div>
</template>
``` 

As `{{}}` suggests two-way binding I thought it works both ways. However it seems it doesn't, at least within `dom-bind`.
Am I doing something wrong? Do you have any plans to support binding negated value back?

I found it quite useful, even for some basic cases like: `<input type="checkbox" checked="{{!booleanSwitch::change}}">`

kevinpschaaf:
Closing as duplicate of https://github.com/Polymer/polymer/issues/1715.

We are considering supporting this, although there are some corner-cases that need to be evaluated; for example, if you bind `{{!stuff.enabled}}` to `checked` and `stuff.enabled` is `undefined`, negating it will send `true` to the checkbox, which will round-trip back and set `stuff.enabled` to `false` (i.e. changing it from its `undefined` value), and there may be cases where that is undesirable.

In the meantime, you can manually perform the two-way negation as follows:

```html
<input type="checkbox" checked="{{!booleanSwitch}}" on-change="checkedChanged">
```
```js
checkedChanged: function(e) {
  this.booleanSwitch = !e.target.checked;
}
```

PR #3352, opened 2016-01-31
Fix negation when a negated binding is changed

### LINKED ISSUE #1715: ! operator doesn't apply to inverse boolean values in bindings upward?
Jsbin: http://jsbin.com/coxucaduyu/1/edit?html,console,output

Given:

    <x-checkbox is-checked="{{!hostChecked}}">

I would expect that if the checkbox sets `this.isChecked = true` then `hostChecked` should be set to false, but this doesn't seem to be the case. Is the `!` operator in bindings not meant to perform this inverse, or am I doing something dumb in the jsbin?

TimvdLippe:
After I fixed the issue, I discovered that https://github.com/Polymer/polymer/tree/1757-kschaaf-negate-2way has fixed it in a similar manner. This Pull-Request also contains a test for the change and is up-to-date with current master.

The `negation` test in `test/notify-path` was removed as all other tests assumed that negated two-way bindings are NOT updated. Therefore the test suite verified that negating a value of 42 would not change the value of the parent to `false`. Since 42 is truthy, the value of the parent would become `false` failing the `assert.equal(value, 42)`.

Fixes #1715


kevinpschaaf:
We'll take another look at this.  IIRC we didn't make a PR and merge that branch due to concerns noted in this comment: https://github.com/Polymer/polymer/issues/2162#issuecomment-126062618.  We just need to spend some time to re-evaluate whether this could break current users.

TimvdLippe:
@kevinpschaaf A situation I could think of where this would go wrong is where there is transformation logic in an `observer`. Although this could lead to unexpected behavior, I think the root issue is a different one: in Polymer/Javascript a boolean can have 3 values. The definition of a boolean is that it can have two values: `true` and `false`. Therefore I think it is more logical to fix the root issue by defaulting booleans to `false` instead of `undefined`. This might be a breaking change and should be investigated. This is also one of the roadmap goals: ["Fix and make the usage of undefined consistent throughout the library"](https://github.com/Polymer/project/blob/master/Roadmap.md#core-library).

azakus:
Looks good after additional testing!

PR #3354, opened 2016-01-31
Fix: There is no effect of kind 'computedAnnotation'

kaste:
Although `computedAnnotation` is the better name, you actually named it `annotatedComputation`.

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -143,7 +143,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var EFFECT_ORDER = {
         'compute': 0,
         'annotation': 1,
-        'computedAnnotation': 2,
+        'annotatedComputation': 2,
         'reflect': 3,
         'notify': 4,
         'observer': 5,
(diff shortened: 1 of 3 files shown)


azakus:

kaste:
Sorry for not giving a description. I just thought this is a simple oversight. 
Did you take a look at the context of this patch? It's about the order effects are applied. Right now order is undefined. Unfortunately, because of this kind of error and how Polymer is written (not in testable units), this feature cannot be tested. AFAIK at least. :wink: 

Bug:
 
fx-order-before

Fix:

fx-order-after

TimvdLippe:
You can test this feature by creating several effects and update a property. Then verify that the effects are propagated in order as expected

kaste:
AFAIK I can't just create effects and plug them to an element. For the
observers it's doable, but the annotations type is.. well complex.

In which file should I put the test?

Tim van der Lippe <notifications@github.com> schrieb am Do., 4. Feb. 2016
16:23:

TimvdLippe:
You can define an element which has a property with an observer. Then a child element with a property also observable and bound to the parent. A computed function on a different parent property and inside the function you can check the order. Then a div with textcontent a computed property. I think you got all the effects then, currently on mobile so cant verify.

I would put the test in test/unit/bind (with dom-modules defined in test/unit/bind-elements)

TimvdLippe:
@kaste @azakus I have submitted a test suite for this PR here: https://github.com/kaste/polymer/pull/1

If Azakus accepts this test suite, Kaste you should merge the PR on your repo and then this PR should be automatically updated with the new commit. Then Azakus could merge it.

kevinpschaaf:
This is great, thank you @TimvdLippe and @kaste!

LGTM :100:

Issue #3353, opened 2016-01-31
dom-if (in restamp mode) should not forward changes as soon as `if` gets false 


kaste:
A completely idiotic example can be found here: http://jsbin.com/tifuciyawo/edit?html,console

B/c of the debounced render, the stamped instance usually gets notified about all changes that happen in the current microtask even if the instance will teardown at the end of the microtask as a result of `render`. 

Low hanging fruit: 

```
_forwardParentProp: function (prop, value) {
  if (this.restamp && !this.if) { return; }
  ...
},
_forwardParentPath: function (path, value) {
  if (this.restamp && !this.if) { return; }
  ...
} 
```

Additionally the effects should be sorted so that child templates (or: effects pointing to a `_parent_...`) are applied at the end.

kevinpschaaf:
Considering this blocked on https://github.com/Polymer/polymer/issues/3343 per comments on https://github.com/Polymer/polymer/pull/3355#issuecomment-179599682

Issue #3343, opened 2016-01-28
Do not require notifyPath or set to update subproperties of Objects


azakus:
Based on user feeback, a confusing and cumbersome part of the databinding system is that changing data-bound objects' subproperties require using the `set` api, or duplicating the path to the subproperty with `notifyPath`.

Instead, we should disable identity checking for objects and always dirty check subproperties for them to allow users to either set the root object directly, or call `notifyPath`, in an easier fashion.

MeTaNoV:
+1

TimvdLippe:
This could be added quite easily in #3238 in [the property dirty check method](https://github.com/Polymer/polymer/pull/3238/files#diff-c8d866d6e3371fd7ff2315aab52c74c6R81)

TimvdLippe:
This has shipped in the mixing [MutableData](https://www.polymer-project.org/2.0/docs/api/mixins/Polymer.MutableData)

PR #3355, opened 2016-02-01
dom-if: Don't forward property changes if instance is about to teardown

kaste:
Because of the debounced render, the instance might be alive until the end of the microtask, thus seeing all property changes within this exact microtask coming from host. See #3353.

### DIFF
--- a/src/lib/template/dom-if.html
+++ b/src/lib/template/dom-if.html
@@ -168,6 +168,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // Called as side-effect of a host property change, responsible for
     // notifying parent.<prop> path change on instance
     _forwardParentProp: function(prop, value) {
+      // B/c of the debounced render, `this._instance` might be alive 
+      // until the end of the microtask.
+      if (this.restamp && !this.if) { return; }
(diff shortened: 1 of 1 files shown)


kevinpschaaf:
The spirit of this is right on.  However, prior to the debouncer running at microtask, if the `if` is changed back to true then data will be out of sync (the purpose of debouncing the template render is to avoid thrashing in this exact case), so I'm not sure we can take this simple approach.

A more robust approach would be to disable all property/path forwarding when `if` is `false` for both the stamp & restamp case, and then "pull" all `_parentProps` back into the Templatizer instance when `if` goes back to true so that all falsy `if`'s benefit from avoiding unnecessary effects.  However, we haven't done this to date because it is not compatible with how we pass array `splices` information via the normal property system; since `splices` information is temporal, if a `dom-repeat` behind a `dom-if` misses some splices, it will forever be out of sync, unfortunately.  There is a path to addressing this via https://github.com/Polymer/polymer/issues/3343, so I would consider resolving this issue effectively blocked on that.

kevinpschaaf:
We'll go ahead and close this PR until we have a solution for #3343, since the solution would be significantly different.

PR #3358, opened 2016-02-01
Support dynamic functions for computed annotations.

kaste:
Alternate take for #3299

### DIFF
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -227,7 +227,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var m = expression.match(/([^\s]+?)\(([\s\S]*)\)/);
       if (m) {
         var sig = { method: m[1], static: true };
-        if (this.properties[sig.method]) {
+        // TODO(kaste): Optimize/memoize `getPropertyInfo`.
+        if (this.getPropertyInfo(sig.method) !== Polymer.nob) {
           sig.static = false;
           sig.dynamicFn = true;
         }
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -171,6 +171,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
               arg);
           }
         }
+        if (sig.dynamicFn) {
+          // trigger=null is sufficient as long as we don't allow paths to be
+          // used. If we change our mind, we must first implement this in the
+          // effects anyway where we basically do a `fn = this[methodName]` at
+          // the moment.
+          this.__addAnnotatedComputationEffect(
+              sig.method, index, note, part, null);
+        }
       }
     },
 
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -79,13 +79,26 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     _addComputedEffect: function(name, expression) {
       var sig = this._parseMethod(expression);
+
+      var dynamicFn = sig.dynamicFn;
+
       for (var i=0, arg; (i<sig.args.length) && (arg=sig.args[i]); i++) {
         this._addPropertyEffect(arg.model, 'compute', {
           method: sig.method,
           args: sig.args,
           trigger: arg,
-          name: name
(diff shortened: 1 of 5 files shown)


kevinpschaaf:
This is super clever!  Of the two approaches, the one in #3302 is going to probably feel slightly less magic to users since the hidden dependency for the dynamic function is... hidden.  We still need to decide whether the extra sugar there will be appropriate, but for now we're going to close this in favor of the other.

kaste:
:ok: Just for the record https://github.com/kaste/polymer/tree/computed-bound-methods has a generalized approach of this: allowing computed, annotatedComputations and observer to have its method/base function be dynamic.

kevinpschaaf:
Reopening per https://github.com/Polymer/polymer/pull/3358#issuecomment-179604592.

Would you mind merging your changes to make all method calls (observers, computed, computed annotations) dependent on themselves?  Will also comment inline in the code on a couple things.

kevinpschaaf:
Oops, I see you just updated the same branch after it was closed, cool.

src/standard/effectBuilder.html
```diff
@@ -193,6 +226,10 @@
       var m = expression.match(/([^\s]+?)\((.*)\)/);
       if (m) {
         var sig = { method: m[1], static: true };
+        if (this.properties[sig.method]) {
```
kevinpschaaf: To be compatible with `behaviors`, I think you can just change this to `this._propertyInfo[sig.method]`.  Could you add tests a `computed`-computed annotation function supplied by a behavior?

test/unit/bind-elements.html
```diff
@@ -486,3 +486,35 @@
     });
   </script>
 </dom-module>
+
+<dom-module id="x-bind-computed-property">
+  <template>
+    <div id="check">[[translate('Hello World.')]]</div>
```
kevinpschaaf: Could you also add a test that shows `translate` called with a non-static argument, e.g. `[[translate(foo)]]` where `foo` has a default value, but `translate` is computed async?  This is to cover the case that a dependency to computing the `translate` function is async (e.g. due to loading a string bundle via an xhr).  Although all args are valued, the method is still undefined, and it should not warn.

I believe your `bailoutEarly` flag covers this case, but would be good to test.

kaste:
I'll look into this tomorrow. It's about midnight in Germany right now ;-)

kaste:
Great, git hates me.

kaste:
I thought I just made a rebase with master? Bah

kaste:
Anyway, `this._propertyInfo` is not available (it is undefined) when this code runs. I had to use `this.getPropertyInfo` to get it going. That call is therefore completely unoptimized, but honestly all this propertyInfo code in `properties.html` looks like someone left the building before finishing it.

kevinpschaaf:
Thanks.  On the [3358-kaste-computed-bound-methods](https://github.com/Polymer/polymer/tree/3358-kaste-computed-bound-methods) branch in https://github.com/Polymer/polymer/commit/87004c901b4a35e6565fb725e91f24ec4b223d3c I moved `_prepPropertyInfo` earlier so it can be used here to avoid the slow lookup, but it's still failing one test.  Will ask @azakus to take a look at it next week.

kaste:
Yeah, I tried moving _prepPropertyInfo as well. It failed on some
hostAttributes thing, and I had no clue b/c there is literally no
dependency between the two. 😯

Kevin Schaaf <notifications@github.com> schrieb am Sa., 13. Feb. 2016 03:39:

azakus:
@kevinpschaaf `_addHostAttributes` happens in `_prepBehaviors` which comes after `_prepPropertyInfo` in your patch at 87004c9

azakus:
Ok, I see the issue now. `_prepPropertyInfo` depends implicitly on behaviors and effects being prep'd first. Kevin's patch is setting up property info before effects, so the effect of hostAttributes being overridden by attributes in composition is missing, at least as far as I understand this codepath.

azakus:
After looking at this for a while, we're fine with the possible performance hit instead of using `_propertyInfo` directly. It's just too intertwined into the property effect and behavior processing.

LGTM!

kaste:
Cheers 🍻

Issue #1303, opened 2015-03-17
Support for on demand databinding


paulgingrich:
I have a use case that is not covered by polymer today.  I have a database that I want to expose as a non visual polymer component.  Widgets can bind to the data in the database using mustache notation or through the polymer editor.  For example,
<input value={{$.db.entry}}> where db is the id of the database polymer component.  Because the database is large in relation to the number of entries that the application will bind to, rather than declaratively publish all entries in the database, I would like to only publish the ones that are used by other widgets on the fly on demand.  Currently, the only way to do this is to duplicate the information and have the user inform the database which entries it is interested in in a declarative fashion.  I would prefer something like a pre-binding event where I can validate that the entry is valid for my database, and create the onEntry() handler to responde to changes after the binding has been created.

Here is a link to further discussions on this topic:
http://stackoverflow.com/questions/28639863/how-to-dynamically-create-properties-in-polymer-as-needed/29106288

TimvdLippe:
@paulgingrich This is quite an old request, but Polymer 2 allows you to customize the bindings parser with [overriding `_parseBindings`](https://www.polymer-project.org/2.0/docs/api/elements/Polymer.Element#method-_parseBindings) Could you confirm/deny that you can implement your usecase with that?

TimvdLippe:
Closing for now as I assume this is possible. Feel free to reopen if you are unable to make it work with Polymer 2!

PR #3362, opened 2016-02-03
Add raw html bind annotation

shawncplus:
This adds a new `{# #}` annotation which will bind to `innerHTML` instead of `textContent`. By design it does not work on compound bindings for some reason I can't really justify. After the 5th or so request in the Polymer slack on advice on how to bind a value with html I threw this together. Please don't ban me for this insanity.

Usage:

```html
<dom-module id="x-foo">
  <template>
    <p class="message">{#messageBody#}</p>
  </template>
  <script>
  Polymer({
    is: 'x-foo',
    properties: {
      messageBody: {
        value: '<b>Some html string here</b>'
      }
    }
  });
  </script>
</dom-module>
```

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -143,7 +143,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           compoundIndex: parts.length,
           value: value,
           mode: mode,
-          safe: m[1][1] === '#',
+          raw: m[1][1] === '#',
           negate: negate,
           event: notifyEvent,
           customEvent: customEvent
@@ -180,12 +180,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // Initialize the textContent with any literal parts
         // NOTE: default to a space here so the textNode remains; some browsers
         // (IE) evacipate an empty textNode following cloneNode/importNode.
-        var safe = parts.length !== 1 ? false : parts[0].safe;
+        var raw = parts.length !== 1 ? false : parts[0].raw;
         node.textContent = this._literalFromParts(parts) || ' ';
         var annote = {
           bindings: [{
             kind: 'text',
-            name: safe ? 'innerHTML' : 'textContent',
+            name: raw ? 'innerHTML' : 'textContent',
             parts: parts,
             isCompound: parts.length !== 1
           }]
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -101,8 +101,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
                                     '(?:' + ARGUMENTS + '?' + ')' +
                                   '\\)\\s*' + ')';
       var BINDING = '(' + IDENT + '\\s*' + ARGUMENT_LIST + '?' + ')'; // Group 3
-      var OPEN_BRACKET = '(\\[\\[|{{)' + '\\s*';
-      var CLOSE_BRACKET = '(?:]]|}})';
+      var OPEN_BRACKET = '(\\[\\[|{{|{#)' + '\\s*';
+      var CLOSE_BRACKET = '(?:]]|}}|#})';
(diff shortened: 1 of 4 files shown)


miztroh:
Doesn't this already work with ``<div inner-h-t-m-l="{{html}}"></div>``?

shawncplus:
Yes, except I, consider that an absolutely horrendous hack.

BLamy:
If it's a `do not use unless you know what you are doing` feature, it makes sense for it to be hidden behind a horrendous hack.

miztroh:
Say what you will, but it least it doesn't add new syntax.

arthurevans:
This says "safe," but I don't think it actually does anything to ensure the bound text doesn't do Bad Things any more than `inner-h-t-m-l`.

shawncplus:
@arthurevans True, as I mentioned to @BLamy in slack. I'm not actually expecting this to be accepted, more start a conversation since binding HTML isn't exactly a niche problem. Without making Polymer opinionated to Caja or something it would rely on the user sanitizing their HTML. Which they should be doing anyway but "You should have done it right" isn't the best design philosophy, I fully admit

jrajav:
Without knowing a whole ton about it, it seems like it would be prohibitively expensive (both in the library size and in runtime compute) to have a syntax that enforces safety, since it would involve processing and ensuring the sanitization of the HTML. However, there are legitimate uses for inserting HTML declaratively. For instance, in my own case, I need to insert a string dynamically based on a Polymer property, but I also need to wrap a portion of the string with HTML tags dynamically to style incremental search highlights. It is easy to style things with CSS when only the overall presentation is changing, but when the tags and the text content need to change orientation dynamically in response to logic, such as in this case, injecting HTML is the only way to accomplish it. And it would be nice to have a clean declarative way to do that with Polymer and not have to set innerHTML.

It might be even nicer to have bindings *within* the injected HTML, akin to `this.injectBoundHTML` from 0.5, but that's a different story.

miztroh:
To update my previous example:

``<div inner-h-t-m-l="{{sanitize(html)}}"></div>``

The current syntax and functionality allows us to do all that's being asked (aside from data bindings in the dynamic HTML).  Just find some JS HTML sanitizer (I found this one from a quick Google search: https://github.com/punkave/sanitize-html) and incorporate it into your custom element as a function, then use the function to sanitize your dynamic HTML on-the-fly.

shawncplus:
A real annotation has a few advantages over `inner-h-t-m-l`:

* consistency: why does binding html content require an attribute when binding text uses an annotation?
* expressiveness: at a glance you can tell "This is binding raw html", it's easy to miss `inner-h-t-m-l` as just another bound attribute then try to use an annotation to that element's content and have not work inexplicably (`<div id="mydiv" hidden$="[[!selectMode]]" inner-h-t-m-l="[[foo]]" class="baz">[[bar]]</div>`
* No "undefined behavior" as with the previous example why does `inner-h-t-m-l` take precedence over the `[[bar]]` annotation? You have no idea why and it's not even remotely documented. When you have an annotation you _can't_ mix the two, you can only use one at a time

rictic:
Hey, this PR is totally safe! As long as no one makes any mistakes!

torch juggling gone wrong


More srsly, this is totally a common request, but I think this is a use case where you want syntactic vinegar rather than a sugary sweet and easy to use footgun in the toolbox. Alternatively, what about a [`caja-html`](https://github.com/google/caja) element?

sorvell:
As @rictic noted, because of the security concerns here, we cannot add additional sugar around binding to raw html to the core library. We plan to allow users to [hook into the binding system](https://github.com/Polymer/polymer/issues/3369) more directly such that we don't need to arbitrate these types of issues at the library level.

jrajav:
@sorvell I can't really tell from the issue you linked - does that mean that it would be possible to write a custom element that would allow injecting dynamic *data-bound* HTML at some point?

sorvell:
No, that will be covered when we address https://github.com/Polymer/polymer/issues/1303.

Issue #1976, opened 2015-06-28
template doesn't work inside svg elements


bendavis78:
For example:

``` html
    <dom-module id="test-element">
      <template>
        <svg class="clock" version="1.1" xmlns="http://www.w3.org/2000/svg">
          <template is="dom-repeat" items="{{positions}}" as="cx">
            <ellipse r="25" cx$="{{cx}}" cy="25" stroke="black"></ellipse> 
          </template>
        </svg>
      </template>
      <script>
        Polymer({
          is: 'test-element',
          ready: function() {
            this.positions = [0, 100, 200, 300];
          },
        });
      </script>
    </dom-module>
```

This produces the following JS error:

```
Uncaught TypeError: Cannot read property 'nodeType' of undefined
```

The problem seems to be that the dom template object doesn't have a content property. This worked fine in Polymer 0.5.

Demo here: http://jsbin.com/wiqowo/31/edit?html,output

bendavis78:
It looks like some code didn't make it over from 0.5 that explicitly addressed this issue. Template elements with a `<svg>` parent have the same `namespaceURI` (`http://www.w3.org/2000/svg`), so the browser doesn't implement `HTMLTemplateElement`. The bootstrap code in 0.5 basically transplanted all child nodes of any svg-namespaced template into a new html-namespaced template, and then inserted that node back in its place.

If anyone else is looking for a workaround, you can put [this snippet](https://gist.github.com/bendavis78/15528ca2f501c44f2fa4) at the top of your element's `<script>` tag (it must be evaluated before call to `Polymer()`). It basically performs the same operation as it was done it 0.5.

nomego:
We also hit this issue, thanks for the workaround @bendavis78 .

Will it be fixed in Polymer?

canufeel:
@bendavis78 your snippet was a day saviour for me but it still has some problem. If you have more than one polymer element on the page and this svg element is not the first one the snippet breaks. This line:
    var root = doc.querySelector('dom-module > template').content;
should be changed so `dom-module` becomes the actual `#id-of-svg-element` which is put in the `<dom-module id="id-of-svg-element">`

kevin-smets:
+1

shmendo:
+1

ocampesato:
Same error in Chrome 49, even with the tweaked version of svg-template-support.js, and also:
svg-template-support.js:4 Uncaught TypeError: Cannot read property 'content' of null

raghav135:
[...]

PR #3372, opened 2016-02-04
Namespace support (a.k.a. SVG handling)

christophe-g:
A first go at allowing custom elements to be defined in other namespaces than html.

This is especially useful when working with SVG. Two main tricks : 
- get the proper native prototype when a namespace is declared (src/micro/extends.html, line 76), and 
- re create template elements defined in appropriate namespace (src/mini/template.html, line 45 and below) .

This allows to declare template like (note **`<g is="d3-enable-zoom">`**):  

``` html
 <template>
    <h2>Testing bindings</h2>
    <p>opacity</p>
    <paper-slider step="0.1" min="0" max="1" value="{{opacity}}"></paper-slider>
    <svg id="svg" width="900" height="450">
      <g is="d3-enable-zoom" test="here we go" opacity="[[opacity]]">
        <g id="content"></g>
      </g>
    </svg>
  </template>
```

and `d3-enable-zoom` declared as 

``` html
<dom-module id="d3-enable-zoom">
  <template>
    <rect id="zoom" class="zoom" data-prevent-focus="true" width="200%" height="200%" x="-50%" y="-50%" opacity$="[[opacity]]" fill="#123"></rect>'
    <g id="zoomGroup" class="zoomGroup">
      <content></content>
    </g>
  </template>
</dom-module>
<script>
(function() {
  'use strict';

  Polymer({
    is: 'd3-enable-zoom',
    extends: 'g',
    namespace: 'http://www.w3.org/2000/svg',
 ....
```

Live example to follow
Cheers,
C.

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -329,7 +329,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // Attribute or property
         var origName = name;
         var kind = 'property';
-        if (name[name.length-1] == '$') {
+        var last = name[name.length-1]
+        if (last == '$' || last == '_') { //hack because setting attribName$ when we create element is not allowed
           name = name.slice(0, -1);
           kind = 'attribute';
         }
--- a/src/lib/polymer-bootstrap.html
+++ b/src/lib/polymer-bootstrap.html
@@ -51,9 +51,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // Note: need to chain user prototype with the correct type-extended
       // version of Polymer.Base; this is especially important when you can't
       // prototype swizzle (e.g. IE10), since CustomElements uses getPrototypeOf
+
       var base = Polymer.Base;
       if (prototype.extends) {
-        base = Polymer.Base._getExtendedPrototype(prototype.extends);
+        base = Polymer.Base._getExtendedPrototype(prototype.extends, prototype.namespace);
       }
       prototype = Polymer.Base.chainObject(prototype, base);
       prototype.registerCallback();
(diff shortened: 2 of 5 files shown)


azakus:

christophe-g:
The problem that this PR is fixing is allowing custom element in other namespace than html. This is mainly useful for dealing with SVG. 
To see it in action : 
http://polymerel.github.io/d3-layout-tree/components/d3-layout-tree/#d3-layout-tree
C.

sorvell:
This is a great PR. Thanks for doing all the heavy lifting here.

We'd like to allow this to be implemented as a behavior rather than integrating it into the core library. The benefit would be that users that don't need svg elements don't have to load this code.

christophe-g:
@sorvell Thanks a lot for the feedback. I am very new to Polymer and really love it - so I am happy to contribute ; ) 

re behavior vs core, as long as it is possible to handle foreign namespace, it doesn't really matter. However I guess some code would need to go to core anyway (e.g. _getNativePrototype in https://github.com/PolymerEl/polymer/commit/3b1c53a0388e47db8306e51f14ac9cb4d4acd2dd)  and the PR is not really heavy in terms of additional code. 

No time to work on tests yet, pretty busy those days. 

Cheers,
C.

binoculars:
:+1:

fooloomanzii:
This should be considered to be a part of Polymer

christophe-g:
Hey, 
Any news re. polymer supporting other namespaces ?
Would it be better supported under Polymer 2.0 ? 
C.

n1ywb:
This only adds ~60 lines of code. It would also presumably permit templating in other namespaces like MathML, XML, etc, not "only" SVG. Seems like core library material to me.

TimvdLippe:
I am closing this PR, as it targets the 1.x codebase. We would be happy to review a PR targeting 2.x which introduces this functionality as a mixin.

renfeng:
For people looking for a working solution,
https://github.com/Polymer/polymer/issues/1976#issuecomment-225454535

PR #3373, opened 2016-02-04
dom-template: parentProps should not override argument based props

kevinpschaaf:
Freshened version of original PR here: https://github.com/Polymer/polymer/pull/1889


kevinpschaaf:
CLA verification passed on https://github.com/Polymer/polymer/pull/1889

PR #3386, opened 2016-02-06
Make behaviors array unique

### LINKED ISSUE #3365: Make behaviors list a unique array
See https://github.com/Polymer/polymer/pull/1912#issue-89357111 for details

TimvdLippe:
Re-submission of #1912 

Fixes #3365

### DIFF
--- a/src/micro/behaviors.html
+++ b/src/micro/behaviors.html
@@ -65,6 +65,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     _desugarSomeBehaviors: function(behaviors) {
+      var behaviorSet = [];
       // iteration 1
       behaviors = this._flattenBehaviorsList(behaviors);
       // iteration 2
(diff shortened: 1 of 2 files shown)


nazar-pc:
I think we don't really need auxiliary array at all.
`behaviorSet.indexOf(b) === i` should be enough, it will become false on each sequential presence of the same element, since every time index of the first occurrence of the element will be returned.
It should also be faster and less code :rocket: 

Otherwise :+1: for handling such cases.

TimvdLippe:
@nazar-pc Then you would return the original array containing the duplicates. Also, since behaviors are right-to-left, `behaviors.lastIndexOf` should be used. Then if we want to remove the duplicates we have to `splice` out the element, which is I think less performant and harder to grasp than a simple extra array. Therefore using an auxilliary array is in my eyes easier and faster.

sorvell:
LGTM, but we're going to do some additional testing before merging this. Because of #3418 it's possible there could be some unintended consequences here.

azakus:
The additional testing looks good. LGTM to merge

PR #3406, opened 2016-02-10
Fix parsing of parenthesis in default of variable declaration

### LINKED ISSUE #3221: Custom properties don't use default values if they contain parenthesis
Pretty easy to repro:

```
background-image: var(--foo-background-image, url(http://placehold.it/400x300));
```

seems like it should work but doesn't. No default background image is displayed.

TimvdLippe:
There were two separate issues which lead to un-parsable default declarations with parenthesis:

1. The regex to match the default declaration of a `var()` did not ignore `(` and matched on every variable declaration. Even if there was 1 pair of parenthesis in the default declaration.
* For extraction of a default declaration with a `colon`, the `String.split` would cut the declaration into three pieces instead of one. Therefore only the first occurrence of the colon should be used to split the declaration.

There is one other separate bug that became apparent: Matching with a default declaration with parenthesis leads to a double declaration of the css rule. The second rule is always `css-property: var( };`. I assume this is a bug, but I am not certain.

Fixes #3221

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -160,12 +160,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             if (m) {
               p = this.valueForProperty(props[m[1]], props);
             } else {
-              var pp = p.split(':');
-              if (pp[1]) {
-                pp[1] = pp[1].trim();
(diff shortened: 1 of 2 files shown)


sorvell:
LGTM. Can you file a separate issue on the unrelated bug? Thanks!

PR #3407, opened 2016-02-11
Disable tap gesture when track gesture is firing for ancestor node

### LINKED ISSUE #3405: Unexpected tap event from child when parent fires track events
Go to http://jsbin.com/bupawoluhu/edit?html,output with Chrome on Android. Drag on the child div with dx, dy < 25. Observe that both track events from the parent and a tap event from the child are fired, but only track events from the parent are expected.

azakus:
Move tap disable into track start

Fixes #3405


azakus:
PTAL @sorvell

kevinpschaaf:
LGTM

Issue #3401, opened 2016-02-10
Computed bindings don't run when an argument is undefined


stephenbunch:
Hey guys,

I'm trying to use a computed binding where one of the input arguments is `undefined`. The computed binding is never executed. However, it runs as expected when there is only one argument.

http://codepen.io/stephenbunch/pen/XXoWbe

MeTaNoV:
This is normal behavior, check that part of the doc: https://www.polymer-project.org/1.0/docs/devguide/data-binding.html#dependent-properties

stephenbunch:
@MeTaNoV Ahh, I see.

> The computing function **is not called until all dependent properties are defined (!=undefined)**.

In that case, it still doesn't make sense that it works when there's only one argument.

MeTaNoV:
you're right, it shouldn't!

MeTaNoV:
I did some spelunking and found out that in https://github.com/Polymer/polymer/blob/master/src/lib/bind/effects.html#L123, you exit like:
```
        if (args.length > 1 && v === undefined) {
          return;
        }
```
where I assumed that we would escape at the first undefined argument? 
Therefore changing to 
```
        if (args.length > 0 && v === undefined) {
          return;
        }
```
solve this issue but might introduce other unwanted effect?
@azakus WDYT?
(I tried running `wct` on master and with or w/o the modification, I ended up with 357 tests failed on Chrome... oO)

MeTaNoV:
[PR #3423](https://github.com/Polymer/polymer/pull/3423) would solve the issue.

TimvdLippe:
Observers run with `undefined` in Polymer 2 as described in [the upgrade guide](https://www.polymer-project.org/2.0/docs/upgrade#update-observers), therefore I am closing this issue.

Issue #3239, opened 2015-12-28
keySplices is never set in array observers (basic snippet in official docs fails too)


mercmobily:
I am trying to figure out how array observers work -- unsuccessfully.
I was writing up some examples, and just couldn't get `changeRecord` to have meaningful values both at init time and when the array changed. In particular, `keySplices` seems to be always `null`. 

In the end, just to make sure, I pasted the exact example [in the guide](https://www.polymer-project.org/1.0/docs/devguide/properties.html#array-observation), and I am getting:

```
Uncaught TypeError: Cannot read property 'indexSplices' of undefined 
```

By digging into the code, I noticed that at init time `this.__data__` only have one key, `users`. With subsequent array operations (for example this.push( 'users', 'merc'), `this.__data__` has both `users` and `splices` as keys, but users.splices.keySplices is still `null`.

If the basic array observer in the docs doesn't work, then I get the feeling that maybe something bad has happened to the array observer...

arthurevans:
Hi @mercmobily. Do you have a jsbin or something that reproduces what you're seeing?

I tried plugging the doc sample into jsbin, and I **am** seeing that the _initial_ call to the observer has an undefined change record—which definitely seems like a bug. But subsequent ones appear to have both the `indexSplices` and `keySplices` properties:

http://jsbin.com/qalema/2/edit?html,console,output

arthurevans:
(Tagging this a bug because at a minimum, the undefined-initial-change-record seems like, well, a bug.)

kaste:
`indexSplices` and `keySplices` get indeed `null` as soon as you use `async` (or debounce). T.i. the changeRecord gets mutated. Ouch.

http://jsbin.com/tikosikomu/edit?html,console,output

mercmobily:
Please accept my apologies for not being as proactive as I should about this one. I just got swamped by work unexpectedly... do you guys want me to look into it more? I saw from the documentation that you are also rewriting the Array section... I suspected that maybe I was using it wrong?

arthurevans:
@kevinpschaaf the issue @kaste is seeing is a little puzzling to the user. It appears to be on purpose, to avoid keeping around (potentially large) change records:

https://github.com/Polymer/polymer/blob/4d45842a40064d509ec97ac897c2f4fb8782ffdc/src/standard/notify-path.html#L448

Should we document this as a limitation? Send the user a shallow copy?

[...]

Issue #2470, opened 2015-09-23
Array splicing and dom-repeat lead to wrong path in deep property observers


sawatsky:
If you have an array that is modified using `this.splice`, `this.push`, and `this.set` from the Polymer library and have a `dom-repeat` iterating over the array, and then do these steps:

1. Add some elements
2. Remove some elements
3. Change a property of any element physically AFTER the elements you removed

You will see that a deep observer of the array is given an incorrect path on the property change.

I believe this has to do with Polymer's use of `delete this.store[key]` instead of, say, `this.store.splice(key, 1)` in the removeKey function. The WeakMap isn't updated, and so future updates to the array are related back to mappings that no longer exist. Although I am not sure.

Example:
```javascript
<dom-module id="test-element">
	<template>
		<template is="dom-repeat" items="{{tests}}">
			<!--Nothing needs to happen in here-->
		</template>
	</template>
			
	<script>
		Polymer({
			is: "test-element",
					
			properties: {
				"tests": {
					type: Array,
					notify: true,
					value: []
				}
			},
					
			observers: [
				"_testsChanged(tests.*)"
			],
					
			"_testsChanged": function(changed) {
				// With the dom-repeat, the wrong path is reported on subproperty change
				// Without dom-repeat, it's fine
				console.log("CHANGED:", changed, "PATH:", changed.path);
			},
					
			"add": function(value) {
				this.push("tests", { "value": value });
			},

			"remove": function(index) {
				this.splice("tests", index, 1);
			},

			"change": function(index, value) {
				this.set("tests."+index+".value", value);
			},
				
			ready: function() {
				this.runTest();
			},
				
			"runTest": function() {
				this.add("a"); // [{ "value": "a" }]
				this.add("b"); // [{ "value": "a" }, { "value": "b" }]
				this.add("c"); // [{ "value": "a" }, { "value": "b" }. { "value": "c" }]
					
				this.remove(0); // [{ "value": "b" }. { "value": "c" }]
				this.remove(0); // [{ "value": "c" }]
					
				this.change(0, "changed"); // [{ "value": "changed" }]
						
				// Expected path: tests.0.value
				// Actual path: tests.2.value
			}
		});
	</script>
</dom-module>
```

Notice the order of operations. Adding some elements, removing any but the last, then changing the property of the last.

azakus:
The item's path in the collection is not the index of the in the array, but it is actually statically assigned when the item is added.

[...]

PR #3423, opened 2016-02-15
Enforce undefined rule for observers

kaste:
Enforce (again) the rule that observers are never called with undefined values. Basically reverts f68c30e21eec7d1fae04dc032c81df497444099e.
- Ensures observers on `*.splices` don't get undefined's initially. See #3239, https://github.com/Polymer/polymer/issues/2470#issuecomment-174199175
- Reduces confusion, e.g. #3401  
- Marks an invariant (dom-repeat sees unexpectedly undefineds) where it belongs, instead of masking it. Removes a confusing dependency between observers `marshalArgs` and `dom-repeat`.

### DIFF
--- a/src/lib/bind/effects.html
+++ b/src/lib/bind/effects.html
@@ -121,9 +121,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _marshalArgs: function(model, effect, path, value) {
       var values = [];
       var args = effect.args;
-      // Actually we should return early as soon as we see an `undefined`,
-      // but dom-repeat relies on this behavior.
-      var bailoutEarly = (args.length > 1 || effect.dynamicFn);
       for (var i=0, l=args.length; i<l; i++) {
         var arg = args[i];
         var name = arg.name;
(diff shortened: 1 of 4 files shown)


azakus:
At first glance, this looks like a breaking change to remove the ability to set `undefined`. At the point we'd be comfortable making a breaking change like this, we will probably reevaluate all of the databinding system.

samccone:
while I like this change, this is going to be wildly breaking I fear

MeTaNoV:
well, that's what is specified... :)

kaste:
@azakus This has nothing to do with removing the ability to set `undefined`. The (simple) property observers are completely unaffected by this change. Intentionally setting a bound value to `undefined` (sic) is jeopardy within Polymer. This behavior is not affected by this PR.

@samccone Sure this will break wildly. The iron-elements people often use the (internally named complex) observer with one dependency instead of the property observer. But from looking at their code it is not even clear they consciously made a decision to opt-in for undefined's. It's more likely they just want to see `null`s or generally falsey values. Maybe they even waste time, doing computations with undefined's.

@MeTaNoV I did not get it down to the spec. Actually just the *first* evaluation of such an computational expression should be delayed until all its dependencies are initialized. But I'm not convinced that this is an actually good idea.

Let me state, that I encountered all three issues/bugs this PR solves in real life. So, this is not a bare hygienic change. **I actually had these issues** using Polymer.

mercmobily:
FWIW, I also encountered _all_ of these problems. Not in real life... But while studying polymer and going through every single sentence in the documentation (with the aim to commit everything to memory).

I am the last kid in the block here and I most certainly don't know the implications for existing code, but I would LOVE to see all of those issued fixed -- if this PR does it and if it's barely acceptable...

kaste:
A puzzle: Write a test that shows how you can get a `undefined` from Polymer in a so-called 'complex' observer with only one dependency. 

- array.splices are out b/c it's a reported bug. 
- Just assigning `undefined` is obviously not allowed as well.

kaste:
By solving my own puzzle I now have a different and actually correct understanding of the problem `dom-repeat` tries to solve. I'll open another PR about undefined's the user wants to handle. Together with my #3439 there will then be 3 PR's around evaluation of complex expressions.

kaste:
The follow-up PR is #3452

kaste:
I'm not convinced anymore that this is a good change. Feel free to close it.

TimvdLippe:
Observers run with `undefined` in Polymer 2 as described in [the upgrade guide](https://www.polymer-project.org/2.0/docs/upgrade#update-observers), therefore I am closing this PR.

Issue #2350, opened 2015-08-23
_notifySplice adds an attribute named 'splices' to arrays 


david-saslawsky:
The following code
````
    Polymer({
        is: 'x-qwe',
        
        properties: {
            data: {
                type: Array,
                value: function() {
                    return [];
                }
            }
        },
        
        attached: function() {
            this.splice('data', 1, 0, { name: 'TEST' });
            console.log(JSON.stringify(this.data.splices));
        }
    });
````

returns `{"keySplices":null,"indexSplices":null}`

Looks like the bug is in `this.set(path + '.splices', change)`:

````
      set: function(path, value, root) {
        var prop = root || this;
        var parts = this._getPathParts(path);
        var array;
        var last = parts[parts.length-1];
        if (parts.length > 1) {
          for (var i=0; i<parts.length-1; i++) {
            var part = parts[i];
            prop = prop[part];
            if (array && (parseInt(part) == part)) {
              parts[i] = Polymer.Collection.get(array).getKey(prop);
            }
            if (!prop) {
              return;
            }
            array = Array.isArray(prop) ? prop : null;
          }
          if (array && (parseInt(last) == last)) {
            var coll = Polymer.Collection.get(array);
            var old = prop[last];
            var key = coll.getKey(old);
            parts[i] = key;
            coll.setItem(key, value);
          }
          prop[last] = value; // <<< HERE
          if (!root) {
            this.notifyPath(parts.join('.'), value);
          }
        } else {
          prop[path] = value;
        }
      },
````

JanaFlauschata:
I'm observing something similar when changing an array via:

    this.push('array', item);

When observing `array.splices` I get a change record with correct `base` and `path` properties, but `keySplices` and `arraySplices` are null. 

Polymer is up to date. Any help is appreciated.

mercmobily:
+1

chen-ye:
Looks like this was fixed in https://github.com/Polymer/polymer/pull/3439, and i can no longer repro in in v1.latest:

https://jsbin.com/voritiduse/1/edit?html,output

PR #3439, opened 2016-02-18
Refactorings around how computational expressions get their arguments

### LINKED ISSUE #2415: Empty `array.splices` object is left on array, which can cause observers to run with null splices
Current implementation sets the `array.splices` change record directly on the array (rather than just passing it in `notifyPath`), because it needs to be `get`able from the `"array.splices"` path when assembling arguments for observers.

This results in

* The possibility of an observer on `array.splices` called with null splice info when a previously mutated array is set to a new property with such an observer
* `for (prop in array)` to return `splices`, which may be unexpected from the user's point of view

### LINKED ISSUE #3179: notifyPath's value doesn't do anything
It seems that calling `notifyPath` will update any observer with the value that exists in the path, not the value given to `notifyPath`. If this is the intended behavior, is there another purpose of the value argument in `notifyPath`?

See [example codepen](http://codepen.io/bedeoverend/pen/ZQYdwQ?editors=101)

kaste:
- Do not set splices property on userland arrays. Fixes #2415, #2350
- Use value sent through the notification system (#3179)
- Don't mutate splices sent into userland. As reported #3239 

Should fix #3179

### DIFF
--- a/src/lib/bind/effects.html
+++ b/src/lib/bind/effects.html
@@ -130,6 +130,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var v;
         if (arg.literal) {
           v = arg.value;
+        } else if (path === name) {
+          v = value;
         } else {
           // Take advantage of the fact that all values sent through the path
           // notifications system are cached on the model. Trivially, all the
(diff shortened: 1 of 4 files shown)


kaste:
I've added three more commits on top of that initial commit. It seemed appropriate to not open a PR for each of them. Anyhow, please review the commits individually. Overall it's a concise change, @azakus give it a P1 ;-)

devinivy:
This is a nice change :+1:

azakus:
LGTM from @kevinpschaaf

PR #3440, opened 2016-02-18
Reduce domif performance overhead

### LINKED ISSUE #3073: Reduce `dom-if` performance overhead
Today there is more runtime overhead for using a `dom-if` than desirable.

The following possible low-hanging fruit have been identified:
- Default `false` values for `restamp` and `if` cause observers & debouncers to queue that ultimately no-op for falsey templates
- Templatizer handling for late-binding of outer parent (host) properties is costly and can be optimized to avoid use of actual getters
- Use of the Templatizer archetype seems like a non-value-add indirection since there is only ever one instance

kevinpschaaf:
### Reference Issue

Fixes #3073
### Description
- Default false values for restamp and if cause removed
- Templatizer handling for late-binding of outer parent (host) properties optimized

### DIFF
--- a/src/lib/template/dom-if.html
+++ b/src/lib/template/dom-if.html
@@ -45,7 +45,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        */
       'if': {
         type: Boolean,
-        value: false,
         observer: '_queueRender'
       },
 
@@ -58,7 +57,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        */
       restamp: {
         type: Boolean,
-        value: false,
         observer: '_queueRender'
       }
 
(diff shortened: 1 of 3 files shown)


src/lib/template/dom-if.html
```diff
@@ -58,7 +57,6 @@
        */
       restamp: {
         type: Boolean,
-        value: false,
```
kaste: I would leave the default of false here. Defining a complex observer (instead of using two property observers) should prevent the execution here until the if has been set. 
Since usually restamp is not set by the user, we would treat undefined as false throughout the code. It's just a small semantic shift.

TimvdLippe:
Closing this issue as the Polymer 2 version has incorporated these changes. @kevinpschaaf please reopen if I misunderstood this PR.

PR #3442, opened 2016-02-19
Restrict early property set to properties that have accessors. This a…

sorvell:
…llows users to set properties in `created` which are listed in `properties` but which have no accessor.


kevinpschaaf:
LGTM

PR #3444, opened 2016-02-19
Serialize before deserialize when configuring attrs. Fixes #3443.

### LINKED ISSUE #3443: Attributes bindings initially set to `false` are configured to `true`
A change was made to master to properly `deserialize` configured attribute bindings. However, the property value must be seralized before being deseralized. For example, `deserialize(false) === true` but `deserialize(serialize(false)) === false`. The latter is correct.

kevinpschaaf:
### Reference Issue
Fixes #3443


sorvell:
LGTM

PR #3450, opened 2016-02-21
Fix undefined class attribute in undefined template scope

### LINKED ISSUE #3364: <template> doesn't bind to class attribute in Shady DOM
Hi!

It seems that when a `<template>` is used without a parent scope, binding a value to `class` attribute doesn't work correctly in Shady DOM. Based on my testing, also the Shadow DOM polyfill works incorrectly. Native Shadow DOM on Chrome **works** perfectly.

As an example,

````html
<body>
  <template is="dom-repeat" items='["foo", "bar"]'>
    <div class$=[[item]] clazz$="[[item]]">[[item]]</div>
  </template>
</body>
````

actually renders:

````html
<div clazz="foo">foo</div>
<div clazz="bar">bar</div>
````

While as:

````html
<body>
  <template is="dom-bind">
    <template is="dom-repeat" items='["foo", "bar"]'>
      <div class$=[[item]] clazz$="[[item]]">[[item]]</div>
    </template>
  </template>
</body>
```` 

renders correctly:

````html
<div class="foo" clazz="foo">foo</div>
<div class="bar" clazz="bar">bar</div>
````

Similarly, any component using templates like `iron-list` fails to render the `class` attribute correctly if it not nested under another component or `dom-bind`:

````html
<body>
  <iron-list items='["foo", "bar"]'>
    <template>
      <div class$=[[item]] clazz$="[[item]]">[[item]]</div>
    </template>
  </iron-list>
</body>
````

I've created a simple jsbin here: http://jsbin.com/mayexiditu/edit?html,output

Tested on Chrome 47, Polymer v1.2.4. (and also v1.0.0, v1.1.0, v1.2.0)

I can manage using a `dom-bind` as a workaround, but any help would be highly appreciated!

p.s. might be related with #2434.

TimvdLippe:
### Reference Issue
Fixes #3364

### Description
If the `dom-repeat` has no template host, `this._rootDataHost` is undefined. Therefore it will not invoke `this._rootDataHost._scopeElementClass`. As fallback, return the plain value received, as Polymer should not perform any scoped styling.

### DIFF
--- a/src/lib/template/templatizer.html
+++ b/src/lib/template/templatizer.html
@@ -395,6 +395,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (host) {
         return host._scopeElementClass(node, value);
       }
+      return value;
     },
 
     /**
--- a/test/unit/bind.html
+++ b/test/unit/bind.html
@@ -16,6 +16,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   <link rel="import" href="../../polymer.html">
   <link rel="import" href="bind-elements.html">
 <body>
+  <template id="class-repeat" is="dom-repeat" items='["class1", "class2"]'>
+    <div class$=[[item]] clazz$="[[item]]">[[item]]</div>
+  </template>
 <script>
 
 suite('single-element binding effects', function() {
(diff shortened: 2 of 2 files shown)


TimvdLippe:
Rebased on current master. PTAL @azakus

azakus:
LGTM, thanks!

PR #3452, opened 2016-02-22
Loosen undefined rule for computational expressions

### LINKED ISSUE #2158: Computed binding not executed if argument is undefined
I've noticed two scenarios in which computed bindings where not executed.
1. Binding to an undefined property:
   
   ``` html
   <span>{{ _computedBinding(blah) }}</span>
   ```
   
   For this is does not matter if `blah` is a declared property or not.
2. Binding to an undefined subproperty if the compute function has more than one argument:
   
   ``` html
   <span>{{ _computedBinding(myString, myObj.blah) }}</span>
   ```
   
   Strangely enough, binding to an undefined property works if it is the only argument:
   
   ``` html
   <span>{{ _computedBinding(myObj.blah) }}</span>
   ```

kaste:
- Let developers opt-in to see undefined's in observers et.al.
- Let computed properties and annotations evaluate to undefined if their dependencies cannot be fulfilled

Can be seen as an add-on to #3423, could be applied individually as well. All in all non breaking change. 

Ref issue #3453, fixes #2158.

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -90,7 +90,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     _bindingRegex: (function() {
-      var IDENT  = '(?:' + '[a-zA-Z_$][\\w.:$\\-*]*' + ')';
+      var IDENT  = '(?:' + '[?]?[a-zA-Z_$][\\w.:$\\-*]*' + ')';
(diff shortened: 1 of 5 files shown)


kaste:
I'm not convinced anymore that this is a good change. I actually doubt it is. Feel free to close it.

TimvdLippe:
Observers run with `undefined` in Polymer 2 as described in [the upgrade guide](https://www.polymer-project.org/2.0/docs/upgrade#update-observers), therefore I am closing this PR. Thanks for your suggestions, they helped defining the Polymer 2 API :smile:

PR #3455, opened 2016-02-23
Expose bare minimum to add effects dynamically.

kaste:
Adds methods to add and remove property effects. Works even on yet uninitialized properties.


kaste:
Dammit I can't reopen.

arthurevans:
Can't reopen on this side either, because the branch was force-pushed or recreated. I think you may need to open a new PR?

kaste:
I had to fix it and thus force-pushed, anyway it's now called #3460. :cake:

Issue #3456, opened 2016-02-23
Expose bare minimum to add property effects dynamically. 


kaste:
As implemented #3460. 

Following API:

```
  Polymer({
    is: 'x-custom-effect'
  });

  var el = document.createElement('x-custom-effect');

  var fx = el.addPropertyEffect('foo', function(path, value, old) {
    // path will be 'foo' or e.g. 'foo.splices' etc.
    // old value will only be avail if path is simple e.g. 'foo'
  });
  // do something
  // then later ... unregister
  el.removePropertyEffect(fx);
```

azakus:
Can you describe what the use cases are for custom property effects with some examples?

kaste:
Well I linked to numerous issues in the PR: #3420, #1796, #2131, #2638, #1778
Note: The PR just _exposes_ the already implemented `functionEffect`. The only missing part is that Polymer does not define the equivalent path effect. 

In a nutshell, this is just the straightforward code for what Polymer calls an `annotationEffect`:  
```
// <x-foo prop="[[ bar ]]"></x-foo>
var el = document.createElement('x-foo');
this.addCustomEffect('bar', function(path, value, oldValue) {
   el.prop = value;
});
```

And since a function effect is pretty low-level, you could rebuild all the effect kinds with it. 

The problem with this PR is, adding elements gets damn easy, but removing them from the document requires that you also unregister all the custom effects, so you still have to keep track of that state. 

The problem Polymer has, is that it somewhat fails in highly _polymorphic situations_: If you can't tell which elements/tags you will need to insert upfront, or if you just don't know which properties for example to bind to etc. Polymer also often forces you to use dom-repeats's or dom-if's although they don't perform that well and are highly opinionated in what they do and how they work and don't work.

kaste:
I think this is a good use-case: #3547.

TimvdLippe:
This has been shipped in Polymer 2: https://www.polymer-project.org/2.0/docs/api/elements/Polymer.Element#method-addPropertyEffect Thanks for the suggestion!

PR #3460, opened 2016-02-23
Expose bare minimum to add effects dynamically.

### LINKED ISSUE #1778: [1.0] Data-binding: Is there any way to do this imperatively?
Hello, considering the below snippet,

```js
  var el = document.createElement("foo-element");
  el.setAttribute("foo", "[[foo]]");
  Polymer.dom(tpl.$.insertion_point).appendChild(el);
```

Full jsbin: (http://jsbin.com/hiloloqoju/1/edit)

In 0.5, I could simply imperatively import a Polymer element, then insert its markup into the DOM. Could I do something like that in 1.0?

Thanks so much,
J

### LINKED ISSUE #1796: Data binding with JavaScript no longer available in 1.0
In 0.5 there was a handy built-in function `this.bindProperty()` as well as `this.bind()` It appears that equivalent functions no longer exist in 1.0. Perhaps this functionality is provided in undocumented `Polymer.Bind` but without documentation it is hard to tell.

### LINKED ISSUE #2131: Polymer.create() - binding issues
I have a fairly large single page app, and consequently only render elements to the DOM if the user navigates to them.

I'm using `Polymer.create()` to achieve this, however I'm having issues setting up bindings using this method. [Here is a jsbin demonstrating the approach and issue](http://jsbin.com/xunone/edit?html,console,output). It gives the following console output:

```
runner-3.34.1.min.js:1 test-elem Ready!
runner-3.34.1.min.js:1 list-elem Ready!
runner-3.34.1.min.js:1 [dom-repeat::dom-repeat]: expected array for `items`, found {{otherThings}}e.(anonymous function) @ runner-3.34.1.min.js:1Polymer._itemsChanged @ polymer.html:3352Polymer.Base.extend._complexObserverEffect @ polymer.html:1282(anonymous function) @ polymer.html:1131Polymer.Bind._modelApi._effectEffects @ polymer.html:1128Polymer.Bind._modelApi._propertySet @ polymer.html:1122setter @ polymer.html:1189Polymer.Base._addFeature._applyEffectValue @ polymer.html:1531Polymer.Base.extend._annotationEffect @ polymer.html:1257(anonymous function) @ polymer.html:1131Polymer.Bind._modelApi._effectEffects @ polymer.html:1128Polymer.Bind._modelApi._propertySet @ polymer.html:1122setter @ polymer.html:1189Polymer.Base._addFeature.create @ polymer.html:1086Polymer.ready @ runner:12Polymer.Base._addFeature._invokeBehavior @ polymer-micro.html:275Polymer.Base._addFeature._doBehavior @ polymer-micro.html:270Polymer.Base._addFeature._readySelf @ polymer-mini.html:76Polymer.Base._addFeature._ready @ polymer-mini.html:63Polymer.Base._addFeature._tryReady @ polymer-mini.html:52Polymer.Base._addFeature._initFeatures @ polymer.html:2906Polymer.Base.createdCallback @ polymer-micro.html:106
runner-3.34.1.min.js:1 error TypeError: Cannot read property 'getKey' of undefined
    at dom-repeat.Polymer._sortAndFilter (http://milestech.net/components/polymer/polymer.html:3434:25)
    at dom-repeat.Polymer._render (http://milestech.net/components/polymer/polymer.html:3402:6)
    at dom-repeat.Polymer.Templatizer._flushTemplates (http://milestech.net/components/polymer/polymer.html:3016:8)
    at Object.Polymer.Debounce.Debouncer.complete (http://milestech.net/components/polymer/polymer.html:960:15)
    at MutationObserver.atEndOfMicrotask (http://milestech.net/components/polymer/polymer.html:924:1)
polymer.html:3434 Uncaught TypeError: Cannot read property 'getKey' of undefinedPolymer._sortAndFilter @ polymer.html:3434Polymer._render @ polymer.html:3402Polymer.Templatizer._flushTemplates @ polymer.html:3016Polymer.Debounce.Debouncer.complete @ polymer.html:960atEndOfMicrotask @ polymer.html:924
```

The meat of the problem is with: `this.create('list-elem', {things: '{{otherThings}}'});`, where `otherThings` is a property of the parent element.

At the moment the only workaround I have is to set up observers in the parent element and explicitly set the values in the child elements, which is quite clunky.

Am I missing something, or is it only possible to bind correctly by rendering everything into the DOM at the same time? If that is the case, is it possible to add support for this in Polymer.create()?

### LINKED ISSUE #2638: set and bind doesn't work from outside
If you try to set a value from outside, it doesn't notify the changes. Also a factoryImpl-Constructor, which binds data by caling set, doesn't bind so that the element notifies it.

``` javascript
<dom-module id="name-tag">

  <template>
    <section on-click="adduser">Click to add a user</section>
    <section id="main">
    </section>
    <ul>
      <template is="dom-repeat" items="{{user}}">
        <li>{{item.name}}</li>
      </template>
    </ul>
  </template>

  <script>
    Polymer({
      is: "name-tag",
      properties: {
        user: {
          type: Array,
          value: function() { return [] },
          notify: true
        }
      },
      ready: function() {
        var dynamicEl = new ChildElement(this.user);
        //var dynamicEl = document.createElement('child-element');
        //this.set('user',this.user,dynamicEl);
        //dynamicEl.notifyPath('user',this.user);
        this.$.main.appendChild(dynamicEl);

        // set some users
        this.push('user', { name: "Shakh", surname: "Samat", age: 15 });
        this.push('user', { name: "Jambyl", surname: "Samat", age: 17 });
        this.push('user', { name: "Aiganym", surname: "Samat", age: 18 });
      },
      adduser: function() {
        // add a user
        this.push('user', { name: "William", surname: "Sulo", age: 19 });
        console.log("added");
      }
    });
  </script>
</dom-module>

<dom-module id="child-element">
  <template>
    <div>
      <ul>
        <template is="dom-repeat" items="{{user}}">
          <li>{{item.name}}</li>
        </template>
      </ul>
    </div>
  </template>
  <script>
    ChildElement = Polymer({
      is: 'child-element',
      factoryImpl: function(users) {
        this.set("user", users);
        //this.notifyPath("user",this.user);
      },
      properties: {
        user: {
          type: Array,
          value: function() { return []; },
          notify: true
        }
      }
    });
  </script>
</dom-module>
```

Link: http://plnkr.co/edit/Ybv7m1ecwzcwSjGXKDpT?p=preview

### LINKED ISSUE #3420: Data binding using Javascript
Hi guys!!! 
I'm trying to create an instance of _service-detail_ element with javascript inside _service-section_ element. 
I've already imported _service-detail_  in _elements.html_. 
For example: 
In **service-section** element: 
```javascript
showDetails: function(e) {
   var detail = document.createElement('service-detail');
   detail.lang = this.lang;
   detail.services = this.services;
}
```
The problem is that if the property **lang** in _service-section_ changes, this change is not propagated to the _service-detail_ element. 

On the other hand, if I use inside the _service-section_ element the tag:
```html
<service-detail lang="{{lang}}" services="{{services}}"></service-detail>
```
When I change the value of the **lang** in _service-section_ it also change in the _service-detail_ . (As it's suppose to work). 

Now the question is, How can I do to bind properties using Javascript and creating the elements dynamically?

Thank you everyone for do the best!

kaste:
Adds methods to add and remove property effects. Basically:

```
  var fx = this.addPropertyEffect(propertyName, function(path, value, old) {});
  /// later
  this.removePropertyEffect(fx);
```

Reference issue #3456 

Fixes #3420, fixes #1796, fixes #2131, fixes #2638, fixes #1778.

### DIFF
--- a/src/standard/notify-path.html
+++ b/src/standard/notify-path.html
@@ -274,10 +274,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var fx$ = this._propertyEffects && this._propertyEffects[model];
         if (fx$) {
           for (var i=0, fx; (i<fx$.length) && (fx=fx$[i]); i++) {
-            // use memoized path functions
-            var fxFn = fx.pathFn;
-            if (fxFn) {
-              fxFn.call(this, path, value, fx.effect);
+            if (!fx.disabled) {
+              // use memoized path functions
+              var fxFn = fx.pathFn;
+              if (fxFn) {
+                fxFn.call(this, path, value, fx.effect);
+              }
             }
           }
         }
@@ -318,6 +320,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         if (this._pathMatchesEffect(path, effect)) {
           Polymer.Bind._annotatedComputationEffect.call(this, path, value, effect);
         }
+      },
+
+      _functionPathEffect: function(path, value, effect) {
+        Polymer.Bind._functionEffect.call(this, path, value, effect);
       },
 
       _pathMatchesEffect: function(path, effect) {
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -79,7 +79,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _effectEffects: function(property, value, effects, old, fromAbove) {
         for (var i=0, l=effects.length, fx; (i<l) && (fx=effects[i]); i++) {
-          fx.fn.call(this, property, value, fx.effect, old, fromAbove);
+          if (!fx.disabled) {
+            fx.fn.call(this, property, value, fx.effect, old, fromAbove);
+          }
         }
       },
 
@@ -111,7 +113,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var propEffect = {
         kind: kind,
         effect: effect,
-        fn: Polymer.Bind['_' + kind + 'Effect']
+        fn: Polymer.Bind['_' + kind + 'Effect'],
+        trigger: property
       };
       fx.push(propEffect);
       return propEffect;
--- a/src/lib/template/templatizer.html
+++ b/src/lib/template/templatizer.html
@@ -282,12 +282,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _createHostPropEffector: function(prop) {
       var prefix = this._parentPropPrefix;
       return function(source, value) {
+        if (source !== prop) return;
         this.dataHost._templatized[prefix + prop] = value;
       };
     },
 
     _createInstancePropEffector: function(prop) {
       return function(source, value, old, fromAbove) {
+        if (source !== prop) return;
         if (!fromAbove) {
           this.dataHost._forwardInstanceProp(this, prop, value);
         }
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -35,6 +35,71 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var prop = Polymer.Bind.addPropertyEffect(this, property, kind, effect);
       // memoize path function for faster lookup.
       prop.pathFn = this['_' + prop.kind + 'PathEffect'];
+      return prop;
+    },
+
+    _ensureOwnEffects: function(property) {
+      // Move the property effects from the prototype to the instance, so
+      // we can mutate them. We use prototypal inheritance here to go cheap.
+      var pE = this._propertyEffects;
+      if (pE) {
+        if (!this.hasOwnProperty('_propertyEffects')) {
+          pE = this._propertyEffects = Polymer.Base.chainObject({}, pE);
+        }
+
+        // Shallow-copy the effects array for this property, to the instance.
+        if (pE[property] && !pE.hasOwnProperty(property)) {
+          pE[property] = pE[property].slice();
+        }
+      }
+    },
+
+    addPropertyEffect: function(property, fn) {
+      this._ensureOwnEffects(property);
+
+      // Because the property descriptor encloses the effects, we recreate
+      // it. Just for the record: In case this would be the first effect of
+      // the property, we had to create the underlying trigger machinery
+      // anyway.
+      var effects = Polymer.Bind.ensurePropertyEffects(this, property);
+      if (!effects._installed) {
+        // The current value will be masked by the descriptor, read it ...
+        var val = this[property];
+        Polymer.Bind._createAccessors(this, property, effects);
+        //                                   closure  ^^^^^^^  alert
+        effects._installed = true;
+        // ... and put it in our data store
+        this.__data__[property] = val;
+      }
+
+      // Theoretically the effects must be sorted again, but `function`-effects
+      // are executed last anyway, so we can skip this for now. NOTE that, if
+      // you add a custom effect inside another effect, t.i. the notification-
+      // loop is already running, the newly added effect will not be executed
+      // before the next notification we have to propagate. This is intended
+      // and a bonus point for adding all custom effects at the end.
+      return this._addPropertyEffect(property, 'function', fn);
+    },
+
+    removePropertyEffect: function(fx) {
+      // Since effects can be removed as a side-effect, t.i. within a running
+      // effects-loop, we must ensure not to change the effects- array
+      // immediately, otherwise the for-loop would break. On the other side
+      // removing an effect must be immediate too. So we use a flag here, and
+      // actually remove the effect in a microtask.
+      fx.disabled = true;
+
+      this.async(function() {
+        var property = fx.trigger;
+        var effects = property && this._propertyEffects &&
+                      this._propertyEffects[property];
+        if (effects) {
+          var index = effects.indexOf(fx);
+          if (index !== -1) {
+            effects.splice(index, 1);
+          }
+        }
+      });
     },
 
     // prototyping
--- a/test/unit/bind-elements.html
+++ b/test/unit/bind-elements.html
@@ -713,3 +713,7 @@
     });
   </script>
 </dom-module>
+
+<script>
+  Polymer({is: 'x-custom-effect'});
+</script>
--- a/test/unit/bind.html
+++ b/test/unit/bind.html
@@ -1023,6 +1023,149 @@ suite('order of effects', function() {
   });
 });
 
+suite('custom user effects', function() {
+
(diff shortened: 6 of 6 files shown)


kaste:
Fixes #3420 
Refs #1778

samccone:
Hey @kaste this is super great and very exciting -- I for one have wanted this feature for quite a while, it opens up the door for all kinds of dynamic element creation while being able to take advantage of the built in data-binding behaviors of polymer! Thanks for the PR

TimvdLippe:
All issues that are fixed by this PR must be included in the PR description. Also they should all be prepended with Fixes. For more information see https://github.com/PolymerElements/ContributionGuide/blob/master/CONTRIBUTING.md#submitting-pull-requests

kaste:
@TimvdLippe Updated the decription.

fooloomanzii:
I have tried the pull request with Array splices and can say it worked.

```
        var dynamicEl = document.createElement('child-element');

        this.$.main.appendChild(dynamicEl);

        this.addCustomEffect('user', function(path, value, oldValue) {
          // console.log(path, value, oldValue);
          if (value && value.indexSplices) {
            var index = value.indexSplices[0].index;
            var addedCount = value.indexSplices[0].addedCount;
            for (var i=index; i < index+addedCount; i++)
              dynamicEl.push("user", value.indexSplices[0].object[i]);
          }
        });
```

Appart from your nice, working solution, the array observation is just so unhandy, that you can loose joy and spoil the delight for yourself and there could be a nicer implementation for array-collections.

kaste:
@fooloomanzii Isn't this what you want: :question: 

        this.addCustomEffect('user', function(path, value, oldValue) {
          // console.log(path, value, oldValue);
          if (path === 'user.splices') {
            dynamicEl.notifySplices('user', value.indexSplices);
          }
        });

fooloomanzii:
@kaste I wasn't aware of `notifySplices`. Thank you, that makes a lot easier. Looking forward for your feature to be pulled

devinivy:
:+1: This is an exciting PR.

kevinpschaaf:
The general concept for this feature is good.  We are doing explorations around a re-layering of Polymer that would expose the underlying machinery of the data system to end-users to enable better customization, and being able to add property effects to instances is definitely something we plan to include in that, so we're on the same wavelength here.

As far as adding this feature to the 1.0 codebase, the `addCustomEffect` / `removeCustomEffect` API is basically ok.  However, I'd probably try to make these changes:
* To avoid the need for code to re-create the accessor on the instance, I would try avoiding closing over the effects in the setter (in `createAccessors`) and instead reference them from the instance to simplify the code; this would need to involve performance testing that change as it is in a data hot-path, but the extra complexity of the current approach may not pay for any performance benefit
* In `removeCustomEffect`, rather than going async (which we always try to avoid), I would instead copy the list & splice the effect (which is safe since `_effectEffects` closes over the original effects list); I'm not as much concerned about penalizing the remove operation with this cost, since if we were going to design the effects system to have good removal characteristics (which was not a goal in this design), we might consider changing everything to use a linked list (could be an option in 2.0)
* Off the top of my head, I'm not keen on the `disabled` gambit; we consider depending on the order of effects running generally bad, and this code seems to imply that if a given custom effect is removed by an effect earlier in the list it _should not_ run, but if it is removed by an effect later in the list it _should_ run.  This has a bad code smell to me, although it is hard to justify without studying the problem more.

@kaste I would say that if you're willing to consider & adjust based on the feedback above to go for it and we can take another look.  As I said we are definitely planning to expose API for instance-time property effect management in the next major version where it will hopefully be a more fully-considered system, but that's still a little further down the road.

kaste:
Currently `_createAccessors` is used by the `Templatizer`. So we rely on prestored effects here. (This is the current version in master. Your newer version that does not create accessors 9a4a8b8bf24fd460bbbbf55725853c8e01be7dfc is not merged yet.)

The side effect of having prestored effects here is that I need the `async` call. B/c there is no other option than mutating the propertyEffects, I cannot reassign a sliced effects-array. So there is that. 

Regarding the `disabled` flag. I actually mimic the exact same behavior native DOM event listeners with their `add/removeEventListener` have. Only two rules here: Adding a new listener (effect) is not immediate, in the sense that this new handler will not be called before the next notification will be dispatched. _Removing_ a listener though MUST be immediate. 

The straightforward implementation for this is as follows::

```
    dispatchEffects() {
        var listeners = this.listeners.slice(); // !
        for (var listener of listeners) {
            if (listener.disabled) continue;
            listener.callback.call(...);
        }
    }

    addEffect(fx) {
        this.listeners.append(fx);
    }

    removeEffect(fx) {
        fx.disabled = true;
        this.listeners.remove(fx);
    }

```

I think I leave this PR as it is, and open a new one with a patched `createAccessors`.

kaste:
#3573 has some of the suggested changes. Implementing a linked list is not among them ;-)

kaste:
Renamed API to `add/removePropertyEffect`.

src/standard/notify-path.html
```diff
@@ -274,10 +274,12 @@
         var fx$ = this._propertyEffects && this._propertyEffects[model];
         if (fx$) {
           for (var i=0, fx; (i<fx$.length) && (fx=fx$[i]); i++) {
```
kaste: Note to self: Must prestore length just like in accessors!

btelles:
Anyone know when this PR is going to be merged and released? It really is a huge improvement. Thanks @kaste.

kaste:
Note: B/c Templatizer makes some fu with propertyEffects I had to fight https://github.com/Polymer/polymer/commit/f9d64e157e07f802fe10dba564012d14c6b25d48#diff-65d1593d7f79b51a64ac3cfc9465a1f4L330 to make instance property effects (this or #3573) work.

kaste:
@btelles The coupling between Templatizer and the Polymer.Bind system, and my veto on removing the disabled flag, -> they need to meet.

joryphillips:
I would greatly benefit from being able to do this! Any time estimates at this point? Thanks.

kaste:
Ping. 

This is now undecided for half a year. It adds value to the 1.x branch and is a non breaking change. You should prefer #3573 of course. On top that ‘linkPath‘ and ‘notifyPathUp‘ can be easily reimplemented as I have shown in other PRs, which will make Polymer faster for everyone.

joh1234:
Update README
https://goo.gl/JBljdR

TimvdLippe:
Sorry for the very late response. Based on this PR and the numerous community requests, this API actually shipped in Polymer 2. You can try it out yourself in http://jsbin.com/toholosuhu/edit?html,console,output For a full documentation of the API you check out the docs: https://www.polymer-project.org/2.0/docs/api/mixins/Polymer.PropertyEffects#method-addPropertyEffect

PR #3470, opened 2016-02-25
Fixes #3461: Only avoid creating a statically scoped stylesheet when …

### LINKED ISSUE #3461: Using var in a css mixin breaks the scoped style of the element
### Description
Using `var` in a css mixin rule causes the element to not define its scoped style. This works fine for Polymer 1.2.4. Did something change or is this no longer allowed?

#### Live Demo

http://jsbin.com/nulepeveku/edit?html,output

#### Steps to Reproduce
1. Create `my-element`
2. Append `my-element` to document.body
3. Define a css mixin containing a `var` on `my-element`

#### Expected Results
image
image

#### Actual Results
image
image

### Versions
- Polymer: v1.3.0
- webcomponents: v0.7.21

sorvell:
…properties are consumed in an element, properly excluding properties produced as a result of consumption.

Fixes #3461

### DIFF
--- a/polymer.html
+++ b/polymer.html
@@ -32,10 +32,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       this._prepConstructor();
       // template
       this._prepTemplate();
-      // styles
+      // styles and style properties
       this._prepStyles();
-      // style properties
-      this._prepStyleProperties();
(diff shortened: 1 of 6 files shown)


kevinpschaaf:
LGTM

PR #3472, opened 2016-02-25
Null debounced callback to set for GC.

kaste:
Please review, maybe I'm stupid. 

AFAIK the side effect 'garbage-collectable' cannot be tested, so there are no tests.

Reported here #3239


azakus:
LGTM

Issue #2606, opened 2015-10-22
Firing an x-changed event is not enough to update the binding


kaste:
Consider:

```
  <x-case value="{{foo}}"></x-case>
```

It is not enough that `<x-case>` fires `this.fire('value-changed', {value: 'new-value'}` to update the bound value/model `foo`. `<x-case>` also has to have a member `value` set accordingly. Respectively if `<x-case>.value === 'bar'` `foo` will be set to `bar` regardless of the `e.detail.value` we sent.

Is it Doc or Bug? At least the coupling between a parent and its child element is not as loose as the documentation suggests.

https://jsbin.com/legigi/edit?html,console

kaste:
The same holds true for the `notifyPath` API. If you listen to `value.*`. You can 

```
            this.value.hello = 'foo';
            this.notifyPath('value.hello', 'quux');
```
or even `this.notifyPath('value.*', 'quux');` whatever that should mean. Either way, a receiving parent elements binding (say `data.*`) gets set, so that `data.hello === 'foo'`, but a change-listener (AKA observer) gets the plain event with the path and value set to whatever you declared, even if that's not meaningful at all.

azakus:
I'm not 100% sure why that doesn't work, but I bet @kevinpschaaf does!

azakus:
FWIW, you can use `this.set('document', 'quux')` and it works, so there is *some* magic happening. Maybe in the event details itself.

kaste:
Actually @sorvell tried to solve this some time ago now (without leaving a comment here), but it didn't work out. Ref: https://github.com/Polymer/polymer/blob/273ab0fbe641209fa3fc3ce8368bb78f83db9b10/src/lib/bind/accessors.html#L222

PR #3473, opened 2016-02-25
Make notifyPath's value argument optional. 

### LINKED ISSUE #3179: notifyPath's value doesn't do anything

kaste:
Fixes #3179. Mentioned here #2606.

Make second and third argument of notifyPath optional. Deprecate them?

An optional `value` argument would be of course useful, as a hint, so that Polymer did not had to look it up. But we have the `_get` call anyway.

`notifyPath` does not return a Boolean, so I removed that comment as well.


azakus:
LGTM

PR #3480, opened 2016-02-26
Only test builds on travis

azakus:
- Reduces up test run length by ~20%
- `gulp switch` should fail fast if there are any issues building


azakus:
@kevinpschaaf @sorvell PTAL

.travis.yml
```diff
@@ -12,12 +12,14 @@ addons:
     - google-chrome-stable
     - g++-4.8
 before_script:
-- npm install -g bower
+- npm install -g bower gulp-cli
```
samccone: i have learned the hard way it is worth locking these installs via gulp-cli@<VERSION> :)

azakus:
RSLGTM from @kevinpschaaf

PR #3485, opened 2016-03-01
Ensure that dom-bind always waits until DOMContentLoaded to render.

### LINKED ISSUE #3484: dom-bind may render before entire document is loaded, including scripts that expect to install api on it
### Description
`dom-bind` may render before the entire document is loaded. This may result in a script which expects to install api on a `dom-bind` running after it renders, resulting in unexpected behavior. This occurs because `dom-bind` relies on RenderStatus.whenReady to decide when to render and this in turn is based on the first `reqestAnimationFrame`. To correct this issue `dom-bind` should make sure not to render while the `document.readyState` is `loading`.

#### Steps to Reproduce
Place a `dom-bind` in a document followed by a lot of HTML followed by a script that installs api on the `dom-bind`.

#### Expected Results
The api is installed on the `dom-bind` before its first render.

#### Actual Results
`dom-bind` may render prior to the script running, due to the parser yielding and rendering.

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions
- Polymer: v1.3.0
- webcomponents: do not load (characteristics change under HTMLImports polyfill

sorvell:
Fixes #3484.

This ensures a script can install api on the dom-bind prior to it rendering. Previously dom-bind waited for first render, but an early parser yield can make this occur unexpectedly early.

### DIFF
--- a/src/lib/template/dom-bind.html
+++ b/src/lib/template/dom-bind.html
@@ -69,7 +69,13 @@ elements to the template itself as the binding scope.
       // have resolved
       var self = this;
       Polymer.RenderStatus.whenReady(function() {
-        self._markImportsReady();
+        if (document.readyState == 'loading') {
+          document.addEventListener('DOMContentLoaded', function() {
+            self._markImportsReady(); 
+          });
+        } else {  
+          self._markImportsReady();
+        }
       });
     },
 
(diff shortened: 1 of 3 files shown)


kevinpschaaf:
LGTM

PR #3501, opened 2016-03-08
Do not reflect properties starting with an uppercase letter

### LINKED ISSUE #3500: reflected properties with leading upper case letter throw exceptions
#### Live Demo

http://jsbin.com/mocakesixa/edit?html,console,output

### Versions

- Polymer: v1.3.1

azakus:
Add tests for uppercase properties

Fixes #3500

### DIFF
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -65,9 +65,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
               event: Polymer.CaseMap.camelToDashCase(p) + '-changed'});
           }
           if (prop.reflectToAttribute) {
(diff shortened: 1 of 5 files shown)


azakus:
Rather than try to work around `setAttribute` not accepting leading '-' characters by trimming the leading character, which could lead to odd collisions if two or more elements shared similar property names, @kevinpschaaf suggested that we instead warn that properties starting with an uppercase character will not reflect to attributes.

Also handle weird attribute bindings like `-u-p-p-e-r-c-a-s-e$=` by warning.

TimvdLippe:
I like that suggestion a lot!

azakus:
@kevinpschaaf PTAL

kevinpschaaf:
LGTM

PR #3509, opened 2016-03-15
Perform element registration tasks at first instance creation rather than registration time

### LINKED ISSUE #3508: Register elements lazily at first instance time
When elements are registered, work is done to create the element's prototype (for example, its template is parsed, styles are shimmed, property accessors are created). If elements are registered but not used, for example, at first paint time, then registration work can delay first paint unnecessarily.

By default, Polymer should perform as many registration tasks as possible when the first instance of an element is created.

sorvell:
Fixes #3508.

Adds a new `lazyRegister: true` global Polymer setting that defers registration tasks such as style shimming, template parsing, and property accessor creation until first instance creation rather than registration time as an optimization.

Also added is an `ensureRegisterFinished()` API to element prototypes returned from `Polymer()`.  Calling this on a prototype when `lazyRegister` is enabled will force that type's registration work to be finished.


kevinpschaaf:
LGTM

PR #3519, opened 2016-03-18
Ensure Annotator recognizes dynamic fn as dependency for parent props.

### LINKED ISSUE #3515: function property binding not working inside dom-if
The binding of properties of type function does not seem to work as expected inside dom-if. The bound function is never called.

Please consider the following example:

```
<dom-module id="test-element">
  <template>
    <h3>Next paragraph will display 'text' (function property outside dom-if):</h3>
    <p>[[translate('text')]]</p>
    <template is="dom-if" if="[[visible]]">
      <h3>However, the next paragraph will be empty (function property inside dom-if):</h3>
      <p>[[translate('text')]]</p>
    </template>
  </template>
</dom-module>

<script>
  Polymer({
    is: 'test-element',

    properties: {
      translate: {
        type: Function,
        value: function () {
          return function(str) {
            return str;
          }
        }
      },
      visible: {
        type: Boolean,
        value: true
      }
    }
  });
</script>
```

By placing a breakpoint in the translate function, I found that it is called in the first paragraph but ignored in the second paragraph, inside the dom-if, which results in it being empty.

### Browsers Affected
- [x] Chrome
- [x] Firefox
- [ ] Edge (possibly)
- [x] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions
- Polymer: v1.3.1
- webcomponents: v0.7.21

kaste:
The annotator sort of 'copies' all dependencies of a computed annotation from a child template to its host (to populate _parentProps). We forgot to include the `method` as a dependency for 'dynamic'/late bound functions.
  
Fixes #3515

### DIFF
--- a/src/standard/annotations.html
+++ b/src/standard/annotations.html
@@ -195,6 +195,9 @@ TODO(sjmiles): this module should produce either syntactic metadata
                   pp[model] = true;
                 }
               }
+              if (p.signature.dynamicFn) {
+                pp[p.signature.method] = true;
+              }
             } else {
               if (p.model) {
                 pp[p.model] = true;
--- a/test/unit/bind-elements.html
+++ b/test/unit/bind-elements.html
@@ -685,3 +685,31 @@
     });
   </script>
 </dom-module>
+
+<dom-module id="x-child-template-with-dynamic-fn">
+  <template>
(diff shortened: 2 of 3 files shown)


azakus:
LGTM

PR #3526, opened 2016-03-22
Makes lazy registration compatible with platforms IE10

### LINKED ISSUE #3525: lazyRegister: not compatible with (IE10) browsers on which custom element prototype is simulated
The `lazyRegister` setting added in Polymer 1.4 depends on the ability to modify a registered prototype after registration. This does not work on IE10 where the custom element prototype is simulated via mixin.

sorvell:
Fixes #3525.


kevinpschaaf:
LGTM

PR #3529, opened 2016-03-23
Add an escape sequence to avoid polymer to bind the content of brackets

theboolean:
Refs https://github.com/Polymer/TemplateBinding/issues/186

This PR is a proposal to add the possibility to escape the brackets without binding the content.
If this kind of behaviour is acceptable, I would be glad to write also unit tests.

**Usage:**

`{{test}}`: normal behaviour
`{{\test}}`: don't bind to test and print literal `{{test}}`
`{{\\test}}`: don't bind to test and print literal `{{\test}}`
and so on... always remove one of the backslashes

`[[test]]`: normal behaviour
`[[\test]]`: don't bind to test and print literal `[[test]]`
`[[\\test]]`: don't bind to test and print literal `[[\test]]`
and so on... always remove one of the backslashes

### DIFF
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -788,7 +788,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
                                 '(?:' + ARGUMENTS + '?' + ')' +
                               '\\)\\s*' + ')';
   const BINDING = '(' + IDENT + '\\s*' + ARGUMENT_LIST + '?' + ')'; // Group 3
-  const OPEN_BRACKET = '(\\[\\[|{{)' + '\\s*';
+  const OPEN_BRACKET = '(\\[\\[\\\\*|{{\\\\*)' + '\\s*';
   const CLOSE_BRACKET = '(?:]]|}})';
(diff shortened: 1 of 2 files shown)


azakus:
This seems reasonable to me. Can you add a test case?

theboolean:
@azakus sorry for the huuuge delay, I finally wrote tests you asked. I also rebased the history on the current master branch, so the feature is available for Polymer 2.0 and the PR can be automatically merged.

TimvdLippe:
After a discussion about this PR, we decided to not include this functionality, mainly because of performance concerns. You can override [`_parseBindings`](https://www.polymer-project.org/2.0/docs/api/elements/Polymer.Element#method-_parseBindings) in your element to implement escaping in the spots you want this feature, which means that regular applications or other elements do not regress.

warpech:
Simple workaround that works in HTML without any overwriting of `_parseBindings`:

Whenever you want to write `{{something}}` on your page, and force Polymer to skip parsing that, put a HTML comment somewhere inside of it, e.g.

```html
{<!-- This comment is needed to prevent Polymer from parsing curly braces -->{something}}
```

PR #3532, opened 2016-03-24
Don't use `translate` as a method for testing

azakus:
`.translate` is some magical property in Safari (and Chrome)
Safari 7 puts this on the instance, and breaks tests for property
effects


kevinpschaaf:
LGTM

PR #3542, opened 2016-03-29
Expose public API to reset mouse cancelling for testing touch

### LINKED ISSUE #3540: Feature request: ability to clear "ignore mouse events" timer in Gestures system
### Description

When dealing with TouchEvents and MouseEvents, Polymer Gestures sometimes sets a timer for ~2.5 seconds during which it will ignore MouseEvents. I would like to propose that a public interface be exposed to reset this timer in order to better aide testing scenarios, where state needs to be reset before each test run.

azakus:
Remove undocument `event.__polymerSyntheticTouch` hack

Fixes #3540

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -231,10 +231,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             }
           }
           // disable synth mouse events, unless this event is itself simulated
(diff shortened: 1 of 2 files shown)


cdata:
LGTM :+1:

azakus:
RSLGTM from @kevinpschaaf

PR #3544, opened 2016-03-30
Set propertyName on parent props for config phase.

kaste:
Ensure `propertyName` is set by the Annotator even for parent properties. 

The side-effect/bug is that an element when it gets configured via simple `node._configValue(name, value)` the name can be otherwise `undefined`. See configure.html `_distributeConfig()` wherein we do `var name = x.effect.propertyName;`.


azakus:
LGTM

PR #3546, opened 2016-03-30
Compute and use correct annotation value during config

kaste:
Two problems fixed here. 

First with `!x.isCompound` we actually never excluded compound annotations from config - `!x.effect.isCompound` would have been correct. They just received the wrong value.

Otherwise we never distributed the correct value the annotation effect would have been propagated downwards if we would have called this effect anyway. E.g. negated bindings first received the plain value (of any type), and after clients were ready the negated (boolean).

By computing the correct value, compound bindings are safe for config. They benefit the most b/c if all their 'parts' can be resolved during config, only one side effect will ever run.


azakus:
LGTM

PR #3547, opened 2016-03-30
Implement linkPath as (custom) property effect.

kaste:
Based on top of #3460. Only https://github.com/Polymer/polymer/pull/3547/commits/644e290f65ccc37f7c4d52bba77037350c0c39f8 is of any interest here. 

If you open function effects to receive (deep) path notifications, you can build `linkPaths` on top of this. This implementation does not rely on dirty checking of values, and thus does not dead loop if we remove them.


TimvdLippe:
I am not sure what the relation of this PR is with the other PRs. I tried to digest and lookup for the equivalent in Polymer 2, but I assume that this was an implementation detail to achieve the other behavior? Given that Polymer 2 shipped the various APIs you suggested, I am closing this PR. Feel free to update the PR if you feel that your suggestion was not incorporated.

PR #3550, opened 2016-03-31
Upwards notifications should not echo downwards

kaste:
Pretty much what `fromAbove` is during downwards data flow, for upwards data propagation. 

Ensures that data that flows up (via DOM events) doesn't come back through the same binding (triggered by annotation effects). 

Should be read commit by commit b/c there is a lot of noise passing the target/origin from the event listener through the whole system down to the annotation effect. The essential code excerpt is: 

```
+
+    _annotationEffect: function(source, value, effect, old, fromAbove,
+                                origin, originalProperty) {
+
+      // Consider:
+      // <that-node that-property="{{ thisPath }}" />
+      var thatNode = this._nodes[effect.index];
+      var thatProperty = effect.name;
+      var thisPath = effect.value;
+
+      // Do not propagate downwards if the origin of this change is this exact
+      // binding
+      if (origin && originalProperty &&
+          origin === thatNode &&
+          originalProperty === thatProperty) {
+        return;
+      }
+
```

:cake:

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -223,7 +223,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _notedListenerFactory: function(property, path, isStructured, negated) {
       return function(target, value, targetPath) {
         if (targetPath) {
-          this._notifyPath(this._fixPath(path, property, targetPath), value);
+          var newPath = this._fixPath(path, property, targetPath);
+          this._notifyPath(newPath, value, false, target, property);
         } else {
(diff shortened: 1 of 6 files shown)


kaste:
Rebased.

TimvdLippe:
I think this issue is solved in Polymer 2 where a bag of properties is used as well as a guard with `lastRun` to ensure that dataflow is performend in waves. @kaste could you confirm that this is the case?

kaste:
Sorry, PR is too old to tell

TimvdLippe:
All right, I will close it then. Feel free to submit a PR for Polymer 2 if you still require changes and we can take another look :)

PR #3558, opened 2016-04-02
Support array path format for linkPaths() and unlinkPaths()

### LINKED ISSUE #3090: `linkPaths` doesn't accept arrays as paths
Most (if not all) functions seem to allow arrays instead of paths (which is absolutely wonderful, as it looks a million times better), however this seems not to be the case for `linkPaths`.

devinivy:
Fixes #3090

Now `linkPaths()` and `unlinkPaths()` support array-formatted paths in addition to string-formatted paths.  Docs and tests have been updated to reflect this.

### DIFF
--- a/src/standard/notify-path.html
+++ b/src/standard/notify-path.html
@@ -332,13 +332,21 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * are routed to the other.
        *
        * @method linkPaths
-       * @param {string} to Target path to link.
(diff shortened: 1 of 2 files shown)


TimvdLippe:
This was implemented in Polymer 2: https://www.polymer-project.org/2.0/docs/api/mixins/Polymer.PropertyEffects#method-linkPaths Sorry for not notifying you sooner :cry:

PR #3560, opened 2016-04-04
Implement initialization guard for computed expressions

### LINKED ISSUE #2158: Computed binding not executed if argument is undefined

### LINKED ISSUE #3239: keySplices is never set in array observers (basic snippet in official docs fails too)

### LINKED ISSUE #3401: Computed bindings don't run when an argument is undefined

### LINKED ISSUE #3453: Loosen undefined rule for computational expressions
As implemented in #3452.
- Let developers opt-in to see undefined's in observers et.al.
- Let computed properties and annotations evaluate to undefined if their dependencies cannot be fulfilled

Can be seen as an add-on to #3423, could be applied individually as well. All in all non breaking change. 

Ad 1: Since users sometimes want to see undefined's, provide a way to mark some of the dependencies of the computation as optional as follows `fn(?foo.bar)`. This usually comes in handy if `foo.bar` becomes unreachable, maybe simply because `foo` no longer has the member `bar`. This is exactly the case of `dom-repeat`'s invariant (See #3423 for that).

Note that there is no notion of a configuration phase for observers et.al. anymore (which Polymer didn't implement anyway). The arguments for an observer can either be fulfilled or the user opts-in and thus guards her code to see undefined's. 

Ad 2: Consider the following: 

```
  <div>[[foo.bar]]</div>
  <div>[[fn(foo.bar, otherProp)]]</div>
```

If `foo.bar` becomes unreachable (thus undefined) the first div will see an update to its textContent, the second one will not update at all (which is a bug #2158). This is hard to work around b/c a computed property wouldn't update as well. The proposal is that for computed properties as well as for computed annotations the assigned value switches to undefined if the dependencies cannot be fulfilled and thus the actual computation cannot be evaluated. ... I hear you cry. Just let that sink for a moment.

The idea here is (of course) that for `fn(foo)` if the computation cannot be evaluated its result is literally _undefined_. You can as above now opt-in and mark some arguments as optional, which allows you to do something like `_OR(?foo.a, foo.b)`, which is otherwise not possible atm.

kaste:
This implements the initialization guard for all computed expressions (like multi observers etc.) as stated in the docs. T.i. we ensure that observers (et.al.) don't get called with undefineds _initially_, but if they did run **once**, they will see undefined values just like any other value.

The second commit fixes the bug that `*.splices` observers did fire with undefined initially. 

Fixes #2158, fixes #3239, fixes #3401.

### DIFF
--- a/src/lib/bind/effects.html
+++ b/src/lib/bind/effects.html
@@ -125,12 +125,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var values = [];
       var args = effect.args;
 
-      var bailoutEarly;
-      if (Polymer.Bind.expressionDidEvaluateOnce(this, effect)) {
-        bailoutEarly = false;
-      } else {
-        bailoutEarly = (args.length > 1 || effect.dynamicFn);
-      }
+      var bailoutEarly = !Polymer.Bind.expressionDidEvaluateOnce(this, effect);
 
       for (var i=0, l=args.length; i<l; i++) {
         var arg = args[i];
--- a/src/lib/bind/effects.html
+++ b/src/lib/bind/effects.html
@@ -62,7 +62,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _complexObserverEffect: function(source, value, effect) {
       var fn = this[effect.method];
       if (fn) {
-        var args = Polymer.Bind._marshalArgs(this.__data__, effect, source, value);
+        var args = Polymer.Bind._marshalArgs.call(this, this.__data__,
+                                                  effect, source, value);
         if (args) {
           fn.apply(this, args);
         }
@@ -79,7 +80,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _computeEffect: function(source, value, effect) {
       var fn = this[effect.method];
       if (fn) {
-        var args = Polymer.Bind._marshalArgs(this.__data__, effect, source, value);
+        var args = Polymer.Bind._marshalArgs.call(this, this.__data__,
+                                                  effect, source, value);
         if (args) {
           var computedvalue = fn.apply(this, args);
           this.__setProperty(effect.name, computedvalue);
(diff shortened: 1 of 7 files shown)


kaste:
I just realized that I committed the wrong thing. The above description though is correct, and this change works and does what I say. I just wanted to implement a different logic.

TimvdLippe:
Observers run with `undefined` in Polymer 2 as described in [the upgrade guide](https://www.polymer-project.org/2.0/docs/upgrade#update-observers), therefore I am closing this PR.

PR #3562, opened 2016-04-05
Prevent tap from firing after a 300ms long touch.

Saulis:
### Reference Issue

Fixes polymer/polymer-gestures#12

Currently, unintentional taps happen easily with elements like `<iron-list>` where a user can first touch scroll down and then back up having `touchend` getting fired inside the same area where touching started.

Adding a simple timeout for the tap event, helps, alternative approach for this issue could to be track the total distance from the track events and have a threshold for that.


azakus:
Sorry for the long delay in addressing this PR:

This really isn't the right way to solve this problem, because the actual issue is not based on time, but on movement in between the `down` and `up`. I think a more correct way to differentiate accidental taps like this would be to measure movement on `touchmove` and cancel the tap if the user has moved greater than some value, probably the same value as checked in the `touchend`.

I'd be open to accepting a PR that uses a movement vector instead.

PR #3570, opened 2016-04-07
Update the package.json name to match the actual npm published package

ChadKillingsworth:
Just a minor tweak to match what has actually been published.

Also, npm doesn't allow mixed-case package names.


azakus:
LGTM

PR #3573, opened 2016-04-10
Bare minimum for custom effects (take 2)

kaste:
Same as #3460. Removed prestored effects in createAccessors as suggested per https://github.com/Polymer/polymer/pull/3460#issuecomment-206599307.

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -224,7 +224,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           // queued events during configuration can theoretically lead to
           // divergence of the passed value from the current value, but we
           // really need to track down a specific case where this happens.
-          value = target[property];
+          value = target.__data__ ? target.__data__[property] : target[property];
 
           if (negated) {
             value = !value;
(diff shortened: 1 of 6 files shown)


kaste:
The last commit here is a minor refactoring in Templatizer that fits on top of your https://github.com/Polymer/polymer/pull/3573/commits/59b5f240dadc61e16fedc98c17dbd9c1c17861af

src/standard/notify-path.html
```diff
@@ -274,10 +274,12 @@
         var fx$ = this._propertyEffects && this._propertyEffects[model];
         if (fx$) {
           for (var i=0, fx; (i<fx$.length) && (fx=fx$[i]); i++) {
```
kaste: Note to self: Must prestore length just like in accessors!

src/lib/bind/accessors.html
```diff
@@ -224,7 +228,7 @@
           // queued events during configuration can theoretically lead to
           // divergence of the passed value from the current value, but we
           // really need to track down a specific case where this happens.
-          value = target[property];
+          value = target.__data__ ? target.__data__[property] : target[property];
```
fearphage: I believe this reads a little clearer. 

```js
value = (target.__data__ || target)[property]
```

TimvdLippe:
Closing per https://github.com/Polymer/polymer/pull/3460#issuecomment-320333064

PR #3575, opened 2016-04-11
Support the devtools console.log api (multiple strings) for polymer logging

### LINKED ISSUE #3574: _log, _warn, _error don't work as console respective methods
`Polymer.Base._log/_warn/_error` should work as the respective `console.log/warn/error` (accept multiple strings rather than only an array as input)
e.g. `Polymer.Base._warn('warn', 'msg')`

http://jsbin.com/xutoyu/2/edit?html,console,output

@azakus PTAL

azakus:
Fixes #3574

### DIFF
--- a/src/lib/base.html
+++ b/src/lib/base.html
@@ -156,17 +156,40 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
     },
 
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
LGTM

PR #3577, opened 2016-04-11
Lazy register extends

### LINKED ISSUE #3345: Add support for element inheritance
Here's a strawman:

```
Polymer({
  is: 'x-base'
});

Polymer({
  is: 'x-foo',
  extends: 'x-base'
});
```

Issues: 
- this will need to be integrated with Polymer's behavior system.
- we'll want to expose a way to modify/merge super/subclass templates.

cdata:
Fixes #3345 .

This branch is a reconciled lineage of the `extends` branch, which fell out of sync with `master` a few months ago. It mostly reconciles an alternate `lazy-register` implementation with the one that eventually landed in master.


TimvdLippe:
Closing this PR as element inheritance is now possible using mixins and the native ES6 classes.

PR #3579, opened 2016-04-12
Allow setting read-only properties using `set`

### LINKED ISSUE #3552: Allow `set()` to set read-only properties
This is a no-brainer. Currently we are forced to use setters like `_setResult()`. Since we have a general purpose `set()`, it should be possible to use it instead `set('result', ...)`. Why the hell?
1. Way easier to find assignments to a property in standard editors. For JS we pretty much rely on '_find all occurrences of this word_' to guide us through the code.
2. One API to rule them all.

kaste:
Fixes #3552

Note: computed properties could be set as well. Nobody should do this.

Can I remove the third argument `root` from `set()`? It is unused within the framework. Since `set` is public API I would like the following signature:

Before `set: function(path, value, root) {`
After `set: function(path, value, fromAbove) {`

### DIFF
--- a/src/standard/notify-path.html
+++ b/src/standard/notify-path.html
@@ -206,8 +206,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             this._notifyPath(parts.join('.'), value);
           }
         } else {
-          // Simple property set
-          prop[path] = value;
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Closing per https://github.com/Polymer/polymer/issues/3552#issuecomment-326162395

PR #3580, opened 2016-04-12
Let read-only properties participate in upwards flowing data.

### LINKED ISSUE #2632: Two-way binding annotations in local template do not update `readOnly` properties
This may be a deliberate design choice, and it's fine if it stays the way it is. However, I would like to submit for consideration the following:

Two-way binding annotations signal an intention from the element author that - even if a property is marked as `readOnly` - the property should be updated when the bound value notifies. Perhaps Polymer should automatically update `readOnly` properties in such conditions.

### LINKED ISSUE #3551: Allow read-only properties to participate in upwards data flow.
Yes, that is participate on `upwards`; restrict like before on `downwards`. 

I think using composition the following element should just work. 

``` html

    <dom-module id="x-ample">
        <template>
            <x-ajax response="{{data}}" url="[[url]]"></x-ajax>
        </template>
        <script>
            Polymer({
                is: 'x-ample',
                properties: {
                    userId: String,
                    url: {
                        computed: '_computeUrl(userId)',
                        readOnly: true
                    },
                    data: {
                        notify: true,
                        readOnly: true
                    },
                },
                _computeUrl: function(userId) {
                    return '/user/' + userId;
                }
            });
        </script>
    <dom-module>

```

Current behavior is, that passing `url` down will work. Yeah, why not, you opted in. But getting the `response` and let it flow up doesn't. That's completely unnecessary b/c that would be opt-in as well.  

Obviously readOnly protects us correctly to not get values from above.

http://jsbin.com/vudovi/edit?html,output

kaste:
Fixes #3551, fixes #2632

Note that from the two possible variants 

```
+    <x-basic id="basic5"
+      notifyingvalue="{{readonlyvalue}}">
+    </x-basic>
+    <x-basic id="basic6"
+      notifyingvalue="{{readonlyobject.foo}}">
+    </x-basic>
```

`basic6` actually was working before, but `basic5` did not. This inconsistency-bug is reported here https://github.com/Polymer/polymer/issues/2632#issuecomment-199011930

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -193,8 +193,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (!model._bindListeners) {
         model._bindListeners = [];
       }
-      var fn = this._notedListenerFactory(property, path,
-        this._isStructured(path), negated);
+      var fn = this._notedListenerFactory(property, path, negated);
       var eventName = event ||
         (Polymer.CaseMap.camelToDashCase(property) + '-changed');
       model._bindListeners.push({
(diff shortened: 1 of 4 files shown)


TimvdLippe:
We can not accept this change right now as it is a breaking change in the data-binding system. As there are small workarounds with computed properties or listening to an event, you can still achieve this functionality yourself.

TimvdLippe:
Oh, I got corrected by @kevinpschaaf, you can actually do it by setting a protected property: http://jsbin.com/lifuzaruro/edit?html,console,output (`this[this.PROPERTY_EFFECT_TYPES.READ_ONLY].prop = false;`) You can also implement this functionality in a mixin to abstract this logic away.

PR #3585, opened 2016-04-13
Fix nextSibling after appendChild

jscissr:
If `ref_node` is undefined instead of `null`, [node.__dom.nextSibling](https://github.com/Polymer/polymer/blob/3bd83871564770551b90926e746c5d651d60942e/src/lib/dom-tree-api.html#L218) is
`undefined` instead of `null` too. Because of this, [Polymer.dom(node).nextSibling](https://github.com/Polymer/polymer/blob/3bd83871564770551b90926e746c5d651d60942e/src/lib/dom-tree-api.html#L107) is the
nextSibling in the browser DOM when it should be `null`.


sorvell:
Fixes https://github.com/Polymer/polymer/issues/3730.

sorvell:
Closed in favor of https://github.com/Polymer/polymer/pull/3731.

PR #3587, opened 2016-04-15
Shim CSS Mixins in terms of CSS Custom Properties

### LINKED ISSUE #3503: The custom style shim should stick the `:root` styles on `html` and not `body`
This is to be able to workaround the `default-theme` issue (https://github.com/PolymerElements/paper-styles/issues/93) by allowing a user to override the `:root` styles present in `default-theme.html` with their own `body {...}` styles. 

I thiiiiink this should work (currently doesn't): http://jsbin.com/revini/edit?html,output

(Sample JSBin using native custom properties: http://jsbin.com/vibadelure/edit?html,output)

azakus:
- Use native css custom properties if supported
  - Rewrite polymer's slightly non-standard syntax to spec
- Rewrite `@apply` to N custom properties

If native css custom properties supported, all style work will happen at registration time (+/- shadydom shimming).

Use of `updateProperties` to modify or create mixins is not supported.

Fixes #3503


azakus:
Fixes #3503

TimvdLippe:
@azakus if you add that line to the PR description it automatically closes the issue. GitHub does not check the comments section for such lines :(

src/lib/apply-shim.html
```diff
@@ -0,0 +1,197 @@
...
+        old = oldSet[i];
+        if (newSet.indexOf(old) === -1) {
+          diff.push(old);
+        }
+      }
+    }
+    return diff;
+  }
+
+  // produce variables in place of a mixin definition
```
sorvell: Let's include an example here: e.g. `--foo { background: red; }` -> `--foo-background: red;`

src/lib/apply-shim.html
```diff
@@ -0,0 +1,197 @@
...
+      v = propMap[p];
+      out.push(mixinName + '-' + p + ': ' + v);
+    }
+    // reset variables that are overriden by the current mixin
+    var resetProps = diffProperties(oldProps, propNames);
+    for (i = 0; i < resetProps.length; i++) {
+      p = resetProps[i];
+      out.push(mixinName + '-' + p + ': initial');
+    }
+    return out.join(';\n') + ';';
```
sorvell: Let's consider omiting the `\n` here. I think it doesn't enhance readability much due to indenting issues.

azakus: This isn't so much about readability, but not breaking following properties on a syntax error.

src/lib/apply-shim.html
```diff
@@ -0,0 +1,197 @@
...
+        if (sp.length > 1) {
+          // some properties may have ':' in the value, like data urls
+          out[sp[0].trim()] = sp.slice(1).join(':');
+        }
+      }
+    }
+    return out;
+  }
+
+  function mixinAssignment(matchText, propertyName, valueProperty, valueMixin) {
```
sorvell: What about `produceMixinProperties`

src/lib/apply-shim.html
```diff
@@ -0,0 +1,197 @@
...
+        parts.push(')');
+        vars.push(parts.join(''));
+      }
+      out = vars.join(';\n');
+    }
+    return out;
+  }
+
+  // replace mixin consumption with variable consumption
+  function mixinApplication(text) {
```
sorvell: What about `consumeMixinProperties`

src/lib/apply-shim.html
```diff
@@ -0,0 +1,197 @@
...
+    }
+    if (!valueMixin) {
+      return matchText;
+    }
+    var mixinAsProperties = mixinApplication(valueMixin);
+    var prefix = matchText.slice(0, matchText.indexOf('--'));
+    var mixinProps = textToProps(mixinAsProperties);
+    var propNames = Object.keys(mixinProps);
+    var oldProps = mapGet(propertyName);
+    mapSet(propertyName, propNames);
```
sorvell: Don't we need to add `propNames` to `oldProps`?

sorvell: Don't we need to add `propNames` to `oldProps`?

src/lib/apply-shim.html
```diff
@@ -0,0 +1,197 @@
...
+  function fixVars(matchText, prefix, value, fallback) {
+    // if fallback doesn't exist, or isn't a broken variable, abort
+    if (!fallback || fallback.indexOf('--') !== 0) {
+      return matchText;
+    }
+    return [prefix, 'var(', value, ', var(', fallback, '));'].join('');
+  }
+
+  // produce variable consumption at the site of mixin consumption
+  // @apply --foo; -> forAll ${propname} in --foo ~> --foo-${propname}: ${value}
```
sorvell: That's not what this does is it?

src/lib/apply-shim.html
```diff
@@ -0,0 +1,197 @@
...
+        f = fallbacks && fallbacks[p];
+        parts = [p, ': var(', mixinName, '-', p];
+        // if a property is defined in this rule, use it as a fallback to this variable
+        if (f) {
+          parts.push(',', f);
+        }
+        parts.push(')');
+        vars.push(parts.join(''));
+      }
+      out = vars.join(';\n');
```
sorvell: Let's consider omitting the `\n`

src/lib/apply-shim.html
```diff
@@ -0,0 +1,179 @@
...
+      return matchText;
+    }
+    var mixinAsProperties = consumeCssProperties(valueMixin);
+    var prefix = matchText.slice(0, matchText.indexOf('--'));
+    var mixinValues = cssTextToMap(mixinAsProperties);
+    var oldProperties = mapGet(propertyName);
+    var combinedProps = mixinValues;
+    if (oldProperties) {
+      combinedProps = Polymer.Base.mixin(oldProperties, mixinValues);
+    } else {
```
sorvell: Why is the else here? Don't we always want to update the property map with the list of all properties that have been used for that mixin?

src/lib/apply-shim.html
```diff
@@ -0,0 +1,179 @@
...
+      mapSet(propertyName, combinedProps);
+    }
+    var out = [];
+    var p, v;
+    // set variables defined by current mixin
+    for (p in mixinValues) {
```
sorvell: Shouldn't this be `combinedProps`?

src/lib/style-properties.html
```diff
@@ -147,10 +152,11 @@
         var parts = property.split(';');
         for (var i=0, p, m; i<parts.length; i++) {
           if ((p = parts[i])) {
-            m = p.match(this.rx.MIXIN_MATCH);
+            this.rx.MIXIN_MATCH.lastIndex = 0;
+            m = this.rx.MIXIN_MATCH.exec(p);
             if (m) {
               p = this.valueForProperty(props[m[1]], props);
-            } else {
+            } else if (!settings.useNativeCSSProperties) {
```
sorvell: Why do we need the check for native here? Shouldn't we never come here if using native?

src/lib/style-properties.html
```diff
@@ -235,55 +241,74 @@
...
+                styleTransformer._transformRuleCss(
+                  rule,
+                  styleTransformer._transformComplexSelector,
+                  scope.is,
+                  hostScope
+                );
+            }
+            selectorToMatch = rule.transformedSelector || hostScope;
+          }
+          if (scope && rule.propertyInfo.properties && selectorToMatch) {
```
sorvell: Why check `scope` here? Perhaps we can bail earlier if `!rule.propertyInfo.properties` (avoiding the unnecessary setup work if we don't use it)

src/lib/apply-shim.html
azakus: Since we are using `mixin`, the reference in the map is always the same and we don't need to `set` again.

src/lib/apply-shim.html
azakus: Yes, and we didn't have a test that exercised this. Adding one now.

src/lib/style-properties.html
azakus: Not anymore it shouldn't, though I think it did when I wrote this a few months ago :(

src/lib/style-properties.html
azakus: `scope` is a leftover of when I replaced the `element` argument, and is now not needed.
I added the check for `propertyInfo.properties` as an early bail-out, and it seems to work.

azakus:
Comments LGTM!

arthurevans:
For re-writing the @apply rules, should we consider something less direct than --mixin-prop to avoid possible collisions with other custom properties?

For example, `paper-menu` has a `--paper-menu` mixin and a `--paper-menu-color` property, so:

```
--paper-menu: { color: red; } 
```

is going to get rewritten as:

```
--paper-menu-color: red;
```

In this case they probably do the same thing, but the collisions could result in unintentional behavior. I'm stretching for an example here, but I can just about imagine a developer using `--my-app-background-color` (for a default background) and `--my-app-background: { color: ... }` (for the text color on a background panel).

Anyway, it seems like we should either a) document this potential funkiness, or b) use a less common character to combine the mixin and prop.

azakus:
@arthurevans That's a reasonable point. How about '$' or '~'?

azakus:
@arthurevans Went with `_-_`, so it will be _super_ obvious. separator is also exposed as `Polymer.ApplyShim._separator`

arthurevans:
:+1:

sorvell:
LGTM, can merge when ready...

PR #3590, opened 2016-04-15
Run each effect exactly once during config

kaste:
Two commits here: First commit ensures that an annotation effect runs exactly once during config, so we don't need to rely on dirty-checks here. Second commit shows the readable, straightforward `distributeConfig` code path, moving the invariants around. 

Note that `serializeValueToAttribute` should be overriden, but x-styling.html already does that.


TimvdLippe:
In Polymer 2 this suggestion has been adopted. For more information see https://github.com/Polymer/polymer/blob/eb170cbb11916cb3717b9cb732bf3f49baba734a/lib/mixins/property-effects.html#L1420-L1422 Once again thanks for the suggestion and sorry that we didn't respond earlier.

PR #3592, opened 2016-04-16
Ensure fromAbove in _forwardParentProp.

kaste:
The general rule here is: `_forwardParentProp/Path` must assume fromAbove
is true; for `_forwardInstanceProp/Path` assume fromAbove is false.


kevinpschaaf:
LGTM, thanks!

Issue #2815, opened 2015-11-17
<property>-change events are not always fired in Polymer 1.2.2/1.2.3


neolequang:
Hi,

I don't know if it's a bug or a new (desired) behavior, but something has changed in `<property>-changed` events in Polymer 1.2.2

I have a custom element element that is a composition of several child custom elements.

Child elements are interested in changes of the value of a property from the parent element.

So I used something similar to what was suggested in issue #2224, and when children are attached, they add an event listener to the parent for the `<property>-changed` event on the parent.

Using this element looks like the following :

```

<host-element>
   <x-parent prop="{{hostProperty}}">
      <x-child></x-child>
      <x-child></x-child>
      ...
   </x-parent>
<host-element>
```

Parent's registration code looks like :
```
    Polymer({
      is: 'x-parent',
      properties: {
        prop: {
          type: Number,
          notify: true,
        }
      },
      ...
```

When I change the `prop` property from the parent, with some code like 
```
this.prop = xxxx;
```
the `prop-changed` event is fired.

But if I change the value by changing the `hostProperty` that is bound to `prop`, then no event is fired. It was fired in Polymer 1.2.1.

I've created a jsbin to reproduce the issue. To see the difference, you will need to switch between 1.2.1 and 1.2.2 in polymer import href.

http://jsbin.com/rupatacuxo/edit?html,console,output

beaudesigns:
:+1: 

We have a similar issue where we're adding a custom event listener for `<property>-changed` events on certain property types. The event listener gets fired the very first time, but not any subsequent times. Digging into it, it looks like the new `_useCache` property in `_notifyChange()` is the culprit.

Manually setting `_useCache` to `false` in our build of Polymer seems to correct the listener.

wburgers:
Adding comment to receive notifications...

I have a child-to-host one-way binding, but I'm not seeing a <property>-changed event (on the host).
Also reflecting back to attribute doesn't seem to work, which is probably based on the same functionality.

wburgers:
Ah never mind my previous comment.
It can be regarded as not sent.
I bound two properties of sibling elements together, one child-to-host and one host-to-child, such that the data would flow from one child to the other via the host.
on the receiving element, I set an attribute ```blaStatus="[[wsStatus]]"```.
This was transferred to the blastatus property on the element, not the blaStatus property.
(Notice the case).
That is why I was not seeing updates.

I did manage to catch the change event eventually.
I added the listener to the wrong object...
[...]

PR #3593, opened 2016-04-16
Ensure fromAbove for downwards bindings.

kaste:
Essentially downwards flowing data change events shouldn't echo upwards.

This is a breaking change since you actually had a test (!) that ensures the opposite.

### DIFF
--- a/src/standard/effectBuilder.html
+++ b/src/standard/effectBuilder.html
@@ -338,10 +338,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         if (pinfo && pinfo.readOnly) {
           return;
         }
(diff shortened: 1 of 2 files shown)


kaste:
Haha, I just saw #3077. 

Note that the main problem here and #2815 was the new cached events systems, which was by far more surprising and breaking than this.

kevinpschaaf:
Going to go ahead and close this due to the breaking change issue, as noted in the code and follow-up to #3077.  As you saw we tried to do this and needed to revert it due to users calling `addEventListener` for `prop-changed` events on bound elements.  That said, the plan is to break this in the next major release of Polymer (which will have a lot of this code refactored).

PR #3594, opened 2016-04-16
Assume attribute changes as coming from above.

kaste:
When using the public API `node.set/removeAttribute` assume that these changes are coming from above, t.i. do not notify upwards.


azakus:
This is no longer an issue in 2.x, as all attribute changes are treated as "from above"

Example: http://jsbin.com/janavozaro/edit?html,console,output

PR #3596, opened 2016-04-17
Remove dirty checks for objects

kaste:
To come to an end here. Assembles the following PRs:
- Ensure fromAbove in _forwardParentProp. #3592
- Ensure fromAbove for downwards bindings. #3593
- Upwards notifications should not echo downwards #3550
- Run each effect exactly once during config #3590
- Bare minimum for custom effects (take 2) #3573
- Implement linkPath as (custom) property effect. #3547

Then we can just remove the dirty check for object values https://github.com/Polymer/polymer/pull/3596/commits/6d24b87dca8f1dc8da868b83038852b7d29dde10.

Related #3343.

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -351,14 +351,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // (properties) are case sensitive. Gambit is to map dash-case to
         // camel-case: `foo-bar` becomes `fooBar`.
         // Attribute bindings are excepted.
-        var propertyName = Polymer.CaseMap.dashToCamelCase(name);
         if (kind === 'property') {
-          name = propertyName;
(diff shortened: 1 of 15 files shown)


kaste:
I think we should have these API changes/additions:

- `set(path, value, fromAbove?)`
- `get(path)`
- `notifyUpwards(path, value?)` or `propagateUpwards`
- `assumeChanged(path, fromAbove?)`

whereby `path` can be simple or deep (`foo`, `foo.bar`)

devinivy:
@kaste I see that many of your ideas, this one included, made their way into the implementation of v2.  Just wanted to say thanks!

TimvdLippe:
As @devinivy pointed out your suggestions have been adopted and are available in Polymer 2 as [the MutableData mixin](https://www.polymer-project.org/2.0/docs/api/mixins/Polymer.MutableData). Thanks!

PR #3602, opened 2016-04-19
add more style[include] doc

### LINKED ISSUE #3601: Make a style[include] doc better
There should be a code example of `<style include="module-name">` in the `custom-style` doc.

We should also note explicitly that multiple `dom-module`s can be `include`d (i.e. `include="module1 module2"`), so people don't have to go searching through the source code or [unit tests](https://github.com/Polymer/polymer/blob/edb59eb6283074ce54f374b30b884c9f9fcf9f5f/test/unit/custom-style-import.html#L23) to figure out if that's possible ;).

### Versions
- Polymer: v1.1.0+

/cc @sorvell

danbeam:
### Reference Issue
Fixes #3601

### DIFF
--- a/src/lib/custom-style.html
+++ b/src/lib/custom-style.html
@@ -54,7 +54,27 @@ Example:
     }
 
   </style>
-
+  
+  
+  <!-- This could easily live in another .html file and be imported instead. -->
+  
+  <dom-module id="shared-style">
+    <template>
+      <style>
+        :root {
+          --shared-background-color: #bada55;
+        }
+      </style>
+    </template>
+  </dom-module>
+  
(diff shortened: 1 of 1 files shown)


src/lib/custom-style.html
```diff
@@ -54,7 +54,27 @@
     }
 
   </style>
-
+  
+  
+  <!-- This could easily live in another .html file and be imported instead. -->
```
sorvell: This actually ends the HTML comment so the example below is not commented.

danbeam: Just removed that comment for now.

sorvell:
LGTM, thx!

Issue #3608, opened 2016-04-21
Use property effects instead of DOM events for upwards propagation


kaste:
WIP. Actually @d1manson brought me to this. Consider the following typical scenario:

```
    <x-mediator>
        <shadow>
            <x-producer people="{{data}}">
            <x-viewer people="{{data}}">
```

The `<x-producer>` populates the data, the mediator passes it to the `<x-viewer>`, after that updates flow between them. In the jsbin I do bursts of 100 updates on each cycle.

Current Polymer http://jsbin.com/xixeta/edit?html,output

![image](https://cloud.githubusercontent.com/assets/8558/14721577/68cc8f08-0807-11e6-928c-221a59722749.png)

Note that on each cycle we have a _long frame_. The selected burst takes 39.5ms, 29.5ms are spent dispatching events.

Replacing the DOM events with property effects http://jsbin.com/tebefe/edit?html,output

![image](https://cloud.githubusercontent.com/assets/8558/14721669/f1f2f9de-0807-11e6-9a13-5c4012c754b9.png)

No long frames. 10ms per cycle. :fireworks:

jfrazzano:
Great plus add. Using typed callbacks for updates to thin larger data sets (100x plus) also lowers notification run time to 25000 key encoded records. Element composition density and nesting Dom-repeats have-based on my limited experience-difficult to calculate, log time issues. 

Is there a way to touch the collection via props not events, to avoid the cross Dom repeat scope synch issue that is shown in model data of object sub props? 

I'm specific, the e.model,__data__
people>{} opened to {a: x, b:y}, 
But showing above as 
a.trueValue, b.y, when changing across Dom repeat scopes?  

In specific, running drag and drop grid cells for data grid/excel/Google sheets proxy, and notify splice fails live update, model for item or element can't can't be used to swap elem as it pulls out a node or swaps a node key to null during event xycle if user drop is slow. 

Waiting loses event data on drag start, and sync of object sub properties on nested Dom repeat is erratic as result of above mentioned prop.a: x, prop{} opened to prop{ a: some other Val in change record}

I managed to map independently and have near zero cycle time but binding is via a generator callback next set get using a--for lack of a better term--pseudo-map tied to a data map tied to the user side arrays 

I ask here primarily because I 1. Lack experience to have certainty on apropos approach, 2. Don't want team work to depend on my hack that goes against coming polymer changes, improvements, and 3. To ask on a related thread if similar model data sub property synching has been experienced. 

[...]

PR #3609, opened 2016-04-21
Use property effects to notify path changes upwards

kaste:
Reference Issue #3608

WIP. Only https://github.com/Polymer/polymer/commit/f9d64e157e07f802fe10dba564012d14c6b25d48 is of any interest here. Uses my custom property effects implementation.


devinivy:
Any word on this?  I'm quite interested in the potential perf improvement here.

sorvell:
Closing this as we're not adding features to Polymer 1.x now, and for Polymer 2.x, we are limited in what we can change since we want to be api-compatible with Polymer 1.x.

For cases where high frequency changes occur and feed into the binding system, it's preferable to debounce the changes and manually notify to avoid the excess work.

Issue #3211, opened 2015-12-17
Inconsistencies observing specific array elements


devinivy:
When I observe a specific array element, it may or may not be notified properly depending on how the element is changed.  This _seems_ to be an issue in 1.2 but not prior to that, though I'm not entirely positive.  Previous to 1.2 I would (referring to the example) observe `arrayProperty.0` rather than `arrayProperty.#0`.

Additionally, I've become very fuzzy on when `#n` element observers need to be used.  I can only get this working with `#n` array element observers, so I'll give an example using that, but any info on this topic would also be very useful.

Ex.

``` js
Polymer({

 /* ... */

  observe: [
    '_watchIt(arrayProperty.#0)'
  ],

  // _watchIt is notified
  ready: function() {
    this.arrayProperty = ['init'];
  },

  // _watchIt is notified
  mutateWithSet: function() {
    this.set('arrayProperty.#0', 'changed with set');
  },

  // _watchIt is not notified
  mutateWithSplice: function() {
    this.splice('arrayProperty', 0, 1, 'changed with splice');
  }

});
```

TimvdLippe:
This is still an issue in 2.x: http://jsbin.com/mucacepibe/edit?html,console,output

PR #3630, opened 2016-05-02
Fix conditions when dom-repeat needs to re-sort or re-filter.

### LINKED ISSUE #3626: dom-repeat doesn't call `_render` when an item changes
### Description

If you have a sorted dom-repeat and a change occurs to an item (such that the path is like `items.#2`), then `_render` will not be called internally (because of [this](https://github.com/Polymer/polymer/blob/master/src/lib/template/dom-repeat.html#L369)).

This seems to be because it assumes any path which isn't `items` or `items.splices` must be a sub-property change, and thus should follow through to the logic for the `observe` property.

However, not all other paths are sub-property changes, clearly. Some can be of the form `items.#2`.

Should this not trigger a re-render? Or somehow check if the properties defined in `observe` have changed?
#### Live Demo

http://plnkr.co/edit/fACxm4GDp1RsLKerazjW?p=preview

Clicking the button will set one of the children to a new object, causing a `items.#2` change notification.

As you can see, a workaround is in this demo, the observer manually calls `render()` if you uncomment it.

kaste:
- For efficiency do not process `items.length` changes
- Ensure observing `foo` is not triggered by a `foobar` etc.
- Ensure assignments like `items.#2` etc. will trigger a resort/-filter

Fixes #3626.

### DIFF
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -342,9 +342,10 @@ Then the `observe` property should be configured as follows:
       this._debounceTemplate(this._render);
     },
 
-    _observeChanged: function() {
-      this._observePaths = this.observe &&
-        this.observe.replace('.*', '.').split(' ');
+    _observeChanged: function(observe) {
+      // All observed paths are handled as wildcard/deep observation. So if
+      // the user specifies a wildcard, we just strip it.
+      this._observePaths = observe && observe.replace('.*', '').split(' ');
     },
 
     _itemsChanged: function(change) {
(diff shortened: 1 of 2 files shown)


kaste:
Note: this goes for least breaking change, and wildcard observing is the only option here, t.i. there is no difference between observing `foo` or `foo.*`.

There is something super-confusing to me. Actually ~~`set('items.#2', obj)`~~`set('items.2', obj)` _should not_ reuse the key, it should splice the array, removing what was at index `2` and adding (at the same index) a new object (with a new key). I mean that's the intent of the whole collections vs. arrays thing here. The keys ought to be stable. But... `set` allows this, and hence we have this issue in the `dom-repeater` not updating the view.

ryanwtyler:
@kaste if it didn't reuse the key I'd think you'd get some unexpected behavior with firebase-collection.  If you make a change locally, this.set('item.1.prop', val), you'd expect item[1] to have same key before and after but it wouldn't because firebase would call this.set('item.1', obj) and splice in a new object with a new key.

43081j:
FYI the new `polymerfire` elements fail badly with `dom-repeat` because of exactly this.

When an item changes, the path is generally `items.#n.prop` and doesn't trigger a filter/sort.

devinivy:
The problem described here is very similar to an issue I logged a while back https://github.com/Polymer/polymer/issues/3211

TimvdLippe:
We would accept a change in 2.x that fixes this behavior, as `dom-repeat` significantly changed between 1.x and 2.x. The change must probably be made in `_handleObservePaths` which should cause a render if the path at that point is an empty string aka it's a set of an array element.

Issue #3632, opened 2016-05-02
Expose bare minimum of a Path library/module


kaste:
Seeing #3626 and #3631 it is really time to factor a `Polymer.Path` module to get these path manipulations right once and for all. Usually you set minutes to hours trying to understand something like https://github.com/Polymer/polymer/pull/3439/commits/970d5763f87d21c41c90189c178e0a3882a84862

What we need:
- head(path)
  `foo.bar` => `foo`
- isDeep(path)
  `foo.bar` => true
  `foo` => false
- isAncestor(base, path)
  `('foo.bar', 'foo')` => true 
  `('foo.bar', 'bar')` => false
- isDescendant(base, path) 
  `('foo.bar', 'foo.bar.baz')` => true 
  `('foo.bar', 'foo.bar')` => false
- translate(base, newBase, path)
  `('foo', 'baz', 'foo.bar')` => 'baz.bar'
- matches(base, wildcard, path)
  `('foo.bar', false, 'foo.bar')` => true
  `('foo.bar', true, 'foo.bar.baz')` => true
  `('foo.bar', false, 'foo.bar.baz')` => false

Note that these functions are pure and can be memoized.

TimvdLippe:
Was fixed in #3633 but not properly closed.

PR #3633, opened 2016-05-02
Expose a Path module/library

kaste:
Reference issue #3632


kevinpschaaf:
Awesome work here @kaste, thanks for the cleanup.  Would you mind fixing the conflict with master, and then we'll merge this.

Actually, one small nit, would you consider changing `Polymer.Path.head()` to `Polymer.Path.root()`?  I prefer that over head.  Otherwise LGTM!

kaste:
Done.

kevinpschaaf:
LGTM!

Issue #3638, opened 2016-05-06
Calling importHref multiple times with the same url adds the link unnecessarily


sorvell:
When an import has already been loaded, we can avoid creating an import and putting it into dom. This is a minor ergonomic issue that would keep the head clean of multiple unnecessary imports.

abdonrd:
@sorvell It is fixed?

sorvell:
Yes, will be included in the next release.

abdonrd:
@sorvell awesome! :)

bastienarm:
is this a real problem to have X many times the same import, or does it have some perf issues? (just for understanding)

Issue #3637, opened 2016-05-06
Attachment timing subtly differs under Shadow and Shady DOM


sorvell:
When an element is registered asynchronously, under Shady DOM attached for hosted elements fires for all nested elements after all nested element observers have run. Under Shadow DOM, attached fires after each nested element's observers run. In general, attachment state is _not_ something that should be relied on in observers (due to the fact that an element may be created outside of dom and attached later); however, the inconsistency between Shady and Shadow DOM is a subtle difference that should be eliminated if possible. Polymer should normalize to the Shady attached behavior.

Behavior under Shady DOM: http://jsbin.com/nudaqozari/1/edit?html,console,output.

Behavior under Shadow DOM: http://jsbin.com/bagoyihoye/edit?html,console,output.

azakus:
This timing change breaks the contract that local dom is distributed before `ready` callback. If we still want this change, it will have to keep that contract.

sorvell:
The original links are now out of date. Here are updated examples that show the issue persists in Polymer 1.x:

* ShadyDOM: http://jsbin.com/vihexahixo/1/edit?html,console,output
* ShadowDOM: http://jsbin.com/canavigudu/1/edit?html,console,output

This issue has been *fixed* in Polymer 2.x

* ShadyDOM: http://jsbin.com/tayagifeja/1/edit?html,console,output
* ShadowDOM: http://jsbin.com/tamugeleyu/1/edit?html,console,output

Closing this issue since it has been corrected in Polymer 2.x and is too difficult to address without breaking changes in 1.x.

PR #3639, opened 2016-05-06
Fix 3637,3638

sorvell:

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -385,6 +385,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           e.target.removeEventListener('error', errorListener);
           return onerror.call(self, e);
         }
+        imprt.addEventListener('error', loadListener);
       }
       // if already loaded/erroed, fire 'fake' load/error event
       if (cached) {
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -385,7 +385,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           e.target.removeEventListener('error', errorListener);
           return onerror.call(self, e);
         }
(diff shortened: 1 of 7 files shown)


src/standard/utils.html
```diff
@@ -363,28 +363,47 @@
...
         }
+        imprt.addEventListener('load', loadListener);
       }
       if (onerror) {
-        l.onerror = function(e) {
+        var errorListener = function(e) {
+          e.target.__firedError = true;
+          e.target.removeEventListener('error', errorListener);
           return onerror.call(self, e);
         }
```
TimvdLippe: The errorListener should be added as listener to the link I suppose?

nazar-pc: No, import can be cached element if present `var imprt = cached || link;`. And below load/error events will be simulated on cached element if it is present or link will be loaded if necessary. So in either case `imprt` is what events will be fired on.

However, I do not understand the purpose of shortening `import` to `imprt` :smile: 
Also link creation might be avoided if cached import is already in place, this will save few wasted CPU cycles.

TimvdLippe: Yes sorry, mistyped link instead of imprt.

TimvdLippe:
@sorvell Would a testcase like this verify the error callback? This should generate a 404 and trigger the `on-error`.

```js
test('importHref error callback is invoked', function(done) {
  Polymer.Base.importHref('non-existing-page.html', null, done);
});
```

cherukumilli:
@sorvell 
@kevinpschaaf 
Any updates on merging this pull request?

sorvell:
https://github.com/Polymer/polymer/issues/3638 was separately closed.

https://github.com/Polymer/polymer/issues/3637 is still open and this change breaks the api contract as discussed in the issue.

Closing this PR since it is no longer relevant.

PR #3641, opened 2016-05-06
Strawman for `setMultipleProperties`

### LINKED ISSUE #3640: Feature requrest: an element method to atomically set multiple properties
As an element author, there are times in which multiple `notify` properties are updated in response to another property changing. Currently these changes will happen in a certain order, and change notifications will happen synchronously and in that order, resulting in observers seeing an inconsistent view of these properties.

It would be nice to be able to set multiple properties atomically, such that there's no externally visible observable time in which some of the properties have been updated and others haven't. 

This appears to be possible if we provided a method that first updates `__data__`, and then does notification.

rictic:
Adds a method to Polymer.Base for setting multiple properties atomically, such
that there's no externally visible time when some but not all of the given
properties have been set.

Fixes #3640

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -87,8 +87,27 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             this.__data__[prop] = undefined;
           }
         }
-      }
+      },
 
+      /**
+       * Updates a set of properties atomically, such that there is no
+       * observable time in which some but not all of the properties are
+       * updated.
+       *
+       * Takes an object mapping property names to their new values. Does not
+       * currently support updating nested properties like 'x.y'.
+       */
+      setMultipleProperties: function(propsObj) {
+        for (var prop in propsObj) {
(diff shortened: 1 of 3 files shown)


rictic:
I expect that if this change is desirable it will need more tests, but I wasn't sure what the most important edge cases would look like.

akc42:
As I have pointed out elsewhere (https://github.com/PolymerElements/app-route/issues/109) and its pull request to resolve it (https://github.com/PolymerElements/app-route/pull/132) it doesn't work because _notifyPathUp generates an event with a "path" key in the detail.  This causes some of the updates to the listening element to be skipped.  The use case I had was that an observer wasn't called.

TimvdLippe:
This has been implemented in Polymer 2 as [`setProperties`](https://www.polymer-project.org/2.0/docs/api/mixins/Polymer.PropertyEffects#method-setProperties) which is also adopted in app-route: https://github.com/PolymerElements/app-route/blob/90d4d8da7d7f7762c68f10d93577ea079035a8e4/app-route.html#L325

Issue #2581, opened 2015-10-16
"use strict"; is not supported in Safari - blocks using ES6 and Babel


azogheb:
When adding "use strict" to the top of a vulcanized and crisper'ed JS file, there are two errors that come up ONLY when using `<template is="dom-repeat">` in Safari.

**TypeError: undefined is not a constructor (evaluating 'new this.ctor(model, this)')**
not sure where this comes from as the backtrace only shows (anonymous function) in _atEndOfMicrotask 

**TypeError: Attempted to assign to readonly property.**
at https://github.com/Polymer/polymer/blob/master/src/lib/template/templatizer.html#L133

This has come up as I am using babel much like @ebidel was in https://github.com/Polymer/polymer/issues/2363

To reproduce this in a nice sandbox I have prepared a branch using @ebidel material playground project as the base. I am not using babel, just simply adding "use strict" to the top of the JS
https://github.com/azogheb/material-playground/tree/useStrictDemo

The command I am using is as such:
`vulcanize --inline-scripts --inline-css elements.html | crisper -h vulcanized.html -j vulcanized.js;
echo "'use strict';"|cat - vulcanized.js > /tmp/out && mv /tmp/out vulcanized.js;`

In the above demo there is another strict mode error with web-animations-js and I have raised a ticket with them as well. https://github.com/web-animations/web-animations-js/issues/46

miztroh:
Is it allowable to "use strict" at the top of a script?  I thought it was only for use inside a function.

azogheb:
@miztroh Yes strict mode is most commonly used for entire scripts. As per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Strict_mode_for_scripts *Strict mode applies to entire scripts or to individual functions. It doesn't apply to block statements enclosed in {} braces; attempting to apply it to such contexts does nothing.* Also see http://babeljs.io/docs/advanced/transformers/other/strict/ for how Babel and ES6 will use strict mode

miztroh:
@azogheb Thanks for the clarification.

ebidel:
@azogheb can put up that modified demo on gh-pages?

ebidel:
Also curious on what version of Safari you're using?

azogheb:
Hey @ebidel I have pushed a updated version to gh-pages here http://azogheb.github.io/material-playground/

I am using Version 9.0 (10601.1.56.2) on OSX 10.10.5 (14F27)

[...]

PR #3647, opened 2016-05-09
`use strict`ify closured Polymer source

jorgecasar:
### Reference Issue

Suggested in https://github.com/Polymer/polymer/issues/2581#issuecomment-185911805

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -61,7 +61,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
  *
  * @class Template feature
  */
-
+(function() {
+  'use strict';
+  
   // null-array (shared empty array to avoid null-checks)
   Polymer.nar = [];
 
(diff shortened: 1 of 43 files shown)


jorgecasar:
`accessors.html` is only file to add `'use strict';` because of un-exist setter on readonly property throw an error.

I thought 2 options, but I'm not sure if there are more o what of this are the best:

1. In `_notedListenerFactory`, check also if `this.properties[path].readonly` before assign the value. [Line 232](https://github.com/jorgecasar/polymer/blob/use-strictify-all-source/src/lib/bind/accessors.html#L232)
2. In `_createAccessors`, create a empty function when the property is readonly, like `defun.set = function(){};`. [Line 175](https://github.com/jorgecasar/polymer/blob/use-strictify-all-source/src/lib/bind/accessors.html#L175)

Is there any other option? What do you think is the best one?

@ebidel solving this question I guess we can close this p1 (#2581).

kaste:
Note: PR #3580 and #3550 already take _notedListenerFactory in that direction.

TimvdLippe:
All files in Polymer 2 should have `use strict` now. Please open a new PR if you find one that we missed. Sorry for the late reply :cry:

PR #3662, opened 2016-05-18
Fixes 3661

sorvell:
Ensure cached style points to applied/active style under Shady DOM. This prevents extra unnecessary styles from spamming the document when changes are made to properties affecting nested elements.


kevinpschaaf:
LGTM

PR #3668, opened 2016-05-20
Allow more whitespace in style include attribute

mgiuffrida:
Style tags can grow beyond 80 characters:

```
<style include="my-custom-element-shared-styles iron-flex iron-flex-reverse
    iron-flex-alignment iron-positioning">
```

which leads to a bunch of warnings (the first one because the supposed module ID includes a newline):

> Could not find style data in module named iron-flex-reverse
> 
> Could not find style data in module named
> Could not find style data in module named
> Could not find style data in module named
> Could not find style data in module named

So split on _any_ whitespace, and ignore empty strings after splitting.

### DIFF
--- a/src/lib/style-util.html
+++ b/src/lib/style-util.html
@@ -163,10 +163,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       },
 
       cssFromModules: function(moduleIds, warnIfNotFound) {
-        var modules = moduleIds.trim().split(' ');
+        var modules = moduleIds.trim().split(/\s/);
         var cssText = '';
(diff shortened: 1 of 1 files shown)


sorvell:
LGTM, but would you mind adding a test? Thanks.

mgiuffrida:
@sorvell Sorry I dropped this, getting back to it now -- but there are a few test files, I'm not sure where a test would be relevant, e.g.:

* test/unit/custom-style.html 
* test/unit/styling-remote-elements.html 
* test/unit/styling-remote.html

freshp86:
Does this fix apply to Polymer 1 too? If so which version includes the fix?

PR #3678, opened 2016-05-24
Fix for changing property to the same value

### LINKED ISSUE #3677: Firing value-changed should not cause repeated value change on element itself
### Description
I've created small behavior for myself to simplify data binding to type-extension element `input[is=cs-input-text]` like following:
```javascript
Polymer.cs.behaviors.value = {
  listeners: {
    change: '_changed',
    input: '_changed'
  },
  _changed: function(){
    this.fire('value-changed');
  }
};
```

The idea is that:
```html
<!-- Now I'm able to write following -->
<input is="cs-input-text" value="{{value}}">
<!-- instead of -->
<input is="cs-input-text" value="{{value::input}}">
```

However, Polymer overrides `input.value` with the same value when data-binding goes downwards and causes cursor to move from its current position to the end of the text.

#### Live Demo
https://jsbin.com/juvimisefu/1/edit?html,output

#### Steps to Reproduce (on Demo)
1. Put cursor inside of `><`
2. Type single character

#### Expected Results
Cursor remains right after typed character

#### Actual Results
Cursor moved to the end of the text

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions
- Polymer: v1.4.0
- webcomponents: v0.7.22

nazar-pc:
Trivial fix, test included, no regressions detected in Chromium and Firefox.

### Reference Issue
Fixes #3677

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -70,7 +70,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var effects = node._propertyEffects && node._propertyEffects[property];
         if (effects) {
           node._propertySetter(property, value, effects, quiet);
-        } else {
+        } else if (node[property] !== value) {
           node[property] = value;
         }
       },
--- a/test/unit/bind.html
+++ b/test/unit/bind.html
@@ -869,6 +869,20 @@ suite('binding corner cases', function() {
     assert.isTrue(el._isAttachedChanged.calledOnce);
     document.body.removeChild(el);
   });
+
+  test('do not override property when going downwards', function() {
+    var value = 5, value_set_again = false, targetObj = {
+      get value () {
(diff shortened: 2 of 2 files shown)


kevinpschaaf:
LGTM

kevinpschaaf:
As this is a more robust approach than what we had previously (to do the dirty check prior to downward binding only when there was a custom event used (e.g. `{{value::input}}`), we can go ahead and remove [that code](https://github.com/Polymer/polymer/blob/1b02e96e77a0ef201976341fade763c617eca0d5/src/standard/effectBuilder.html#L328-L332).  I made a [separate PR](https://github.com/Polymer/polymer/pull/3799) for that.

Issue #3684, opened 2016-05-30
dom-if negated undefined expression does not compute as true


samuelli:
Test case: http://jsbin.com/hajekiluqo/edit?html,console,output

Expected: undefined is displayed
Actual: undefined is not displayed

Having `<template is="dom-if" if="[[!null]]">` computes to `true`. The same is not true with `undefined`. Seems inconsistent.

TimvdLippe:
We have discussed this use-case and decided that it is working as intended. As the data-binding system of Polymer relies on a push-model (e.g. when you set values, side effects like bindings run), a value that has yet to be set (`undefined`) causes no side effects to occur. Thus, if you want to "push" the value and update the binding, you should use a value other than `undefined`. Luckily this should be straightforward with any other falsy value, such as `null`.

We do agree that this notion is different than the falsiness of `undefined` in JavaScript, but this is a consequence of the push model used in data-bindings. Thus I am closing this issue as working as intended.

PR #3685, opened 2016-05-30
fix-binding undefined value of property

alonhar:
### Reference Issue
https://github.com/Polymer/polymer/issues/3684

the problem is that old at the beginning is undefined. If value===undefined 
then old===value. 
We have to consider the situation  that 
effect.negate===true  && value == undefined -> !undefined ===true  

but old===value so not change occur.


alonhar:
I signed it!

PR #3686, opened 2016-05-30
fix-binding undefined value of property

alonhar:
### Reference Issue

https://github.com/Polymer/polymer/issues/3684

the problem is that old at the beginning is undefined. If value===undefined 
then old===value. 
We have to consider the situation  that 
effect.negate===true  && value == undefined -> !undefined ===true  

but old===value so not change occur.


TimvdLippe:
I am more inclined to let bindings that are negated always run. However, if we would make this change now it would break existing applications who are relying on the fact that the dom-if does not trigger. Therefore I don't think we can make this change.

TimvdLippe:
Closing per above comment. If you want to explicitly trigger the binding, you can choose a different falsy value like `false`, `''` or `null`.

PR #3693, opened 2016-06-06
Support preventDefault() on touch

### LINKED ISSUE #3567: preventDefault only works in a tap handler if the source was a click, not a touch
If you call event.preventDefault() in a tap handler, it will only prevent default if the source event was a click, but not if the source was a touch.

This means that if I have a tap handler on a site with touch input and I want to prevent default, I need to add a touchend handler and call event.preventDefault there.  This leads to having two handlers for the same event (a tap and a touchend handler) in many places across the site, which isn't ideal.

In gestures.html (https://github.com/Polymer/polymer/blob/master/src/standard/gestures.html#L395), preventDefault is called for click events only, because for a touch the sourceEvent is Touch (changedTouches[0]) rather than something from the browser's event chain, like touchend.

azakus:
### Reference Issue

Expose setting to enable mouse handlers on touch-only devices

Fixes #3567


azakus:
@sorvell PTAL

azakus:
@kevinpschaaf PTAL

kevinpschaaf:
LGTM

PR #3708, opened 2016-06-10
Ensure custom styles updated after adding custom-style async. Fixes #3705.

### LINKED ISSUE #3705: Lazy-loaded `<style is="custom-style">` should trigger Polymer.updateStyles
### Description
Lazy-loaded `<style is="custom-style">` should trigger global updateStyles.  If you currently async load an import that contains a top-level `custom-style` (or more likely, an element that depends on a `custom-style`, such as `paper-card`), although the `custom-style` gets registered and the global `Polymer.StyleDefaults` properties get updated, they do not cascade down the tree.

#### Live Demo
Works (paper-card loaded sync): http://jsbin.com/raboji/edit?html,output
Fails (paper-card loaded on demand): http://jsbin.com/ruhuqo/edit?html,output

#### Steps to Reproduce
See failing example above.  Note, it's important to have a host element that consumes custom properties between the lazy-loaded element consuming lazy-loaded styles and the root, otherwise the custom property styles are chained lazily back up to the root and it will work.

#### Expected Results
Styles applied by global `<style is="custom-style">` should apply after being lazy-loaded.

#### Actual Results
Styles applied by global `<style is="custom-style">` do not apply after being lazy-loaded.

### Browsers Affected

- [X] Chrome

### Versions
- Polymer: v1.5.0
- webcomponents: n/a

kevinpschaaf:
### Reference Issue
Fixes #3705


sorvell:
LGTM'd by @kevinpschaaf

azakus:
custom-style-async tests are broken in Firefox with `useNativeCssProperties` because the `x-client` boots before the `custom-style`. https://travis-ci.org/Polymer/polymer/builds/139643120

PR #3721, opened 2016-06-17
Fixes #3555. Ensure selectors including `::content` without a prefix …

### LINKED ISSUE #3555: Local DOM rules with multiple selectors are shimmed incorrectly
### Description

Shared styles to global scope if we used multiple selectors (one usual, and one with `::content`) in one CSS rule.
Or i have to write only selectors `::content selector` if i'll use dynamically added content (elements).
Maybe related to this fix https://github.com/Polymer/polymer/commit/d9f3dda
Also it's only for `shady` DOM ;)

#### Live Demo

http://plnkr.co/edit/9RFUOiZNOsL0WNJ16iqT?p=preview

#### Steps to Reproduce

1. Use/Write multiple selectors in one rule, in second selector without `:host` before `::content`.
  - example: `button, ::content button {}`
2. You'll see affected usual non-components elements and dynamically added elements inside components.

#### Expected Results

```css
button.custom-element-2:hover,  custom-element-2 button:hover {
  background-color: cadetblue;
}
```

#### Actual Results

```css
button.custom-element-2:hover,  button:hover {
  background-color: cadetblue;
}
```

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 9
- [x] Safari 8
- [x] IE 11

### Versions

- Polymer: v1.4.0
- webcomponents: v0.7.20

sorvell:
...inside a complex selector are shimmed correctly and do not leak styling to the global scope.

Fixes #3555.

### DIFF
--- a/src/lib/style-transformer.html
+++ b/src/lib/style-transformer.html
@@ -168,10 +168,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             p$[i] = transformer.call(this, p, scope, hostScope);
           }
         }
-        // ensure reasonable whitespace between compound selectors
-        var selector = p$.join(COMPLEX_SELECTOR_SEP);
-        selector = selector.replace(/\s+/g, ' ').replace(/,([^\s])/g, ', $1');
-        return selector;
+        return p$.join(COMPLEX_SELECTOR_SEP);
       },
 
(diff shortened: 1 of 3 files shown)


azakus:
LGTM!

PR #3723, opened 2016-06-17
Make sure effect functions receive latest values

### LINKED ISSUE #3722: Multiple Observers for a property receive stale values
After https://github.com/polymer/polymer/commit/677f10c02bf372689b810b3be486aac73ec6fe9b, property observers can receive stale values in the observer function if they are not the first observer.
This can lead to competing validator observers and infinite loops in the property effects.

An example of this issue is visible with https://github.com/Polymer/polymer/blob/a54c1f23458167a9895a7b3b37f4b241e1c8197d/test/smoke/bad-property.html

azakus:
Fixes #3722


azakus:
@sorvell @kevinpschaaf PTAL

nazar-pc:
Don't you think function signature should be changed, since `value` argument is not used anymore? This is private method, so should not be a big problem.

sorvell:
LGTM.

PR #3724, opened 2016-06-18
Fixes #3530. When `updateStyles` is called and an element is not atta…

### LINKED ISSUE #3530: Calling updateStyles before an element is attached does nothing and is confusing
Calling `updateStyles` manually provokes the custom properties shim to update. Calling it before an element is attached has no effect. This is confusing to users and should either be fixed or produce a warning. For example, a developer might write an observer that calls `this.updateStyles`. This needs to be gated on attachment to function properly.

sorvell:
Fixes #3530.

…ched, invalidate its styling so that when it is attached, its custom properties will be updated.

Also fixed a related issue with scope style cache disablement when :host function rules are used:
 * "property prepping" and decoration are now after static shimming so that transformed selectors are available to calculate :host function,
 * the detection of :host function properly accounts for type extension,
 * since "property prepping" was moved after shimming, apply shim was separated from this and moved before shimming so that its output can be used there.


azakus:
LGTM!

azakus:
LGTM, #3708 is what broke the tests.

PR #3731, opened 2016-06-22
Fixes #3730, nextSibling after appendChildFixes

### LINKED ISSUE #3730: Polymer.dom(node).nextSibling incorrect after appendChild if node is not last child in composed tree
If `ref_node` is `undefined` instead of `null`, `node.__dom.nextSibling` is
`undefined` instead of `null` too. Because of this, `Polymer.dom(node).nextSibling` is the
`nextSibling` in the browser DOM when it should be `null`.

Here's an example: http://jsbin.com/qokeho/edit?html,output.

sorvell:
Inspired by https://github.com/Polymer/polymer/pull/3585. Fix is in a slightly different place and added one more test.


kevinpschaaf:
LGTM

PR #3735, opened 2016-06-24
Fixes #3734: address HI/CE timing issue in importHref. Fixes upgrade …

### LINKED ISSUE #3734: Lazy-loaded `<style is="custom-style">` containing mixin not shimmed correctly
### Description
Fix for #3705 left one failing test on Firefox which has native CSS properties but not HTML imports.  In this configuration when `useNativeCSSProperties` is used, the following test fails (not caught due to intermittent selenium failures on CI):
https://github.com/Polymer/polymer/blob/master/test/runner.html#L75

Failure happens because lazy-loaded import containing script that calls `Polymer({...})` causes synchronous registration & upgrade of an already-parsed element before a `<custom-style>` it depends on upgrades.  This violates the timing requirements of the `useNativeCSSProperties` optimization, which requires that custom property bags be defined before they're used.

#### Live Demo / Steps to Reproduce
Run test highlighted below on Firefox:
https://github.com/Polymer/polymer/blob/master/test/runner.html#L75

### Browsers Affected

- [ ] Chrome
- [X] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions

- Polymer: v1.5.0+master

sorvell:
…time dependencies of scripts on previous elements in async imports.

Fixes #3734.

### DIFF
--- a/src/lib/dom-module.html
+++ b/src/lib/dom-module.html
@@ -95,6 +95,15 @@
   // loaded by this point. In addition the HTMLImports polyfill should be
   // changed to upgrade elements prior to running any scripts.)
   var cePolyfill = window.CustomElements && !CustomElements.useNative;
+  // NOTE: Under polyfilled CE/HI, if script and html are separate, then
+  // for dom modules to be found, script should be executed as follows:
+  //   HTMLImports.whenReady(function() {
+  //    CustomElements.ready = false;
+  //    // registrations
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
LGTM.

`pr` build still fails with `Error: Unable to connect to selenium`, but more complete `push` build passes.

PR #3737, opened 2016-06-25
Even lazier, lazyRegister

### LINKED ISSUE #3736: The lazyRegister option can be even lazier providing additional performance
When `lazyRegister: true` is used, Polymer defers registration tasks until first-instance time. Currently, behaviors are mixed in at registration time. This could be done at first-instance time for an additional performance gain.

sorvell:
Fixes #3736. Defers behavior mixin until first-instance time.


test/unit/behaviors.html
```diff
@@ -53,13 +53,17 @@
...
-    var el = document.createElement('behavior-as-is');
-    assert.equal(el.is, 'behavior-as-is');
-    assert.ok(el.$.content);
-    el.prop = 42;
-    assert.isTrue(el.propChanged1Called);
-    assert.isTrue(el.propChanged2Called);
-    assert.isTrue(el.hasAttribute('attr'));
+    if (Polymer.Settings.lazyRegister === 'max') {
+      assert.isTrue(window.behaviorAsIsFail, 'Setting `is` in a behavior did not fail when lazyRegister: `max` set');
+    } else {
```
kevinpschaaf: Shouldn't be else?  Shouldn't the rest of the test still work in `max` mode?

kevinpschaaf: Sorry, got it.  `is` is defined in a behavior, which is the thing not supported with `max` mode.

kevinpschaaf:
LGTM

PR #3738, opened 2016-06-25
DRY on same code

mitogh:
Used on multiple instances on the same file.


TimvdLippe:
Closing this PR as `Polymer.Collection` is removed in Polymer 2: https://www.polymer-project.org/2.0/docs/upgrade#remove-key-paths-and-polymer-collection

PR #3740, opened 2016-06-25
Fixes #3739: correctly shim `:host(.element-name)` as `element-name.e…

### LINKED ISSUE #3739: `:host(.element-name)` leaks globally
Recently, `:host(element-name)` shimming was changed to work correctly as `element-name` but the fix incorrectly shims `:host(.element-name)` has `.element-name`. This leaks globally.

Example: http://jsbin.com/fikame/edit?html,output.

sorvell:
…lement-name`.

Also converts `:host(not-element-name)` to a non-matching selector.

Fixes #3739.


kevinpschaaf:
LGTM

PR #3772, opened 2016-07-08
Fix css shady build mistakenly matching root rules as host rules

azakus:
Fallback to parsedSelector when matching host rules to catch host-function and host-context rules corectly

Fixes tests in polymer-css-build


sorvell:
LGTM

PR #3777, opened 2016-07-12
Fix a typo in dom-if.html.

krozycki:
This fixes a small typo.

### DIFF
--- a/src/lib/template/dom-if.html
+++ b/src/lib/template/dom-if.html
@@ -13,7 +13,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <script>
 
   /**
-   * Stamps the template iff the `if` property is truthy.
+   * Stamps the template if the `if` property is truthy.
    *
    * When `if` becomes falsey, the stamped content is hidden but not
    * removed from dom. When `if` subsequently becomes truthy again, the content


bicknellr:
I think this is being used as shorthand for "if and only if" - there are [a few more instances](https://github.com/Polymer/polymer/search?q=iff) of 'iff' laying around.

krozycki:
Well, will it be a good idea than, if I go on and unwrap these shorthands?

arthurevans:
Yes, I think that's a good idea. "If and only if" is much clearer for the average reader.

ebidel:
"iff" is pretty well known in the technical/scientific  community as "if and only if".

krozycki:
So should I change this?

On Thu, Jul 21, 2016 at 17:22, Eric Bidelman <notifications@github.com> wrote:
"iff" is pretty well known in the technical/scientific community as "if and only if".

kevinpschaaf:
I'll go ahead and rule that the [shorthand](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=define%3A%20iff) is fine as-is.

arthurevans:
"Pretty well known" is not a great standard for terminology in API docs. Especially in a community like JavaScript developers, where many people are self-taught or don't have a traditional CS background. It's also extremely easy to overlook the repeated "f". 

From my own perspective, I've been coding and doing developer docs for 25 years now, but I'm largely self-taught as a coder. And the Polymer source is the *very first place* I have knowingly encountered this shorthand. And I only found the definition, by accident, after replacing several instances with "if," because I thought it was a typo.

I may be an anomaly, or I might not. But making it clear to the Arthurs of the world only costs 11 characters here, and I think that's a good tradeoff.

krozycki:
Imho we should chamge it. It may not be obvious to everyone.

On Thu, Jul 21, 2016 at 18:27, Arthur Evans <notifications@github.com> wrote:
"Pretty well known" is not a great standard for terminology in API docs. Especially in a community like JavaScript developers, where many people are self-taught or don't have a traditional CS background. It's also extremely easy to overlook the repeated "f".

From my own perspective, I've been coding and doing developer docs for 25 years now, but I'm largely self-taught as a coder. And the Polymer source is the very first place I have knowingly encountered this shorthand. And I only found the definition, by accident, after replacing several instances with "if," because I thought it was a typo.

I may be an anomaly, or I might not. But making it clear to the Arthurs of the world only costs 11 characters here, and I think that's a good tradeoff.

krozycki:
https://github.com/Polymer/polymer/pull/3794

kevinpschaaf:
Ok, that's fair @arthurevans.  I stand overruled, and merged #3794!

PR #3780, opened 2016-07-12
Ensure properties override attributes at upgrade time. Fixes #3779.

### LINKED ISSUE #3779: When an attribute and property are present at upgrade time, property should win
### Description
If both an attribute and its associated property are set at upgrade time, the property value should win.  

The current Polymer behavior is (unexpectedly) that attributes take precedence over properties. Based on the fact that attributes should generally be used for static initial configuration, and properties for dynamic changes thereafter, it is reasonable to assume that in the case an element has both an attribute and a property at upgrade time, the property was likely set last and should take precedence.

Normally with Polymer data-binding, the state of having both an attribute value and a property value at startup never happens since binding annotations are removed from attribute values before the element upgrades.  However, other data-binding systems may leave stale data or even binding annotations in the attribute, but proceed to update property values.

#### Live Demo
The issue came up in the context of this Angular1+Polymer demo:
https://plnkr.co/edit/F95ascEJWCtAP1xxmbKP?p=preview

#### Steps to Reproduce
Open the Plunkr

#### Expected Results
Polymer values are correct.

#### Actual Results
On Chrome, the Polymer values are displayed as expected.
On polyfill, the Polymer values are wrong. 

### Browsers Affected

All, although the particular incantation shown in the Plunkr above affects CE polyfilled browsers.

### Versions
- Polymer: v1.6.0
- webcomponents: v0.7.22

kevinpschaaf:
### Reference Issue
Read instance values into `_aboveConfig` rather than `config`, so that they override attribute values.  Requires handling instance values & defaults as separate loops, and required stubbing the new `_configureInstanceProperties` in `dom-bind` since the `hasOwnProperty` check would pass for `dom-bind`'s instance-time accessors, causing `undefined` from the getter to be configured into `_aboveConfig`.  `dom-bind` has its own early instance property handling.

Fixes #3779

### DIFF
--- a/src/standard/configure.html
+++ b/src/standard/configure.html
@@ -92,6 +92,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       this._configureAnnotationReferences();
       // save copy of configuration that came from above
       this._aboveConfig = this.mixin({}, this._config);
+      // save instance properties that may have been bound prior to upgrade
+      this._configureInstanceProperties(this._aboveConfig);
       // get individual default values from property configs
       var config = {};
       // mixed-in behaviors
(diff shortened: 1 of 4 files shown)


src/standard/configure.html
```diff
@@ -113,18 +115,23 @@
       }
     },
 
-    _configureProperties: function(properties, config) {
-      for (var i in properties) {
-        var c = properties[i];
+    _configureInstanceProperties: function(config) {
```
sorvell: Let's consider having this call `_configValue` to get the `readOnly` check. That will also mean this method should be called before `_aboveConfig` is made.

sorvell:
Needs a test =)

sorvell:
LGTM

PR #3679, opened 2016-05-24
Fixes #3676: retain <style> in <template preserve-content/>

jcmoore:
Fixes #3676

- check for `preserve-content` attribute on `<template/>`s in when gathering css

PR #3785, opened 2016-07-19
Fixes #3676: retain <style in <template preserve-content/> (re-attempt of #3679 to -- automatically -- appease CLA overlords)

### LINKED ISSUE #3676: "preserve-content" attribute does not protect style tags nested in templates
A `<template preserve-content/>` like the following #template:
```
<template preserve-content id="template">
  <div id="retained"></div>
  <style id="lost"></style>
</template>
``` 
will retain the `<div/>` but will lose the `<style/>` when nested inside of the main `<template/>` of a `<dom-module>`.

See http://jsbin.com/fixotaquno/edit?html,output

I think it would be preferable for any `<template preserve-content>` -- perhaps aside from the main `<template/>` of a `<dom-module>` -- to preserve its contents in their entirety.

This functionality seems achievable by adding some conditions around https://github.com/Polymer/polymer/blob/409ad8347b25ab029a72d821c27d9cccdc9f5ac7/src/lib/style-util.html#L145

I will follow with a PR.

jcmoore:
Fixes #3676

- check for `preserve-content` attribute on `<template/>`s in when gathering css

### DIFF
--- a/src/lib/style-util.html
+++ b/src/lib/style-util.html
@@ -142,7 +142,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           e = e$[i];
           // look inside templates for elements
           if (e.localName === 'template') {
-            cssText += this.cssFromElement(e);
+            // retain css content when specified,
+            if (!e.hasAttribute('preserve-content')) {
+              cssText += this.cssFromElement(e);
+            }
           } else {
             // style elements inside dom-modules will apply to the main document
             // we don't want this, so we remove them here.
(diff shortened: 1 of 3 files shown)


src/lib/style-util.html
```diff
@@ -142,7 +142,12 @@
           e = e$[i];
           // look inside templates for elements
           if (e.localName === 'template') {
-            cssText += this.cssFromElement(e);
+            // retain css content when specified,
+            // unless the template is subordinate to a <dom-module>
+            if (!e.hasAttribute('preserve-content') ||
+                !element.localName === 'dom-module') {
```
sorvell: Seems like we should remove the check for 'dom-module'. Putting `preserve-content` on a top level template in a `dom-module` is odd, but it seems like we should respect the user's ask here. Perhaps there's an alternate way to supply a template for some reason.

jcmoore:
Ok @sorvell , changed it up

sorvell:
LGTM. Thanks!

PR #3790, opened 2016-07-20
Fix mouse input delay on systems with a touchscreen

### LINKED ISSUE #3784: Tap event doesn't fire on touchscreen tap followed by mouse click
### Description

Tap event listeners do not fire when the user does a touchscreen tap followed by a mouse click. This issue is ported from https://crbug.com/618968.

#### Live Demo
https://jsbin.com/xewatatevi/edit?html,console,output

#### Steps to Reproduce
1. Open JSBin link above on a touchscreen laptop
2. Tap button one on a touchscreen
3. Click button two with a mouse

#### Expected Results
The event listener should fire for both events, printing 'tap' to the console twice.

#### Actual Results
The event listener only fires for the first tap, the second one is ignored.

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge

Untested:
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions

- Polymer: 1.6.0
- webcomponents: latest

azakus:
Use event.sourceCapabilities to allow non-synthetic mouse events to be
processed by the gesture system

Devices such as touchscreen Chromebooks can have input from mouse and
touch and the 2500ms delay is not necessary.

Fixes #3784

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -41,12 +41,24 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   // Check for touch-only devices
   var IS_TOUCH_ONLY = navigator.userAgent.match(/iP(?:[oa]d|hone)|Android/);
 
+  // Check for sourceCapabilities, used to distinguish synthetic events
+  var HAS_SOURCE_CAPS = 'sourceCapabilities' in UIEvent.prototype;
+
   // touch will make synthetic mouse events
   // `preventDefault` on touchend will cancel them,
   // but this breaks `<input>` focus and link clicks
   // disable mouse handlers for MOUSE_TIMEOUT ms after
   // a touchend to ignore synthetic mouse events
   var mouseCanceller = function(mouseEvent) {
+    if (HAS_SOURCE_CAPS) {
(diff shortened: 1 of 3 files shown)


src/standard/gestures.html
```diff
@@ -41,12 +41,24 @@
...
+  // Check for sourceCapabilities, used to distinguish synthetic events
+  var HAS_SOURCE_CAPS = 'sourceCapabilities' in UIEvent.prototype;
+
   // touch will make synthetic mouse events
   // `preventDefault` on touchend will cancel them,
   // but this breaks `<input>` focus and link clicks
   // disable mouse handlers for MOUSE_TIMEOUT ms after
   // a touchend to ignore synthetic mouse events
   var mouseCanceller = function(mouseEvent) {
+    if (HAS_SOURCE_CAPS) {
```
kevinpschaaf: Is this `if` only to save the `if` on line 58? You already test for existence of `sourceCapabilities` on that line, so seems like extra code for no gain.

azakus: Yeah, I'll rip that out, and `HAS_SOURCE_CAPS`

kevinpschaaf:
LGTM

PR #3793, opened 2016-07-21
Add a `noUrlSettings` flag to Polymer.Settings

### LINKED ISSUE #3786: Add a way to make Polymer not configurable via runtime settings
Probably via something like this:

`Polymer = {noFlags: true};`

This will prevent users from being able to opt in to Shadow DOM via `?dom=shadow` on sites that are not compatible with Shadow DOM.

sorvell:
Fixes #3786.


kevinpschaaf:
LGTM

Issue #3094, opened 2015-11-26
`filter` on `<dom-repeat>` doesn't use JS filter syntax as specified in documentation


David-Mulder:
The documentation states 

 > A function that can be used to filter items out of the view. This property should either be provided as a string, indicating a method name on the element's host, or else be an actual function. **The function should match the sort function passed to Array.filter**. Using a filter function has no effect on the underlying items array.

First of all it should probably state `callback function` instead of `sort function`. Secondly and more importantly the calls [here](https://github.com/Polymer/polymer/blob/master/src/lib/template/dom-repeat.html#L473) and [here](https://github.com/Polymer/polymer/blob/master/src/lib/template/dom-repeat.html#L558) miss the `index` and `array` arguments which one would expect on `Array.filter`'s callback function:

 > Function to test each element of the array. Invoked with arguments (element, index, array). Return true to keep the element, false otherwise.

Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

And yes, this is extremely useful, as Polymer doesn't support `myArray.0` bindings right now, so anytime you need to do that you have to actually use a  `dom-repeat` just to have the binding system work '-_- (and yes, I am aware of `array-selector` but that would be even more confusing)

kevinpschaaf:
Closing per https://github.com/Polymer/docs/issues/1381; website documentation is updated.  See discussion on https://github.com/Polymer/polymer/issues/2187 for background.

ktiedt:
When this was closed... the actual documentation in the code was not fixed.

https://github.com/Polymer/polymer/blob/master/src/lib/template/dom-repeat.html#L161-L168

SMEmp:
Also here: https://github.com/Polymer/polymer/blob/master/src/lib/template/dom-repeat.html#L72-L75

PR #3798, opened 2016-07-22
Missing piece to fixing #3094

ktiedt:
Missing piece to fixing #3094 

Update the documentation in the actual dom-repeat element.


SMEmp:
There's another misleading statement in the same file, located here: https://github.com/Polymer/polymer/blob/master/src/lib/template/dom-repeat.html#L72-L75

ktiedt:
Pulled the similar text from the /docs repo and updated that section as well. Thanks for catching that @SMEmp

kevinpschaaf:
LGTM, thanks!

PR #3806, opened 2016-07-25
updateNativeStyles should only remove styles set by updateNativeStyles

### LINKED ISSUE #3800: `updateStyles` incorrect when `nativeCSSCustomProperties` is true
The current code attempts to unapply all existing inline style properties and then sets the given properties. Instead, it should unapply any properties previously set with `updateStyles` and then apply the given properties.

azakus:
Fixes #3800


sorvell:
LGTM.

PR #3807, opened 2016-07-26
Rerun Apply Shim when mixins with consumers are redefined

### LINKED ISSUE #3802: ApplyShim can produce incorrect properties when elements are shared
See http://jsbin.com/qipodaq/edit?html,output. Inside `x-host2`, `my-element` should have a green background. 

In this case `x-mixin` contains a rule `@apply --mixin`. It is used within `x-host1` and `x-host2` but each of these defines a different set of properties for `--mixin`. Whichever of these elements stamps last will have incorrect properties applied via `@apply --mixin`. This is because we "bake" the stylesheet for `x-mixin` at the time its first instance is created.

Here's a proposal for how to fix this:

- For every mixin name, the ApplyShim keeps a map of elements that depend on (`@apply`) that mixin.
- Whenever the set of properties for a mixin is added to, each of the elements in the elements map is invalidated.
- Whenever an element is created, if it has already run the ApplyShim, it checks to see if any of its mixin dependencies are invalid and if so, it re-runs the ApplyShim. For ShadyDOM, this can replace the element's stylesheet. For ShadowDOM, it is probably good enough to fix the stylesheet in the element's template thus affecting all subsequent instances.

Sketch:

When `x-host1` is defined:

```
MixinMap:
  '--mixin': {
    properties: {color: true},
    elements: {'x-mixin': true}
  }
```

When `x-host2` is defined:

Add `background: green` to `--mixin` and invalidate `x-mixin`
 
```
MixinMap:
  '--mixin': {
    properties: {color: true, background: true},
    elements: {'x-mixin': false}
  }
```

When the next `x-mixin` is created:

- IF it has already run the ApplyShim, 
- then get the list of mixins it uses [`--mixin`], 
- check if any of these are invalid `(MixinMap['--mixin'].elements['x-mixin'] === false)`
 - if yes, re-run the ApplyShim

azakus:
When element stamps, make its mixins are valid, and if not, run the
Apply Shim again.

Flag is stored on element prototype to make the "valid" check extremely
fast

Only new instances of consuming elements will be updated

Fixes #3802

### DIFF
--- a/polymer.html
+++ b/polymer.html
@@ -72,7 +72,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       this._setupShady();
       this._registerHost();
       if (this._template) {
-        this._validateMixins();
+        this._validateApplyShim();
         // manage local dom
         this._poolContent();
         // host stack
--- a/polymer.html
+++ b/polymer.html
@@ -72,6 +72,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       this._setupShady();
       this._registerHost();
       if (this._template) {
+        this._validateMixins();
         // manage local dom
         this._poolContent();
         // host stack
@@ -88,7 +89,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // acquire instance behaviors
       this._marshalBehaviors();
       /*
-      TODO(sorvell): It's *slightly() more efficient to marshal attributes prior 
+      TODO(sorvell): It's *slightly() more efficient to marshal attributes prior
       to installing hostAttributes, but then hostAttributes must be separately
       funneled to configure, which is cumbersome.
       Since this delta seems hard to measure we will not bother atm.
--- a/src/lib/apply-shim.html
+++ b/src/lib/apply-shim.html
@@ -142,8 +142,10 @@ Polymer.ApplyShim = (function(){
       // NOTE: since we use mixin, the map of properties is updated here
       // and this is what we want.
       combinedProps = Polymer.Base.mixin(mixinEntry.properties, mixinValues);
+      var currentProto = ApplyShim.__currentElementProto;
+      var currentElementName = currentProto && currentProto.is;
       for (var elementName in mixinEntry.dependants) {
-        if (elementName !== ApplyShim.__currentElementProto.is) {
+        if (elementName !== currentElementName) {
           mixinEntry.dependants[elementName].__mixinsInvalid = true;
         }
       }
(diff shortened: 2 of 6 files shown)


polymer.html
```diff
@@ -72,6 +72,7 @@
       this._setupShady();
       this._registerHost();
       if (this._template) {
+        this._validateMixins();
```
sorvell: Let's name this more explicitly, what about '_validateApplyShim'

src/lib/apply-shim.html
```diff
@@ -133,12 +136,17 @@
...
     var mixinValues = cssTextToMap(mixinAsProperties);
-    var oldProperties = mapGet(propertyName);
     var combinedProps = mixinValues;
-    if (oldProperties) {
+    var mixinEntry = mapGet(propertyName);
+    if (mixinEntry) {
       // NOTE: since we use mixin, the map of properties is updated here
       // and this is what we want.
-      combinedProps = Polymer.Base.mixin(oldProperties, mixinValues);
+      combinedProps = Polymer.Base.mixin(mixinEntry.properties, mixinValues);
```
sorvell: We only have to invalidate if `mixinValues` adds new keys. Do you think it's worth detecting that?

azakus: I think that can be done in the loop lower that generates `initial` values. I'll give it a shot.

src/lib/apply-shim.html
```diff
@@ -133,12 +136,17 @@
...
+      for (var elementName in mixinEntry.dependants) {
+        if (elementName !== ApplyShim.__currentElementProto.is) {
```
sorvell: Is it too inconvenient to thread through the current element name here?

src/lib/apply-shim.html
```diff
@@ -118,6 +121,16 @@
     return out;
   }
 
+  function invalidateMixinEntry(mixinEntry) {
+    var currentProto = ApplyShim.__currentElementProto;
+    var currentElementName = currentProto && currentProto.is;
+    for (var elementName in mixinEntry.dependants) {
+      if (elementName !== currentElementName) {
+        mixinEntry.dependants[elementName].__mixinsInvalid = true;
```
sorvell: Let's also make this flag more explicit, maybe `__applyShimInvalid`

src/lib/custom-style.html
```diff
@@ -209,6 +209,7 @@
       }
       var styleRules = styleUtil.rulesForStyle(e);
       if (!targetedBuild) {
+        applyShim.__currentElementProto = this.__proto__;
```
sorvell: What use case does this support?

sorvell:
LGTM

Issue #3811, opened 2016-07-26
template dom-repeat does not iterate over array-like objects


pmaudsley:
I know there is an issues regarding repeating over object keys in https://github.com/Polymer/polymer/issues/1385

However, I am working with array-like objects and these also do not get stamped out either due to the Array.isArray() check.

I'll add a PR which implements a further check on objects passed through dom-repeat has the required properties and methods. (https://www.polymer-project.org/1.0/docs/devguide/properties#array-mutation)

mattyclarkson:
To provide more background: We have a web API that we models objects as a sparse array:

```
{
  "elements": {
    "713": {
      "type": "abc"
    },
    "16": {
      "type": "foo"
    }
  }
```

These are represented as an array-like `Object` as so:

```
console.log(api.elements.length);  // => 2
console.log(api.elements[0]._id);  // => 713
console.log(api.elements[1]._id);  // => 16
api.elements.push({'type': 'foo'});
api.elements[2].loadId().then(console.log);  // => unique id (say, 199)
api.elements.splice(1, 1);  // Removes the object with ID 16 in the API and locally
console.log(api.elements[0]._id);  // => 713
console.log(api.elements[1]._id);  // => 199
```

For all intents an purposes, `api.elements` acts like and array. It implements all the functions that a ES2015 array implements and when objects are added/removed does the necessary grunt work to keep the API up-to-date. Representing the API as an array, makes Polymer integration easy as `dom-repeat` works really well in this situation, except...it doesn't work because it has a `Array.isArray` check that discards our object and doesn't create a collection from it.

PR #3812, opened 2016-07-26
Allow dom-repeat for array-like Objects

pmaudsley:
Implement a further check on objects passed through dom-repeat to check if they have the required properties and methods (https://www.polymer-project.org/1.0/docs/devguide/properties#array-mutation) to allow stamping out of array-like objects on a template dom-repeat.
### Reference Issue

https://github.com/Polymer/polymer/issues/3811

### DIFF
--- a/src/lib/template/dom-repeat.html
+++ b/src/lib/template/dom-repeat.html
@@ -359,9 +359,17 @@ Then the `observe` property should be configured as follows:
         this.observe.replace('.*', '.').split(' ');
     },
 
+    _isArrayLike: function(o) {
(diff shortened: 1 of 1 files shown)


TimvdLippe:
The problem is a little bit more subtle as the code assumes that `items` is always an array. I think instead it should switch to `[Symbol.iterator]` instead to also be compatible with `Map` and `Set`.

TimvdLippe:
Closing per above comment. Feel free to publish a new PR that uses `[Symbol.iterator]` as a check. Right now there a lot of internal checks in `dom-repeat` that rely on the value being an Array, so the PR might grow a lot bigger.

PR #3815, opened 2016-07-27
add importHref listeners to non-cached imports

### LINKED ISSUE #3814: importHref only triggers onload and onerror if first invocation has listeners
### Description

`Polymer.Base.importHref` only triggers `onload` and `onerror` if first invocation has listeners. If another call to the same URI is made with load or error listeners these will never be called.

Imagine a scenario where two different components lazyload the same import, one of which cares for load events and one who just progressively enhance the experience. If the one which doesn't need to know whether or not the load is complete is initiated before the component that cares no load event will be triggered at all.
#### Live Demo

Example using a delayed prefetch for `paper-toast`. The button _Your toast_ must be clicked at least once before prefetching is complete to be able to show a toast using that button.
https://jsbin.com/behoyu/15/edit?html,output
#### Steps to Reproduce
1. Call `Polymer.Base.importHref('a.html', null, null, true)`
2. Wait for load to complete.
3. Call `Polymer.Base.importHref('a.html', e => console.log('onload', e), e => console.log('onerror', e), true)`

Neither `onload` nor `onerror` will be shown in the console.
#### Expected Results

Ordering of the two calls in the example should be of no consequence. The `load` or `error` listeners should be called in either case.
##### Expected Results - Example
1. Call `Polymer.Base.importHref('a.html', null, null, true)`
2. Wait for load to complete.
3. Call `Polymer.Base.importHref('a.html', e => console.log('onload', e), e => console.log('onerror', e), true)`

and
1. Call `Polymer.Base.importHref('a.html', () => {}, () => {}, true)`
2. Wait for load to complete.
3. Call `Polymer.Base.importHref('a.html', e => console.log('onload', e), e => console.log('onerror', e), true)`

should generate the same console output.
#### Actual Results

If first call is with no listeners the load and error callbacks in subsequent calls will never fire.
### Browsers Affected

All
### Versions
- Polymer: my guess is since #3638 got fixed.

TimLuq:
### Reference Issue

One solution that fixes #3814

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -375,20 +375,24 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var cached = list[link.href];
       var imprt = cached || link;
       var self = this;
-      if (onload) {
+      if (onload || !cached) {
         var loadListener = function(e) {
           e.target.__firedLoad = true;
           e.target.removeEventListener('load', loadListener);
-          return onload.call(self, e);
-        }
+          if (onload) {
+            return onload.call(self, e);
+          }
+        };
         imprt.addEventListener('load', loadListener);
       }
-      if (onerror) {
+      if (onerror || !cached) {
         var errorListener = function(e) {
           e.target.__firedError = true;
           e.target.removeEventListener('error', errorListener);
-          return onerror.call(self, e);
-        }
+          if (onerror) {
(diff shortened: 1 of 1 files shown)


azakus:
PR looks good to me. Can you add a test with a small reproduction?
Thanks!

PR #3816, opened 2016-07-27
Fixes #3801.

### LINKED ISSUE #3801: Style caching can be incorrect in document when `style is=custom-style` is used
#### Live Demo

test/smoke/custom-style-cache.html

#### Expected Results
First element has a font-size of 32px
Second element has a font-size of 12px

#### Actual Results
Both elements have a font-size of 32px

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 9
- [x] Safari 8
- [x] IE 11

### Versions

- Polymer: v1.6.0

sorvell:
Fixes #3801. Ensure style host calculates custom properties before elment. This ensures the scope's styles are prepared to be inspected by the element for matching rules.

### DIFF
--- a/src/standard/x-styling.html
+++ b/src/standard/x-styling.html
@@ -118,6 +118,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _updateStyleProperties: function() {
         var info, scope = this._findStyleHost();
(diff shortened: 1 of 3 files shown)


azakus:
LGTM

PR #3819, opened 2016-07-27
[property shim] Make sure "initial" and "inherit" behave as they would natively

### LINKED ISSUE #3803: Custom CSS variable shim applies invalid custom variables
Hooray for [red herrings](https://github.com/Polymer/polymer/issues/3802)! What does this do?

```
:host {
  --my-color: inherit;
  color: var(--my-color, red); /* Spoiler: Should resolve to red. */
}
```

To [substitute a var()](https://www.w3.org/TR/css-variables/):

* If the value of the custom property named by the first argument to the var() function is **anything but the initial value**, replace the var() function by the value of the corresponding custom property.
* Otherwise, if the var() function has a fallback value as its second argument, **replace the var() function by the fallback value.**

[Without useNativeCSSProperties, this doesn't work](https://jsfiddle.net/mgiuffrida/xhc8cdtd/) in a dom-module.

I tried testing in a `<style is="custom-style">` without a dom-module, but [NO variables seem to work in a custom-style](https://jsfiddle.net/mgiuffrida/29z1Lvhr/) without useNativeCSSProperties.

azakus:
Fixes #3803

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -91,10 +91,15 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         } else {
           var m, rx = this.rx.VAR_ASSIGN;
           var cssText = rule.parsedCssText;
+          var value;
           var any;
           while ((m = rx.exec(cssText))) {
             // note: group 2 is var, 3 is mixin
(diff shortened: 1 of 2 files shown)


sorvell:
LGTM.

PR #3821, opened 2016-07-27
[apply shim] Track dependencies for mixins before creation

### LINKED ISSUE #3802: ApplyShim can produce incorrect properties when elements are shared

azakus:
Fixes #3802 (again)

### DIFF
--- a/src/lib/apply-shim.html
+++ b/src/lib/apply-shim.html
@@ -133,20 +133,11 @@ Polymer.ApplyShim = (function(){
 
   function produceCssProperties(matchText, propertyName, valueProperty, valueMixin) {
     // handle case where property value is a mixin
-    var possibleVariableRedefine = false;
     if (valueProperty) {
       // form: --mixin2: var(--mixin1), where --mixin1 is in the map
       styleUtil.processVariableAndFallback(valueProperty, function(prefix, value) {
         if (value && mapGet(value)) {
           valueMixin = '@apply ' + value + ';';
-          // because the mixinMap is global, we mark this rule
-          // as we might conflict with a different scope's simple variable
-          // Example:
-          // some style somewhere:
-          // --foo:{ ... }
-          // some other element:
-          // --foo: red;
-          possibleVariableRedefine = true;
         }
       });
     }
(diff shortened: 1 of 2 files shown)


src/lib/apply-shim.html
```diff
@@ -178,6 +188,9 @@
     if (mixinEntry) {
       mixinEntry.properties = combinedProps;
     }
+    if (possibleVariableRedefine) {
```
sorvell: This flag seems unnecessary. Can't you just check if `valueProperty` here?

test/unit/styling-cross-scope-apply.html
```diff
@@ -736,6 +776,17 @@
...
+
+  test('mixins can be depended on before creation', function() {
+    var e = document.createElement('x-apply-depend-before-create');
+    document.body.appendChild(e);
+    CustomElements.takeRecords();
+    var e2 = document.createElement('x-create-after-apply');
+    document.body.appendChild(e2);
+    CustomElements.takeRecords();
+    assertComputed(e2.$.child, '2px');
+    assertComputed(e2.$.child, 'rgb(0, 0, 255)', 'color');
```
sorvell: Let's make a separate test for this assertion since it's testing that variable values that are mixins elsewhere function correctly.

sorvell:
LGTM.

Issue #1506, opened 2015-05-08
[0.9] Uncaught TypeError: Invalid value used as weak map key


teckays:
This error is thrown in `collection.html:131` if a binding like `<span>{{value}}</span>` contains an `!` character.

kevinpschaaf:
Could you please provide a quick reproduction case?  Thanks.

teckays:
I'm trying to replicate the error and can't get it done. In my code this is thrown it inside a template iteration through an ajax response object. I'll keep trying, hopefully I can reproduce it! Thanks.

teckays:
Still not able to replicate the case in jsbin. Would you like me to provide a link to the actual app so you could investigate the error from there?

teckays:
http://demo:UQDRNC7Vs1Ubmw@www.coders.pub/search/?location=moldova&skill=c&page=8

I have consoled the object received from ajax for you to check the values. I think that on this particular page the error is thrown by iterating `response.result.skills`. Here's the code responsible for iterations:

```html
...
<template is="dom-repeat" items="{{response.result.items}}">
  ...
      <div class="skills">
          <template is="dom-repeat" items="{{item.skills}}" as="s">
              <div class="skill">
                  <a href="{{linkBySkill(s)}}">{{s}}</a>
              </div>
          </template>
      </div>
    ...
</template>
...

<script>
    Polymer({
        is: "my-component",

        ...
        linkBySkill: function(skill){
            return '/search/?location=' + this.location + "&skill=" + skill;
        },
        ...
    });
</script>
```

zclark:
You can reproduce this error message by creating a listener using
```js
listeners: {
  "noexist.tap": "_handler"
}
```
while having no node with id of noexist in your template, possibly the cause?

Draccoz:
Polymer version 1.2.3 still has the above issue. Reproducible by attaching event listener to element inside dom-if template, example:

```html
<dom-module id="login-form">
  <template>
    <template is="dom-if" if="[[!loggedIn]]">
      <form id="login" action="." method="post">
        <button>Log in</button>
      </form>
    </template>
  </template>
</dom-module>
```

```js
listeners: {
  "login.submit": "_handler"
}
```

Pretty annoying and seems as old thing, would be nice if its fixed :).

azakus:
There's actually two separate issues here. The first is related to `dom-repeat`, and the second is related to the `listeners` block.

In the `dom-repeat` case, we'll need a working reproduction to investigate.

[...]

PR #3823, opened 2016-07-28
Better error message for bad listeners entry.

arthurevans:
Currently, if you specify a bad node in the `listeners` object, the result is a cryptic message about a missing key in a weak map. This can result if you use the ID of an element that's inside a `dom-if`, for example.
### Test case

Load the following element. It should log a more useful message when  it encounters the listeners key `foo.click`.

```
<dom-module id="a-thing">
  <template>
    <template is="dom-if" if="{{test}}">
    <div id="foo">
        Foo
      </div>
  </template>
  <div id="bar">
      Bar
  </div>
  </template>

  <script>
    Polymer({
      is: 'a-thing',
      listeners: {
        'foo.click': 'clickFoo',
        'bar.click': 'clickBar'
      },
      clickFoo: function() {
        console.log('foo clicked.');
      },
      clickBar: function() {
        console.log('bar clicked.');
      }
    });
  </script>
  </dom-module>
```
### Reference Issue

Related to: https://github.com/Polymer/docs/issues/1689, https://github.com/Polymer/polymer/issues/1506#issuecomment-175850553
### Note

Adding a `continue;` after line 71 lets the other listeners register correctly and lets the element finish initializing. Currently, the error appears to break the element entirely.  Not sure which behavior is preferable, so I left the `continue` out here.

### DIFF
--- a/src/standard/events.html
+++ b/src/standard/events.html
@@ -67,6 +67,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         } else {
           name = eventName.split('.');
           node = this.$[name[0]];
+          if (!node) {
+            console.warn('listeners: could not find node with ID', name[0]);
+          }
           name = name[1];
         }
         this.listen(node, name, listeners[eventName]);


TimvdLippe:
Closing per https://github.com/Polymer/docs/issues/1689#issuecomment-319837797

PR #3824, opened 2016-07-28
Handle mixins with property values of inherit and initial

### LINKED ISSUE #3818: ApplyShim does not correctly handle values of `inherit` or `initial` in mixins
The apply shim uses `initial` as a special value to help it forward default values to properties. Consider:

```
.foo {
  border: 2px solid red;
  @apply --foo;
}
```

In this case, if `--foo` does not contain a border property, elements matching `.foo` will have a 2px red border. Under the apply shim, if one setting of `--foo` contains `border` and another does not, we ensure that all `--foo` declarations contain at least `border: initial`. This allows us to write a rule like this:

```
.foo {
  border: 2px solid red;
  /*  shim for @apply --foo; */
  border: var(--foo_-_border, 2px solid red);
}
```

This works *unless* the user wants to set a value of `initial` for `border` in `--foo`. In this case the value is incorrect. In the native case, the computed value is `initial` and in the shim'd case it is the fallback.

See: http://jsbin.com/veqebo/edit?html,output.

azakus:
Fixes #3818

### DIFF
--- a/src/lib/apply-shim.html
+++ b/src/lib/apply-shim.html
@@ -80,6 +80,7 @@ Polymer.ApplyShim = (function(){
   // match var(--a, --b) to make var(--a, var(--b));
   var BAD_VAR = /var\(\s*(--[^,]*),\s*(--[^)]*)\)/g;
   var APPLY_NAME_CLEAN = /;\s*/m;
+  var INITIAL_INHERIT = /^\s*(initial)|(inherit)\s*$/;
 
   // separator used between mixin-name and mixin-property-name when producing properties
   // NOTE: plain '-' may cause collisions in user styles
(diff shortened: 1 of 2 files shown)


src/lib/apply-shim.html
```diff
@@ -102,19 +103,51 @@
...
+  function getInitialValueForProperty(property) {
+    if (!measureElement) {
+      measureElement = document.createElement('meta');
+      measureElement.style.all = 'initial';
+      document.head.appendChild(measureElement);
+    }
+    return window.getComputedStyle(measureElement).getPropertyValue(property);
+  }
+
+  function replaceInitialOrInherit(property, value) {
```
sorvell: put on ApplyShim object so it's easier to override in the build.

test/unit/styling-cross-scope-apply.html
```diff
@@ -811,6 +876,35 @@
     CustomElements.takeRecords();
     assertComputed(e, '20px');
   });
+
+  test('Mixins can set "inherit" for a property', function() {
+    if (navigator.userAgent.match(/Trident/)) {
```
sorvell: Add a comment: "IE does not support `inherit`"

test/unit/styling-cross-scope-apply.html
```diff
@@ -811,6 +876,35 @@
...
+      this.skip();
+    }
+    var e = document.createElement('x-inherit');
+    document.body.appendChild(e);
+    CustomElements.takeRecords();
+    assertComputed(e.$.child, 'rgb(0, 255, 0)', 'color');
+    if (!Polymer.Settings.useNativeCSSProperties) {
+      assertComputed(e.$.child, '2px');
+    } else {
+      // this is not supported for the apply shim
```
sorvell: Let's be more explicit so we remember this: "Setting `inherit` on a non-inheriting property is not supported under the apply shim."

test/unit/styling-cross-scope-apply.html
```diff
@@ -811,6 +876,35 @@
...
+      assert.throws(function() {
+        assertComputed(e.$.child, '2px');
+      })
+    }
+  });
+
+  test('Mixins can set "initial" for a property', function() {
+    if (navigator.userAgent.match(/Trident/)) {
```
sorvell: Add a comment: "IE does not support `initial`"

test/unit/styling-cross-scope-apply.html
azakus: Done.

sorvell:
LGTM

PR #3825, opened 2016-07-29
Do not resolve urls with leading slash and other prototcols

### LINKED ISSUE #2448: Polymer 1.0 rewrites links to be relative to source document rather than page.
In polymer 1.0, my elements that have links in the template dom are re-written to be relative to where the element is loaded rather than relative to the current browser location.

That is, if (in my element) I have:

```html
  <a href="/about">About us</a>
```

and the page is at `http://widgets.com/` but the static content is served from `http://cdn.com/path/to/content/`, the link will point to `http://cdn.com/about` instead of `http://widgets.com/about`.

This appears to be occurring due to the urlResolver code:

```
    function resolve(url, ownerDocument) {
        if (url && url[0] === '#') {
            return url;
        }
        var resolver = getUrlResolver(ownerDocument);
        resolver.href = url;
        return resolver.href || url;
    }
    function getUrlResolver(ownerDocument) {
        return ownerDocument.__urlResolver || (ownerDocument.__urlResolver = ownerDocument.createElement('a'));
    }
```

which is getting applied to my element.

A workaround at the moment is to inject the url via data-binding to a computed function, e.g.:

```html
  <a href="{{str('/about')}}">About us</a>
```

with

```js
  str: function(s) { return s; },
```

in my element... but that's just silly.

azakus:
Fixes #2448

### DIFF
--- a/src/lib/resolve-url.html
+++ b/src/lib/resolve-url.html
@@ -11,17 +11,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   (function() {
 
-    // path fixup for urls in cssText that's expected to 
+    // path fixup for urls in cssText that's expected to
     // come from a given ownerDocument
     function resolveCss(cssText, ownerDocument) {
       return cssText.replace(CSS_URL_RX, function(m, pre, url, post) {
-        return pre + '\'' + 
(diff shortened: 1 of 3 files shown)


sorvell:
LGTM.

PR #3834, opened 2016-08-03
Make sure mixins are applied no matter the ordering of definition

### LINKED ISSUE #3833: Ordering of @apply and custom variable definitions
TL;DR: [Why are these different?](https://jsfiddle.net/mgiuffrida/y4hettan/)

```
      :host {
        @apply(--gets-applied);
        --gets-applied: { color: red; };
      }
      :host {
        @apply(doesnt-get-applied);
      }
      :host {
          --doesnt-get-applied: { font-size: 100px; };
      }
```

Filtering should happen before cascading:
> [In order to find the declared values, implementations must first identify all declarations that apply to each element.](https://drafts.csswg.org/css-cascade-4/#filtering)

So that should find all declarations of the mixins.

Then cascading happens. The `@apply`s are dependent on their mixin properties being defined. On the other hand:

> [For the purposes of the cascade, the @apply rule must be treated as if it were replaced by the properties in the custom property set that is the value of the custom property it references.](http://tabatkins.github.io/specs/css-apply-rule/)

So before the cascading, both `--gets-applied` and `doesnt-get-applied` are identified. Shouldn't both `@apply`s work, then?

The motivating example includes an `@apply` from a `<style include>` and then tries to set that mixin from within the element:

https://jsfiddle.net/mgiuffrida/4LL8weq9/

but doesn't work with `useNativeCSSProperties: true`.

azakus:
Fixes #3833

### DIFF
--- a/src/lib/apply-shim.html
+++ b/src/lib/apply-shim.html
@@ -289,12 +289,20 @@ Polymer.ApplyShim = (function(){
     _separator: MIXIN_VAR_SEP,
     transform: function(styles, elementProto) {
       this.__currentElementProto = elementProto;
-      styleUtil.forRulesInStyles(styles, this._boundTransformRule);
-      elementProto.__applyShimInvalid = false;
+      styleUtil.forRulesInStyles(styles, this._boundFindDefinitions);
+      styleUtil.forRulesInStyles(styles, this._boundFindApplications)
+      if (elementProto) {
+        elementProto.__applyShimInvalid = false;
+      }
       this.__currentElementProto = null;
     },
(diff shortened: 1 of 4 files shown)


sorvell:
Can you add a test for this case in `custom-style`?

azakus:
Looks like custom-style is pretty hard to get right with this setup:
```html
<style is="custom-style">
#foo {
  @apply --mixin;
}
</style>
<style is="custom-style">
:root {
  --mixin: { border: 10px solid black; };
}
</style>
```
WDYT @sorvell ?

PR #3841, opened 2016-08-05
Do not insert semicolon when fixing var() syntax

### LINKED ISSUE #3836: var() functions rewritten incorrectly with useNativeCSSProperties=true
Polymer rewrites

> color: var(--empty-prop, var(--my-set-color, --another-empty-prop));

as

> color: var(--empty-prop, var(--my-set-color,var(--another-empty-prop));)

which is invalid syntax.

Polymer also rewrites:

> color: var(--empty-prop, --my-set-color);

which should set `color` to `--my-set-color` (an INVALID value) as:

> var(--empty-prop,var(--my-set-color))

which sets `color` to `var(--my-set-color)` which can be substituted with an actual color when the `--my-set-color` property is set.

Polymer Fiddle: https://jsfiddle.net/mgiuffrida/rhr7m5ad/
Without Polymer: https://jsfiddle.net/mgiuffrida/5wmme0xa/

azakus:
Fixes #3836

### DIFF
--- a/src/lib/apply-shim.html
+++ b/src/lib/apply-shim.html
@@ -219,10 +219,10 @@ Polymer.ApplyShim = (function(){
   }
 
   // fix shim'd var syntax
-  // var(--a, --b) -> var(--a, var(--b));
+  // var(--a, --b) -> var(--a,var(--b))
   function fixVars(matchText, varA, varB) {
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
LGTM

Issue #3418, opened 2016-02-13
Merging of properties between behaviors and element is inconsistent


sorvell:
- effects are added for all behaviors so if you have an `observer` in a behavior and then override that property, you still get the observer. This is probably the right behavior.
- if you `notify: true` the same property multiple times, you get multiple notifications. This is undesirable and needs to be fixed.
- if you `readOnly: true`, then you get the readOnly setter based on first property object found in  prototype first then oldest to youngest behavior (this order is probably wrong, but this is also the answer you get in `getPropertyInfo`). However, deserialization and avoid binding readOnly props based on 'first truthy value in a property object' in order behavior oldest to youngest then proto (standard behavior order for other stuff).

sorvell:
Related to #2493

TimvdLippe:
As a Polymer user, I would expect that a property defined by a child overrides configuration of a parent. However, since observers do not interfere between classes in the inheritance chain, I would expect that every observer of the property in the inheritance chain to be called.

The reason that notify is called multiple times is because [PropertyEffects](https://github.com/Polymer/polymer/blob/master/src/standard/effectBuilder.html#L52-L63) are added for every property in the chain. That means that if a parent and child both define `notify: true`, the effect is added twice. The fix would be to let `_addPropertyEffects` iterate over [`_propertyInfo`](https://github.com/Polymer/polymer/blob/master/src/micro/properties.html#L140) which is a merged array of all properties.

kevinpschaaf:
Related repro from #2242: http://jsbin.com/xebehe/edit?html,console,output

TimvdLippe:
Will this be incorporated into 2.0 while you have the window of opportunity

sorvell:
Since these issues have been addressed in Polymer 2.x we're going to close this issue. The rules are detailed [here](https://github.com/Polymer/polymer/blob/master/lib/mixins/element-mixin.html#L150).

Since the changes that would not be breaking to 1.x are minor performance enhancements, we're electing not to address them.

PR #3856, opened 2016-08-11
Do notify once per property at most

### LINKED ISSUE #1791: Too many property change notifications fired for a property if the property is defined by multiple behaviors
If an element implements `BehaviorA` and `BehaviorB` and both behaviors declare a `foo` property that is `notify:true`, the element will get two `foo-changed` events when the `foo` property is changed.

kaste:
If you have multiple behaviors defining the same property, all setting notify
to true, Polymer just blindly added multiple effects. This resulted in multiple
DOM events sent upwards.

The fix is to just check if there is already a notify effect registered for a
given property, before we register the effect.

Reference issue #3418.
Fixes #1791.
Fixes PolymerElements/paper-input#66

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -118,6 +118,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       return propEffect;
     },
 
+    hasPropertyEffectOfKind: function(model, property, kind) {
+      var fx$ = model._propertyEffects && model._propertyEffects[property];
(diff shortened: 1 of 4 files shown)


dman777:
I hit this bug that this solves with paper-input `on-value-change` firing twice for every key press...whatever became of this since it has not been merged?

TimvdLippe:
Closing per https://github.com/Polymer/polymer/issues/1791#issuecomment-321436929

PR #3857, opened 2016-08-11
Update the CSS transforms to allow colons within Simple Selectors.

jordanyaker:
Updated the CSS transforms to allow colons within Simple Selectors.  This prevents the issues discovered in Polymer#3852.


TimvdLippe:
CSS transformation has been moved to https://github.com/webcomponents/shadycss Please reopen a PR there if you are still encountering your issue.

PR #3858, opened 2016-08-11
Use this instead of self.

### LINKED ISSUE #3840: issue in whenHostOrRootRule function
https://github.com/Polymer/polymer/blob/c1ebf1efba484c525c7cacaeea7fef8a480d0f95/src/lib/style-properties.html

 whenHostOrRootRule: function(scope, rule, style, callback) {
        if (!rule.propertyInfo) {
          **self.decorateRule(rule);  // Not correct. "self" should be "this".**
        }
        if (!rule.propertyInfo.properties) {
          return;
        }

kaste:
`self` refers the global window object here. 

Fixes #3840 

Notes:
1. This code branch is not covered by tests, otherwise it would have thrown early.
2. If we set `browser: false` in eslintrc, eslint would have thrown early as well.

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -287,14 +287,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       whenHostOrRootRule: function(scope, rule, style, callback) {
         if (!rule.propertyInfo) {
-          self.decorateRule(rule);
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Closing per https://github.com/Polymer/polymer/issues/3840#issuecomment-322948721

PR #3866, opened 2016-08-15
Fix for leaking taps, fixes #3704

### LINKED ISSUE #3704: Tap events leak within nodes
### Description

`tap` event leaks to elements underneath an opaque element when the top element gets hidden on tap.
This issue is causing problems like this one https://github.com/PolymerElements/paper-dialog/issues/59
#### Live Demo

http://jsbin.com/kugepa/3/edit?html,output
#### Steps to Reproduce
- open http://jsbin.com/kugepa/3/ in Chrome in iPhone 6 mode (using the developer toolbar)
- press the element on top (in the area intersecting the bottom element)
- notice that the bottom element gets the focus
#### Expected Results

When using `tap` listener, the bottom element gets the focus event
#### Actual Results

The bottom element should not get the focus.
### Browsers Affected

- [x] Chrome (Android Nexus 5x)
- [x] Chrome (in phone simulator mode)

rubenstolk:
Fixes #3704, will cover tests tomorrow.

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -430,10 +430,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   Gestures.register({
     name: 'downup',
-    deps: ['mousedown', 'touchstart', 'touchend'],
+    deps: ['mousedown', 'touchstart', 'touchend', 'focus'],
     flow: {
       start: ['mousedown', 'touchstart'],
-      end: ['mouseup', 'touchend']
+      end: ['mouseup', 'touchend', 'focus']
     },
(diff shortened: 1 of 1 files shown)


azakus:
Hmm, this looks like it would disable all focus events. Was that the intent?

rubenstolk:
@azakus what else do you propose? Users of Polymer-based apps really report this issue as a severe UX-glitch...

azakus:
Turning off focus entirely is not the right way to fix this. It will severely impact users with accessibility needs.

What should happen instead is that the ghost-click prevention code should also capture and block the next focus event.

PR #3871, opened 2016-08-19
Fix binding to field in object with slash in key name

### LINKED ISSUE #3870: Binding to properties having '/' in the key doesn't work
https://jsfiddle.net/tqcrqran/65/

TimvdLippe:
Added the slash character to the variable binding regex.
### Reference Issue

Fixes #3870


TimvdLippe:
Superseded by #4768, we continue the discussion there.

PR #3872, opened 2016-08-19
Do not parse bindings inside script tags

### LINKED ISSUE #3726: Binding notations inside script/style are removed
### Description

When putting a literal binding inside a `<script>` or `<style>`, the bindings are removed and replaced with empty strings. This is especially a prominent issue when using `marked-element` with `<script type="text/markdown">` that has bindings in it (to show Polymer code).
#### Live Demo

https://jsbin.com/wamocuyapu/edit?html,console,output
#### Steps to Reproduce
1. Log a literal with a binding (`{{value}}`)
#### Expected Results

`"Value is: {{value}}"`
#### Actual Results

`"Value is: "`
### Browsers Affected

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11
### Versions

- Polymer: v1.5.0
- webcomponents: v0.7.22

TimvdLippe:
Scripts are now ignored when parsing bindings. This makes it impossible to craft weird functions like `returnStringWithBindingInScript`.
### Reference Issue

Fixes #3726

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -224,7 +224,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // add annotations gleaned from children of `root` to `list`, `root`'s
     // `annote` is supplied as it is the annote.parent of added annotations
     _parseChildNodesAnnotations: function(root, annote, list, stripWhiteSpace) {
-      if (root.firstChild) {
+      if (root.firstChild && root.localName !== 'script') {
         var node = root.firstChild;
         var i = 0;
         while (node) {
--- a/test/unit/bind-elements.html
+++ b/test/unit/bind-elements.html
@@ -55,6 +55,15 @@
       <div id="compound2">
         literal1 {{cpnd1}} literal2 {{cpnd2}}{{cpnd3.prop}} literal3 {{computeCompound(cpnd4, cpnd5, 'literal')}} literal4
       </div>
+      <script type="text/markdown" id="markdownScript">
+        This is a markdown script which shows a Polymer {{binding}}.
+      </script>
+      <script>
+      /*exported returnStringWithBindingInScript*/
+        function returnStringWithBindingInScript() {
+          return 'Value is: {{value}}';
+        }
+      </script>
       <span id="boundWithDash">{{objectWithDash.binding-with-dash}}</span>
   </template>
   <script>
--- a/test/unit/bind.html
+++ b/test/unit/bind.html
@@ -282,6 +282,12 @@ suite('single-element binding effects', function() {
     assert.equal(el.$.boundChild.computedWildcard, 15);
   });
 
+  test('does not process bindings in script tags', function() {
+    assert.equal(el.$.markdownScript.textContent.trim(), 'This is a markdown script which shows a Polymer {{binding}}.');
+    /*global returnStringWithBindingInScript*/
+    assert.equal(returnStringWithBindingInScript(), 'Value is: {{value}}');
+  });
+
   test('binding with dash', function() {
     el.objectWithDash = {
       'binding-with-dash': 'yes'


TimvdLippe:
The failure on Firefox is a result of `<script>` not being executed inside a `<template>` when stamped. See https://github.com/webcomponents/webcomponentsjs/issues/470

Therefore you can craft the incorrect script behavior in Chrome, but you can not in Firefox.

TimvdLippe:
You can see the bug live here btw: https://timvdlippe.github.io/blog/#/b1-carbon-route
It shows `<app-location route="&#123;&#123;route&#125;&#125;"></app-location>` instead of `<app-location route="{{route}}"></app-location>`

jfrazzano:
As a side note, this free-style parsing also becomes an issue with string literals as they are layered and parsed en masse at browser discretion; issues include ui collapse in favor of traditional don-insertions—elements placed on the dom at a previous time, removed, placed in a mobile ui, and then called (textarea was the villain) fall back to their old pointer location, and wipe current screen, solid garbage collection fixes this issue.  On the “in script tag issue”, soon we will all be “stringing” things together, and the location and timing of the parse will be a challenge at best to identify… I sort of have an inkling their is a larger order  of operations issue that `$${`maybe solved by slot assignment???`.italics()}`… Don’t know if this helps or is a valid concern.  Just an issue I have experienced, when carelessly mixing traditional dom placement and parsing with on the fly es6 versions of same.  Items and elements get lost.  I have been mapping for myself and it helps a ton.


Also, Tim, on the library work: There are some simple element related parsing patterns with ternaries that make api building way easier: tabs, numbered sets of questions, all lists, list like columns, and /poorman’s tables/ fall under a one three element set, columned numbered lined animated text, drop down menus, toggle buttons, animated inputs with dropdown menus, dropdown menu cells that can animate to full pages—tested at sets of 50 dropdown items in 2500 inputs sharing the same dropdown, but filling it with different data, fall into the second. 

The ternaries can get dicey if dealing with poor data sets on the way in, but if regularized in a professional setting, they could ad a dynamism, universality, polymorphism, and standards adherence (they endlessly destroy themselves and recreate themselves) that the more declarative, and state aware— i do miss that awareness, but that’s the opposite of the standard—approach offered.

Again, you guys can have anything I have done, and I only want to offer some help, if its possible.

Thanks again. 

No response needed.

Keep up the awesome work.

Sincerely, 

Jason

tomalec:
@TimvdLippe Please note that, unfortunately, my workaround for 

> The failure on Firefox is a result of <script> not being executed inside a <template> when stamped. 

May result in another bug
https://github.com/webcomponents/webcomponentsjs/issues/470#issuecomment-261238583

TimvdLippe:
@tomalec I am sorry but I do not understand what you mean. Could you maybe elaborate on what you want to say?

tomalec:
I have found a workaround for the fact that "a result of `<script>` is not being executed inside a `<template>` when stamped". It's described at https://github.com/webcomponents/webcomponentsjs/issues/470#issuecomment-167591113

Thanks to that, I believe the fix you provided is even more useful. Even though it fails in Firefox, authors would be able to run it there with this workaround.


Then, what i tried to say in [the comment above](https://github.com/Polymer/polymer/pull/3872#issuecomment-261238825), is that today, I found a buggy side-effect of my workaround, which is described at https://github.com/webcomponents/webcomponentsjs/issues/470#issuecomment-261238583 :(.

TimvdLippe:
@tomalec Ah I understand now. Yes that sounds like a good fix, hopefully you can find a solution for the side-effect so we can get this merged in :tada:

TimvdLippe:
@tomalec What is the status of your PR on the polyfill? Since the v1 polyfill is now being actively developed, will the v0 polyfill still be updated to work around this bug?

tomalec:
I haven't made a PR for that, yet. Also, I haven't tested it with v1, as back then I found more bugs in wc.js stopping me to get there. I would try to take another look somewhere in the middle of February.

What's more the ongoing redesign of HTML Imports polyfill behavior https://github.com/webcomponents/html-imports/pull/1 may change the story drastically. Especially, that imported document may no longer be... er... a `Document` ;) https://github.com/webcomponents/html-imports/issues/3

TimvdLippe:
I will update this PR to both ignore styles and scripts, as that is really unintended. I will update the tests as well to work on non-Chrome browsers.

TimvdLippe:
Superseded by #4841

PR #3883, opened 2016-08-24
Variables invalid at compute time should have the value of "unset"

### LINKED ISSUE #3881: CSS variables without a fallback should use unset as the fallback
### Description

CSS variables that have not be defined should be treated as unset. They are currently ignored. See: https://drafts.csswg.org/css-variables/#invalid-at-computed-value-time

This behavior causes problems when using `useNativeCSSProperties` since the behavior differs between browsers that do and do not support custom properties.
#### Live Demo

https://jsbin.com/viyarefiqo/1/edit?html,output
#### Steps to Reproduce
1. Give an element `color: red`
2. Give the element `color: var(--my-nonexistent-color)` in a rule with more specificity or that appears after in the stylesheet
#### Expected Results

The element inherits the color from the parent
#### Actual Results

The element uses `color: red`
### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 9
- [x] Safari 8
- [x] IE 11
### Versions
- Polymer: v1.6.1
- webcomponents-lite: 0.7.22

azakus:
Fixes #3881

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -96,8 +96,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           while ((m = rx.exec(cssText))) {
             // note: group 2 is var, 3 is mixin
             value = (m[2] || m[3]).trim();
-            // value of 'inherit' is equivalent to not setting the property here
-            if (value !== 'inherit') {
+            // value of 'inherit' or 'unset' is equivalent to not setting the property here
+            if (value !== 'inherit' && value !== 'unset') {
               properties[m[1].trim()] = value;
             }
             any = true;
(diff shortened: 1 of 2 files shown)


src/lib/style-properties.html
```diff
@@ -158,12 +158,15 @@
...
+              if (!value) {
+                return prefix + suffix;
+              }
               var propertyValue = self.valueForProperty(props[value], props);
               // if value is "initial", then the variable should be treated as unset
               if (!propertyValue || propertyValue === 'initial') {
                 // fallback may be --a or var(--a) or literal
                 propertyValue = self.valueForProperty(props[fallback] || fallback, props) ||
-                fallback;
+                fallback || 'unset';
```
sorvell: For browsers that do not support `unset` don't we need to place `inherit` or `initial` explicitly?

azakus: That would require knowing which properties are `inherit` and `initial` by default, including all shorthands.
Should we bother?

sorvell: This seems no worse than before so let's defer that until needed.

sorvell:
LGTM after we get the tests passing...

azakus:
Looks like older browsers like Safari 7 don't handle `unset` like I thought :(

azakus:
This is a dead-end for 1.x because we have moved on. This is implemented in webcomponents/shadycss for 2.x

PR #3944, opened 2016-09-12
Fixes #3938

### LINKED ISSUE #3938: `hostAttributes` objects are applied backwards
_From @bicknellr on September 8, 2016 21:2_

`hostAttributes` applied in an element definition are overridden by `hostAttributes` from behaviors they use.

The order of the two calls in this function might be the problem:
https://github.com/PolymerLabs/alacarte/blob/8f237fc635cc56092e83172832bac22ee671c8b0/src/legacy/class.html#L104
And I think the test here is wrong:
https://github.com/PolymerLabs/alacarte/blob/fe0f812a89f356496e1418e44a05cc929c8d652c/test/unit/behaviors.html#L406

_Copied from original issue: PolymerLabs/alacarte#116_

sorvell:
Fixes #3938 

* Separate application of listeners and host attributes. 
* Adds `_applyListeners` and `_ensureAttributes` as override points. 
* Legacy `Polymer` impl ensures attributes subclass before superclass so that a subclass gets first crack as to what attribute value should exist (This matches the behavior of Polymer 1.0).

### DIFF
--- a/src/elements/legacy-element.html
+++ b/src/elements/legacy-element.html
@@ -46,16 +46,23 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         this.created();
       }
 
-      _applyConfigMetaData() {
-        this._applyConfigMetaDataFrom(this.constructor._ownConfig);
+      _applyListeners() {
+        this._applyConfigListeners(this.constructor._ownConfig);
       }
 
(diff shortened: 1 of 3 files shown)


kevinpschaaf:
LGTM

Issue #3930, opened 2016-09-10
`this.$` is undefined on `attached()` with native ShadowDom


kevinpschaaf:
_From @valdrinkoshi on August 26, 2016 22:35_

`x-1` contains a `x-2` in its shadowRoot.
When `x-2`'s attached is invoked, `this.$` is `undefined` (native ShadowDom)

```html
<dom-module id="x-1">
  <template>
    <style>
      :host {
        display: block;
        padding: 10px;
        border: 1px solid;
      }
    </style>
    <x-2>
      <content></content>
      <slot></slot>
    </x-2>
  </template>
  <script>
    Polymer({
      is: 'x-1'
    });
  </script>
</dom-module>


<dom-module id="x-2">
  <template>
    <div id="wrapper">
      <content id="content"></content>
      <slot id="slot"></slot>
    </div>
  </template>
  <script>
    Polymer({
      is: 'x-2',
      attached: function() {
        console.log(this.$);
      }
    });
  </script>
</dom-module>

<x-1>
  <div> hello </div>
</x-1>
```

_Copied from original issue: PolymerLabs/alacarte#96_

kevinpschaaf:
_From @valdrinkoshi on August 26, 2016 22:43_

@sorvell is this related to https://github.com/PolymerLabs/alacarte/issues/71?

kevinpschaaf:
_From @sorvell on August 26, 2016 23:29_

yes

sorvell:
Will be addressed via https://github.com/Polymer/polymer/issues/3933.

kevinpschaaf:
Resolved via https://github.com/Polymer/polymer/pull/3947

PR #3947, opened 2016-09-13
Ready/attached ordering

### LINKED ISSUE #3933: Consider deferring `attached` to emulate 1.0 callback timing
_From @sorvell on August 29, 2016 21:27_

Related issues: #71, #96.

In 1.0, elements were ensured to be readied before being attached. This allowed use of local dom and (at lease naive) measurement in attached.

This was achieved specifically by deferring `attached` until after `ready` if the system tried to call it too soon.

We could retain this guarantee without explicitly deferring `attached` by readying (flushing) an element before attaching its local dom. This would change the guarantees around `ready`. Specifically, it would be unsupported to look at `shadowRoot` in ready.

Perhaps we should simply add back the defer `attached` behavior from 1.0 to ensure as little breakage for users of the 'legacy' api as possible.

_Copied from original issue: PolymerLabs/alacarte#102_

sorvell:
Fixes #3933, #3930.

Ensure dom is attached after is has been readied. In Polymer 1.0, the `attached` call was specially deferred until after `ready`. Instead, we now slightly loosen the definition of `ready` to obviate the need for deferring `attached`. An element has `this.root` in ready and this dom fragment contains all of the the element's "readied" dom; however, it does not yet have `this.shadowRoot`. Directly after calling `ready` the content of `this.root` is added to `this.shadowRoot` and then `this.root` is set to the shadowRoot. 

With this change we've also added an overridable `_attachDom(dom)` method which can be used to place an element's local dom in some other location than its shadowRoot.

This change should maintain enough of the `attached` guarantee that Polymer 1.x had while also naturally playing better with just directly using `connectedCallback` as `Polymer.Element` users shall do.

### DIFF
--- a/src/elements/element.html
+++ b/src/elements/element.html
@@ -197,10 +197,22 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         hostStack.endHosting(this);
       }
 
+      /**
+       * Attach an element's stamped dom to itself. By default,
+       * this method creates a `shadowRoot` and adds the dom to it.
+       * However, this method may be overridden to allow an element
+       * to put its dom in another location.
+       *
+       * @method _attachDom
+       * @param {NodeList} dom to attach to the element.
+       * @returns {Node} node to which the dom has been attached.
+       */
       _attachDom(dom) {
         if (this.attachShadow) {
           if (dom) {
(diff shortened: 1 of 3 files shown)


kevinpschaaf:
LGTM

PR #3951, opened 2016-09-13
Support more expressive `:root` and `html` selectors

### LINKED ISSUE #3902: :root[dir="rtl"] does not work with shimed CSS properties
#### Live Demo

With `useNativeCSSProperties`
http://jsbin.com/zoloru/edit?html,output

Without:
http://jsbin.com/qezehez/edit?html,output

### LINKED ISSUE #3924: The :root -> html change doesn't work in 1.0
_From @notwaldorf on August 23, 2016 19:14_

This does not work in 1.0 (but does in 2.0)
```html
<style is="custom-style">
  html {
 }
</style>
```

This does not work in 2.0 (but does in 1.0)
```html
<style is="custom-style">
  :root {
 }
</style>
```

JSBin: http://jsbin.com/bunihoc/edit?html,output

_Copied from original issue: PolymerLabs/alacarte#83_

### LINKED ISSUE #3991: `:root` and `:host > *` should produce same styling
In Polymer 2.x `:root` has been replaced inside elements with `:host > *`. This is an indication to the custom property shim that any custom properties used in that selector should apply anywhere in the host. Polymer should make this work in 1.x to ease forward compatibility with 2.x. There are a few issues with the current behavior:

* Under ShadowDOM, there is an exception when using `:host > *` (http://jsbin.com/gafare/edit?html,output).

* Under ShadowDOM, `:root` is not converted to `:host > *` (http://jsbin.com/yiwevak/edit?html,output)

* Under ShadyDOM, `:root` is converted to `.my-element-0 .my-element:root`which does not match (http://jsbin.com/vivitec/edit?html,output).

azakus:
Selectors like `:root[dir="rtl"]` and `html.foo` are now supported

Support `html` selector for `custom-style`

Fixes #3902
Fixes #3924
Fixes #3991

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -285,8 +285,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         return {properties: props, key: o};
       },
 
+      _rootSelector: /(?:^:root)|(?::host\s*>\s*\*)/,
+
       _checkRoot: function(hostScope, selector) {
-        return selector.indexOf(':root') === 0 ||
+        return Boolean(selector.match(this._rootSelector)) ||
           (hostScope === 'html' && selector.indexOf('html') === 0);
       },
 
@@ -302,7 +304,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         'html';
         var parsedSelector = rule.parsedSelector;
         var isRoot = this._checkRoot(hostScope, parsedSelector);
-        var isHost = parsedSelector.indexOf(':host') === 0;
+        var isHost = !isRoot && parsedSelector.indexOf(':host') === 0;
         // build info is either in scope (when scope is an element) or in the style
         // when scope is the default scope; note: this allows default scope to have
         // mixed mode built and unbuilt styles.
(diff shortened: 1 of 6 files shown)


src/lib/style-properties.html
```diff
@@ -292,11 +292,14 @@
         if (!rule.propertyInfo.properties) {
           return;
         }
+        var checkRoot = function(hostScope, selector) {
```
sorvell: why is this defined in the function?

src/lib/style-properties.html
```diff
@@ -331,6 +334,9 @@
           // parsedSelector fallback for 'shady' css build
           selectorToMatch = rule.transformedSelector || rule.parsedSelector;
         }
+        if (isRoot && hostScope === 'html') {
```
sorvell: any reason not to always fallback from `transformedSelector` to `parsedSelector`

azakus: `:root` for elements is supposed to match against the element itself, but the rule's transformedSelector will be something like `.x-foo:root`.

Should this be fixed in the selector transformer then? What about `:host > *`?

src/lib/style-properties.html
azakus: Good point, moved it out into a separate function.

src/lib/style-properties.html
sorvell: I think what you have makes sense. I missed the fact that we start with matching the host scope and then only conditionally match the others

And... won't `:root` in an element now be transformed to `:host > *`?

azakus: I think the `:host > *` transform only happens in native ShadowDOM

src/lib/style-properties.html
```diff
@@ -285,6 +285,13 @@
         return {properties: props, key: o};
       },
 
+      _rootSelector: /(?:^:root)|(?::host\s*>\s*\*)/,
```
sorvell: Shouldn't we never have `:root` at this time? also, this is a full (potentially complex) selector so starting form the beginning only is potentially bad and unnecessary, right?

azakus: It is being tested on the parsedSelector, so `:root` will probably be there because we don't replace it before the parsing.

azakus: Yeah, there's no need for `:root` to be first in the selector, removed with 63433c8

sorvell:
@azakus : Let's fix the edge tests before merging

PR #3954, opened 2016-09-14
TypeScript decorators for defining Polymer elements

justinfagnani:
This adds `@customElement()`, `@property()`, `@observe()`, `@query()`, and `@queryAll()` decorator factories which let authors define elements with TypeScript classes using class properties.

`@property` and `@observe` work by writing to a private configuration property and defining a static 'config' getter that returns that property for Polymer.

`@query` and `@queryAll` are intended to replace `$` and `$$` in a way that's somewhat type-safe: by defining a field decorated with @query and with the right type annotation, TypeScript will know the return type of the query, eliminating the need for casts.

`@customElement` simply defines the class, but it also sets `is` for use by Polymer.

I added a test of a compiled TypeScript element. The dependency loading of the test isn't great, but brings relatively little tooling dependencies, it uses <script> tags to load and IMD for a module registry.

### DIFF
--- a/custom_typings/dom.d.ts
+++ b/custom_typings/dom.d.ts
@@ -1,0 +1,11 @@
+declare interface Window {
+  customElements: CustomElementRegistry;
+}
+
+declare class CustomElementRegistry {
+  define(name: string, definition: {prototype: any}): void;
+}
+
+declare interface HTMLElement {
+  shadowRoot: DocumentFragment;
+}
--- a/package.json
+++ b/package.json
@@ -8,6 +8,8 @@
     "test": "test"
   },
   "devDependencies": {
+    "@types/chai": "^3.4.32",
+    "@types/mocha": "^2.2.31",
     "del": "^2.2.0",
     "dom5": "^1.3.1",
     "eslint-plugin-html": "^1.3.0",
@@ -21,10 +23,11 @@
     "polyclean": "^1.2.0",
     "run-sequence": "^1.1.0",
     "through2": "^2.0.0",
+    "typescript": "^2.0.2",
     "web-component-tester": "^4"
   },
   "scripts": {
-    "build": "gulp",
+    "build": "tsc",
     "test": "gulp lint && wct",
     "test-build": "gulp switch && wct && gulp restore"
   },
@@ -37,5 +40,8 @@
   "bugs": {
     "url": "https://github.com/Polymer/polymer/issues"
   },
-  "homepage": "https://github.com/Polymer/polymer"
+  "homepage": "https://github.com/Polymer/polymer",
+  "dependencies": {
+    "reflect-metadata": "^0.1.8"
+  }
 }
--- a/src/typescript/decorators.d.ts
+++ b/src/typescript/decorators.d.ts
@@ -1,0 +1,54 @@
+/// <reference path="../../node_modules/reflect-metadata/Reflect.d.ts" />
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+/**
+ * A TypeScript class decorator that defines a custom element with name
+ * `tagname` and the decorated class.
+ */
+export declare function customElement(tagname: string): (clazz: any) => void;
+export interface PropertyOptions {
+    notify?: boolean;
+}
+/**
+ * A TypeScript property decorator factory that defines this as a Polymer
+ * property.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export declare function property<T>(options?: PropertyOptions): (proto: any, propName: string) => any;
+/**
+ * A TypeScript property decorator factory that causes the decorated method to
+ * be called when a property changes. `targets` is either a single property
+ * name, or a list of property names.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export declare function observe(targets: string | string[]): (proto: any, propName: string) => any;
+/**
+ * A TypeScript property decorator factory that converts a class property into a
+ * getter that executes a querySelector on the element's shadow root.
+ *
+ * By annotating the property with the correct type, element's can have
+ * type-checked access to internal elements.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export declare const query: (selector: string) => (proto: any, propName: string) => any;
+/**
+ * A TypeScript property decorator that converts a class property into a getter
+ * that executes a querySelectorAll on the element's shadow root.
+ *
+ * By annotating the property with the correct type, element's can have
+ * type-checked access to internal elements. The type should be NodeList
+ * with the correct type argument.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export declare const queryAll: (selector: string) => (proto: any, propName: string) => any;
--- a/src/typescript/decorators.ts
+++ b/src/typescript/decorators.ts
@@ -1,0 +1,125 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+// This file requires the reflect-metadata package to be loaded.
+/// <reference path="../../node_modules/reflect-metadata/Reflect.d.ts" />
+
+/**
+ * A TypeScript class decorator that defines a custom element with name
+ * `tagname` and the decorated class.
+ */
+export function customElement(tagname: string) {
+  return (clazz: any) => {
+    clazz.is = tagname;
+    window.customElements.define(tagname, clazz);
+  }
+}
+
+export interface PropertyOptions {
+  notify?: boolean;
+};
+
+/**
+ * A TypeScript property decorator factory that defines this as a Polymer
+ * property.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export function property<T>(options?: PropertyOptions) {
+  return (proto: any, propName: string) : any => {
+    const notify : boolean = options && options.notify;
+    const type = Reflect.getMetadata("design:type", proto, propName);
+    const config = _ensureConfig(proto);
+    config.properties[propName] = {
+      type,
+      notify,
+    };
+  }
+}
+
+/**
+ * A TypeScript property decorator factory that causes the decorated method to
+ * be called when a property changes. `targets` is either a single property
+ * name, or a list of property names.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export function observe(targets: string|string[]) {
+  return (proto: any, propName: string) : any => {
+    const targetString = typeof targets === 'string' ? targets : targets.join(',');
+    const config = _ensureConfig(proto);
+    config.observers.push(`${propName}(${targetString})`);
+  }
+}
+
+
+/**
+ * A TypeScript property decorator factory that converts a class property into a
+ * getter that executes a querySelector on the element's shadow root.
+ *
+ * By annotating the property with the correct type, element's can have
+ * type-checked access to internal elements.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export const query = _query(
+    (target: NodeSelector, selector: string) => target.querySelector(selector));
+
+/**
+ * A TypeScript property decorator that converts a class property into a getter
+ * that executes a querySelectorAll on the element's shadow root.
+ *
+ * By annotating the property with the correct type, element's can have
+ * type-checked access to internal elements. The type should be NodeList
+ * with the correct type argument.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export const queryAll = _query(
+    (target: NodeSelector, selector: string) => target.querySelectorAll(selector));
+
+
+interface Config {
+  properties: {[name: string]: PropertyDefinition};
+  observers: string[];
+}
+
+interface PropertyDefinition {
+  notify?: boolean;
+  type: Function;
+}
+
+function _ensureConfig(proto: any): Config {
+  const ctor = proto.constructor;
+  if (ctor.hasOwnProperty('__polymer_ts_config')) {
+    return ctor.__polymer_ts_config;
+  }
+
+  Object.defineProperty(ctor, 'config', {
+    get() { return ctor.__polymer_ts_config; }
+  });
+
+  const config: Config = ctor.__polymer_ts_config = ctor.__polymer_ts_config || {};
+  config.properties = config.properties || {};
+  config.observers = config.observers || [];
+  return config;
+}
+
+function _query(queryFn: (target: NodeSelector, selector: string) => Element|NodeList) {
+  return (selector: string) => (proto: any, propName: string): any => {
+    Object.defineProperty(proto, propName, {
+      get() {
+        return queryFn(this.shadowRoot, selector);
+      },
+      enumerable: true,
+      configurable: true,
+    });
+  };
+}
--- a/src/typescript/polymer.d.ts
+++ b/src/typescript/polymer.d.ts
@@ -1,0 +1,89 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+interface Constructor<T> {
+  new (...args: any[]): T;
+}
+
+/**
+ * An interface to match all Objects, but not primitives.
+ */
+interface Base {}
+
+/**
+ * A subclass-factory style mixin that extends `superclass` with a new subclass
+ * that implements the interface `M`.
+ */
+type Mixin<M> =
+    <C extends Base>(superclass: Constructor<C>) => Constructor<M & C>;
+
+/**
+ * The Polymer function and namespace.
+ */
+declare var Polymer: {
+
+  /**
+   * The "Polymer function" for backwards compatibility with Polymer 1.x.
+   */
+  (definition: any): void;
+
+  /**
+   * A base class for Polymer custom elements that includes the
+   * `Polymer.MetaEffects`, `Polymer.BatchedEffects`, `Polymer.PropertyEffects`,
+   * etc., mixins.
+   */
+  Element: PolymerElementConstructor;
+
+  ElementMixin:  Mixin<PolymerElement>;
+
+  PropertyEffects: Mixin<PolymerPropertyEffects>;
+
+  BatchedEffects: Mixin<PolymerBatchedEffects>;
+};
+
+declare interface PolymerElementConstructor {
+  new(): PolymerElement;
+}
+
+declare class PolymerElement extends PolymerMetaEffects {
+  static finalized: boolean;
+  static finalize(): void;
+  static readonly template: HTMLTemplateElement;
+
+  ready(): void;
+  updateStyles(properties: string[]): void;
+}
+
+declare class PolymerPropertyEffects extends HTMLElement {
+  ready(): void;
+  linkPaths(to: string, from: string): void;
+  unlinkPaths(path: string): void;
+  notifySplices(path: string, splices: any[]): void;
+  get(path: string|(string|number)[], root: any): any;
+  set(path: string|(string|number)[], value: any): void;
+  push(path: string, ...items: any[]): any;
+  pop(path: string): any;
+}
+
+declare class PolymerBatchedEffects extends PolymerPropertyEffects {
+  // _propertiesChanged(currentProps, changedProps, oldProps): void;
+  // _setPropertyToNodeFromAnnotation(node, prop, value): void;
+  // _setPropertyFromNotification(path, value, event): void;
+  // _setPropertyFromComputation(prop, value): void;
+  // _enqueueClient(client): void;
+  // _flushClients(): void;
+  setProperties(props: any): void;
+}
+
+declare class PolymerMetaEffects extends PolymerBatchedEffects {
+  // _clearPropagateEffects(): void;
+  // _createPropertyFromInfo(name: string, info): void;
+  // _setPropertyDefaults(properties): void;
+}
--- a/test/unit/typescript/decorators.html
+++ b/test/unit/typescript/decorators.html
@@ -1,0 +1,49 @@
+<!doctype html>
+<!--
+@license
+Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<html>
+  <head>
+    <meta charset="utf-8">
+    <script src="../../../../webcomponentsjs/webcomponents-lite.js"></script>
+    <script src="../../../../web-component-tester/browser.js"></script>
+    <!--
+      This test uses IMD for simplicity and less tools, but browserify,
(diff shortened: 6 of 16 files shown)


justinfagnani:
Example class:

```typescript
import {customElement, property, query, queryAll, observe} from '../polymer/typescript/decorators.js';

@customElement('test-element')
class TestElement extends Polymer.Element {

  @property({notify: true})
  foo: number = 42;

  @property({notify: true})
  bar: string = 'yes';

  @query('h1')
  header: HTMLHeadingElement;

  @queryAll('input')
  allInputs: HTMLInputElement[];

  @observe('foo')
  private _fooChanged(newValue: number) {
    console.log(`foo is now: ${newValue}`);
  }

  @observe(['foo', 'bar'])
  private _fooBarChanged(newFoo: number, newBar: string) {
    console.log(`foo is now: ${newFoo}, bar is now ${newBar}`);
  }

}
```

jfrazzano:
Justin, do you think the decorators could be 'blobbed' into assemble-able nodes containing state, through the location object and a merge sort?  
I have been working on a composition strategy using history, and it seems like this would be pretty cool to use 

Thanks in advance. Have always admired your work 

Sincerely, 

Jason

justinfagnani:
@jfrazzano can you describe what you mean a little more? I'm not sure how the location object and merge sort fit into the story here.

These decorators are simply an alternate way to declare the data that would be in the static config getter, so that instead of:

```javascript
class A extends Polymer.Element {
  static get is() { return 'a-element'; }
  static get config() { return {
    properties: {
      'foo': String,
    },
    observers: ['_fooChanged(foo)'],
  }};

  _fooChanged(newFoo) { ... }
}
customElements.define('a-element', A);
```

You write:

```typescript
@customElement('a-element')
class A extends Polymer.Element {
  @property()
  foo: string;

  @observe('foo')
  _fooChanged(newFoo) { ... }
}
```

jfrazzano:
@justinFagani

Am building a large education app, state of user on each use is important, my design focuses on a few small elements that use utf symbols

A dynamic custom form/drop down/context menu that is re-re sugared and case switched on use

This elem is bound often as a searchable dropdown to a composed input that sometimes will load as many as 1000 options--the ui is composed of moving panels and widgets, all composed from the input as text area read only, the menu, text, a dynamic grid, and a header with again utf symbols to replace buttons arrows menu icons tool gears etc

Data is managed by a prototype data manager which is then set on a dataManager behavior as a single instance of shared through the behavior 

Views, are handled by a dashboard object which is a map of all the ui's composed based on prototypes for user types and sub classes of sub object 

Titles and labels are pulled from global "propertyName" to "label" and "label" to "property" name functions all defendant on camel case; 

General API components take the top level name of the user type and role (student profiles) for example and take a view specific identifier from the view type (inputs, review, charts and data, analysis etc). 

Each view is not determined by a natural link path but rather by the user actions, a compiled set of data to beat help kids learn, and data/use records of user specific API preferences. 

This meant needing a way to carry state and sugar, or just really archived history into the app every re-entry 

We were running about 120,000 records with a 1.5 second load, all new additions and any measuring on window size or device change had to be carefully done and passed to children and js objects to maintain the speed. 

We have been playing with building a routing semantic that reflects this data, and the release of the 8/30 document by w3 made our desire to find a way to hold state outside of the polymer compositions and child to parent Dom assembly paramount 

In first run native tests we crawled or just stopped 

It's been 18 months of work so we decided to get creative and use the 650k or so in each location/nav state object to hide our sugar  and state 

If each view is state object only recording the most recent ui change, than a shared master worker that creates more workers to reassemble the composed Dom data can work at any quiet time off thread to assemble state, send it to db store it in indexed database and construct handle forward looking architecture etc. 

It's complicated, for sure, and was an adopted plan only as a result of the wait till connected  (not sure what happens when disconnected--if you fall back into the shadow do you have to start again?we are testing that tomorrow)

At any rate out thinking was to  use blobs to try and circumvent some restrictions, and give us the ability to manage our own composition based on data only we know and which is user and user state specific. 

If the static methods can be filled dynamically via typed script, well... We don't lose all the work, and we have a cool, kinda interesting use case for the undefined stateObject API. 

No rules can sometimes be the best rules

I still am nautilus on no camel case properties. 

Anyway, Thanks for asking and I am sorry I went on so long, but I have been really into the possibilities of polymer since I found it like 16 months ago.  I literally learned to code because of it, and the possibilities it offered which three years of hired developers could not. All we ever heard was "that can't be done".  

Now, it seems more like anything can be done and learned which is more or less what we have been teaching our kids at our center the last ten years. 

I don't want to fail them, and I sure as heck am not letting a little spec make that so

Again, I appreciate all everyone on this project has done.  I cannot imagine the frustration you guys must feel as other parties refuse to play nice on the specs. 

Thanks 

Jason

src/typescript/decorators.d.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+/**
+ * A TypeScript class decorator that defines a custom element with name
+ * `tagname` and the decorated class.
+ */
+export declare function customElement(tagname: string): (clazz: any) => void;
+export interface PropertyOptions {
+    notify?: boolean;
```
rictic: readOnly? observer?

src/typescript/decorators.d.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+/**
+ * A TypeScript property decorator factory that defines this as a Polymer
+ * property.
+ *
+ * This function must be invoked to return a decorator.
```
rictic: Here and elsewhere, I think giving a good and a bad example of use would be better documentation than this caveat. It strikes me as only making sense to people who already understand decorators well enough to not need the documentation. :)

src/typescript/decorators.d.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export declare function property<T>(options?: PropertyOptions): (proto: any, propName: string) => any;
```
rictic: What does the returned function return? Can we say anything more than `any`? Maybe it doesn't matter though

src/typescript/decorators.ts
```diff
@@ -0,0 +1,125 @@
+/**
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
```
rictic: Is it worth noting somewhere that decorators.js and decorators.d.ts are generated from this file, and checked into source control for convenience?

src/typescript/decorators.ts
```diff
@@ -0,0 +1,125 @@
...
+  properties: {[name: string]: PropertyDefinition};
+  observers: string[];
+}
+
+interface PropertyDefinition {
+  notify?: boolean;
+  type: Function;
+}
+
+function _ensureConfig(proto: any): Config {
```
rictic: We might be able to do better than `any` here.

```typescript
interface PolymerElementConfig {
  properties: {
    [propertyName: string]: PropertyDefinition;
  };
  observers: string[];
}

interface PolymerElementPrototype<T> {
  constructor: {
    new () => T;
    __polymer_ts_config?: PolymerElementConfig;
    readonly config: PolymerElementConfig
  }
  
}
```

src/typescript/decorators.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * A TypeScript property decorator factory that causes the decorated method to
+ * be called when a property changes. `targets` is either a single property
+ * name, or a list of property names.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export function observe(targets: string|string[]) {
+  return (proto: any, propName: string) : any => {
```
rictic: why return type of any rather than void?

src/typescript/polymer.d.ts
```diff
@@ -0,0 +1,89 @@
...
+ * An interface to match all Objects, but not primitives.
+ */
+interface Base {}
+
+/**
+ * A subclass-factory style mixin that extends `superclass` with a new subclass
+ * that implements the interface `M`.
+ */
+type Mixin<M> =
+    <C extends Base>(superclass: Constructor<C>) => Constructor<M & C>;
```
rictic: Arg, we talked about this last week I think and I've already forgotten about the limitations of this technique. Or maybe it's a new technique?

src/typescript/polymer.d.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+interface Constructor<T> {
+  new (...args: any[]): T;
```
rictic: Can element constructors take arguments?

I guess so long as they're all always optional?

src/typescript/polymer.d.ts
```diff
@@ -0,0 +1,89 @@
...
+  unlinkPaths(path: string): void;
+  notifySplices(path: string, splices: any[]): void;
+  get(path: string|(string|number)[], root: any): any;
+  set(path: string|(string|number)[], value: any): void;
+  push(path: string, ...items: any[]): any;
+  pop(path: string): any;
+}
+
+declare class PolymerBatchedEffects extends PolymerPropertyEffects {
+  // _propertiesChanged(currentProps, changedProps, oldProps): void;
```
rictic: Are these commented out as a TODO or because there was some problem with typing them?

package.json
```diff
@@ -21,10 +23,11 @@
     "polyclean": "^1.2.0",
     "run-sequence": "^1.1.0",
     "through2": "^2.0.0",
+    "typescript": "^2.0.2",
     "web-component-tester": "^4"
   },
   "scripts": {
-    "build": "gulp",
+    "build": "tsc",
```
rictic: `tsc && gulp`?

test/unit/typescript/decorators.html
```diff
@@ -0,0 +1,49 @@
...
+      webpack, or rollup could probably be used relatively easily to build a
+      single JS bundle
+    -->
+    <script src="../../../../imd/imd.js"></script>
+    <script src="../../../../chai/chai.js" as="chai"></script>
+    <script
+      src="../../../node_modules/reflect-metadata/Reflect.js">
+    </script>
+    <script>
+      define('reflect-metadata', () => {
```
rictic: Any thoughts on what to do about this line and the need for the reflect-metadata script tag in the long run?

test/unit/typescript/test-element.js
```diff
@@ -0,0 +1,72 @@
+/**
```
rictic: Test is failing in firefox: https://travis-ci.org/Polymer/polymer/builds/159791491#L1052

`The module "http://localhost:2000/components/polymer/test/unit/typescript/test-element.js" has not been loaded`

rictic:
@jfrazzano thanks for the enthusiastic endorsement!

I'm not sure I totally follow you though! It kinda sounds like you're talking about a very dynamic element lifecycle? Maybe?

jfrazzano:
Just trying to reuse elements in repeaters change their appearance and some functionality based on user responses and the scope the user is currently in

A silly example is our call center ui. 

I like a searchable field with a dropdown and an icon that chamges to globe for perfect match plus for close ? Mark for uncertain and a red x for a pattern matching break

Now if view one of the call intake asks are you a returning caller in a current service 

A lot of calculation and preparation can be done before the animated page turns 

If those elements (the drop down with its specific data, filtered and sorted from choice one) can't be accessed we lose the speed gain and background responsiveness. 

I guess a lot is doable if you separate the data from the element and move more into the tiered js CSS HTML world 

But we started parsing our own CSS strings from the style text string, faster and easier than CSS 

Won't leak and needs not application save a second parse by polymer or chrome. In essence taking the data as the view made the CSS and repeat updates way smoother 

It's what you guys do, and it is conflict free. 

And from my perspective (new code junkie) it's way easier than leaking selectors, etc and can be done while changing attributes or making the related data changes 

But our project in house. 

We set the rules of use, as far as that goes. What you guys do is impossible, in comparison. 

Really, your work and the compromises you have had to make silently, without appreciation 

You guys are the publics code defenders. More people should realize it.  

This is tough work and you do a great job 
Thank u. 

Sincerely 

Jason.

robwormald:
@justinfagnani really cool stuff. this should in theory allow for a really, really slick angular2 integration with polymer elements - we should chat about how we're using decorators and TS with metadata 👍

custom_typings/dom.d.ts
```diff
@@ -0,0 +1,11 @@
+declare interface Window {
+  customElements: CustomElementRegistry;
```
robwormald: you might see about getting these added to typescript's lib files, so a user could do something like in their tsconfig
```
{ "types" : [ "es2015", "dom", "customElements" }
```
cc @DanielRosenwasser

jfrazzano:
This is just a curiosity: how does type script interact with wai-aria property objects roles etc? is there a cross over of roles.  No reason to thing there is.  Just curious.

custom_typings/dom.d.ts
justinfagnani: Definitely, and these typings are anemic too. Since they're part of the DOM spec, I think they should just go into "dom".

src/typescript/polymer.d.ts
justinfagnani: This is a slight variation, with the`Mixin` type, but the limitations are the same: it casts to an intersection type, and requires a separate declaration of the interface.

src/typescript/polymer.d.ts
justinfagnani: I need to ask if they're private or protected.

test/unit/typescript/decorators.html
justinfagnani: I want to try a webpack setup rather than IMD. That might improve things here. As for reflect-metadata, I need to learn a lot more about it...

package.json
justinfagnani: The gulp build is out of date and didn't do anything

justinfagnani:
@robwormald thanks, I'd love to talk about decorators and integration!

jfrazzano:
It will be significantly more difficult to slide meta-data in now, I was working through the element factories I had used and their meta-data behavior; if the elements can’t become elements until they are “connected”, its a challenge;  The one good thing is that once the element is registered it need not be registered again.  

It could be attached pulled, and reused. 

But the aria spec says the elem can’t change roles.  Does role approximate type? Is the native input, which we transform and which has historically been a chimera, violate the spec? And is that why apple won’t use is? It could be changed as a second script could be “linked” to the document.  The link though covers that eventuality, as “links are self explanatory based in the text…”  Whatever the text is that is the link role…

Interface is the word I had originally chosen too, but its a tricky one as I go through spec requirements, it almost eliminates any semantic state data.

Sort of a tough call on how to move things.  Definitely can see the difficulties.

Sincerely, 

Jason

custom_typings/dom.d.ts
DanielRosenwasser: That looks reasonable to me. I think we'd be willing to take that. Any objections @mhegazy? Would this go in https://github.com/Microsoft/TSJS-lib-generator @zhengbli?

zhengbli: I agree, this looks like a reasonable addition.

jfrazzano:
How is the type accessed? Access to the registry cache to the author agent, particularly of an author of a non-application role `widget` could lesson some of the typing weight.  If the author agent is in control of his state and role and has access to the registry cross browser, well— the onus should shift,  and not leak back.

custom_typings/dom.d.ts
justinfagnani: I see that that tool pulls DOM definitions from Edge. Does Edge pick up IDL for features it hasn't implemented yet, like Shadow DOM and Custom Elements?

jfrazzano:
I am asking if a descriptor wouldn’t be helpful in meeting standard, upholding a dedication to semantic web and user state, all while letting the person running the app, handle the details. Since custom elements are hanging around till connect, so like three plus call backs why not do it there.

justinfagnani:
@jfrazzano your discussion seems to be more about Polymer and HTML elements in general, rather than about this PR. Maybe the polymer-dev@googlegroups.com list would be a better venue for this, and you might find some people interested in picking up the thread with you. Let's try to keep the comments here on-topic. Thanks!

custom_typings/dom.d.ts
mhegazy: We will need to check with the Edge folks about this and get back to you.

mhegazy: but either ways we could start adding them to the library and get better/more accurate declarations as Edge picks them up.

jfrazzano:
the type had spec specific meaning, to include it off thread, is.. well.. it’s relevant, as silent inclusion represents acceptance in many venues and becomes an avenue for later issues.  But, I said what I had to say.  I just practiced enough law in the last 20 years to know silence is anything but golding an international disability standards.

custom_typings/dom.d.ts
justinfagnani: Anything I can do to help?

mhegazy: A [PR](https://github.com/Microsoft/TSJS-lib-generator#contribution-guidelines) would be appreciated. We will need to add the types to [addedTypes.json overrides](https://github.com/Microsoft/TSJS-lib-generator/blob/master/inputfiles/addedTypes.json).

src/typescript/decorators.d.ts
```diff
@@ -0,0 +1,54 @@
+/// <reference path="../../node_modules/reflect-metadata/Reflect.d.ts" />
```
bicknellr: Is [this](https://github.com/rbuckton/ReflectDecorators) the spec defining `Reflect.getMetadata`?

src/typescript/decorators.ts
justinfagnani: Yeah, or possible not checking them in if we publish a special branch for Bower. @azakus ?

src/typescript/decorators.ts
justinfagnani: Let me check on this, since this code has lived in several places. I thought I had gotten type warnings if I didn't have the signature right...

src/typescript/decorators.d.ts
justinfagnani: Maybe? Typescript decorators have diverged from the current proposals for decorators, so "spec" is probably a loose term at this point.

bicknellr: Ok, and its use here is a requirement of interfacing with TypeScript, right? I'm completely opposed to that 'spec'; I believe it's very poorly thought out. But if it's *required* by TypeScript, then it makes sense to include in a PR for interfacing with it.

src/typescript/decorators.ts
gertcuykens: Doh! Makes me realize I understand even less than I previously thought I understood haha. I was looking at this file with, "what the ... is this js thing doing in this typescript map"

gertcuykens:
We need to discuss `super.connectedCallback();` before this merge I think?

https://github.com/Polymer/polymer/tree/2.0-preview#20-es6-class-based-syntax

image

example: https://github.com/gertcuykens/hello-world/blob/2.x/hello-world.html

src/typescript/decorators.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * A TypeScript property decorator factory that defines this as a Polymer
+ * property.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export function property<T>(options?: PropertyOptions) {
+  return (proto: any, propName: string) : any => {
+    const notify : boolean = options && options.notify;
```
gertcuykens:
```ts
const notify : boolean = options && options.notify;                                                                   
             ~~~~~~                                                                                                          
                                                                                                                             
tmp/decorators.ts(36,11): error TS2322: Type 'boolean | undefined' is not assignable to type 'boolean'.                      
  Type 'undefined' is not assignable to type 'boolean'. 
```
const notify = (options && options.notify) as boolean;

src/typescript/decorators.ts
```diff
@@ -0,0 +1,125 @@
...
+  config.properties = config.properties || {};
+  config.observers = config.observers || [];
+  return config;
+}
+
+function _query(queryFn: (target: NodeSelector, selector: string) => Element|NodeList) {
+  return (selector: string) => (proto: any, propName: string): any => {
+    Object.defineProperty(proto, propName, {
+      get() {
+        return queryFn(this.shadowRoot, selector);
```
gertcuykens:
```ts
return queryFn(this.shadowRoot, selector);                                                                       
                           ~~~~                                                                                              
                                                                                                                             
tmp/decorators.ts(116,24): error TS2683: 'this' implicitly has type 'any' because it does not have a type annotation.        
   
```

src/typescript/decorators.ts
```diff
@@ -0,0 +1,125 @@
...
+ * A TypeScript property decorator factory that converts a class property into a
+ * getter that executes a querySelector on the element's shadow root.
+ *
+ * By annotating the property with the correct type, element's can have
+ * type-checked access to internal elements.
+ *
+ * This function must be invoked to return a decorator.
+ */
+export const query = _query(
+    (target: NodeSelector, selector: string) => target.querySelector(selector));
```
gertcuykens:
```ts
(target: NodeSelector, selector: string) => target.querySelector(selector));                                            
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~                                              
                                                                                                                             
tmp/decorators.ts(71,3): error TS2345: Argument of type '(target: NodeSelector, selector: string) => Element | null' is not a
ssignable to parameter of type '(target: NodeSelector, selector: string) => Element | NodeList'.                             
  Type 'Element | null' is not assignable to type 'Element | NodeList'.                                                      
    Type 'null' is not assignable to type 'Element | NodeList'.  
```

joh1234:
Update README
https://goo.gl/JBljdR

rictic:
We can soon drop the declarations of shadow dom and custom elements APIs here, as they've landed upstream at https://github.com/Microsoft/TSJS-lib-generator/pull/151 and https://github.com/Microsoft/TSJS-lib-generator/pull/150

src/typescript/decorators.d.ts
43081j: I don't think decorator factory functions return anything, so this can probably be `void`

43081j:
You can drop the custom typings now (`dom.d.ts`) as they have landed in stable.

You can also make all your decorator factories return void functions rather than `any`.

A suggestion for you, too, is to add a `computed` decorator:
```
@computed
myComputedProperty(foo, bar) {
```

Where the actual computed string is produced using the method's parameters or by passing them in (`@computed("foo,bar")`).

Reason being, we maintain strong typing for a computed property then, whereas this would result in some weakness:
```
@property({ type: Boolean, computed: '_computeMyComputedProperty(foo, bar)' })
public myComputedProperty: boolean;
```

Because typescript wouldn't know that `_computeMyComputedProperty` returning anything but a `boolean` is an error...

src/typescript/decorators.ts
43081j: is there a reason we don't just have two functions? seeing as they do different things (querySelector vs querySelectorAll)

src/typescript/polymer.d.ts
43081j: can you make use of typescript 2.2's mixin support here?

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html

nbransby:
With 2.0 RC now released is there plans to release typescript support anytime soon?

gertcuykens:
Note that typescript will always be supported in the sense of if you know what you are doing you can use this PR as a example and make it work yourself. But it's the whole amd modeling / requirejs part needed to make this work that make it so complex that everybody else in the polymer team would not even touch typescript. My guess is they are all waiting for es6 modules to come out in browsers before one would jump on the typescript boat. For me browser vendors are keeping typescript hostage if they don't implement modules.

justinfagnani:
An update on this PR: Now that 2.0 is in a more final state and we have better support in the Analyzer, I'm working on a .d.ts generator for Polymer packages and have some initial results for the core library here: https://github.com/Polymer/polymer/pull/4377/files

I'll be able to get back to the generator in a week or so.

As for the decorators, they don't need to change much, but my current thinking is to put them into a separate repo, since they're not required to use Polymer 2.0 and decorators are still non-standard, with TypeScript's version not aligned with the latest TC39 proposal.

gertcuykens:
Separate polymer-decorators also fine for me, just try your best to convince your colleagues typescript is more important to make web applications then they think it is :P I can't even image writing plain javascript anymore without a compiler telling me nope bad idea all the time. Plain javascript is almost the same as giving me a uglified minification file and then ask me to change a property in it... You just don't use plain javascript anymore like you don't use assembler anymore. 

> TypeScript's version not aligned with the latest TC39 proposal.

Wait what? And what did Microsoft had to say about that? As far as I can tell they look the same? Maybe some details or something TC39 changed? No doubt Typescript will adapt to it for sure I mean its their number one priority to be standard compatible

lanthaler:
Any update on this?

justinfagnani:
@lanthaler I'll get back to this after the Polymer Summit

lanthaler:
Awesome. That's great news!

justinfagnani:
Closing in favor of https://github.com/Polymer/polymer-decorators and a future PR that will be generated from https://github.com/Polymer/gen-typescript-declarations

PR #3964, opened 2016-09-19
3430 - ie memory leak fixes - disable event caching, fixed resolver u…

### LINKED ISSUE #3430: Polymer will lead to memory leaks in Internet Explorer 11.
Polymer will lead to memory leaks in Internet Explorer 11. 
1. e.g  "open site "https://www.polymer-project.org/1.0/" in  Internet Explorer 11
2. then refresh page some times ， observe the system ram， would raise up constantly

bbsteventill:
### Reference Issue

…rl adding to root doc, and weak map ie issues

### DIFF
--- a/src/lib/bind/accessors.html
+++ b/src/lib/bind/accessors.html
@@ -23,7 +23,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         value = value === undefined ? this[source] : value;
         event = event || Polymer.CaseMap.camelToDashCase(source) + '-changed';
         this.fire(event, {value: value},
-          {bubbles: false, cancelable: false, _useCache: true});
+          {bubbles: false, cancelable: false, _useCache: Polymer.Settings.eventDataCache || !CustomElements.isIE});
       },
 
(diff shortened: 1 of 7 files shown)


bbsteventill:
I signed it!

Bloomberg has added me to their cla

reply@reply.github.com " data-digest="From: reply@reply.github.com " style="">

domrongpon-tanpaibul:
Our team found this problem and this patch helps to solve the problem (partially).

Another issue that causes the leak is HTML Import and we do a workaround by copying components.html to the main page.

Using both solutions together can significantly reduce the leak in IE, even though not totally fixed. 

If this patch can be released at earliest, that would be great.

bbsteventill:
FYI https://github.com/webcomponents/webcomponentsjs/pull/644 fixes the memory leaks within web components

domrongpon-tanpaibul:
thanks @bbsteventill !

sorvell:
@azakus can you please review this? Thanks.

jay8t6:
@azakus @sorvell when is the next release coming out? I'd love to see the performance improvement in IE.

Issue #3377, opened 2016-02-05
notifySplices fails to notify about changes in array of primitives


tomalec:
I have an array of Strings that's being changed:

``` js
var model = {
    people: [
        "Alice",
        "Bob"
    ]
};
// ...
model.people[0] = "Anna";
```

So, I would like to notify my `dom-bind` about this change using `notifySplices`:

``` js
domBind.notifySplices('organization.people', [{
    index: 0,
    removed: removed,
    addedCount: 1,
    object: model.people,
    type: 'splice'
}]);
```

Live example here
http://jsbin.com/kocowalimi/1/edit?html,console,output

In result, entire content of `dom-repeat` disappears and throws an error:

```
"TypeError: Cannot read property 'isPlaceholder' of undefined
    at dom-repeat.Polymer._detachInstance (http://polygit.org/components/polymer/polymer.html:4415:10)
    at dom-repeat.Polymer._detachAndRemoveInstance (http://polygit.org/components/polymer/polymer.html:4430:17)
    at dom-repeat.Polymer._applySplicesArrayOrder (http://polygit.org/components/polymer/polymer.html:4406:6)
    at dom-repeat.Polymer._render (http://polygit.org/components/polymer/polymer.html:4259:6)
    at Object.Debouncer.complete (http://polygit.org/components/polymer/polymer-mini.html:2099:15)
    at boundComplete (http://polygit.org/components/polymer/polymer-mini.html:2078:6)
    at Object.Polymer.Async._atEndOfMicrotask (http://polygit.org/components/polymer/polymer-mini.html:2055:1)
    at MutationObserver.window.MutationObserver.observe.characterData (http://polygit.org/components/polymer/polymer-mini.html:2070:15)"
```

If I change string names to objects like `{name: "Alice"}..` it works, but I cannot change my database models, just to fit this UI.

d1manson:
I'm trying to understand `notifySplices` myself, the docs aren't super clear.  Reading the code for `dom-repeat` is vaguely helpful, but I haven't found everything I needed yet.

@tomalec - I noticed one actual mistake in your jsbin - you need to pass an array as the second arg to `notifySplices`, rather than a raw object.   But that alone doesn't make it work. 

Here's a link to my edit of your jsbin - http://jsbin.com/juwiyodefu/1/edit?html,console,output. For simplicity I made `people` a property of the model directly, rather than a child of `organisation`, though that also didn't help.  However I did find the following:

```javascript
// This does work...
[...]

Issue #3062, opened 2015-11-20
Duplicate strings in array while using 'dom-repeat' throws error: 'Uncaught Error: unexpected key undefined'


chigur:
Following is the element I made which throws an error mentioned in the title.

``` html
<link rel="import" href="../polymer/polymer.html" />

<dom-module id="data-list">
    <template>
        <template is="dom-repeat" items={{data}}>
            <div>{{item}}</div>
        </template>
    </template>
    <script>
        Polymer({
            is: 'data-list',

            properties: {
                data: {
                    type: Array,
                    value: function() {
                        return ['Random', 'Random'];
                    },
                    notify: true
                }
            },

            attached: function() {
                this.splice('data', 0, 1);
                this.splice('data', 0, 1);
            }
        });
    </script>
</dom-module>
```

This is caused by storing values of an array as keys in the object [`pmap`](https://github.com/Polymer/polymer/blob/master/src/lib/collection.html#L47).

fredefox:
I have noticed this too. Only after seeing this issue did I realize that it happens when you use non-unique values in the array. Do you have a work-around?

chigur:
@fredefox Using objects instead of primitive values should do the trick.

fredefox:
Right you are! Thank you.

chigur:
I would really love to contribute a PR for this bug if somebody guides me on how to go about it. It would be my first ever PR on GitHub :).

MeTaNoV:
Just go ahead by following this: https://github.com/Polymer/project/blob/master/Contributing.md

StokeMasterJack:
I too am getting this error. I am using an array of object which may contain duplicates. It happens on this line:

this.splice('selected', indexToRemove, 1);

and it seems t happen when deleting a duplicate for the second time.

electrovir:
I have an issue related to this but I am not getting an error thrown.  Instead, when I use dom-repeat to bind an array with non-unique strings to input boxes, editing the text inside any input besides the first (using normal user interaction) changes the value in the first input box.

TimvdLippe:
This has been fixed in Polymer 2: http://jsbin.com/fuqarizedo/edit?html,console,output

Issue #1913, opened 2015-06-19
dom-repeat: Duplicate values result in unsynchronized data


kevinpschaaf:
The collection-based key abstraction currently assumes unique array element identity.

Problem can be observed here (type `bbbbb` in the top input, and observe data out of sync):
http://jsbin.com/rabize/edit?html,output

zmoshansky:
+1, this would be really nice to have.
The workaround as suggested https://github.com/Polymer/polymer/issues/2667#issuecomment-167875621 is to wrap primitives in an object. 
`['analgesic', null, 'analgesic' ]` -> `[{k: 'analgesic'}, {k: null}, {k: 'analgesic'}]`

PSG-Luna:
I've come across this issue as well, it's quite a big problem when having an array of any non object type

rwestlund:
I just ran into this with 1.4.0, but I don't believe I have duplicate values in my string array.

I have this tap handler:
```
addStep: function() {
    this.push("recipe.directions", "");
}
```
and I'm using paper-inputs for each `recipe.directions` element within dom-repeat.  I'm adding text to each step before I add the next one, so I don't think there are ever duplicate values.

I found the following workaround:
```
this.push("recipe.directions", new String());
```

This likely works because `typeof new String() === 'object'`

Pupix:
@kevinpschaaf are you still working on this? Would be nice to not have such a constrain on value uniqueness, especially when talking about binding on form inputs.

TimvdLippe:
Original JSBin is now working correctly with Polymer 2, thus I am closing this issue.

PR #3970, opened 2016-09-20
Eliminate collection. Fixes #3946

### LINKED ISSUE #3946: dom-repeat doesn't stamp correctly
### Description
`dom-repeat` should update the values when setting `t.items = t.items` and update the dom with the new values. This currently happens only on the first setting of the array.

#### Live Demo
http://jsbin.com/rumiko/2/edit?html,console,output

#### Steps to Reproduce

1. click `add` button

#### Expected Results

Dom should have 2 paragraphs with `item 0`, `item 1`...as many times as add was clicked

kevinpschaaf:
### Reference Issue
Fixes #3946: dom-repeat doesn't stamp correctly (which is a `2.x`-specific issue; see below for related `1.x` issues that this will also solve).

### Description
This is a significant change that eliminates the `Polymer.Collection` abstraction from Polymer 2.x.  

This abstraction aimed to assign unique keys to Arrays via a behind-the-scenes store associated per array via WeakMap, along with integration into Polymer's notifying array mutation API's.  The abstraction allowed for two-way binding into elements repeated from arrays by virtue of stable keys providing unique paths to bound array items, with efficient updating of the rendered items via splices.  

However, the system was inherently complex and in the end the theoretical efficiency gained did not pay for itself in terms of the cognitive overhead for things like key-based paths (`items.#5.value`), and a long tail of issues related to lack of support for duplicate array items and primitives via this approach.  Likewise, as #3946 pointed out, it also was fundamentally incompatible with a major goal of the 2.0 release, which is to eliminate the _requirement_ to use `set`/`notifyPath` (although the API remains and will still be more optimal in cases) by avoiding use of (default) object identity in Polymer's dirty-checking, since Collections were associated to arrays via object identity.

This change results in `dom-repeat` simply reacting to `items` or `items.splices` by iterating the items array and re-assigning item, index, and origIndex (index of item in original array prior to applying user sort/filter) for each change.  This approach was already basically best-in-class on the [dbmon benchmark](http://mathieuancelin.github.io/js-repaint-perfs/) which always took the "full refresh" code path, and still leaves room for more straightforward optimizations in the future.

Path notifications are still handled, but using the original index in the path where previously key was used (which is possible since the new rendering stores the `itemsIdx`<->`instIndex` mapping upon each change to `items`).  The main breaking changes are that:
a. `Polymer.Collection` is now removed, and all array notifications are now sent directly with the index rather than the key
b. `array.splices` observers will still receive an object that only contains `indexSplices`, but will no longer contain `keySplices`

Note, this change also resolves (or makes moot) the following 1.x issues; it is likely that these will only be resolved in the 2.x branch due to the nature of the breaking changes required to resolve:
* #1913: dom-repeat: Duplicate values result in unsynchronized data
* #3062: Duplicate strings in array while using 'dom-repeat' throws error: 'Uncaught Error: unexpected key undefined'
* #3377: notifySplices fails to notify about changes in array of primitives


sorvell:
LGTM

PR #3974, opened 2016-09-21
Add support for slot->content transformation.

kevinpschaaf:
This PR is (a start) to enable writing 1.0 compatible elements that also target the 2.x compatibility layer.  Specifically, it transforms `<slot name="foo">` to `<content select="[slot=foo]">`, enabling users to write to slot and fallback to content in 1.0.

Need to bikeshed opt-in attribute (currently "auto-content")


sorvell:
LGTM

PR #3984, opened 2016-09-22
Ensure template is not upgraded multiple times. Fixes #3977

### LINKED ISSUE #3977: fwdHostPropToInstance synchronously called
This obliges user to declare variable then assign it. E.g. click twice on the `stamp` button, notice the error on the second click

```html
<dom-module id="x-sample">
  <template>
    <template id="to-stamp">
      <div opened$="[[opened]]">[[opened]]</div>
    </template>
    <div>
      opened is [[opened]]
    </div>
    <slot></slot>
  </template>

  <script>
    const templatizer = new Polymer.Templatizer();

    class XSample extends Polymer.Element {
      static get is() {
        return 'x-sample';
      }

      static get config() {
        return {
          properties: {
            opened: {
              type: Boolean,
              value: false
            }
          }
        };
      }

      stampNewInstance() {
        // Seems like fwdHostPropToInstance is invoked synchronously.
        // Declaring the variable and then assigning would fix the issue...
        const instance = new(templatizer.templatize(this.$['to-stamp'], {
          fwdHostPropToInstance: function (host, prop, value) {
            instance && instance.forwardProperty(prop, value, host);
          }
        }))(this);
        this.appendChild(instance.root);
      }
    }

    customElements.define(XSample.is, XSample);
  </script>
</dom-module>

<button onclick="sample.stampNewInstance()">stamp</button>
<x-sample id="sample"></x-sample>

kevinpschaaf:
### Reference Issue
Fixes #3977

`DataTemplate.upgradeTemplate` was running multiple times when calling `templatizer` multiple times for the same template, which runs the property effects constructor which initializes `__data` and `__dataPending` with info from `__dataProto`. This was leaving `__dataPending` in an un-flushed state on the second call.


sorvell:
LGTM

PR #3987, opened 2016-09-23
Check value not ownProperty to save accessor value. Fixes #3972 and #…

### LINKED ISSUE #3972: Methods defined in behaviors do not exist when used in an HTML binding
```html
<html lang="en">
<head>
  <script src="../webcomponentsjs/webcomponents-lite.js"></script>
</head>
<body>
  <dom-module id="my-element">
    <template>
      <style>
      span[foo=yes] {
        font-weight: bold;
      }
      </style>
      <span foo$="[[_computeFoo(bar)]]">Hello world</span>
    </template>

    <script>
    var MyBehavior = {

      properties: {
        bar: {
          type: String,
          value: 'yes'
        }
      },

      _computeFoo: function(bar) {
        return bar;
      }

    };

    Polymer({
      is: 'my-element',

      behaviors: [MyBehavior]
    });
    </script>
  </dom-module>

  <my-element></my-element>
</body>
</html>
```

## Expected
`_computeFoo` is defined on `<my-element>`, and `<span>` has attribute `foo="yes"`.

## Actual
`_computeFoo` is `undefined`, and `<span>` is missing the attribute.

kevinpschaaf:
### Reference Issue
Fixes #3972 and #3972

The `hasOwnProperty` check was not robust against mixins/behaviors adding methods further down the prototype chain.

### DIFF
--- a/src/properties/property-accessors.html
+++ b/src/properties/property-accessors.html
@@ -17,13 +17,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   'use strict';
 
   function saveAccessorValue(model, property) {
-    if (model.hasOwnProperty(property)) {
+    let value = model[property];
+    if (value !== undefined) {
       if (!model.__dataProto) {
         model.__dataProto = {};
       } else if (!model.hasOwnProperty('__dataProto')) {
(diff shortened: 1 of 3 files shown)


sorvell:
LGTM

PR #3996, opened 2016-09-26
`Polymer.dom(element).observeNodes(callback)`: ensure callback is cal…

sorvell:

### DIFF
--- a/src/utils/polymer.dom.html
+++ b/src/utils/polymer.dom.html
@@ -149,7 +149,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // update cache
       this.effectiveNodes = newNodes;
       if (info.addedNodes.length || info.removedNodes.length) {
-        this.callback(info);
+        this.callback.call(this.target, info);
       }
     }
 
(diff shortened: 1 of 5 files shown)


src/utils/polymer.dom.html
```diff
@@ -244,7 +244,7 @@
       let list = [];
       for (let i=0, l=c$.length, c; (i<l) && (c=c$[i]); i++) {
         if ((c.nodeType === Node.ELEMENT_NODE) &&
-            Polymer.Utils.matchesSelector(this.node, selector)) {
+            Polymer.Utils.matchesSelector(c, selector)) {
```
bicknellr: If you're going down the same rabbit hole as me right now, I think https://github.com/Polymer/polymer/blob/fd052da7f8e4aa7432a9ed36e55345d77e2aa0f3/src/utils/polymer.dom.html#L24 is going to end up being your next change. :) `node.childNodes` -> `(node.root ? node.root : node).childNodes` or something similar.

sorvell: I believe we shouldn't need to check `.root` there unless there is a Shady DOM bug. Let's deal with that separately if so.

bicknellr: Ok, I'll open a bug.

bicknellr: Polymer/polymer#4000

PR #3997, opened 2016-09-26
Fix IE style cache performance

### LINKED ISSUE #3965: Huge performance degradation between Polymer 1.4 and 1.6 on IE11
Hi,

We're seeing a huge performance degradation between Polymer 1.4.x and Polymer 1.6.x for a specific component of ours (GE Digital's Polymer-based design system): https://github.com/PredixDev/px-data-table

This only happens in IE11, I've built a couple of examples to show this below. Behind each of those github pages is a minimal demo showing the behaviour.

#### Live Demo
Fast: https://mdwragg.github.io/polymer14-data-table/
Slow: https://mdwragg.github.io/polymer16-data-table/

and repos:

https://github.com/mdwragg/polymer14-data-table
https://github.com/mdwragg/polymer16-data-table

#### Steps to Reproduce
Load up the `polymer14` url, note how it is snappy on IE11, now load up the `polymer16` URL and note how slow it is...

#### Expected Results
Performance should be comparable between the two.

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [x] IE 11

Thanks!
Martin.

azakus:
Continue workaround from 80be0df for `@media` staleness, but only refresh
the stylesheet if it came from the cache and has `@media` queries.

Fixes #3965

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -494,9 +494,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             }
           // shady and cache hit but not in document
           } else if (!style.parentNode) {
+            if (IS_IE && cssText.indexOf('@media') > -1) {
+              // @media rules may be stale in IE 10 and 11
+              // refresh the text content of the style to revalidate them.
+              style.textContent = cssText;
+            }
             styleUtil.applyStyle(style, null, element._scopeStyle);
           }
-
         }
         // ensure this style is our custom style and increment its use count.
         if (style) {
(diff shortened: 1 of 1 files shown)


azakus:
Anecdotally tested with https://mdwragg.github.io/polymer14-data-table/ and https://mdwragg.github.io/polymer16-data-table/ vs a checkout with this branch.

sorvell:
LGTM

PR #4001, opened 2016-09-26
Briefly explain how to split element definition

dandv:


PRIMER.md
```diff
@@ -467,7 +467,8 @@ Example:
 We say that an element definition has an imperative and declarative portion. The imperative
 portion is the call to `Polymer({...})`, and the declarative portion is the `<dom-module>`
 element. The imperative and declarative portions of an element's definition may be placed
-in the same html file or in separate files.
+in the same HTML file or in separate files. They are matched mased on the `id` and `is`
```
kevinpschaaf: s/mased/based

Also suggest clarifying:
> They are matched based on the `dom-module`'s `id` having the same value as the element's `is` property.

PR #4007, opened 2016-09-27
Hashtag spelt wrong (how embarrassing)

dllndv:
This isn't anything to do with code, I just noticed that your hashtag #usetheplatform was spelt wrong. I hope this will save you from ridicule and thus boost your self esteem and be more motivated to write great code!

And sorry, I had to close the previous PR and resubmit this one. CLA was giving me trouble, and this is my first pull request with anything Google!


sorvell:
LGTM, we feel better already. Thanks!

PR #4027, opened 2016-10-01
Update gulp build to transpile to es5.

kevinpschaaf:
For now, uses `polymer-build` and builds to `dist/bundled` and `dist/unbundled`, and links _both_ `polymer.html` and `polymer-element.html` to `dist/bundled/polymer.html`.  Source can be restored with `gulp restore-src`.

To run natively in Chrome, currently requires loading `custom-elements/src/native-shim.js`.

Decision on what to actually distribute in releases is tbd.  Ideally if we distribute bundled versions, our two main entry points (`polymer.html` and `polymer-element.html`) are built to have a shared bundle (e.g. `polymer.html` would import `polymer-element.html`).


gulpfile.js
```diff
@@ -11,112 +11,146 @@
...
+   .pipe(project.splitHtml())
+   // add compilers or optimizers here!
+   .pipe(gulpif(/\.js$/, babel({presets: ['es2015']/*, plugins: ['external-helpers']*/})))
+   .pipe(gulpif(/\.js$/, uglify()))
+   .pipe(project.rejoinHtml())
+   .pipe(htmlmin({removeComments: true}));
+
+ // process dependencies
+ const dependencies = project.dependencies()
+   .pipe(project.splitHtml())
```
TimvdLippe: Since sources and dependencies are processed the same, maybe extract a function to improve maintainability?

gulpfile.js
```diff
@@ -11,112 +11,146 @@
...
-  return gulp.src([mini + '.bak', micro + '.bak', max + '.bak'])
+gulp.task('restore-src', function(cb) {
+  const files = ENTRY_POINTS.map(f=>`${f}.src`);
+  gulp.src(files)
     .pipe(rename(function(p) {
       p.extname = '';
     }))
-    .pipe(gulp.dest('.'));
+    .pipe(gulp.dest('.'))
+    .on('end', ()=>Promise.all(files.map(f=>del(f))).then(()=>cb()));
```
TimvdLippe: Insert newlines into the callback to make it readable

PR #4057, opened 2016-10-11
Clear paths upon object invalidation. Fixes #3998.

### LINKED ISSUE #3998: computed binding function is passed with an old dependency
- checkout shop 2.0-preview branch:
`git clone git@github.com:Polymer/shop.git; cd shop; git checkout 2.0-preview; git pull; bower install;`
- run the app
- click on Men's Outerwear tab; it should show the correct list of items
- click on Ladies Outerwear tab; it should again show the correct list of items
- now click on Men's Outerwear tab again; the items are still showing Ladies Outerwears.

If you put a breakpoint at shop-list.html#L161, you can see `items` is referring the old value.

kevinpschaaf:
This re-institutes the 1.x [`clearPath`](https://github.com/Polymer/polymer/blob/master/src/lib/bind/accessors.html#L87-L93) behavior, which clears cached path values from `__data` when an object root is set.

This is a more correct fix than the one previously discussed (unconditionally read `Polymer.path.get` for all `arg.structured` [in marshalArgs](https://github.com/Polymer/polymer/blob/2.0-preview/src/properties/property-effects.html#L489-L490), since that would still leave invalid cached paths lying around.  I believe it is safe to keep the current optimization in `marshalArgs` that was causing the problem (only read from the object if no path is cached) as long as paths are being cleared when the object is invalidated, but it may be safer to also add that change to make the code less brittle -- to be discussed.

That said, needing to add the `clearPath` code back to the fast path is unfortunate.  Since the entire path notification system is now unified with the property system via `_setProperty`, it would be a significant change to remove all path caching, since queueing a path in `__dataPending` is the unified way to introduce a change into the system.  To be investigated further, but recommending unblocking the issue with this change first.

### DIFF
--- a/src/properties/property-effects.html
+++ b/src/properties/property-effects.html
@@ -629,6 +629,18 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         return this.__dataPending && (prop in this.__dataPending);
       }
 
+      _setPendingProperty(prop, value) {
+        // clear cached paths
+        if (typeof value == 'object') {
+          for (var p in this.__data) {
+            if (Polymer.Path.isDescendant(prop, p)) {
+              this.__data[p] = undefined;
+            }
+          }
+        }
+        return super._setPendingProperty(prop, value);
+      }
+
       // Prototype setup ----------------------------------------
 
       _addPropertyEffect(path, type, effect) {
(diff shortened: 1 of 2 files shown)


sorvell:
LGTM

PR #4069, opened 2016-10-13
[2.x] Treat non-Polymer `<template is="smth">` as any other element. 

tomalec:
Treat non-Polymer `<template is="smth">` as any other element. 
Stamp non-`dom-*` template untouched, do not change it to `<smth>`

### Reference Issue
Polymer/polymer/issues/4060
Polymer/polymer/issues/4068


kevinpschaaf:
LGTM, thanks

tomalec:
Thanks!

PR #4071, opened 2016-10-14
Minor feature additions/improvements

kevinpschaaf:
- Add Polymer.mix and utils tests (with mixin tests).
- Set dom-repeat initialCount default to 50 (makes incremental rendering default, so that users must "opt-out" to worse performance)
- Allow single-argument `set` to provoke changes (facades to `notifyPath`)


kevinpschaaf:
Closing, out of date.

PR #4082, opened 2016-10-21
Upgrade shorthand type to {type:Type} when flattening. Fixes #4066.

### LINKED ISSUE #4066: Shorthand property types not respected during deserialization in 2.x
### Description
Shorthand property types are not respect during deserialization in 2.x.

```js
// works
properties: {
  longhand: {
    type: Boolean
  }
}
```
```js
// no longer works
properties: {
  shorthand: Boolean
}
```

#### Live Demo

http://jsbin.com/zibapig/edit?html,console

### Versions
- Polymer: v2.x

kevinpschaaf:
### Reference Issue
Fixes #4066

### DIFF
--- a/src/elements/element.html
+++ b/src/elements/element.html
@@ -10,7 +10,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 <link rel="import" href="../utils/boot.html">
 <link rel="import" href="../legacy/dom-module.html">
-<link rel="import" href="../utils/utils.html">
 <link rel="import" href="../utils/case-map.html">
 <link rel="import" href="../events/event-listeners.html">
 <link rel="import" href="../template/template-stamp.html">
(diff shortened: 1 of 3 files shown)


jfrazzano:
You have the little one yet kev? Congrats!

src/elements/element.html
```diff
@@ -49,10 +60,10 @@
       // (2) element default values
       static get _flattenedProperties() {
         if (!this.hasOwnProperty('__flattenedProperties')) {
-          this.__flattenedProperties = this._ownConfig.properties || {};
+          this.__flattenedProperties = flattenProperties({}, this._ownConfig.properties);
```
sorvell: Let's add a comment here that we need to scrutinize this when optimizing. There's an extra loop that we didn't need before and the custom `flattenProperties` prevents us from potentially using `Object.assign`. It's probably nbd but I feel safer putting a bookmark in the code that we made a decision here.

kevinpschaaf: Done, added comment.

kevinpschaaf:

PR #4083, opened 2016-10-21
Refactor dom-if & dom-repeat using Polymer.Element.

kevinpschaaf:
Refactor dom-if & dom-repeat using Polymer.Element.  Eliminates dependency on legacy Polymer API and related bulk.

### DIFF
--- a/src/templatizer/dom-if.html
+++ b/src/templatizer/dom-if.html
@@ -8,7 +8,7 @@ Code distributed by Google as part of the polymer project is also
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 -->
 
-<link rel="import" href="../../polymer-legacy.html">
+<link rel="import" href="../../polymer-element.html">
 <link rel="import" href="templatizer.html">
 <link rel="import" href="../utils/debounce.html">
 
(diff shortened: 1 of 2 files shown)


src/templatizer/dom-if.html
```diff
@@ -30,63 +30,78 @@
    * Set the `restamp` property to true to force the stamped content to be
    * created / destroyed when the `if` condition changes.
    */
-  Polymer({
-
-    is: 'dom-if',
+  class DomIf extends Polymer.Element {
+
+    static get is() { return 'dom-if' }
```
sorvell: You didn't use this when defining and there's no template. We should be able to omit it here then right?

src/templatizer/dom-repeat.html
```diff
@@ -108,170 +108,179 @@
 
   var templatizer = new Polymer.Templatizer();
 
-  Polymer({
+  class DomRepeat extends Polymer.Element {
 
-    is: 'dom-repeat',
+    static get is() { return 'dom-repeat' }
```
sorvell: Same point as `dom-if`, can we remove this?

src/templatizer/dom-repeat.html
```diff
@@ -397,31 +406,29 @@
...
           }
         }
       }
-    },
+    }
 
-    _debounceRender: function() {
-      Polymer.Templatizer.enqueueDebouncer(this.debounce('_render', this._render));
-    },
+    _debounceRender(fn, delay) {
```
sorvell: In `dom-if` a similar function is called `queueRender`. Any reason to name this differently?

src/templatizer/dom-repeat.html
kevinpschaaf: Done, removed both.

src/templatizer/dom-repeat.html
kevinpschaaf: Done, renamed `dom-if`'s `_queueRender` -> `_debounceRender`

kevinpschaaf:

PR #4090, opened 2016-10-25
Move resolveUrl to Polymer.Element, get `is` from ctor. Fixes #3992

### LINKED ISSUE #3992: resolveUrl missing from 2.x
Doesn't appear to be in the prototype for 2.x. The omission isn't noted in the current release notes.

kevinpschaaf:
### Reference Issue
Fixes #3992

### Details
The reason `resolveUrl` didn't work even on `class extend LegacyElement` is because in the class-based syntax the `static get is()` getter is only on the constructor, and never gets put on the prototype/instance.

For this change, I just made `resolveUrl` look at `this.constructor.is` instead of `this.is`.  However, we might want to re-consider if we really want to lose `is` from the prototype/instance or not (was this an intentional change?).

### DIFF
--- a/src/elements/element.html
+++ b/src/elements/element.html
@@ -281,6 +281,26 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         }
       }
 
+      /**
+       * Rewrites a given URL relative to the original location of the document
+       * containing the `dom-module` for this element.  This method will return
+       * the same URL before and after vulcanization.
+       *
+       * @method resolveUrl
+       * @param {string} url URL to resolve.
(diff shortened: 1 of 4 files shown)


sorvell:
Placing `is` on the class rather than the prototype was intentional but it's arguable. The idea is that this is class related metadata (`is`, `properties`, `template`) so it makes sense to be there and not unnecessarily polluting the prototype.

kevinpschaaf:
Ok, so we'll leave it as-is.  Just wanted to ensure it was intentional. The legacy `Polymer()` API retains `is` on the prototype, so we're covered there.

kevinpschaaf:

PR #4092, opened 2016-10-25
Fixes #4084: distribution on binded slot name.

### LINKED ISSUE #4084: Binded slot name do not distribute content as expected in Polymer#1.7.0
### Description

Binded slot name do not correctly distribute slotted elements on non-chromium browser
#### Live Demo

http://jsbin.com/denumizugi/edit?html,console,output
#### Steps to Reproduce

Create a slot using binding for the name : `<slot name="[[slotName]]"></slot>`
#### Expected Results

Nothing should be displayed
#### Actual Results

'Should not be assigned' is displayed on non-chromium browser
### Browsers Affected

- [ ] Chrome
- [x] Firefox (flashes)
- [x] Edge
- [x] Safari 10
- [x] Safari TP Release 15 (Safari 10.1, WebKit 12603.1.8)
- ? Safari 9
- ? Safari 8
- [x] IE 11
### Versions

- Polymer: v1.7.0
- webcomponents: v0.7.22

I wanted to do `<slot name$="[[slotName]]"></slot>` but is failed converting to `<content select$="[slot=[[slotName]]]"></slot>`. So the converted binding looks like `<content select="[slot=[[slotName]]]"></slot>`. Therefore, when the binding occurs, it sets 'select' as a property instead of an attribute on non-chrome browser (that do not natively support v0 web components). During distribution, the `_matchesContentSelect` function 

``` javascript
var select = contentElement.getAttribute('select');
if (!select) {
  return true;
}
select = select.trim();
if (!select) {
  return true;
}
```

returns always true.

hypnoce:

### DIFF
--- a/src/mini/shady.html
+++ b/src/mini/shady.html
@@ -354,7 +354,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       },
 
       _matchesContentSelect: function(node, contentElement) {
-        var select = contentElement.getAttribute('select');
+        // Check attribute first. Fallback to property to handle cases where
+        // 'slot' is binded to a property <slot name="[[slotName]]"></slot>
+        // and gets converted to <content select="[slot=[[slotName]]]"></content>.
+        // Converting binded attribute (<slot name$="[[slotName]]"></slot>) cannot be acheived
+        // since attribute with $ in the name cannot be set to an element,
+        // despites they can be parsed by the HTML pareser.
+        var select = contentElement.getAttribute('select') || contentElement['select'];
         // no selector matches all nodes (including text)
         if (!select) {
           return true;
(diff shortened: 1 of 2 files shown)


hypnoce:
Hi team,

does this PR look good to you ?

Thanks

hypnoce:
Hi team,

Should we close or merge this PR ?
We are waiting for it to switch to 2.0.

Thanks

François

web-padawan:
@hypnoce I think you should use attribute binding: `<slot name$="[[slotName]]">`  
If it works, this PR is obsolete.

TimvdLippe:
@web-padawan is correct. You should use `name$=` to get this behavior: http://jsbin.com/quhihetafo/edit?html,output

Issue #4096, opened 2016-10-26
'_showHideChildren is not a function' thrown with dom-repeat and dom-if


nik3daz:
### Description

Having certain dom-repeat and dom-if structures causes
`Uncaught TypeError: this._instances[i]._showHideChildren is not a function` on new data sometimes.
#### Live Demo

https://jsbin.com/lehajuvito/1/edit?html,console,output
#### Steps to Reproduce

1. Go to JSBin
2. Mash the Refresh button
3. Error pops into the console occasionally
#### Expected Results

_showHideChildren should not be called when the instance is a placeholder.
#### Actual Results

It does get called, but it doesn't exist, throwing an error.
### Versions

- Polymer: v1.X.X, Probably v2.X.X too

web-padawan:
Seems that you are missing using `as` which is required for nested `dom-repeat`.
Have a look how this should look like:
https://jsbin.com/meqogakice/edit?html,console,output

nik3daz:
It seems to work both ways for me (on Chrome). Either way, the error still appears when you hit refresh a few times.

kevinpschaaf:
Resolved by https://github.com/Polymer/polymer/pull/4097

arthurevans:
Running the original JSBin against 1.7.1, I'm not seeing the original error message, but I am seeing this new console error appear occasionally instead:

```
"Error: The ref_node to be inserted before is not a child of this node
    at DomApi.insertBefore (https://polygit.org/components/polymer/polymer-mini.html:709:7)
    at HTMLTemplateElement._insertInstance (https://polygit.org/components/polymer/polymer.html:5042:25)
    at HTMLTemplateElement._render  
    ...
```

web-padawan:
Any update on this? Occasionally faced this issue as well in a rarely reproducible case.

chen-ye:
I can reliably reproduce this with the following test-case:
https://jsfiddle.net/Lq2mczh6/

Specific parameters to trigger this bug:
Child dom-repeat is nested inside a dom-if.
Child dom-repeat is configured with initial-count = x
Items is set to an 2d array where the size of the outer array is n, and the size of the inner array is m != x.
Page is rendered/flushed.
Items is set to a 2d array where the size of the outer array < n (in this case, empty).

PR #4097, opened 2016-10-26
Prevent _showHideChildren from being called on placeholders.

nik3daz:
### Reference Issue

This CL fixes issue #4096 where certain cases could cause
_showHideChildren to get called on template placeholders.


danbeam:
@tjsavage can you add source: hooli?  this is affecting chrome's history page

danbeam:
also /cc @kevinpschaaf @sorvell

PR #4099, opened 2016-10-26
2.0 preview: extending templates

mkazlauskas:
- Fixes a bug with extended templates, where both base and super class has bind listeners (both reference the same existing listeners array). When a super class template is parsed it would add new bind listeners to base class, referencing nodes that do not exist in base class template.
- Adds basic tests for extending element's template.

Steps to reproduce:
1. Pull my fork
2. Comment out line 169 in `src/elements/element.html`
3. `$ wct test/unit/polymer.element.html`


mkazlauskas:
I've tested latest 2.0-preview commit and this PR is still relevant.

sorvell:
Closed in favor of https://github.com/Polymer/polymer/pull/4493.

PR #4105, opened 2016-10-27
Dom-repeat... fixed when to throw message about absent template

MartinsThiago:
Fixed small bug where absent `<template>` in `<dom-repeat>` was causing `NullPoint`


kevinpschaaf:
Resolved on master via other changes. Thanks for PR!

PR #4107, opened 2016-10-27
Add Polymer.Element base class

### LINKED ISSUE #4103: Add Polymer.Element class for Hybrid mode
### Description

If somebody already using ES6 class for the polymer elements then it would be nice if we could add the `extends Polymer.Element` which does not need to do anything with 1.0, only just handle all of the new methods which are in the Custom Elements v1 and get the properties and observers from the config.

In this case we could prepare for the 2.0 our elements which are already using ES6 class, so we need minimal migration (behaviours) to jump from hybrid mode to 2.0 when it's released.
#### Current Markup

Based on https://www.polymer-project.org/1.0/blog/es6 article the current ES6 markup is this:

``` javascript
class MyElement {
    get behaviours() { return []; }
    beforeRegister() {
        this.is = 'my-element';
        this.properties = {};
    }

    // moarh code...
}

// Initialize ES6 Class Element
Polymer( MyElement );
```

Currently im using a bit upgraded version from it:

``` javascript
class MyElement {
    static get is() { return 'my-element'; }

    // Keeping for 1.0 and hybrid mode
    static get properties() { return this.config.properties||{}; };

    // Setter is needed if behaviours are exist
    static set properties( values ) {
        if(this._config!=undefined) this._config.properties = values;
    }

    // Keeping for 1.0 and hybrid mode
    get behaviours() { return []; }

    // Preparing for 2.0
    static get config() {
        // Return stored configuration
        if(this._config!=undefined) return this._config;

        // Create configuration if needed
        this._config = {
            properties: {}
        };

        return this._config;
    }

    // moarh code...
    ready() {
        // ...
    }

    attached() {
        // ...
    }
}

// Initialize ES6 Class Element
Polymer( MyElement );
```
#### Expected Markup

``` javascript
class MyElement extends Polymer.Element {
    static get is() { return 'my-element'; }

    // Keeping for 1.0 and hybrid mode
    static get properties() { return this.config.properties||{}; };

    // Setter is needed if behaviours are exist
    static set properties( values ) {
        if(this._config!=undefined) this._config.properties = values;
    }

    // Keeping for 1.0 and hybrid mode
    get behaviours() { return []; }

    // Preparing for 2.0
    static get config() {
        // Return stored configuration
        if(this._config!=undefined) return this._config;

        // Create configuration if needed
        this._config = {
            properties: {}
        };

        return this._config;
    }

    // moarh code...
    ready() { // or constructor()
        // super()
        // ...
    }

    attached() { // or connectedCallback()
        // super.connectedCallback()
        // ...
    }
}

// Initialize ES6 Class Element
Polymer( MyElement );
```
### Versions
- Polymer: v1.7.0

justinfagnani:
### Reference Issue

Fixes #4103

### DIFF
--- a/src/lib/element.html
+++ b/src/lib/element.html
@@ -1,0 +1,56 @@
+<!--
+@license
+Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<link rel="import" href="base.html">
+<script>
+(function() {
+
+  'use strict';
+
+  /**
+   * A base class for class-based Polymer elements.
+   *
+   *     class MyElement extends Polymer.Element {
+   *       static get is() { return ‘my-element’; }
+   *       static get config() { return {
+   *         properties: {
+   *           // the properties you know and love from Polymer
+   *         },
+   *         observers: [
+   *           // the observers you know and love from Polymer
+   *         ]
+   *       }}
+   *       get listeners() { return {
+   *         // the listeners you know and love from Polymer
+   *       }}
+   *     }
+   *
+   * Element classes should extend directly from Polymer.Element. Property,
+   * observer, and listener declarations on superclasses will not be processed
+   * by Polymer.
+   */
+  function PolymerElement() {}
+  PolymerElement.prototype = {
+    get is() { return this.constructor.is; },
+    get properties() {
+      var config = this.constructor.config;
+      return config && config.properties;
+    },
+    get observers() {
+      var config = this.constructor.config;
+      return config && config.observers;
+    },
+  };
+  PolymerElement.prototype.constructor = PolymerElement;
+  Object.setPrototypeOf(PolymerElement.prototype, Polymer.Base);
+  Object.setPrototypeOf(PolymerElement, HTMLElement);
+
+  Polymer.Element = PolymerElement;
+})();
+</script>
--- a/src/lib/polymer-bootstrap.html
+++ b/src/lib/polymer-bootstrap.html
@@ -37,7 +37,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var options = {
         prototype: prototype
       };
-      // NOTE: we're specifically supporting older Chrome versions here 
+      // NOTE: we're specifically supporting older Chrome versions here
       // (specifically Chrome 39) that throw when options.extends is undefined.
       if (prototype.extends) {
         options.extends = prototype.extends;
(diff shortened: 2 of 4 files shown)


arthurevans:
Just a thought—this is going to end up with the whole `Polymer.Base` API surface, which makes it equivalent to `Polymer.LegacyElement` in 2.0.

kevinpschaaf:
Looks good, so let's...
* add a unit test
* add note/example about `behaviors` (would be similar to `listeners`, where you can do an instance getter `get behaviors() { return [...] }`, which is not forward compatible to 2.0 classes
* add note: class expression mixins can be used, but a `config` getter there will overwrite the superClass (similar to your note about subclassing)

src/lib/element.html
```diff
@@ -0,0 +1,56 @@
...
+      return config && config.properties;
+    },
+    get observers() {
+      var config = this.constructor.config;
+      return config && config.observers;
+    },
+  };
+  PolymerElement.prototype.constructor = PolymerElement;
+  Object.setPrototypeOf(PolymerElement.prototype, Polymer.Base);
+  Object.setPrototypeOf(PolymerElement, HTMLElement);
```
kevinpschaaf: Eh, let's go ahead and use `Polymer.Base.chainObject` (same as https://github.com/Polymer/polymer/pull/4107/files#diff-7838a926e8a75d4e50eb319e1284438dR61), since there's no good reason not to...

rictic:
Hm, is there a way to write a hybrid element with a base class?

Maybe something like:

```javascript
class SmoothSlider extends Polymer.LegacyElement { /* ... */ }

if (/^1\./.test(Polymer.version)) {
  Polymer(SmoothSlider);
} else {
  customElements.define(SmoothSlider.is(), SmoothSlider);
}
```

justinfagnani:
I think it'd be good to wrap up the logic so the user doesn't write a conditional. Maybe something like:

```javascript
Polymer.LegacyElement.define(MyElement);
```

Since statics inherit, we could conceivably do:

```javascript
MyElement.define();
```

though that's a bit harder to statically analyze and get the same kind of detection we get for CEv1 vanilla elements, but if we analyze solely on the prototype chain (or jsdoc tag) that won't be an issue.

jfrazzano:
Moving along these lines, could you implement a library of const string=attributes=>`someHTML in a string literal with replacements like ${items.map(item=>`${item.someProperty} more dom`);} `;
mySetup=[{object properties} ,{object properties 2}…];
mySetup.forEach(obj=>`${obj.someProperty===selectedValue?`&#x2705;`:`&#x2b1b;`}`).join(‘’);

var myStringComponent=string(mySetup);
let nameEtAl=` any internal settings for the anchor wrapper, finished by leaving off the closing ‘ quotation mark ‘ for the \name\ attribute being “overfilled”
var nextComponent=myStringComponent.anchor(nameEtAl);

document.body.innerHTML+=nextComponent;  

I have just been experimenting with this approach, and am now attempting to add the js for the components via URIs that accord with the respective anchor document fragments.  I don’t know if it will work, or if I am even on the right track a little.  But the previous work that I tried, when using regex well, and splitting up the dom into little pieces to be updated via the regex and re-representation of small string fragments, has seemed to work well.  I don’t know if this is what you were broadly suggesting, Peter, but it seems in accord with your talks over the years and your general design preferences in your work.  

Again, I am no expert and am experimenting to add functionality similar to factory methods to component like implementations that can share behavior like methods… 

If this makes sense, and is worth pursuing discussing.  Awesome, I am game.  If its nonsense.  Please just ignore it and I apologize. 

Hope I was helpful, 

sincerely, 

Jason

with sub replacements for the replacements handled vi

azakus:
Closing because we'd rather not add this to Polymer 1.x.
I've documented a somewhat simple way to accomplish this task this in user code: https://github.com/Polymer/polymer/issues/4103#issuecomment-331312682

PR #4110, opened 2016-10-28
support text node distribution to default slot

### LINKED ISSUE #4109: Text nodes are not distributed to default slot
When entering text nodes in the host element, they are not distributed to the default slot.
#### Live Demo

http://codepen.io/oridan/pen/rrgvQY?editors=1000
#### Expected Results

The live demo should show two rows as follows:
[Icon] item 1
[Icon] item 2
#### Actual Results

I see just two icons:
[Icon]
[Icon]
### Browsers Affected

- [x] Chrome
- [x] Firefox
- [?] Edge
- [x] Safari 9
- [?] Safari 8
- [?] IE 11
### Versions

- Polymer: v1.7.0
- webcomponents: v0.7.22

TomK:
Fixes #4109

### DIFF
--- a/src/mini/shady.html
+++ b/src/mini/shady.html
@@ -364,6 +364,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         if (!select) {
           return true;
         }
+        // default slot selector can match text nodes
+        if(select === ':not([slot])' && node.nodeType === Node.TEXT_NODE) {
+          return true;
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
Closing per https://github.com/Polymer/polymer/issues/4109#issuecomment-257019369, since there's no way to make native Shadow DOM match this behavior.

PR #4136, opened 2016-11-07
Added/updated API docs (WIP)

kevinpschaaf:
Start of work to add API documentation, starting with `property-accessors` and `property-effects`.  Some minor API refactoring was done along the way.

### DIFF
--- a/src/properties/property-accessors.html
+++ b/src/properties/property-accessors.html
@@ -16,7 +16,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   'use strict';
 
-  // Save map of native properties
+  // Save map of native properties; this forms a blacklist or properties
+  // that won't have their values "saved" by `saveAccessorValue`, since
+  // reading from an HTMLElement accessor from the context of a prototype throws
   const nativeProperties = {};
   let proto = HTMLElement.prototype;
   while (proto) {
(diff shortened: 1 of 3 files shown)


src/properties/property-accessors.html
```diff
@@ -27,6 +29,19 @@
     proto = Object.getPrototypeOf(proto);
   }
 
+  /**
+   * Used to save the value of a property that will be overridden with
```
arthurevans: Used to save => Saves (for consistency with other methods).

"Will be overridden" is vague because it's not clear whether this is something that Polymer's doing, or something the user's doing.  ... Looks like if called from `_createPropertyAccessor` it's called immediately before Polymer overrides the accessor. (In which case, perhaps, "before Polymer overrides it with an accessor.")

But this is also a non-underscored method, which implies that people might call it directly under some circumstance...?

Also a little clearer if it sticks to present tense: will be => are/is 

Also it struck me as slightly inconsistent that property name was specified second here, but the other methods that operate on properties specify the property name first. (But this doesn't have "property" in the method name, either, so maybe that's irrelevant.)

arthurevans:
:tada:

src/properties/property-accessors.html
kevinpschaaf: This is a `@private` function in the closure, not accessible outside this file. Right now I'm erring on the side of documenting/annotating everything in prep for making Polymer buildable with jsCompiler.

arthurevans: Ah, awesome. Thanks for clarifying!

PR #4187, opened 2016-11-30
Make tests more x-browser compatible

sorvell:


kevinpschaaf:
LGTM.

Edge remains flaky, investigating but won't block this forward movement.

PR #4190, opened 2016-12-01
Fix unreadable logging in dark theme Chrome console

n0wak:
### Reference Issue

When using the dark theme in Chrome the default coloured console logging is unreadable (light text on yellow). This is easily fixed by forcing the text colour to be dark.


TimvdLippe:
`Polymer.Base._warn` and friends are now in legacy mode and all logging is now passed into the regular `console` methods. Therefore I am closing this PR.

PR #4191, opened 2016-12-01
Observer property pass

nochev:
Pass `property` name as third argument in observer callback. It's very useful to reuse observer logic when we know the name of changed property.


nochev:
I signed it!

TimvdLippe:
Polymer 2 does not have this atm, WDYT @kevinpschaaf @sorvell @azakus ?

TimvdLippe:
This can be implemented in Polymer 2by overriding [`_propertiesChanged`](https://www.polymer-project.org/2.0/docs/api/elements/Polymer.Element#method-_propertiesChanged) which provides a bag of `changedProps` that contain the property names and its new values.

PR #4199, opened 2016-12-05
Fix for #4150

### LINKED ISSUE #4150: Unclear exception when binding a property to <textarea> (IE11)
### Description
When assigning an attribute to <textarea> via `=` instead of `$=`, IE11 throws the following exception:
`Unable to get property '_propertyInfo' of undefined or null reference.`

Chrome, Firefox, Safari, and Edge handle this situation. Assigning an attribute via `$=` to an `<input>` element works as well on IE11.

It would be useful to have a better error message as it's rather difficult to debug this bug.

More detailed description: [Polymer + IE11 fail: Unable to get property ‘_propertyInfo’](https://medium.com/collaborne-engineering/polymer-ie11-fail-unable-to-get-property-propertyinfo-e3bef228919d#.haaq8c3ka)

#### Live Demo
https://jsbin.com/sehimorilu/edit?html

#### Steps to Reproduce
1. Add `<textarea>` element
2. Bind attribute via `=`
3. Load in IE11

#### Expected Results
- Meaningful warning
- Textarea shows

#### Actual Results
- Exception is thrown
- Textarea is not shown

### Browsers Affected
- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [x] IE 11

### Versions
- Polymer: v1.7.0
- webcomponents: v0.7.23

proxyLaw:
Fixes #4150

The source of the bug is a problem from IE 10/11 that causes `placeholder` attribute value to be copied as textarea content. It caused an additional and improper(?) binding that was found at _parseChildNodesAnnotations.

### DIFF
--- a/src/standard/annotations.html
+++ b/src/standard/annotations.html
@@ -114,6 +114,15 @@ TODO(sjmiles): this module should produce either syntactic metadata
 
 */
 
+(function () {
+  var ifTextAreaHasPlaceholderBug = (function () {
+    var textArea = document.createElement('textarea'),
+        placeholderValue = "bugged";
+    textArea.setAttribute('placeholder', placeholderValue);
+    return textArea.textContent === placeholderValue;
+  }());
+
+
   Polymer.Base._addFeature({
 
     // registration-time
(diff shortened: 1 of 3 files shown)


TimvdLippe:
Closing this PR per https://github.com/Polymer/polymer/issues/4150#issuecomment-410455256

Issue #3908, opened 2016-09-09
importHref fails to fetch resource when it's used offline, then while online


Gaubee:
### Description

Page returned from the offline to the online state, the original request error page(404) can not be requested again.
#### Live Demo

https://jsbin.com/xukike/edit?html,output
### Browsers Affected

- [x] Chrome
- [ ] Safari 8
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] IE 11
### Versions

- Polymer: v1.6.1
- webcomponents: vX.X.X

ebidel:
Please post a jsbin that repos the issue and reopen when you do.

TimvdLippe:
@ebidel https://jsbin.com/foduvubici/edit?html,console,output

ebidel:
Thanks for that. Looks like this is because `imporHref` maintains an internal cache (https://github.com/Polymer/polymer/blob/06bc9dbe34042268b94e60374910b6d6ce1ae3d7/src/standard/utils.html#L395).

PR #4200, opened 2016-12-05
More practical caching strategy for importHref

cdata:
### Reference Issue
Fixes https://github.com/Polymer/polymer/issues/3908

Previously, importHref would cache failed results. This caused failures
to be cached indefinitely, even when the failure was the result of
conditions that are subject to change over time (such as poor network
connectivity).

This implementation no longer caches failures. Additionally, it relies
on the link tag's presence in the <head> of the document as an
indication that it is loading or has loaded, rather than keeping an
object map of URLs to link tags.

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -416,7 +416,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // will be automatically created again the next time `importHref` is
         // called.
         if (link.parentNode) {
-          link.parentNode.removeChild(this);
+          link.parentNode.removeChild(link);
         }
         if (onerror) {
           whenImportsReady(function() {
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -17,16 +17,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   'use strict';
 
-  // run a callback when HTMLImports are ready or immediately if
-  // this api is not available.
-  function whenImportsReady(cb) {
-    if (window.HTMLImports) {
-      HTMLImports.whenReady(cb);
-    } else {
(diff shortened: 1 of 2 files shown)


src/standard/utils.html
```diff
@@ -367,50 +367,52 @@
...
         }
+      };
+
+      link.addEventListener('load', loadListener);
+      link.addEventListener('error', errorListener);
+
+      if (link.parentNode == null) {
+        link.href = href;
         document.head.appendChild(link);
+      } else if (link.__dynamicImportLoaded || !link.__dynamicImport) {
```
cdata: Rationale here is that imports that were loaded dynamically and finished, or weren't loaded dynamically to begin with, should fire a synthetic `load` event. This might fall apart for imports that were loaded outside of `importHref`, but aren't fully loaded yet by the time we get here. Thoughts?

src/standard/utils.html
```diff
@@ -367,50 +367,52 @@
...
+      if (link == null) {
+        link = document.createElement('link');
+        link.rel = 'import';
+        link.__dynamicImport = true;
       }
-      if (onerror) {
-        imprt.addEventListener('error', errorListener);
+
+      if (optAsync) {
+        link.setAttribute('async', '');
```
cdata: This conditional branch exists on its own to satisfy the constraints of a test. The test asserts that a link is imported async by checking for an `async` attribute on a link. If a link is imported synchronously at first, and then later imported asynchronously (as is the case in the test), then we must always set the `async` attribute even if we are relying on a "cached" (that is, already-done-loading) `link` tag for state.

TimvdLippe:
These changes have been incorporated in #4305, but this PR was not properly closed.

PR #4209, opened 2016-12-09
Refactor importHref

gronke:
- uses altering test assets to ensure tests do not affect each other
- ensure importHref always returns a link element
- ensure onload() and onerror() functions are only called once when the same element was loaded again

Blocked by https://github.com/webcomponents/webcomponentsjs/pull/655 because the load event is invoked twice when a failed import attempt is repeated.


gronke:
@cdata 🎉 would you please have a look at this changes that were build on top of your pull-request #4200 to improve importHref.

src/standard/utils.html
```diff
@@ -370,47 +370,57 @@
...
       var errorListener = function(event) {
-        link.removeEventListener('load', loadListener);
-        link.removeEventListener('error', errorListener);
+        cleanup();
         // In case of an error, remove the link from the document so that it
         // will be automatically created again the next time `importHref` is
         // called.
-        document.head.removeChild(link);
+        link.__loaded = true;
+        link.__error = true;
```
cdata: The intention of the change is to discard this `link`. What is the benefit of tracking state on the `link` if it is about to be discarded?

cdata:
Can you reference the bug in the polyfill that results in the Firefox error?

src/standard/utils.html
```diff
@@ -370,47 +370,57 @@
...
         if (onerror != null) {
           return onerror.call(self, event);
         }
       };
 
+      if (link == null) {
+        link = document.createElement('link');
+        link.rel = 'import';
+        link.href = href;
+        link.__loaded = false;
```
gronke: That's true 😳

test/unit/dynamic-import.html
```diff
@@ -94,36 +71,117 @@
...
+
+      test('importHref does not cache failed results', function(done) {
+
+        var a, b;
+
+        a = Polymer.Base.importHref('does_not_exist.html', function() {
+          throw new Error('Load handler should not be called.');
+        }, function() {
+          b = Polymer.Base.importHref('does_not_exist.html', function() {
+            throw new Error('Load handler should still not be called.');
```
gronke: Unless https://github.com/webcomponents/webcomponentsjs/pull/655 is merged this test fill fail on browsers without native html imports support (e.g. Firefox on Travis CI).

gronke:
Since the base branch https://github.com/Polymer/polymer/tree/refactor-import-href was not rebased for a while, the new PR #4348 is pointing against master directly. Closing.

PR #4221, opened 2016-12-14
2.0 data updates

### LINKED ISSUE #3910: Breaking change: `x-changed` event fired before internal bindings dependent on `x` are run.
_From @bicknellr on July 27, 2016 17:47_

It's no longer safe to assume that after receiving a change notification event (`x-changed`) from an element that all of that element's internal observers that are dependent on the changed value (`x`) have run.

For example, one of the tests for iron-image waits on `loaded-changed` to check that the iron-image's placeholder is hidden. Because the placeholder's visibility is determined by a multi-property binding to its class attribute, the `loaded-changed` event is now fired _before_ the placeholder's class list has been updated causing the test (which uses `getComputedStyle`) to inspect styles too early and fail.

_Copied from original issue: PolymerLabs/alacarte#12_

### LINKED ISSUE #3932: Changing value in <property>-changed listener fires observers
_From @e111077 on August 29, 2016 17:44_
# Description

Changing a value in a `<property-name>-changed` event listener seems to fire pending observers resulting in a really-confusing order of operations.
# Repro Case

``` html

<dom-module id="asdf-asdf">
  <template>

  </template>
  <script>
    Polymer({
      is: 'asdf-asdf',
      properties: {
        value: {
          type: String,
          notify: true
        },

        linkedValue: {
          type: String,
          notify: true,
          observer: 'linkedValueChanged'
        }
      },

      linkedValueChanged: function() {
        console.log('change value in observer');
        this.set('value', this.linkedValue);
      }
    });
  </script>
</dom-module>

<asdf-asdf id="asdf"></asdf-asdf>

<script>
  asdf.addEventListener('linked-value-changed', function() {
    console.log('change value in event listener');
    asdf.value = 'THIS IS WRITE';
    console.log(asdf.value);
  });

  asdf.set('linkedValue','THIS IS WRONG!!!');
  console.log(asdf.value);
</script>
```
## Expected Result

Optimal, prints:

```
change value in observer
change value in event listener
THIS IS WRITE!!!
THIS IS WRITE!!!
```

Still makes sense, but not-really-expected:

```
change value in event listener
THIS IS WRITE
change value in observer
THIS IS WRONG!!!
```
## Actual Result

```
change value in event listener
change value in observer
THIS IS WRONG!!!
THIS IS WRONG!!!
```
# Workaround

Do not use listeners on `<property>-changed` but instead create an observer and do all changes in the observer. This doesn't really work well because you can't add an observer to an element nicely after it has been created.

_Copied from original issue: PolymerLabs/alacarte#100_

### LINKED ISSUE #4061: Consider ways to remove path work from hot paths
Currently path key/values are stored with other property data. However, it is necessary to be able to clear all paths for a given property. To do this, we currently need to iterate the entire data object.  In 2.0 paths and properties are unified. This has some benefits but also introduces costs like this. Consider ways to factor out the inefficiencies and re-evaluate the overall cost/benefit of tracking path information.

See also https://github.com/Polymer/polymer/pull/4057.

### LINKED ISSUE #4163: [2.0-preview] Path-based changes to array sub-key only notifies first time per index
### Description
I was attempting to manipulate a sub-key of an item in an array (a simple boolean property). I noticed that only the first change to an item at a given index would actually notify of the change to the sub-key.

If I had an element in `fooArray`, and changed it like this:

```javascript
this.set('fooArray.0.bool', true);
```

The change would notify. As an eventual side-effect of this notification, the first element is removed from the array (so 'fooArray.0' should refer to a new object). A subsequent change to the array using the same path did not result in a notification.

#### Live Demo
http://jsbin.com/tubajo/edit?html,console,output

#### Steps to Reproduce
 1. Open the live demo URL
 2. Click the button labeled "Acknowledge next item", note the console output
 3. Click the button labeled "Add item"
 4. Click the button labeled "Acknowledge next item", note the console output

#### Expected Results
The complete console output should read:

```
"Array path changed: fooArray"
"Acknowledging item at index 0..."
"Array path changed: fooArray.0.acknowledged"
"Removing acknowledged item!"
"Array path changed: fooArray.splices"
"Array path changed: fooArray.length"
"Array path changed: fooArray.splices"
"Array path changed: fooArray.length"
"Added item to fooArray, now it has length 1"
"Acknowledging item at index 0..."
"Array path changed: fooArray.0.acknowledged" <-- Important!
"Removing acknowledged item!"
"Array path changed: fooArray.splices"
"Array path changed: fooArray.length"
```

#### Actual Results
The console output actually reads:

```
"Array path changed: fooArray"
"Acknowledging item at index 0..."
"Array path changed: fooArray.0.acknowledged"
"Removing acknowledged item!"
"Array path changed: fooArray.splices"
"Array path changed: fooArray.length"
"Array path changed: fooArray.splices"
"Array path changed: fooArray.length"
"Added item to fooArray, now it has length 1"
"Acknowledging item at index 0..."
```

NOTE: The change notification for `fooArray.0.acknowledged` does not occur.

### Versions
- Polymer: 2.0-preview

kevinpschaaf:
### Description
Set of data updates to primarily improve backward-compatibility with 1.0:
* Removes the `typeof value == 'object'` passthru in the `_shouldPropUpdate` dirty check, and replaced it with a "temporary data cache" (`__dataTemp`) which is used for dirty checking objects and is cleared at the end of each turn. This is a safer way to allow re-setting objects with deep changes to the same property and have all effects dependent on that object's sub-properties re-evaluated, since the temporary cache prevents the possibility of infinite loops.
* A similar approach is used to remove all path caching (and associated clearing from `__data`).  Paths are also temporarily cached in `__dataTemp` and then cleared, which avoids leaving invalidated index-based array paths in the cache following mutations which can lead to errors.
* The re-introduction of temporary object/path caching allows the notification effect to be moved to the end of `_propertiesChanged`, for better compatibility with 1.0 without leading to n^2 propagation when notifying objects _up_ the tree (dirty-check stops propagation back down the branch from which the notification came).
* Collapses `batched-effects` into `property-effects` for code hygiene, since the `property-effects` layer was not completed as a usable un-batched subset as it was conceived (had no support for configuration)

### Reference Issues
Resolves the following issues:
* Fixes #4061 - Consider ways to remove path work from hot paths
* Fixes #4163 - [2.0-preview] Path-based changes to array sub-key only notifies first time per index
* Fixes #3932 - Changing value in <property>-changed listener fires observers
* Fixes #3910 - Breaking change: `x-changed` event fired before internal bindings dependent on `x` are run


src/properties/property-accessors.html
```diff
@@ -146,17 +148,40 @@
        * @protected
        */
       _setPendingProperty(property, value) {
-        let old = this.__data[property];
-        if (this._shouldPropChange(property, value, old)) {
+        let isPath = Polymer.Path.isPath(property);
+        let isObject = typeof value == 'object';
```
kevinpschaaf: This change adds a `Polymer.Path` dependency to `property-accessors` (up till now accessors was agnostic to path keys).  Refactor and pull up to `property-effects`?

src/templatizer/dom-repeat.html
```diff
@@ -566,20 +567,23 @@
...
+          var inst = this._instances[instIdx];
+          if (inst) {
+            inst.forwardProperty(this.as + (itemPath ? '.' + itemPath : ''), value);
             inst.flushProperties();
-            return true;
-          }        
-        }
+          }
+        }        
+        return true;
```
kevinpschaaf: This change ensures that a item path notification, e.g. `items.3.prop` causes this function to return `true` regardless of whether there was an instance created and logged in the map for that item yet; the previous code would only return `true` if the instance for that path was found.  Returning `true` causes `dom-repeat` to do a full-refresh, which shouldn't happen for an item path notification.

src/properties/property-effects.html
```diff
@@ -26,6 +26,11 @@
   'use strict';
 
   const CaseMap = Polymer.CaseMap;
+  const mixin = Polymer.Utils.mixin;
+
+  // Monotonically increasing unique ID used for de-duping effects triggered
+  // from multiple properties in the same turn
+  let effectUid = 0;
```
kevinpschaaf: Moved to `property-effects` from `batched-effects`

src/properties/property-effects.html
```diff
@@ -77,21 +82,59 @@
...
+      let id = effectUid++;
+      for (let prop in props) {
+        if (runEffectsForProperty(inst, effects, id, prop, props[prop],
+          oldProps && oldProps[prop])) {
+          ran = true;
+        }
       }
     }
+    return ran;
+  }
```
kevinpschaaf: Moved to `property-effects` from `batched-effects` (replaces p-e version)

src/properties/property-effects.html
```diff
@@ -77,21 +82,59 @@
...
+          (!fx.info || fx.info.lastRun !== id)) {
+          fx.fn(inst, prop, Polymer.Path.get(inst, prop) || value, old, fx.info, fromAbove);
+          if (fx.info) {
+            fx.info.lastRun = id;
+          }
+          ran = true;
+        }
+      }
+    }
+    return ran;
```
kevinpschaaf: Moved to `property-effects` from `batched-effects` (called from `runEffects` and `runNotifyEffects`)

src/properties/property-effects.html
```diff
@@ -77,21 +82,59 @@
...
+   * on an instance.
    *
    * @param {Object} inst The instance with effects to run
-   * @param {string} property Name of property
-   * @param {*} value Current value of property
-   * @param {*} old Previous value of property
-   * @param {Object<string,Array>} effects List of effects
+   * @param {string} type Type of effect to run
+   * @param {Object} props Bag of current property changes
+   * @param {Object} oldProps Bag of previous values for changed properties
```
kevinpschaaf: Moved from `batched-effects` to `property-effects` (replaces p-e version)

src/properties/property-effects.html
```diff
@@ -77,21 +82,59 @@
...
+   * Runs a list of effects for a given property.
+   *
+   * @param {Object} inst The instance with effects to run
+   * @param {Array} effects Array of effects
+   * @param {number} id Effect run id used for de-duping effects
+   * @param {string} prop Name of changed property
+   * @param {*} value Value of changed property
+   * @param {*} old Previous value of changed property
+   * @private
+   */
```
kevinpschaaf: Moved to `property-effects` from `batched-effects` (called from `runEffects` and `runNotifyEffects`)

src/properties/property-effects.html
```diff
@@ -77,21 +82,59 @@
...
+  function runEffectsForProperty(inst, effects, id, prop, value, old) {
+    let ran;
+    let rootProperty = Polymer.Path.root(prop);
+    let fxs = effects[rootProperty];
+    if (fxs) {
+      let fromAbove = inst.__dataFromAbove;
+      for (let i=0, l=fxs.length, fx; (i<l) && (fx=fxs[i]); i++) {
+        if (Polymer.Path.matches(fx.path, prop) &&
+          (!fx.info || fx.info.lastRun !== id)) {
+          fx.fn(inst, prop, Polymer.Path.get(inst, prop) || value, old, fx.info, fromAbove);
```
kevinpschaaf: The `Polymer.Path.get` call here is a change from previous implementation which just did `inst.__data[prop]`, because paths are no longer persisted in `__data`; using `get` is more correct (better immunity against changes to path in earlier effect run this turn), but `inst.__data[prop] || value` passes all tests... revert?

src/properties/property-effects.html
```diff
@@ -117,6 +160,82 @@
...
+   * changed was a path and the root property for that path didn't have a
+   * "notify" effect.  This is to maintain 1.0 behavior that did not require
+   * `notify: true` to ensure object sub-property notifications were
+   * sent.
+   *
+   * @param {Object} inst The instance with effects to run
+   * @param {Object} props Bag of current property changes
+   * @param {Object} oldProps Bag of previous values for changed properties
+   * @private
+   */
```
kevinpschaaf: Moved to `property-effects` from `batched-effects` and given a better name (was previously `notifyProperties`); simplified from previous implementation since re-entry stuff was removed, as it is called

src/properties/property-effects.html
```diff
@@ -117,6 +160,82 @@
...
+   * Dispatches {property}-changed events to indicate a property (or path)
+   * changed.
+   *
+   * @param {Element} inst The element from which to fire the event
+   * @param {string} eventName The name of the event to send ('{property}-changed')
+   * @param {*} value The value of the changed property
+   * @param {string=} path If a sub-path of this property changed, the path
+   *   that changed (optional).
+   * @private
+   */
```
kevinpschaaf: Moved to `property-effects` from `batched-effects`, minor refactoring

src/properties/property-effects.html
```diff
@@ -129,12 +248,74 @@
...
+  }
+
+  /**
+   * Adds a 2-way binding notification event listener to the node specified
+   *
+   * @param {Object} node Child element to add listener to
+   * @param {Object} inst Host element instance to handle notification event
+   * @param {Object} info Listener metadata stored via addAnnotatedListener
+   * @private
+   */
```
kevinpschaaf: Same function as before, moved up to collate notify-related functions together.

src/properties/property-effects.html
```diff
@@ -129,12 +248,74 @@
...
+   * the event contained path information, translate that path to the host
+   * scope's name for that path first.
+   *
+   * @param {Event} e Notification event (e.g. '<property>-changed')
+   * @param {Object} inst Host element instance handling the notification event
+   * @param {string} property Child element property that was bound
+   * @param {string} path Host property/path that was bound 
+   * @param {boolean} negate Whether the binding was negated
+   * @private
+   */
```
kevinpschaaf: Same function as before, moved up to collate notify-related functions together.

src/properties/property-effects.html
```diff
@@ -129,12 +248,74 @@
...
+  }
+
+  /**
+   * Called by 2-way binding notification event listeners to set a property
+   * or path to the host based on a notification from a bound child.
+   *
+   * @param {string} path Path on this instance to set
+   * @param {*} value Value to set to given path
+   * @protected
+   */
```
kevinpschaaf: Was instance method for overriding, moved to private function.

src/properties/property-effects.html
```diff
@@ -169,6 +350,36 @@
...
+   * there are no more newly computed properties.  This ensures that all
+   * properties that will be computed by the initial set of changes are
+   * computed before other effects (binding propagation, observers, and notify)
+   * run.
+   *
+   * @param {Object} inst The instance the effect will be run on
+   * @param {Object} changedProps Bag of changed properties
+   * @param {Object} oldProps Bag of previous values for changed properties
+   * @return {Object} Bag of newly computed properties from "computed" effects
+   */
```
kevinpschaaf: Moved to `property-effects` from `batched-effects`, no changes

src/properties/property-effects.html
```diff
@@ -183,7 +394,46 @@
   function runComputedEffect(inst, property, value, old, info) {
     var result = runMethodEffect(inst, property, value, old, info);
     var computedProp = info.methodInfo;
-    inst._setPropertyFromComputation(computedProp, result);
+    if (inst._hasPropertyEffect(computedProp)) {
+      inst._setPendingProperty(computedProp, result);
+    } else {
+      inst[computedProp] = result;
+    }
+  }
```
kevinpschaaf: 👆 No longer calls out to instance method for overriding.

src/properties/property-effects.html
```diff
@@ -983,25 +1217,6 @@
         }
       }
 
-      /**
-       * Adds to the default implementation in `PropertyAccessors` by clearing
-       * any locally cached path values if a root object has been set, as that
-       * invalidates any descendant paths of that object.
-       *
-       * @override
-       */
```
kevinpschaaf: Clear path no longer needed; replaced with temp `__dataTemp` cache clearing in `_flushProperties`

src/properties/property-effects.html
```diff
@@ -1131,23 +1346,32 @@
...
        * @param {string} path Path to set
        * @param {*} value Value to set
+       * @param {boolean} fromPath If the value being set was from a path; in
+       *   this case the value was shared, so no dirty check is performed.
        * @return {?string} If the root property was managed, the normalized
        *   string representation of the path, otherwise returns falsey.
        * @protected
        */
-      _setPathOrUnmanagedProperty(path, value) {
+      _setPathOrUnmanagedProperty(path, value, fromPath) {
```
kevinpschaaf: Changed to perform a dirty check of a path against the actual object it refers to, and skip setting & return null if the value hasn't changed.  This puts the dirty check for path changes at the entry point (this is only called from public entry points), allowing per-instance caching of paths to be eliminated.

The `fromPath` argument indicates this method is being called as a result of a path notification, meaning it's a set to a shared object and has hence the dirty check should be skipped as it was already dirty checked upon entry.

src/properties/property-effects.html
```diff
@@ -1176,6 +1400,60 @@
       }
 
       /**
+       * Enqueues the given client on a list of pending clients, whose
+       * pending property changes can later be flushed via a call to
+       * `_flushClients`.
+       *
+       * @param {Object} client PropertyEffects client to enqueue
+       * @protected
+       */
```
kevinpschaaf: Moved from `batched-effects` to `property-effects`. No changes.

src/properties/property-effects.html
```diff
@@ -1176,6 +1400,60 @@
...
+          this.__dataPendingClients.set(client, true);
+        }
+      }
+
+      /**
+       * Flushes any clients previously enqueued via `_enqueueClient`, causing
+       * their `_flushProperties` method to run.
+       *
+       * @protected
+       */
```
kevinpschaaf: Moved from `batched-effects` to `property-effects`. Only change is to reset `__dataPendingClients` before iterating.

src/properties/property-effects.html
```diff
@@ -1254,16 +1532,35 @@
...
+       * a specific order (compute, propagate, reflect, observe, notify).
        *
        * @override
        */
       _propertiesChanged(currentProps, changedProps, oldProps) {
-        for (let p in changedProps) {
-          let effects = this.__propertyEffects[p];
-          runEffects(this, p, changedProps[p], oldProps[p], effects);
-        }
+        // ----------------------------
```
kevinpschaaf: Moved from `batched-effects` to `property-effects`.  Simplified to remove handling of notify re-entry support, and moves notify to end.

src/templatizer/dom-bind.html
```diff
@@ -59,7 +59,7 @@
           this._flushProperties(this);
         }
         this._insertChildren();
-        this.dispatchEvent(new CustomEvent('dom-change'));
+        this.dispatchEvent(new CustomEvent('dom-change', {bubbles: true}));
```
kevinpschaaf: In 1.0 this was a bubbling event.

src/templatizer/dom-repeat.html
sorvell: Rename to `handleItemPath`: returns true if item path was handled. Also rename `_checkObservedPaths` to `_handleObservedPaths` and give that the same contract.

src/properties/property-accessors.html
```diff
@@ -146,17 +148,40 @@
...
+        // Dirty check against temporary cache for paths & objects, and normal
+        // data storage for primitive properties; the dirty-check is important
+        // to prevent loops due to two-way notification, but the temporary cache
```
kevinpschaaf: Done

src/properties/property-accessors.html
```diff
@@ -146,17 +148,40 @@
...
+        // is cleared at end of turn since:
+        // a.) any cached array paths (e.g. 'array.3.prop') may be invalidated
+        //     due to array mutations like shift/unshift/splice; this is fine
+        //     since path changes are dirty-checked at user entry points like `set`
```
kevinpschaaf: Done. https://github.com/Polymer/polymer/issues/4227

src/properties/property-effects.html
```diff
@@ -282,8 +549,18 @@
       // Property binding
       let prop = info.name;
       if (node._hasPropertyEffect && node._hasPropertyEffect(prop)) {
-        inst._setPropertyToNodeFromAnnotation(node, prop, value);
-      } else if (inst._shouldPropChange(prop, node[prop], value)) {
+        setPropertyToNodeFromAnnotation(inst, node, prop, value);
```
kevinpschaaf: Done. `Polymer.sanitizeDOMValue`

src/properties/property-effects.html
```diff
@@ -282,8 +549,18 @@
...
+      // The `else` clause is for interop: binding to a non-Polymer element's
+      // property. Some native element properties have side-effects when
+      // re-setting the same value (e.g. setting <input>.value resets the
+      // cursor position), so we do a dirty-check before setting the value.
+      // However, for better interop with non-Polymer custom elements that
+      // accept objects, we explicitly re-set object changes coming from the
+      // Polymer world (which may include deep object changes without the
+      // top reference changing), erring on the side of providing more
+      // information
+      } else if (inst._shouldPropChange(prop, node[prop], value) ||
```
kevinpschaaf: Done

src/properties/property-effects.html
```diff
@@ -1131,23 +1346,32 @@
...
        *   string representation of the path, otherwise returns falsey.
        * @protected
        */
-      _setPathOrUnmanagedProperty(path, value) {
+      _setPathOrUnmanagedProperty(path, value, fromPath) {
         let rootProperty = Polymer.Path.root(Array.isArray(path) ? path[0] : path);
         let hasEffect = this._hasPropertyEffect(rootProperty);
         let isPath = (rootProperty !== path);
-        if (!hasEffect || isPath) {
+        if ((!hasEffect || isPath) && !fromPath) {
```
kevinpschaaf: Done

src/templatizer/dom-repeat.html
kevinpschaaf: Done

src/properties/property-effects.html
```diff
@@ -1131,30 +1373,99 @@
...
         let hasEffect = this._hasPropertyEffect(rootProperty);
         let isPath = (rootProperty !== path);
-        if (!hasEffect || isPath) {
+        if (!hasEffect) {
+          Polymer.Path.set(this, path, value);
+        } else if (isPath && !fromPath) {
+          let old = Polymer.Path.get(this, path);
           path = Polymer.Path.set(this, path, value);
-        }
+          if (!this._shouldPropChange(path, value, old)) {
```
sorvell: Let's add a comment here describing what's going on!

src/properties/property-effects.html
```diff
@@ -1131,30 +1373,99 @@
...
+       * user code); we could introduce a "fixup" for temporarily cached array
+       * paths if needed: https://github.com/Polymer/polymer/issues/4227
+       *
+       * @override
+       */
+      _setPendingProperty(property, value) {
+        let isPath = Polymer.Path.isPath(property);
+        let isObject = typeof value == 'object';
+        let prevProps = isPath || isObject ? this.__dataTemp : this.__data;
+        if (this._shouldPropChange(property, value, prevProps[property])) {
```
sorvell: Does this make sense to expose anymore?

sorvell:
LGTM 👍

PR #4223, opened 2016-12-16
replace CustomElements.takeRecords with customElements.flush

notwaldorf:
Because @kevinpschaaf said so.


notwaldorf:
The 1 failed test seems borked on master too

notwaldorf:
YASSSSS thanks

Issue #3336, opened 2016-01-27
scopeSubtree duplicates classes on successive runs


miztroh:
I've been developing a WYSIWYG editor with undo / redo management.  As part of the setup, I call `this.scopeSubtree(node, true)` where `this` is the editor custom element and `node` is its local DOM `contenteditable` node.  This helps ensure that the editable content is properly scoped for styles.

First, I start with some basic content:
`<p><a href="#">Hello</a>, World!</p>`

The scopeSubtree MutationObserver does its thing and updates the content to this:
`<p class="style-scope wysiwyg-editable"><a href="#" class="style-scope wysiwyg-editable">Hello</a>, World!</p>`

So far, so good.  I keep this version as the first in history and I discard the unscoped version.  However, look at what happens when I make my first edit by selecting this content and bolding it:

`<p class="style-scope wysiwyg-editable style-scope wysiwyg-editable"><b class="style-scope wysiwyg-editable"><a href="#" class="style-scope wysiwyg-editable style-scope wysiwyg-editable style-scope wysiwyg-editable style-scope wysiwyg-editable">Hello</a>, World!</b></p>`

As you can see, the classes `style-scope wysiwyg-editable` are getting applied in an additive fashion, so older nodes have multiple occurences of them.  `scopeSubtree` needs to allow for only one occurrence of each class per node.

miztroh:
Here's a JSBin: http://jsbin.com/kuyirimiti/edit?html,output

nazar-pc:
If you specify second argument as `true` - you need to call it only once on parent element and everything inside will be automatically handled by MutationObserver afterwards. Code itself doesn't have any protection from repeated calls (I've being looking for this a while ago as well).

miztroh:
Already aware of the second argument, as mentioned above.

miztroh:
Ping.

n1ywb:
Can confirm, this behavior is insane

![image](https://cloud.githubusercontent.com/assets/316961/21355768/4199bf28-c69d-11e6-9e6c-61d4a63914a6.png)

TimvdLippe:
This should be fixed in ShadyCSS: http://jsbin.com/jedorakoxu/edit?html,console,output working in both Firefox and Chrome.

PR #4229, opened 2016-12-20
3336 use classList.add to avoid duplicate classes

n1ywb:
### Reference Issue

Fixes https://github.com/Polymer/polymer/issues/3336

in `scopeSubtree()` use `classList.add('class')` instead of `setAttribute('class', ...)` when setting scoped style classes to avoid adding duplicate classes

duplicates can happen when the observer fires e.g. when D3 does weird things

### DIFF
--- a/src/standard/styling.html
+++ b/src/standard/styling.html
@@ -166,12 +166,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var self = this;
         var scopify = function(node) {
           if (node.nodeType === Node.ELEMENT_NODE) {
-            var className = node.getAttribute('class');
-            node.setAttribute('class', self._scopeElementClass(node, className));
+            node.classList.add(Polymer.StyleTransformer.SCOPE_NAME, self.is);
             var n$ = node.querySelectorAll('*');
             for (var i=0, n; (i<n$.length) && (n=n$[i]); i++) {
-              className = n.getAttribute('class');
-              n.setAttribute('class', self._scopeElementClass(n, className));
+              n.classList.add(Polymer.StyleTransformer.SCOPE_NAME, self.is);
             }
           }
         };


n1ywb:
I authored the commits under my jeff@jefflaughlinconsulting.com email address; please let me know if I need to do anything else WRT CLA (preferably you can add my address to it)

n1ywb:
I added a CLA for my jeff@jefflaughlinconsulting.com address

arthurevans:
I believe the  use of `setAttribute` was added to make `scopeSubtree` work for SVG, so we'd have to make sure this fix passes the existing unit tests on all browsers.

Original fix that introduced `setAttribute` is here, for what it's worth:

https://github.com/Polymer/polymer/pull/3085

IE11 lacks support for `classList` on SVG and MathML elements:

http://caniuse.com/#search=classList

Maybe we could conditionally use `classList` on non-SVG elements? (Or conditionally use `setAttribute` when we need it.) WDYT @azakus?

n1ywb:
or emulate it by tokenizing the class? Polyfill even?

On Dec 21, 2016 4:24 PM, "Arthur Evans" <notifications@github.com> wrote:

I believe the use of setAttribute was added to make scopeSubtree work for
SVG, so we'd have to make sure this fix passes the existing unit tests on
all browsers.

Original fix that introduced setAttribute is here, for what it's worth:

#3085 <https://github.com/Polymer/polymer/pull/3085>

IE11 lacks support for classList on SVG and MathML elements:

http://caniuse.com/#search=classList

Maybe we could conditionally use classList on non-SVG elements? (Or
conditionally use setAttribute when we need it.) WDYT @azakus
<https://github.com/azakus>?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Polymer/polymer/pull/4229#issuecomment-268643771>, or mute
the thread

TimvdLippe:
Closing per https://github.com/Polymer/polymer/issues/3336#issuecomment-330706945 This has been fixed in ShadyCSS.

PR #4232, opened 2016-12-21
Clicks on labels don't check their corresponding inputs

tmaroschik:
Example here https://jsbin.com/dolayuhatu/edit?html,js,console,output

If you click on the checkbox one click event is triggered. If you click on the label, two click events are fired. One on the label, one separate on the input field. If default is prevented on the input field, the checkbox won't be checked.

In Polymer in the Gesture handler, the mouseCanceler takes care of canceling ghost clicks. As the input field is most likely not in the event path of the label, the input click event default is prevented which results in a prevented checked state, at least in Safari on iOS10.

The fix resolves the forward target of the label click and saves it until the second click event arrives. If it's stored it just skips further canceling.

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -72,7 +72,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     if (mouseEvent.type === 'click') {
       var path = Polymer.dom(mouseEvent).path;
       for (var i = 0; i < path.length; i++) {
-        if (path[i] === POINTERSTATE.mouse.target) {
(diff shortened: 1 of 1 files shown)


tmaroschik:
The CLA: I signed it! Our corporation is DFAU.

tmaroschik:
This pull request is still valid although it's targeted for 1.x. Is it interesting? I really need this stuff in our projects.

TimvdLippe:
If I recall correctly this has been fixed along the way. Given this PR has a large merge conflict, I am closing this for now. Feel free to reopen a PR targeting either 2.x or 3.x if it has not.

PR #4249, opened 2017-01-05
Fixes #4247: limit simple property observer callbacks

### LINKED ISSUE #4247: Polymer 2.0 -- simple property observers are called too eagerly
### Description
Simple property observers get called when modifying any multipart path rooted in that property.
( may be related to #4072 )

#### Live Demo

Demo (and proposed fix -- uncomment the line with the `patch()`) here https://jsbin.com/wukekanevi/edit?html,output

Should it be necessary, the original raw html is below:
```
<!doctype html>
<html>
<head>
  <base href="https://polygit.org/polymer+:2.0-preview/webcomponentsjs+:v1/shadydom+webcomponents+:master/shadycss+webcomponents+:master/custom-elements+webcomponents+:master/components/">
  <script src="webcomponentsjs/webcomponents-lite.js"></script>
  <link rel="import" href="polymer/polymer.html">
</head>
<body>

  <dom-module id="x-el">
    <template>
      <div id="log">objChanged Log:
        <template is="dom-repeat" items="{{log}}">
          <li>{{item}}</li> 
        </template>
      </div>
    </template>
    <script>
      
      //debugger;
      //patch() // TODO: uncomment this line to apply probable fix
      
      function patch() {
        
        Polymer.Base._createObservedProperty = function (property, methodName) {
          this._addPropertyEffect(property, Polymer.Base.PROPERTY_EFFECT_TYPES.OBSERVE, {
            fn: runObserverEffectPatched,
            info: {
              strictlyMatchProperty: property,
              methodName: methodName
            }
          });
        };
        
        function runObserverEffectPatched (inst, property, value, old, info) {
          let fn = inst[info.methodName];
          if (fn) {
            if (info.strictlyMatchProperty === property) {
              fn.call(inst, value, old);
            }
          } else {
            console.warn('observer method `' + info.methodName + '` not defined');
          }
        }
      }
      
      Polymer({
        is: 'x-el',
        properties: {
          log: {
            type: Array,
            value: function () {
              return [];
            },
          },
          obj: {
            type: Object,
            value: function () {
              return {deeply:{nested:{field:-1}}};
            },
            observer:'objChanged'
          },
        },

        objChanged: function(young, old) {
          this.push("log", [
            "objChanged("+arguments.length+")",
            JSON.stringify(young) || "undefined",
            JSON.stringify(old) || "undefined",
          ].join(" , "))
        },
      });
    </script>
  </dom-module>
  
  <x-el></x-el>
  
  <script>
    //HTMLImports.whenReady(function() {
      let el = document.querySelector('x-el');
      let id = !"topLevelChange" ?
        setInterval(function () {
          let value = 0|el.obj.deeply.nested.field
          el.obj = {deeply:{nested:{field:value + 3}}};
        }, 1000) :
        setInterval(function () {
          let value = 0|el.obj.deeply.nested.field;
          el.set("obj.deeply.nested.field", String(value + 1));
        }, 1000);
      
      setTimeout(function () { clearInterval(id) }, 10000)
    //});
  </script>
  
</body>
</html>
```
#### Steps to Reproduce
1. Load the demo
2. Wait several seconds (~10seconds)
3. Observer the browser output panel on the right (which will generally display the arguments passed to the `_objChanged()` "observer" method)

#### Expected Results
```
objChanged Log:
* objChanged(2) , {"deeply":{"nested":{"field":-1}}} , undefined
```

#### Actual Results
```
objChanged Log:
* objChanged(2) , {"deeply":{"nested":{"field":-1}}} , undefined
* objChanged(2) , "0" , undefined
* objChanged(2) , "1" , undefined
* objChanged(2) , "2" , undefined
* objChanged(2) , "3" , undefined
* objChanged(2) , "4" , undefined
* objChanged(2) , "5" , undefined
* objChanged(2) , "6" , undefined
* objChanged(2) , "7" , undefined
* objChanged(2) , "8" , undefined
* objChanged(2) , "9" , undefined
```

### Versions
- Polymer: 2.0-preview
- webcomponents: v1

---

Under the assumption that the intended behavior in fact matches what my `patch()` provides (i.e. the existing behavior is not expected nor does it reflect any changes in the [data-system](https://github.com/Polymer/polymer/tree/2.0-preview#data-system)), I'll open up a PR with the proposed fix.

jcmoore:
-	prevent deep/multi-part-path modification from triggering simple observer callbacks

### Reference Issue
Fixes #4247

### DIFF
--- a/src/properties/property-effects.html
+++ b/src/properties/property-effects.html
@@ -167,7 +167,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   function runObserverEffect(inst, property, value, old, info) {
     let fn = inst[info.methodName];
     if (fn) {
-      fn.call(inst, value, old);
+      if (info.property === property) {
+        fn.call(inst, value, old);
+      }
     } else {
       console.warn('observer method `' + info.methodName + '` not defined');
     }
@@ -2007,6 +2009,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         this._addPropertyEffect(property, TYPES.OBSERVE, {
           fn: runObserverEffect,
           info: {
+            property: property,
             methodName: methodName
           }
         });
--- a/test/unit/notify-path-elements.html
+++ b/test/unit/notify-path-elements.html
@@ -81,6 +81,15 @@
     Polymer({
       is: 'x-stuff',
       properties: {
+        simplyObserved: {
+          type: Object,
+          value: function () {
+            return {
+              nestedField: "originalArbitraryValue",
+            };
+          },
+          observer: "simplyObservedChanged",
+        },
         computedFromPaths: {
           computed: 'computeFromPaths(a, nested.b, nested.obj.c)'
         }
@@ -121,6 +130,7 @@
         this.aChanged = sinon.spy();
         this.xChanged = sinon.spy();
         this.yChanged = sinon.spy();
+        this.simplyObservedChanged = sinon.spy();
         if (this.$) {
           this.$.compose.resetObservers();
           this.$.forward.resetObservers();
--- a/test/unit/notify-path.html
+++ b/test/unit/notify-path.html
@@ -430,6 +430,31 @@ suite('path effects', function() {
     assert.equal(el.$.boundChild.computedFromPaths, 100);
   });
 
+  test('simple property observer with shallow modification', function() {
+    // Setup
+    var simplyObservedOld = el.simplyObserved;
+    var simplyObserved = {
+      nestedField: "differentArbitraryValue"
+    };
+    el.resetObservers();
+    // Do the thing
+    el.simplyObserved = simplyObserved;
+    // Verify
+    assert.equal(el.simplyObservedChanged.callCount, 1);
+    assert.equal(el.simplyObservedChanged.firstCall.args[0], simplyObserved);
+    assert.equal(el.simplyObservedChanged.firstCall.args[1], simplyObservedOld);
+  });
+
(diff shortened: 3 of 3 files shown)


jcmoore:
Hey @kevinpschaaf , I noticed that this general behavior also exists for all "method-effects" (methodObservers, methodBindings, and computedProperties).

First of all, do you know if someone is actively working on this (or is the current behavior the expected behavior)?  If so, I'm happy to stop now -- but if not...

I think I have a fix for this if you will take a look at https://github.com/jcmoore/polymer/commit/018e6c9cab2e3a224135a393c4788ddd0e9033d8

Also, in regards to the changes I'm proposing, can you please evaluate whether or not some of your `BREAKME` comments* should be modified/removed (they confused me a bit)?

If the changes are acceptable, I will include those changes in this PR.

---
(*) specifically ["TODO/BREAKME(kschaaf): non-wildcard object observers now called for deep path changes"](https://github.com/Polymer/polymer/blob/2ca05290556b0c02846dc5cc7cf15943a23ffd18/test/unit/notify-path.html#L57-L58), ["BREAKME(kschaaf): order of args doesn't make a difference when using setProperties"](https://github.com/Polymer/polymer/blob/2ca05290556b0c02846dc5cc7cf15943a23ffd18/test/unit/notify-path.html#L355), and ["BREAKME(kschaaf): arg order doesn't matter with set properties"](https://github.com/Polymer/polymer/blob/2ca05290556b0c02846dc5cc7cf15943a23ffd18/test/unit/notify-path.html#L462)

jcmoore:
So things are a bit in flux still then.  I understand, thank you

PR #4255, opened 2017-01-13
depends on test-fixture ^2.0.1

e111077:
Trying to delete that branch on `test-fixture` which is now merged into `2.0.0`


kevinpschaaf:
@e111077 good to go

PR #4256, opened 2017-01-13
Make Polymer 2.0 compatible with Closure Compiler

azakus:
With this PR, Polymer is now compatible with Closure Compiler's ADVANCED optimization setting.
However, Closure Compiler is not yet compatible with Polymer's use of ES6 class mixins, and currently warn.
Once that is implemented in Closure Compiler, we will likely see more warnings.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -33,7 +33,10 @@ const ENTRY_POINTS = [POLYMER_LEGACY, POLYMER_ELEMENT];
 
 const polymer = require('polymer-build');
 const PolymerProject = polymer.PolymerProject;
-const project = new PolymerProject({ shell: DEFAULT_BUILD_TARGET });
+const project = new PolymerProject({
+  sources: ['./polymer.html'],
+  shell: './polymer.html'
+});
 const fork = polymer.forkStream;
 
 const mergeStream = require('merge-stream');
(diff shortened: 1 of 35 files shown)


src/elements/element.html
```diff
@@ -113,16 +118,16 @@
...
             Polymer.telemetry.register(proto);
           }
           let config = this._ownConfig;
           if (config) {
             this._finalizeConfig(config);
           }
-          if (this.template) {
-            let template = this.template.cloneNode(true);
-            this._finalizeTemplate(template);
+          let template = this.template;
```
sorvell: remove this symbol and do `this.template.cloneNode(true)` where `_finalizeTemplate` is called?

src/elements/element.html
```diff
@@ -141,11 +146,15 @@
...
+        return Polymer.DomModule.import(this.is, 'template');
+      }
+
       static get template() {
-        if (!this.hasOwnProperty('_template')) {
+        if (!this.hasOwnProperty(goog.reflect.objectProperty('_template', this))) {
           // TODO(sorvell): support more ways to acquire template.
           // this requires `is` on constructor...
-          this._template = Polymer.DomModule.import(this.is, 'template') ||
+          this._template = this._getTemplate() ||
```
sorvell: remove the TODO here... seems outdated.

src/properties/property-effects.html
```diff
@@ -1956,23 +1972,27 @@
...
         if (arguments.length == 1) {
           // Get value if not supplied
           let info = {};
           value = Polymer.Path.get(this, path, info);
-          path = info.path;
+          propPath = info.path;
         } else if (Array.isArray(path)) {
           // Normalize path if needed
-          path = Polymer.Path.normalize(path);
+          propPath = /** @type {string} */(Polymer.Path.normalize(path));
```
sorvell: Let's consider making `normalize` always return a string to avoid this cast.

azakus: Done, added `split` to do the work of old `normalize(path, true)`

src/utils/polymer.dom.html
```diff
@@ -36,6 +36,11 @@
   class EffectiveNodesObserver {
 
     constructor(target, callback) {
+      /** @type {MutationObserver} */
```
kevinpschaaf: Question: Why doesn't the type need to be `{MutationObserver=}` or something to say it can be undefined (given that the property is clearly undefined for some time since it doesn't gain a value in the constructor)?  FWIW, these can just be nulled in the ctor if that is more idiomatic.

azakus: The `=` in `{MutationObserver=}` is only valid for function parameters. I'm not 100% sure why this is ok in the type system of Closure, but I'd guess that Closure really only tracks assignments. Given that I've only declared the variable, I guess Closure just takes my word for it.

src/utils/boot.html
```diff
@@ -12,4 +12,13 @@
   window.Polymer = window.Polymer || {};
   window.Polymer.version = '2.0-preview';
 
+  /* eslint-disable no-unused-vars */
```
azakus: Done

gulpfile.js
```diff
@@ -8,10 +8,9 @@
  * subject to an additional IP rights grant found at http:polymer.github.io/PATENTS.txt
```
azakus: Done

src/template/annotations.html
```diff
@@ -217,7 +217,7 @@
     // TODO(sjmiles): is this for non-ELEMENT nodes? If so, we should
```
azakus: Done

PR #4268, opened 2017-01-21
Fix `strip-whitespace` for nested templates.

sorvell:


kevinpschaaf:
Error is due to webcomponents regression; to be fixed separately.

Issue #4181, opened 2016-11-29
Provide an api for converting a 1.x behavior into a 2.x mixin


sorvell:
Users writing 2.x elements in class based syntax may want to use Polymer 1.x behaviors. This could be made possible via an api that wraps the behavior in a class expression mixin.

sorvell:
This has been exposed as `Polymer.mixinBehaviors([...behaviors...], baseClass)`. 

Leaving this open so it can be documented.

kevinpschaaf:
Closing, as we have a separate tracker for pending API documentation (and it's already been documented in our API documentation branch: https://github.com/Polymer/polymer/compare/2.0-preview...2.0-doc-annotations#diff-ac5771b3a44bf972bca119c60ec3baecR46)

PR #4280, opened 2017-01-30
2.0 mixin behaviors

sorvell:
Adds `Polymer.MixinBehaviors` to allow behaviors to be used with classes and optimizes legacy `Polymer({...})` syntax somewhat.


miztroh:
@sorvell Can you provide an example of how to utilize existing behaviors inside a new class-based element definition?

arthurevans:
@miztroh I believe this would be like this:

```
class XClass extends Polymer.mixinBehaviors([MyBehavior, MyBehavior2], Polymer.Element) {
  static get is() { return 'x-class'}

  ...
}
customElements.define(XClass.is, XClass);
```

The `mixinBehavior` function also mixes in the Legacy APIs, the same as if you extended `Polymer.LegacyElement`. These APIs are required since since hybrid behaviors depend on them.

miztroh:
@arthurevans Thanks!

arthurevans:
I think this should resolve issue #4181? @kevinpschaaf

PR #4281, opened 2017-01-30
[2.x] Data-system perf micro-optimizations

kevinpschaaf:
Handful of data-subsystem performance optimizations

### DIFF
--- a/src/properties/property-effects.html
+++ b/src/properties/property-effects.html
@@ -1143,9 +1143,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    */
   function notifySplices(inst, array, path, splices) {
     let splicesPath = path + '.splices';
-    inst._setPendingPropertyExternal(splicesPath, { indexSplices: splices });
-    inst._setPendingPropertyExternal(path + '.length', array.length);
-    inst._invalidateProperties();
+    inst.notifyPath(splicesPath, { indexSplices: splices });
+    inst.notifyPath(path + '.length', array.length);
     // Null here to allow potentially large splice records to be GC'ed.
     inst.__data[splicesPath] = {indexSplices: null};
   }
@@ -1986,7 +1985,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         } else {
           propPath = /** @type{string} */(path);
         }
-        this._setProperty(propPath, value);
+        if (this._setPendingPropertyExternal(path, value, true)) {
+          this._invalidateProperties();
+        }
       }
 
       /**
--- a/src/properties/property-effects.html
+++ b/src/properties/property-effects.html
@@ -176,9 +176,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    */
   function runNotifyEffects(inst, props, oldProps) {
     // Notify
-    let hasPaths;
     let fxs = inst.__notifyEffects;
-    if (fxs || (hasPaths = inst.__dataHasPathsCounter)) {
+    let hasPaths = inst.__dataHasPathsCounter;
+    if (fxs || hasPaths) {
       let notified;
       let id = effectUid++;
       // Try normal notify effects; if none, fall back to try path notification
@@ -312,9 +312,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     let detail = event.detail;
     if (detail && detail.queueProperty) {
       if (!inst._hasReadOnlyEffect(path)) {
-        if (inst._setPendingPropertyExternal(path, value, Boolean(detail.path))) {
-          inst._invalidateProperties();
-        }
+        inst._setPendingPropertyExternal(path, value, Boolean(detail.path));
       }
     } else {
       inst.set(path, value);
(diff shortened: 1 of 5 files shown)


src/properties/property-effects.html
```diff
@@ -190,26 +176,28 @@
...
+      // "notify" effects AND (2) if we knew there were no paths to notify. This
+      // 2nd piece of info is not currently available but could be added perhaps
+      // in `_setProperty`.
+      for (let prop in props) {
+        if (notifyEffects && runEffectsForProperty(inst, notifyEffects, id,
+          prop, props, oldProps)) {
+          notified = true;
+        } else if (notifyPath(inst, prop, props)) {
+          notified = true;
+        }
```
TimvdLippe: Can be shortened to

```js
if (notifyEffects) {
  notified = runEffectsForProperty(inst, notifyEffects, id, prop, props, oldProps);
} else {
  notified = notifyPath(inst, prop, props);
}
```

It is slightly more readable and formatted imo

kevinpschaaf: The intention is that `notified` latches on if _any_ property in the loop over `props` resulted in notification; this code would re-assign `notified` (potentially back to `false`) each prop.

The only purpose of the `nofified` flag is to prevent a frivolous call to `host_flushProperties`, so I might benchmark eliminating that bookkeeping and see if its worth keeping around though.

src/properties/property-effects.html
```diff
@@ -190,26 +176,28 @@
...
-    // 2nd piece of info is not currently available but could be added perhaps
-    // in `_setProperty`.
-    for (let prop in props) {
-      if (notifyEffects && runEffectsForProperty(inst, notifyEffects, id,
-        prop, props[prop], oldProps && oldProps[prop])) {
-        notified = true;
-      } else if (notifyPath(inst, prop, props[prop])) {
-        notified = true;
+    let notifyEffects = inst.__notifyEffects;
+    if (notifyEffects) {
```
kevinpschaaf: To investigate: seems like we shouldn't be able to get away with this perf optimization because the 1.x contract was that path notifications always fire (regardless of existence of a notify effect for the root property, which is the whole reason notify effects are special cased in this function).  I put this in provisionally while I was looking at separating path & property code, but need to look into why this isn't causing tests to fail.  I think `if (notifyEffects || this.__dataHasPathsCounter)` is probably enough to avoid the `for` loop over props if there are no notify effects.

src/properties/property-effects.html
TimvdLippe: Ah okay thanks for the explanation! For the record: my snippet the `=` should then have been updated to `|=`

src/properties/property-effects.html
```diff
@@ -1397,33 +1380,35 @@
        *   notification of change to a shared path).
        * @protected
        */
-      _setPathOrUnmanagedProperty(path, value, fromPath) {
+      _setPendingPropertyExternal(path, value, fromPath) {
```
sorvell: Add note about who calls this. This is the entry point where paths (slow path) can be switched on.

src/properties/property-effects.html
```diff
@@ -1397,33 +1380,35 @@
...
+              if (!path || !super._shouldPropertyChange(path, value, old)) {
+                return false;
+              }
+            }
+            this.__dataHasPaths = true;
+          }
+          return this._setPendingProperty(path, value);
         } else {
-          return undefined;
+          Polymer.Path.set(this, path, value);
```
sorvell: use `isPath` here to gate `set` and otherwise just set the property `this[path] = value`

src/properties/property-effects.html
```diff
@@ -1397,33 +1380,35 @@
...
-          path = /** @type {string} */(Polymer.Path.set(this, path, value));
-          // Use property-accessor's simpler dirty check
-          if (!super._shouldPropertyChange(path, value, old)) {
-            return undefined;
-          }
-        }
         if (hasEffect) {
-          return /** @type {string} */(path);
+          if (isPath) {
+            if (!fromPath) {
```
sorvell: Consider changing to `isPathNotification`. If this flag is false then we need to set the path.

src/properties/property-effects.html
```diff
@@ -1523,11 +1508,11 @@
        */
       _invalidateProperties() {
         if (this.__dataInitialized) {
-          if (this._asyncEffects) {
-            super._invalidateProperties();
-          } else {
+          // if (this._asyncEffects) {
```
sorvell: Remove commented code. Add note: propertyAccessors takes changes async and propertyEffects does not.

src/properties/property-effects.html
```diff
@@ -117,30 +115,18 @@
    * @param {*} old Previous value of changed property
    * @private
    */
-  function runEffectsForProperty(inst, effects, id, prop, value, old) {
+  function runEffectsForProperty(inst, effects, id, prop, props, oldProps) {
     let ran;
     // TODO(kschaaf) ideally a system exists to parse path information once
     // and send structured information through the system for better perf
-    let rootProperty = Polymer.Path.root(prop);
+    let rootProperty = inst.__dataHasPathsCounter ? Polymer.Path.root(prop) : prop;
```
sorvell: Consider instead threading `hasPaths` through where needed rather than maintaining a counter.

PR #4283, opened 2017-01-31
Flush wrapper before template & default dom-if to sync (async optional).

### LINKED ISSUE #2712: Elements inside dom-if receive data binding updates even if the update is about to result in the dom-if deleting or hiding them
If a dom-if with restamp="true" (which should delete its children while false) is used to guard elements from receiving data they can't or shouldn't be seeing, it works for the initial pass, but if the data is updated such that a dom-if that evaluated to true (and already created its child elements) now evaluates to false, the new data is still passed to its children just before they are deleted. This results in an error condition if one is depending on the dom-if as a guard. This jsbin demonstrates the issue - if the dom-if deleted its children as soon as it evaluated to false, before they received any update, neither Larry nor Moe should ever sneeze:

http://jsbin.com/kesexa/3/edit?html,console,output

```
<!DOCTYPE html>
<html>
  <head>
    <title>polymer</title>
    <script src="https://rawgit.com/webcomponents/webcomponentsjs/master/webcomponents-lite.js"></script>
    <base href="http://polygit.org/polymer/components/">
    <link href="polymer/polymer.html" rel="import">
  </head>
  <body>

    <dom-module id="x-test">
      <template>

        <template is="dom-if" if="{{isLily(flower)}}" restamp="true">
          <x-larry flower="{{flower}}"></x-larry>
        </template>

        <template is="dom-if" if="{{isDaisy(flower)}}" restamp="true">
          <x-moe flower="{{flower}}"></x-moe>
        </template>

      </template>
    </dom-module>

    <script>
        Polymer({
          is: 'x-larry',
          properties: {
            flower: {
              type: String,
              observer: 'sniff'
            }
          },
          sniff: function (flower) {
            if (flower === 'daisy') {
              console.log('Achoo!') // Larry is allergic to daisies
            }
          }
        });
        Polymer({
          is: 'x-moe',
          properties: {
            flower: {
              type: String,
              observer: 'sniff'
            }
          },
          sniff: function (flower) {
            if (flower === 'lily') {
              console.log('Achoo!') // Moe is allergic to lilies
            }
          }
        });
        Polymer({
          is: 'x-test',
          properties: {
            flower: {
              type: String,
              value: 'daisy'
            }
          },
          ready: function () {
            var that = this
            setTimeout(function () {
              that.flower = 'lily'
            }, 1000);
          },
          isLily: function (flower) {
            return flower === 'lily'
          },
          isDaisy: function (flower) {
            return flower === 'daisy'
          }
        });
      });
    </script>

    <x-test></x-test>
  </body>
</html>
```

kevinpschaaf:
### Reference Issue
Fixes #2712

### Description
In order to ensure template wrapper elements like `<dom-if>` process their changes (e.g. changes to the `if` property) prior to `_host_...` properties being forwarded to instances, we override `_flushProperties` on  the `<template>` to explicitly flush the wrapper first.  In the context of `_flushClients`, if the template happens to go before the wrapper, the wrapper will be flushed first via this mechanism, and no-op later when `_flushClients` gets to it in its order.

Along with this change, `dom-if` stamps synchronously by default, and gains an `async` flag for backwards-compatible timing if needed for performance.  Async stamping is assumed to be much less necessary in 2.0 now that changes are batched.  Synchronously processing the `if` by default means that in the `restamp` case we ensures the instance is torn down before host properties are forwarded to it.

Considerations to discuss during review:
* Requiring sync & restamp to fix #2712 is not strictly necessary, however this dramatically simplifies things
  * Host property forwarding happens sync; squelching forwarding based on `if: false` means we may need to re-sync all (?) host properties in the case `if` becomes true again and the instance is reused (either due to `restamp: false` or if `if` becomes true again before the debouncer runs with `restamp: true`)
  * The current threading seems good: Host property forwarding is only squelched when there is no instance to forward to, but with `restamp: true` and the default sync behavior, the instance is removed before the forwarding occurs
* The wrapper check is added in the flush hot-path, but should be minimal
* Want to consider whether `templatize` should set `__wrapper`

### DIFF
--- a/src/templatizer/dom-if.html
+++ b/src/templatizer/dom-if.html
@@ -65,7 +65,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
            */
           restamp: {
             type: Boolean,
-            observer: '_debounceRender'
+          },
+
+          /**
+           * By default, `dom-if` will stamp (or hide/remove) nodes synchronous to
+           * its `if` value changing.  To cause it to stamp asynchronously at
+           * microtask timing, set `async` to true.  This should generally only
+           * be used if the `if` value may change multiple times in a turn,
+           * to debounce unnecessary work.
+           */
+          async: {
+            type: Boolean
           }
 
         }
@@ -80,9 +90,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
 
     _debounceRender() {
-      this._renderDebouncer =
-        Polymer.Debouncer.debounce(this._renderDebouncer, this._render, undefined, this);
-      Polymer.Templatizer.enqueueDebouncer(this._renderDebouncer);
+      if (this.async) {
+        this._renderDebouncer =
+          Polymer.Debouncer.debounce(this._renderDebouncer, this._render, undefined, this);
+        Polymer.Templatizer.enqueueDebouncer(this._renderDebouncer);
+      } else {
+        this._render();
+      }
     }
 
     disconnectedCallback() {
@@ -102,11 +116,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
 
     /**
-     * Forces the element to render its content. Normally rendering is
-     * asynchronous to a provoking change. This is done for efficiency so
-     * that multiple changes trigger only a single render. The render method
-     * should be called if, for example, template rendering is required to
-     * validate application state.
+     * Forces the element to render its content (and any nested templates) when
+     * the `async` flag is in use.
      */
     render() {
       Polymer.Templatizer.flush();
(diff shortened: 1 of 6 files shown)


src/templatizer/dom-if.html
```diff
@@ -150,10 +161,10 @@
             observer.observe(this, {childList: true});
             return false;
           }
-          template.__domIf = this;
+          template.__wrapper = this;
```
sorvell: How is this different from `__dataHost`?

src/templatizer/templatizer.html
```diff
@@ -39,6 +39,15 @@
         new DataTemplate();
         PatchedHTMLTemplateElement._newInstance = null;
       }
+      _flushProperties() {
```
kevinpschaaf: Revised gambit: Don't flush wrapper early, instead squelch forwarding based on `if` and add logic for capturing invalidated properties while `if` is false and re-syncing them upon `true`.

kevinpschaaf: Implemented in https://github.com/Polymer/polymer/issues/2712

kevinpschaaf:
Closing in favor of https://github.com/Polymer/polymer/issues/2712 with fewer breaking changes.

PR #4291, opened 2017-02-06
set assetpath

valdrinkoshi:
Addresses https://github.com/webcomponents/html-imports/issues/13


src/legacy/dom-module.html
```diff
@@ -57,6 +75,7 @@
         modules[id] = this;
         lcModules[id.toLowerCase()] = this;
         this._styleOutsideTemplateCheck();
+        this._setAssetpath();
```
sorvell: I'm not sure it's worth it to always set this attribute and it's likely a small performance cost. Maybe instead we can give `dom-module` an imperative api that returns the assetpath?

valdrinkoshi: Yes, it could just be a getter. `Polymer.Element.resolveUrl` would need to be updated accordingly.

valdrinkoshi: Done, PTAL 🍻

PR #4298, opened 2017-02-08
Deduplicate mixins correctly

kevinpschaaf:


azakus:
IE 11 failure may be Set polyfill not copying correctly?

PR #4311, opened 2017-02-13
Ensure disable-upgrade elements are not "configured". Fixes #4302

### LINKED ISSUE #4302: disable-upgrade on an element with data-binding results in `Cannot read property` error if there are multiple custom elements in the template
### Description
The following error is thrown when there are 2 custom elements in a template and one has data-binding & the `disable-update` attribute applied to it:

```
polymer.html:2033 Uncaught TypeError: Cannot read property 'foo' of undefined
    at HTMLElement._configValue (polymer.html:2033)
    at HTMLElement._distributeConfig (polymer.html:2094)
    at HTMLElement._configure (polymer.html:2053)
    at HTMLElement._beforeClientsReady (polymer.html:2039)
    at HTMLElement._ready (polymer-mini.html:63)
    at HTMLElement._tryReady (polymer-mini.html:56)
    at HTMLElement._initFeatures (polymer.html:4133)
    at HTMLElement.__initialize (polymer-micro.html:233)
    at HTMLElement.createdCallback (polymer-micro.html:216)
    at window.Polymer (polymer-micro.html:62)
```

#### Live Demo
http://codepen.io/JohnRiv/pen/RKeJNx?editors=1010

See Live Demo for full code, but here's the markup that causes the error:
```html
  <dom-module id="x-bar">
    <template>
      <x-baz></x-baz> <!-- Remove <x-baz></x-baz> and there are no errors -->
      <x-foo disable-upgrade foo="[[foo]]"></x-foo>
    </template>
  </dom-module>
```

#### Steps to Reproduce
1. Define 2 basic custom elements, one of which should define a property
2. Create a custom element with a defined property and place the 2 custom elements in the element's template
3. Data-bind the property of the parent custom element to the child custom element in the template
4. Add the `disable-upgrade` attribute to the custom element 

#### Expected Results
No error is thrown

#### Actual Results
Error is thrown.
If you either remove the other custom element from the template, or remove the data-binding, there is no error thrown.

### Browsers Affected
- [x] Chrome
(I haven't tested other browsers yet)

### Versions
- Polymer: v1.8.0
- webcomponents: not applied, using Chrome Version 57.0.2987.21

kevinpschaaf:
Ensure disable-upgrade elements are not "configured".
### Reference Issue
Fixes #4302

### DIFF
--- a/src/lib/base.html
+++ b/src/lib/base.html
@@ -67,6 +67,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (settings.disableUpgradeEnabled) {
         if (this.hasAttribute('disable-upgrade')) {
           this._propertySetter = disableUpgradePropertySetter;
+          this._configValue = null;
           this.__data__ = {};
           return;
         } else {
@@ -314,6 +315,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (!this.__hasInitialized && name === 'disable-upgrade') {
         this.__hasInitialized = true;
         this._propertySetter = Polymer.Bind._modelApi._propertySetter;
+        this._configValue = Polymer.Base._configValue;
         this.__initialize();
       }
       origAttributeChangedCallback.call(this, name, oldValue, newValue);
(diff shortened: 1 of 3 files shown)


sorvell:
Edge timing out in sauce only so merging.

PR #4318, opened 2017-02-18
Continuation of 2.0 API & TODO scrub

kevinpschaaf:


jsilvermist:
What is the meaning of the double underscore vs the old single underscore?

kevinpschaaf:
@jsilvermist We're still finalizing the API documentation and final scrub over all of the documentation, but single-underscore is intended to be "protected" API intended for use by subclassers/mixin users (but not end element users) and thus still subject to semver, whereas double-underscore is intended to be private.  We will try and make this clear in the final 2.0 documentation.

kevinpschaaf:
Closing; superseded by #4331

PR #4323, opened 2017-02-21
2.0 render status compat

sorvell:
Change `Polymer.RenderStatus.afterNexRender` to have the same signature as it does in Polymer 1.x.

### DIFF
--- a/src/utils/render-status.html
+++ b/src/utils/render-status.html
@@ -54,14 +54,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (!scheduled) {
         schedule();
       }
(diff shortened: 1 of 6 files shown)


src/utils/render-status.html
```diff
@@ -32,9 +32,13 @@
   }
 
   function flushQueue(queue) {
-    for (let i=0; i<queue.length; i++) {
+    for (let i=0, q, context, callback, args; i<queue.length; i++) {
       try {
-        queue[i]();
+        q = queue[i];
```
azakus: `let [context, callback, args] = queue[i];`

azakus: Just a nit, LGTM otherwise.

PR #4327, opened 2017-02-22
1.x style include

sorvell:
Adds a setting `preserveStyleIncludes` that, when using native css properties, shadow dom, and a css build targeted at shadow dom, will expand styles included with the `include` attribute into separate style elements in shadowRoots. This has 2 benefits:
1. reduces build file size when shared styles are used
2. allows browsers to optimize shared styles by using the same style element content for them


azakus:
LGTM

PR #4328, opened 2017-02-23
Exclude SD polyfill tests for Edge due to lack of workarounds for Edg…

kevinpschaaf:
Get back to green on tests.


kevinpschaaf:
LGTM'ed from @sorvell.

Issue #3994, opened 2016-09-26
Feature request (2.0): make template getter accept string ( inline templates )


Hotell:
#### Steps to Reproduce

Currently when using module bundlers like webpack, I would like to write component template inline ( like in other frameworks Angular/React ).
In polymer 2.0, this is possible but forces to create unnecessary boilerplate. Instead it should be allowed/handled by Polymer itself.

This is also useful if user wants to have external template in separate `*.html` file and import it directly to the class via webpack loader.
#### Expected Results

``` typescript
export class MyButton extends Polymer.Element {

  static get is() { return 'my-button' }

  static get template() {
   return `<template>
        <button><slot>Hello button!</slot></button>
    </template>`;
  }
}
```
#### Actual Results

``` typescript
export class MyButton extends Polymer.Element {

  static get is() { return 'my-button' }

  static get template() {
    const template = document.createElement( 'template' );
    const content = `
        <button><slot>Hello button!</slot></button>
    `;
    template.innerHTML = content;
    return template;
  }
}
```
### Versions

- Polymer: v2.0.0-preview
- webcomponents: v1

sorvell:
Added via https://github.com/Polymer/polymer/pull/4332.

PR #4332, opened 2017-02-24
2.0 config

sorvell:
For simplicity, replace `Polymer.ElementMixin's` static `config` property with `properties` and `observers`.

Also:
* removes reliance on `static is` getter for polyfilling style scoping.
* adds support for string based templates, addressing #3994


kevinpschaaf:
Need to update README.md in a bunch of spots if we decide to go forward with this.

PR #4344, opened 2017-02-27
Enable gathering styles from link rel="import" type="css" inside element dom-module

### LINKED ISSUE #3837: Un-deprecate external styles
`<link rel="import" type="css">` is supported by Polymer, but for whatever reason is deprecated.
Now it is in spec and Chromium (in regular form `<link rel=stylesheet>`): https://chromium.googlesource.com/chromium/src/+/f0cecf3ac0c7adf7a7c78eeb02cc75d380a9ce4d
So I'd like it to become un-deprecated in Polymer as well, hopefully, with default `<link rel=stylesheet>` syntax: https://www.polymer-project.org/1.0/docs/devguide/styling#external-stylesheets

sorvell:
Fixes #3837. Enable support for obtaining styles via `<link rel="import" type="css">` inside an element `dom-module`.

In the future we may support actual `link rel="stylesheet"` inside templates; however, for now we will retain support for the Polymer 1.x method of including an external stylesheet in an element.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -26,10 +26,10 @@
   },
   "devDependencies": {
     "web-component-tester": "^5",
-    "test-fixture": "PolymerElements/test-fixture#safari9-constructor"
+    "test-fixture": "PolymerElements/test-fixture#^3.0.0-rc.1"
   },
   "private": true,
   "resolutions": {
-    "test-fixture": "safari9-constructor"
+    "test-fixture": "3.0.0-rc.1"
   }
 }
--- a/bower.json
+++ b/bower.json
@@ -26,7 +26,7 @@
   },
   "devDependencies": {
     "web-component-tester": "^5",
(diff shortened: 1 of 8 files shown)


arthurevans:
@sorvell Are we calling this un-deprecated now? Or are we just ensuring it lives for another release?

Nitin-Prabhakar:
[Should `<link rel="stylesheet" href="...>` be allowed?](https://github.com/w3c/webcomponents/issues/530)

Looks like the good old way of injecting stylesheets is now scoped to the connected element's shadow tree.
What does this mean for Polymer? 
would we still need HTML Imports to load an external stylesheet?
Also,
If  we can use the `rel=stylesheet` tag, would we need explicit deduplication in case many elements load the same stylesheet? or would the browser be doing it

PR #4345, opened 2017-02-28
Fixes #4213

### LINKED ISSUE #4213: Polymer rewrites anchor hrefs
### Description
A relative url inside the href of an anchor element inside a Polymer element gets rewritten to be relative to the import.

For instance, if my element lives in `src/elements`

Something like `<a href="foo">` becomes `<a href="/src/elements/foo">`. While this behavior is desirable for things like CSS and images, it is not desirable for anchors as it messes up the ability to do routing with relative paths. This is especially important because Polymer's rewrite happens in a way that ignores the `<base>` element, meaning if you want to serve your project from a subdirectory (like GH pages) you can't easily use `<base>` to tack on the subdir.

cc @sorvell @justinfagnani

sorvell:
Only rewrite url's in templates that correspond to inline loadable resources (inline style and src attribute). This avoids scoping things like <a> href to template import location which is almost always not what the user wants. When this automated re-writing fails, users should fix url's via binding or imperative code.


sorvell:
Blocked on https://github.com/webcomponents/html-imports/issues/39.

sorvell:
Looks good with https://github.com/webcomponents/html-imports/pull/38.

sorvell:
Let's hold off on this change for now. We need to figure out how Polymer's hybrid mode works with this before proceeding.

kevinpschaaf:
Superseded by https://github.com/Polymer/polymer/pull/4360

PR #4348, opened 2017-02-28
Improve importHref()

gronke:
- uses altering test assets to ensure tests do not affect each other
- ensure importHref always returns a link element
- ensure onload() and onerror() functions are only called once when the same element was loaded again

Blocked by https://github.com/webcomponents/webcomponentsjs/pull/655 because the load event is invoked twice when a failed import attempt is repeated.

duplicate of #4209 that points against master instead of the branch behind #4200 ([refactor-import-href](https://github.com/Polymer/polymer/tree/refactor-import-href))

### DIFF
--- a/src/standard/utils.html
+++ b/src/standard/utils.html
@@ -400,7 +400,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         link = document.createElement('link');
         link.rel = 'import';
         link.href = href;
-        link.__loaded = false;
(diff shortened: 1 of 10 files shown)


TimvdLippe:
These changes have been incorporated in #4305. Please republish this PR if you find any changes we missed there.

gronke:
Wow, thanks. I didn't even have to rebase 🎂

PR #4352, opened 2017-03-01
2.0 connect order

### LINKED ISSUE #4228: 2.0 shadowRoot is not available in initial call to observers
### Description
In Polymer 2.0, `this.shadowRoot` is not yet available the first time observers are being flushed.  This is ultimately due to the fact that we cannot use `connectedCallback` as the signal to flush pending properties in the data system, due to timing infidelities in the custom elements polyfill (using async MutationObservers to detect connection, discussed in https://github.com/webcomponents/custom-elements/issues/34). Instead, we need to flush local clients before they are actually connected, and then append the children to the shadow root after.

Once https://github.com/webcomponents/custom-elements/issues/34 is resolved, we should be able to refactor the data initialization flow ala the changes prototyped in this branch: https://github.com/Polymer/polymer/tree/2.0-sr-before-observer

#### Live Demo
Current: http://jsbin.com/xawubor/edit?html,console,output

Expected (from above POC branch): http://jsbin.com/pixaquj/edit?html,console,output

#### Expected Results
```
has this.shadowRoot: true
localDom: prop: one
has this.root: true
localDom: prop: one
--- ready ---
has this.shadowRoot: true
localDom: prop: two
has this.root: true
localDom: prop: two
```

#### Actual Results
```
has this.shadowRoot: false
has this.root: true
localDom: prop: one
--- ready ---
has this.shadowRoot: true
localDom: prop: two
has this.root: true
localDom: prop: two
```

### Versions
- Polymer: v2.0-preview

sorvell:
Fixes #4228. Introduces in `PropertyEffects` a `_readyClients` method which is called the first time client elements need to be flushed. It can be used (and is in ElementMixin) to attach element dom to the element to ensure that element dom connected to the element by the time element property observers, ready, and connected are called.


sorvell:
@kevinpschaaf LGTM'd.

govis:
This commit may have affected the ready/attached(connected) order between parent and child elements (under the native shadow DOM at least). The parent's ready used to be called before the child's attached and it is now reversed.

There are quite a few instances across PolymerElements where child elements fire 'xxx-register' event from their attached callback and parent adds them to a collection (a property that's initialized when parent becomes ready).

PR #4360, opened 2017-03-02
Fixes #4357. 

### LINKED ISSUE #4357: Change URL handling in element templates
Users have requested more flexibility about how urls are dealt with in element templates. See, for example, issue #4213. We plan to change this behavior in Polymer 2.x.

The root problem stems from the following, given: 
```
<template>
  <img src="foo/bar.jpg">
</template>
```

What URL should this `src` be resolved against?

There are multiple answers that make sense: (1) the URL of the HTMLImport in which the element is defined, (2) the main document, (3) some other URL where all page resources are.

Polymer 1.x and related tools had the following, somewhat odd, rules. Inside element templates, the following URLs were re-written to be relative to the HTMLImport in which the element was defined:
* URLs in `<style`> elements
* URLS in attributes: `src`, `style`, `href`, `url`
* URLS in `<form>`'s attribute: action

URLs inside bindings were not re-written. Polymer 1.x provided a `resolveUrl` method that could be called to create a binding with an element HTMLImport document relative URL.

Specifically users have [complained](https://github.com/Polymer/polymer/issues/4213) that it especially does not make sense to re-write `href` attributes for `<a>` elements as this is almost never what the user wants.

sorvell:
Adds `importPath` and `rootPath` properties to Polymer.Element. The `importPath` property defaults to the path part of the url containing the dom-module matching the element's static `is` property. The `rootPath` property is set to the `Polymer.rootPath` property and defaults to the path part of the main document url. These properties can be used in bindings to url's. The `importPath` property is used to scope url's in element styles. [Note, this depends on this [html-imports pr](https://github.com/webcomponents/html-imports/pull/44) ].

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -492,6 +492,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        *       return memoizedTemplate;
        *     }
        *   }
+       *
+       * @returns {HTMLTemplateElement|string}
        */
       static get template() {
         if (!this.hasOwnProperty(goog.reflect.objectProperty('_template', this))) {
@@ -511,7 +513,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * used to create bindings relative to the import path.
        * Defaults to the path matching the url containing a `dom-module` element
        * matching this element's static `is` property.
-       * Note, this path does not contain a trailing `/`.
+       * Note, this path should contain a trailing `/`.
+       *
+       * @returns {string}
        */
       static get importPath() {
         if (!this.hasOwnProperty(goog.reflect.objectProperty('_importPath', this))) {
@@ -520,7 +524,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             Object.getPrototypeOf(this.prototype).constructor.importPath;
         }
         return this._importPath;
-
       }
 
       constructor() {
@@ -551,7 +554,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // avoid dependence on `is` for polyfilling styling.
         if (this._template && !this._template.__polymerFinalized) {
           this._template.__polymerFinalized = true;
-          finalizeTemplate(this.__proto__, this._template, importPath + '/',
+          const baseURI =
+            importPath ? Polymer.ResolveUrl.resolveUrl(importPath) : '';
+          finalizeTemplate(this.__proto__, this._template, baseURI,
             this.localName);
         }
         super._initializeProperties();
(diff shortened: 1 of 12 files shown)


lib/mixins/element-mixin.html
```diff
@@ -501,6 +504,25 @@
...
 
+      /**
+       * Path matching the url from which the element was imported.
+       * This path is used to resolve url's in template style cssText.
+       * The `importPath` property is also set on element instances and can be
+       * used to create bindings relative to the import path.
+       * Defaults to the path matching the url containing a `dom-module` element
+       * matching this element's static `is` property.
+       * Note, this path does not contain a trailing `/`.
+       */
```
kevinpschaaf: `@returns {string}`

lib/mixins/element-mixin.html
```diff
@@ -392,13 +392,16 @@
      * style scoping.
      * @param {HTMLElement} proto
      * @param {HTMLTemplateElement} template
+     * @param {string} baseURI URL against which to resolve urls in
+     * style element cssText.
```
kevinpschaaf: `@returns {HTMLTemplateElement}`

lib/mixins/element-mixin.html
```diff
@@ -524,13 +546,18 @@
...
         if (this._template && !this._template.__polymerFinalized) {
           this._template.__polymerFinalized = true;
-          finalizeTemplate(this.__proto__, this._template, this.localName);
+          finalizeTemplate(this.__proto__, this._template, importPath + '/',
+            this.localName);
         }
         super._initializeProperties();
+        // set path defaults
+        this.rootPath = Polymer.rootPath;
+        this.importPath = importPath;
```
kevinpschaaf: Reminder to discuss why it was bad if these went before initializeProperties

lib/utils/boot.html
```diff
@@ -18,6 +21,12 @@
   window.Polymer = function(info) {
     return window.Polymer._polymerFn(info);
   }
+
+  // support user settings on the Polymer object
+  if (userPolymer) {
+    Object.assign(Polymer, userPolymer);
+  }
```
kevinpschaaf: Just noting, this fixes the ability to set `Polymer.sanitizeDOMValue` before loading polymer, closing a loophole where it's conceivable that main-document e.g. `dom-repeat`'s could upgrade before the sanitization hook is installed. 👍

lib/utils/resolve-url.html
```diff
@@ -13,99 +13,81 @@
...
+          workingURL = (u.href === 'http://a/c%20d');
+        } catch (e) {
+          // silently fail
+        }
+      }
+      if (workingURL) {
+        return (new URL(url, baseURI)).href;
+      }
+      // Fallback to creating an anchor into a disconnected document.
+      let doc = tempDoc;
```
kevinpschaaf: Why this indirection?

kevinpschaaf:
Also, looks like there's some Linux 41 issue

PR #4363, opened 2017-03-02
Adds `restamp` mode to dom-repeat.

### LINKED ISSUE #1713: dom-repeat is not properly updating elements on filter/sort
When you utilize the filter/sort attributes the repeated nodes are not updating properly. This is the same behavior that was exhibited in the [Polycast](https://www.youtube.com/watch?v=2UKPRbrw3Kk&list=PLOU2XLYxmsII5c3Mgw6fNYCzaWrsM3sMN&index=5) on using core-list.

Rob gave a workaround for this by using `data-index="{{index}}"`. This seemed to fix the issue for his use case but when using the filter and sort attributes things get weird really quick.

I threw together a quick demo of this (it utilizes `iron-image` and `iron-ajax` so I opted to just give you a repo instead of a jsbin) it is [here](https://github.com/jshcrowthe/dom-repeat-bug) (Pardon the whimsical API). A link to the running demo: http://jshcrowthe.github.io/dom-repeat-bug/ (After the data loads note the position and color/images utilized then mess with the filters).

kevinpschaaf:
Fixes #1713 

When `restamp` is true, template instances are never reused with different items.  Instances mapping to current items are moved to their new location, new instances are created for any new items, and instances for removed items are discarded.  This mode is generally more expensive than `restamp: false` as it results in more instances to be created and discarded during updates and disconnection/reconnection churn.  By default, object identity is used for mapping instances to items.  Set `restampKey` to provide key based identity. 

When `restamp: true` is used, `restampKey` can be used to provide a path into the item object that serves as a unique key for the item rather than using object identity, for purposes of mapping instances to items.  Instances for items with the same key will be reused, while all others will be either restamped or discarded.


azakus:
@kevinpschaaf please rebase

evorapa:
Is this PR still in progress? I see that the tests pass on chrome but timeout on firefox. Is there any ETA for this being functional?

TimvdLippe:
@azakus Rebased and fixed the tests

sushantbs:
Any updates about when this will be available? Have been waiting for this for a while.

TimvdLippe:
The change was reverted as @kevinpschaaf and @sorvell wanted to improve the API, but wasnt in the PR. We should be able to get the ball rolling again soon!

sushantbs:
@TimvdLippe, @sorvell Would be nice if you could let us the reasons for reverting the PR and what are the use cases that the API needs to cater to. 
Maybe I, or someone from the community, can possibly take this up?

Thanks for all the great work...

TimvdLippe:
@sushantbs Paging that to @kevinpschaaf and @sorvell who wanted to have some more discussion and tackle some issues. I am not aware of the enhancements they want to make to this PR :smile:

skortchmark9:
Would be nice to know what happened here.

leebickmtu:
@TimvdLippe @kevinpschaaf @sorvell 
So is restamp planned to be re-added at any point?

kevinpschaaf:
This implementation was put on hold, out of concern that the algorithm that was used to maintain key-instance identity was too simplistic and could have pathologically-bad performance as a result.

In the interim, a more complicated but [performance-optimized algorithm](https://github.com/Polymer/lit-html/pull/501) was implemented in `lit-html`'s `repeat` directive.  Back-porting that to Polymer would be a significant amount of work and at this point in the Polymer library's [lifecycle](https://www.polymer-project.org/blog/2018-05-02-roadmap-faq) we realistically probably wouldn't implement such a significant new feature.

PR #4365, opened 2017-03-03
Remove use of node.shadyRoot in testing

azakus:
node.shadowRoot works in both native and polyfill


sorvell:
Failures are expected based on HTMLImports polyfill... will be fixed via https://github.com/Polymer/polymer/pull/4360.

PR #4389, opened 2017-03-07
Don't override the goog namespace if it already exists

ChadKillingsworth:
When used on the same site as closure-library, the `goog` namespace is overwritten by the reflection shim.

Alternatively, you have the option to use the specially named `JSCompiler_renameProperty` function which has the exact same functionality. The compiler recognizes either signature but `JSCompiler_renameProperty` is much less likely to be used by other projects. It's specific to soy templates mainly.


azakus:
@ChadKillingsworth is that function something that can be set with a closure flag, or just hardcoded as `JSCompiler_renameProperty(string, object)`?

ChadKillingsworth:
@azakus It's hard coded and can't ever change due to historical Google use cases: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/NodeUtil.java#L63

PR #4397, opened 2017-03-09
add `behaviors` to instances too

### LINKED ISSUE #4396: 2.x class-based elements don't have behaviors defined in the constructor
A legacy element (`Polymer({ is: 'my-el' })`) will have `behaviors` defined for each instance.

A class-based element (`class MyEl extends Polymer.Element`) will instead have `behaviors` defined in the constructor, so each instance must access via `this.constructor.behaviors`.
http://jsbin.com/fufikun/1/edit?html,console,output

How should hybrid element authors go about it?

valdrinkoshi:
Fixes #4396, https://github.com/PolymerElements/iron-a11y-keys-behavior/issues/69

### DIFF
--- a/lib/legacy/class.html
+++ b/lib/legacy/class.html
@@ -54,7 +54,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (!Array.isArray(behaviors)) {
         behaviors = [behaviors];
       }
-      let superBehaviors = klass.behaviors;
+      let superBehaviors = klass.prototype.behaviors;
       // get flattened, deduped list of behaviors *not* already on super class
       behaviors = flattenBehaviors(behaviors, null, superBehaviors);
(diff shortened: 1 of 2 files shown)


lib/legacy/class.html
```diff
@@ -63,6 +63,7 @@
         behaviors = superBehaviors.concat(behaviors);
       }
       klass.behaviors = behaviors;
```
sorvell: Let's remove this off the class if we're putting it on the prototype.

lib/legacy/class.html
```diff
@@ -63,6 +63,7 @@
...
+      klass.prototype.behaviors = behaviors;
```
valdrinkoshi: not sure if this should belong to the mixin or the legacy element, same for `properties`.
Should I just have two getters in the LegacyElementMixin?

PR #4399, opened 2017-03-09
Implements `disable-upgrade` 

### LINKED ISSUE #4398: Implement `disable-upgrade` in Polymer 2.x
This feature was added in Polymer 1.8 and should be implemented in Polymer 2.x

sorvell:
Fixes #4398.

Implements `disable-upgrade` attribute which prevents readying an element until the attribute is removed.


lib/mixins/lazy-upgrade-mixin.html
```diff
@@ -0,0 +1,66 @@
...
+          node.if = true;
+        }
+        node.removeAttribute(DISABLE_UPGRADE);
+        node.removeAttribute(LAZY_UPGRADE);
+      }
+
+      return class LazyUpgrade extends base {
+        ready() {
+          super.ready();
+          this.__lazyUpgradeDeadline = 16;
```
sorvell: Let's make this protected.

lib/mixins/lazy-upgrade-mixin.html
```diff
@@ -0,0 +1,66 @@
...
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<link rel="import" href="../utils/boot.html">
+<link rel="import" href="../utils/mixin.html">
+<link rel="import" href="../utils/render-status.html">
+<script>
+  (function(){
+    'use strict';
+
+    Polymer.LazyUpgradeMixin = Polymer.dedupingMixin(function(base) {
```
sorvell: Needs test.

lib/mixins/element-mixin.html
```diff
@@ -673,7 +678,18 @@
...
+        // process `disable-upgrade` specially
+        if (name === DISABLED) {
+          const disabled = value !== null;
+          if (this.__upgradeFlush && !disabled) {
+            this._flushProperties = this.__upgradeFlush;
+            this.__upgradeFlush = null;
+            this._flushProperties();
+          } else if (disabled) {
+            this.__upgradeFlush = this._flushProperties;
+            this._flushProperties = function() {};
```
kevinpschaaf: So this will kind of just pause property flushing if you apply `disable-upgrade` a second time... Don't think it's particularly problematic, just noting (we made it a one-shot in 1.x, but not sure that's particularly bad). Hmm, it may be problematic from a `splices` path notification point of view, now that I think about it...

lib/mixins/lazy-upgrade-mixin.html
```diff
@@ -0,0 +1,131 @@
...
+      /** @polymerMixinClass */
+      return class LazyUpgrade extends base {
+        // add a configurable
+        static get properties() {
+          return {
+            /**
+             * Instance-level property for configuring the frame budget for lazy-upgrading elements.
+             * Defaults to 16ms
+             */
+            _lazyUpgradeBudget: {
```
kevinpschaaf: `dom-repeat` uses a `targetFramerate` API (which is basically inverse)... worth matching API?

lib/mixins/lazy-upgrade-mixin.html
```diff
@@ -0,0 +1,131 @@
...
+            _lazyUpgrading: {
+              type: Boolean
+            }
+          }
+        }
+
+        ready() {
+          super.ready();
+          // check for dom-repeat and dom-if elements stamping lazy-upgrade nodes
+          this.addEventListener('dom-change', (ev) => {
```
kevinpschaaf: Would be slightly safer (with no downsides) to move the listener setup to `constructor` (or at least, before `ready` super call) to remove dependency on elements that fire `dom-change` doing so asynchronously.

kevinpschaaf: Also just curious... was there a use case for needing support for repeat/if templates stamping `lazy-upgrade` elements? Or is this just "just in case"?  I can maybe picture scenarios where it might be good, but seems cornery.

lib/utils/render-status.html
```diff
@@ -32,7 +32,10 @@
   }
 
   function flushQueue(queue) {
-    while (queue.length) {
+    const max = queue.length;
+    let i=0;
+    while (queue.length && i < max) {
+      i++;
```
kevinpschaaf: I assume this change is from a realization that calling `RenderStatus.afterNextRender` from within an `afterNextRender` callback should _not_ go in the current turn, right?  However, seems like `beforeNextRender` called within a `beforeNextRender` _should_ go in the current turn, and this breaks it since `flushQueue` is used for both?

Also, keeping the while construction makes it hard to grok; this would read cleaner: `for (let i=0, max=queue.length; i<max; i++)`.

kevinpschaaf:
Also, we have some Chrome 41 and Edge failures.

lib/mixins/lazy-upgrade-mixin.html
azakus: I'm not 100% sure that we need to support that feature. WDYT @sorvell?

lib/mixins/lazy-upgrade-mixin.html
azakus: I'm not sure I understand the `targetFramerate` API in dom-repeat. Is it the desired # of elements to be stamped every second?

PR #4417, opened 2017-03-12
Remove "private": true from bower.json

FluorescentHallucinogen:


FluorescentHallucinogen:

notwaldorf:
This change isn't needed.`private` is only used by the bower catalog, and we don't publish directly to it.

FluorescentHallucinogen:
Ok, but, BTW, why Polymer don't publish to Bower catalog directly?

PR #4420, opened 2017-03-12
[1/2] Update authors

FluorescentHallucinogen:


FluorescentHallucinogen:

notwaldorf:
This change isn't needed, since we're already linking to the authors file in all of the license headers ([example](https://github.com/webcomponents/webcomponentsjs/blob/master/src/pre-polyfill.js#L5)). Also, FYI, we use the "Polymer" not "Polymer Project" name.

FluorescentHallucinogen:
@notwaldorf
> Also, FYI, we use the "Polymer" not "Polymer Project" name.

But you use "Polymer Project": https://github.com/Polymer/polymer/blob/master/package.json#L41

Please correct it in `package.json` in this case.

Also I've used HTTPS URL instead of HTTP.

FluorescentHallucinogen:
@notwaldorf

And you already use linking to the authors file in `bower.json`: https://github.com/Polymer/polymer/blob/master/bower.json#L17

Please correct it in `bower.json` in this case.

PR #4421, opened 2017-03-12
[2/2] Update authors

FluorescentHallucinogen:


FluorescentHallucinogen:

notwaldorf:
This change isn't needed, since we're already linking to the authors file in all of the license headers ([example](https://github.com/webcomponents/webcomponentsjs/blob/master/src/pre-polyfill.js#L5)). Also, FYI, we use the "Polymer" not "Polymer Project" name.

notwaldorf:
This change isn't needed, since we're already linking to the authors file in all of the license headers ([example](https://github.com/webcomponents/webcomponentsjs/blob/master/src/pre-polyfill.js#L5)). Also, FYI, we use the "Polymer" not "Polymer Project" name.

Also, *please* stop spamming all the repos with the same kind of automatic change until you hear back feedback on one of them (or, alternatively, send one PR with all your changes to each repo). This makes it impossible to triage issues and PRs from anyone else.

PR #4480, opened 2017-03-31
Failing unit test to demonstrate shadycss#69

TomK:
Regarding http://github.com/webcomponents/shadycss/issues/69

I created a unit test within shadycss which passed.  So I wrote the same test against polymer which fails.

This appears to be a time based issue.  I have abstracted the assertions into two functions to show this.
`assertProp` - ensure the mixin style is exploded into properties correctly
`assertMix` - ensure the applied styles are correctly set from the properties

the test named `both assertions in a single test` fails if run first, but passes if run last.


TomK:
interesting.... :/

TomK:
closing, this was addressed via a different pr

PR #4485, opened 2017-04-01
test more permutations

### LINKED ISSUE #4484: 2.x should test also wc-ce=true only & wc-shadydom=true only
Currently tests are run forcing both ShowDOM & CustomElements polyfills. Permutations of ShadyDOM only and CustomElements only should be supported/tested as well

valdrinkoshi:
Fixes #4484

### DIFF
--- a/test/runner.html
+++ b/test/runner.html
@@ -100,26 +100,26 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     // test shadowdom/custom elements polyfills together
     // preferring both if possible.
     var matrix = [''];
-    var webcomponents = '';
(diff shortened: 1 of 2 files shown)


azakus:
@valdrinkoshi @sorvell Still want to make progress on this?

sorvell:
@valdrinkoshi Would you mind updating the failing Polymer tests? If they are green, I think we can take this.

valdrinkoshi:
@sorvell tests fixed by moving the `remove, append domif` test to be executed after the `clients persist/clients recreated` ones, and taking into account the increment of uid at that spot. PTAL

PR #4492, opened 2017-04-04
Introducing an `insertBefore` method on Templatize instances

### LINKED ISSUE #4447: Improper ordering of the ready events in v2 in some scenarios
### Description

Element's ready callback is called after the parent elements ready callback when the parent is rendered in another element's light dom and there is a binding between the parent it's containing element. I think.

#### Live Demo

http://jsbin.com/komavinoza/edit?html,console,output

#### Steps to Reproduce

I've done my best to narrow it down in the linked JS Bin. There's a lot required for the setup. I'll do my best to explain:

There are three elements: `parent-element --> child-element --> grandchild-element`. The child is rendered inside a `dom-repeat` and a property rendered by the child is bound to the parent. The child renders the grandchild element in it's local dom.

Rendering the child in the parent's local dom, removing the rendering of the property in the child, or removing the binding between the child and the parent will solve the problem.

#### Expected Results

The ready event of `grandchild-element` is fired before the ready event of `child-element`.

#### Actual Results

The ready event of `child-element` is fired before the ready event of `grandchild-element`.

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions

- Polymer: v2.0.0-rc.3

Polymer v1.x does not exhibit this issue.

sorvell:
Fixes #4447. This api should be used to insert instances into the dom. Using it ensures that sub elements are "readied" before host elements.

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -580,7 +580,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
       let beforeRow = this.__instances[instIdx + 1];
       let beforeNode = beforeRow ? beforeRow.children[0] : this;
-      inst.insertBefore(this.parentNode, beforeNode);
+      inst.connect(this.parentNode, beforeNode);
       this.__instances[instIdx] = inst;
       return inst;
     }
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -580,7 +580,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
       let beforeRow = this.__instances[instIdx + 1];
       let beforeNode = beforeRow ? beforeRow.children[0] : this;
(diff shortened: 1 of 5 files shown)


lib/elements/dom-repeat.html
```diff
@@ -580,7 +580,7 @@
       }
       let beforeRow = this.__instances[instIdx + 1];
       let beforeNode = beforeRow ? beforeRow.children[0] : this;
-      this.parentNode.insertBefore(inst.root, beforeNode);
+      inst.insertBefore(this.parentNode, this);
```
kevinpschaaf: Should be `inst.insertBefore(this.parentNode, beforeNode);` (which will also solve your eslint error!!)

sorvell:
Closing this PR since a better fix for the behavior that does not introduce new API is in the works.

PR #4503, opened 2017-04-06
Fix custom-style-late tests

### LINKED ISSUE #4501: custom-style-late tests broken in shadycss master
Looks like https://github.com/webcomponents/shadycss/commit/33512fc4a6257dc01c5092d855f5cbf484bcb94b caused the custom-style-late tests to fail.

These tests are weird holdovers from Polymer 1.x and should be rewritten, but the issue may persist after rewriting.

azakus:
The included element was never upgrading because it is an extension
element, and CustomStyleInterface boots up async due to
webcomponents/shadycss@33512fc and misses the resync.

This is solved by making a real element.

Also, this test wasn't really "late", so the `<custom-style>` is
asynchronous now

Fixes #4501


test/unit/custom-style-late.html
```diff
@@ -14,19 +14,48 @@
...
+  </custom-style>
+
+  <template id="late">
+  <custom-style>
+    <style is="custom-style">
+      :root {
+        --cs-blue: {
+          border: 8px solid blue;
+        }
+        ;
```
sorvell: put this on the line before?

test/unit/custom-style-late.html
```diff
@@ -14,19 +14,48 @@
...
+      test('late defined properties applied to custom-style', function(done) {
+        Polymer({
+          is: 'x-input'
+        });
+        var template = document.querySelector('template#late');
+        document.body.appendChild(document.importNode(template.content, true));
+        setTimeout(function() {
+          assertComputed(input, '8px');
+          done();
+        }, 125);
```
sorvell: why the specific timeout here?

test/unit/custom-style-late.html
azakus: I think this was trying to test some broken syntax, but I'm fine cleaning it up.

test/unit/custom-style-late.html
azakus: I can use some `flush` functionality probably.

PR #4432, opened 2017-03-15
Separate binding-specific code from template stamp. Expose override points

kevinpschaaf:

PR #4510, opened 2017-04-06
2.0 Improvements to binding API

kevinpschaaf:
- Adds override points for _parseBindings and _evaluateBinding
- Adds support for runtime template binding
- Moves ready(), _hasAccessor tracking, and instance property swizzle at ready time to `PropertyAccessors`
- Other refactoring and cleanup

### DIFF
--- a/lib/mixins/property-accessors.html
+++ b/lib/mixins/property-accessors.html
@@ -79,7 +79,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    * the standard `static get observedAttributes()`, implement `_propertiesChanged`
    * on the class, and then call `MyClass.createPropertiesForAttributes()` once
    * on the class to generate property accessors for each observed attribute
-   * prior to instancing.  Any `observedAttributes` will automatically be
+   * prior to instancing.  Last, call `this._flushProperties()` once to enable
+   * the accessors.
+   *
+   * Any `observedAttributes` will automatically be
    * deserialized via `attributeChangedCallback` and set to the associated
    * property using `dash-case`-to-`camelCase` convention.
    *
@@ -137,6 +140,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         this.__dataOld = null;
         if (this.__dataProto) {
           this._initializeProtoProperties(this.__dataProto);
+          this.__dataProto = null;
         }
         // Capture instance properties; these will be set into accessors
         // during first flush. Don't set them here, since we want
@@ -166,6 +170,22 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         for (let p in props) {
           this._setProperty(p, props[p]);
         }
+      }
+
+      /**
+       * Called at ready time with bag of instance properties that overwrote
+       * accessors when the element upgraded.
+       *
+       * The default implementation sets these properties back into the
+       * setter at ready time.  This method is provided as an override
+       * point for customizing or providing more efficient initialization.
+       *
+       * @param {Object} props Bag of property values that were overwritten
+       *   when creating property accessors.
+       * @protected
+       */
+      _initializeInstanceProperties(props) {
+        Object.assign(this, props);
       }
 
       /**
@@ -444,7 +464,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * @protected
        */
       _invalidateProperties() {
-        if (!this.__dataInvalid) {
+        if (!this.__dataInvalid && this.__dataInitialized) {
           this.__dataInvalid = true;
           microtask.run(() => {
             if (this.__dataInvalid) {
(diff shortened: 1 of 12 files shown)


arthurevans:
I am so excite! And also confuse! 

How can people do runtime template binding? By that do you mean identifying an instance-specific template and then calling `_bindTemplate` on it? (I note there's an `addBinding` method, but AFAICT that is private and not exposed on `PropertyEffects`.)

Can you bind a template more than once? (i.e., could you mutate and re-bind the template) Can you bind multiple templates?

Inquiring minds want to know...

Thanks,
Arthur

lib/mixins/property-accessors.html
```diff
@@ -436,12 +463,41 @@
        * @protected
        */
       _flushProperties() {
-        let oldProps = this.__dataOld;
-        let changedProps = this.__dataPending;
-        this.__dataPending = null;
-        this.__dataCounter++;
-        this._propertiesChanged(this.__data, changedProps, oldProps);
-        this.__dataCounter--;
+        if (!this.__dataInitialized) {
```
sorvell: Now that we have `__dataInitialized` it probably makes sense to prevent `_invalidateProperties` from scheduling a flush if this flag is not true.

lib/mixins/property-effects.html
```diff
@@ -462,37 +456,43 @@
...
-              compoundIndex: part.compoundIndex,
-              event: part.event,
-              customEvent: part.customEvent,
-              negate: part.negate
+          let dependencies = part.dependencies;
+          let info = { index, binding, part, evaluator: constructor };
+          part.compoundIndex = i;
+          for (let j=0; j<dependencies.length; j++) {
+            let trigger = dependencies[j];
+            if (typeof trigger == 'string') {
```
sorvell: Bookmark to consider if we want to assume we should `parseArg` the `trigger` if it's a string.

lib/mixins/property-effects.html
```diff
@@ -506,36 +506,29 @@
...
+  function runBindingEffect(inst, path, props, oldProps, info, hasPaths, nodeList) {
+    let node = nodeList[info.index];
+    let binding = info.binding;
+    let part = info.part;
     // Subpath notification: transform path and set to client
     // e.g.: foo="{{obj.sub}}", path: 'obj.sub.prop', set 'foo.prop'=obj.sub.prop
-    if (hasPaths && (path.length > info.value.length) &&
-        (info.kind == 'property') && !info.isCompound &&
-        node.__propertyEffects && node.__propertyEffects[info.name]) {
+    if (hasPaths && part.source && (path.length > part.source.length) &&
```
sorvell: Add a comment explaining why path processing is not included with `_evaluateBinding`.

lib/mixins/property-effects.html
```diff
@@ -2141,37 +1975,123 @@
...
+       * `ShadyCSS.prepareTemplate` may only be called once per element. As such,
+       * any styles required by in runtime-stamped templates must be included
+       * in the main element template.
+       *
+       * @param {HTMLTemplateElement} template Template to stamp
+       * @return {DocumentFragment} Cloned template content
+       * @protected
+       */
+      _stampBoundTemplate(template) {
+        let dom = super._stampTemplate(template);
```
sorvell: `super` => `this`

lib/mixins/property-effects.html
```diff
@@ -2285,26 +2193,147 @@
...
+                dependencies.push(arg);
+              }
+            }
+            let dynamicFns = templateInfo.dynamicFns;
+            if (dynamicFns && dynamicFns[methodName] || signature.static) {
+              dependencies.push(methodName);
+            }
+          } else {
+            // Property or path
+            dependencies.push(source);
```
sorvell: Should `parseArg(source)` so it doesn't auto-wildcard.

kevinpschaaf:
@arthurevans You can call `_stampBoundTemplate(template)` on a template (with bindings, events, etc.) of your choosing as many times as you like, and the doc frag returned will have elements that are bound to the element.  `_bindTemplate` is only for prototypical binding, not runtime binding (`_stampBoundTemplate` stamps & binds) for more efficient stamping of the element's main template -- don't expect end users to use that unless they're building their own base classes since `ElementMixin` does this for you already.

`_stampBoundTemplate` doesn't support mutating the template and stamping it again... for that you'd just need to clone the template before mutating it, since template metadata from parsing it are memoized on the template. Should probably add that limitation to the docs.  There's also a limitation with `shadycss` that I did mention in there, which is that since `shadycss` doesn't support runtime addition of `<style>`s, any runtime-stamped templates can't stamp new styles-- their styles need to be in the main (prototypical) template for the element.

Note one review comment is to change `_stampBoundTemplate` back to an override of `_stampTemplate`, so the name will likely change back.

This PR (and the previously merged https://github.com/Polymer/polymer/pull/4432) wasn't outright intending to add new features in the RC; the main goal here was to refactor the existing code to make the whole binding system more extensible,however the ability to stamp multiple templates sort of fell out pretty easily, so we'll probably go ahead and add this feature before the final release.

Still need to add tests and address review feedback... If anything starts looking too risky we may trim this back. Will keep you in the loop.

PR #4514, opened 2017-04-07
Correct changelog markdown title

Ffloriel:
Add space between ## and version. The title was previously not displaying correctly because of the lack of space.  
Could not find where the problem was but it is probably in the gen-changelog.sh file

(CLA should be ok now)


kevinpschaaf:
Thanks, @azakus will fix the changelog script.

PR #4524, opened 2017-04-11
add nestedObserve feature

mrLeung:

### DIFF
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -1442,6 +1442,32 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
 
       /**
+       * TODO(wahson):
+       * 1、if property is not an object
+       * 2、support for array
+       * @param {object or Array} property
+       * @param {string} path
+       */
+      _initProxy(property, path) {
+        let handler = (path) => {
+          return {
+            set: (obj, prop, value) => {
+              let _path = `${path}.${prop}`;
+              if (value instanceof Object) {
+                value = new Proxy(value, handler(_path));
+              }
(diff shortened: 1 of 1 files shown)


lib/mixins/property-effects.html
```diff
@@ -1442,6 +1442,32 @@
...
+       * @param {object or Array} property
+       * @param {string} path
+       */
+      _initProxy(property, path) {
+        let handler = (path) => {
+          return {
+            set: (obj, prop, value) => {
+              let _path = `${path}.${prop}`;
+              if (value instanceof Object) {
+                value = new Proxy(value, handler(_path));
```
web-padawan: Isn't `Proxy` [slow](http://thecodebarbarian.com/thoughts-on-es6-proxies-performance.html)? What about IE support? I wonder if Polymer should use such stuff.

kevinpschaaf:
Please see the [CONTRIBUTING.md](https://github.com/Polymer/polymer/blob/master/CONTRIBUTING.md#contributing-pull-requests) guide regarding minimum requirements for pull requests (referring issue, tests, etc.).

That said, we've intentionally designed Polymer 2.0 to be much more extensible via base class override points (single-underscore'ed and documented methods such as `_setPendingProperty` are `@protected` and we intend to include their API's in the semver contract so that they are safe to use).  As such, we would encourage this type of "add-on" feature for Polymer to be implemented and distributed yourself as a class mixin or a new base class that extends `Polymer.Element`.

We are unlikely to grow the core features provided by the property effect system, but instead want to embrace a third-party ecosystem of add-ons like this in the future (and then could consider pulling these in after observing the community reaction).

PR #4533, opened 2017-04-14
Use `_importPath` in `resolveUrl` so it available early. Fixes #4532

### LINKED ISSUE #4532: Polymer 1.9 resolveUrl does not work as intended
### Description

`this.resolveUrl` in a properties block is simply just not working as intended.

#### Live Demo

Wrong: http://jsbin.com/ridukodefe/2/edit?html,console
Intended: https://jsbin.com/sekeyeheso/1/edit?html,console

#### Steps to Reproduce

1. Create element
2. Add a property in the properties block
3. set `value: function() { return this.resolveUrl('./relativeUrlThatExists')}`
4. set the url of `resolveUrl` to something that exists.
5. observe that it does not find the proper url

#### Expected Results

it should return the actual absolute url

#### Actual Results

it just returns the same relative url entered

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions

- Polymer: v1.9.0
- webcomponents: v0

kevinpschaaf:
Use `_importPath` in `resolveUrl` so it available early. Fixes #4532

### DIFF
--- a/test/unit/resolveurl.html
+++ b/test/unit/resolveurl.html
@@ -71,19 +71,30 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       test('Urls in styles and attributes', function() {
         var el = document.createElement('p-r');
-        var rx = /sub\/foo\.z/;
-        assert.match(el._styles[0].textContent, rx, 'url not relative to main document');
-        assert.match(el.$.div.getAttribute('style'), rx, 'style url not relative to main document');
-        assert.match(el.$.img.src, rx, 'src url not relative to main document');
-        assert.match(el.$.a.href, rx, 'href url not relative to main document');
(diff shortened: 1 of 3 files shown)


test/unit/resolveurl.html
```diff
@@ -138,6 +147,13 @@
         assert.equal(el.$.hash.getAttribute('href'), '#foo.z', 'hash-only url should not be resolved');
         assert.equal(el.$.absolute.getAttribute('href'), '/foo.z', 'absolute urls should not be resolved');
         assert.equal(el.$.protocol.getAttribute('href'), 'data:foo.z', 'urls with other protocols should not be resolved');
+        el.$.if.render();
```
sorvell: Would be an improvement to fix the `rx` to help ensure the url is fully resolved correctly and not just that it contains the `sub` folder.

PR #4539, opened 2017-04-17
Closure Compilation support

azakus:


lib/mixins/property-accessors.html
```diff
@@ -563,6 +568,8 @@
 
     }
 
+    return PropertyAccessors;
```
kevinpschaaf: Investigate if separating return from class definition is necessary; not used above.

lib/utils/templatize.html
```diff
@@ -40,7 +40,12 @@
     }
 
     // Base class for TemplateInstance's
-    class TemplateInstanceBase extends Polymer.PropertyEffects(class{}) {
+    /**
+     * @constructor
+     * @mixes {Polymer_PropertyEffects}
```
kevinpschaaf: `@implements`

lib/mixins/property-accessors.html
azakus: This is actually necessary :(

kevinpschaaf:
LGTM

PR #4541, opened 2017-04-18
Guard against overwriting bound values with hasOwnProperty. Fixes #4540

### LINKED ISSUE #4540: [2.x] Upgrade after binding results in default value for properties with no effects
### Description
When an element is lazily upgraded after its host binds values to it, the bound value will incorrectly revert to its default value _if the property has no property effects_.  This behavior is correct for properties that do have property effects.

#### Live Demo
https://jsbin.com/hepidek/edit?html,console,output

#### Expected Results
```
init: host
ready: host
```

#### Actual Results
```
init: host
ready: client
```

### Browsers Affected
All

### Versions
- Polymer: v2.0.0-rc.5

kevinpschaaf:
Fixes #4540

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -597,14 +597,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         }
         for (let p in p$) {
           let info = p$[p];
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
cc/ @frankiefu

PR #4545, opened 2017-04-18
add gulp task to lint for closure warnings

azakus:


azakus:
@kevinpschaaf done with 25cfc882

PR #4546, opened 2017-04-18
Process paths regardless of accessor, & loop on computeLinkedPaths.

### LINKED ISSUE #4542: Transitive linkPaths broken in 2.0
### Description

Calling linkPaths(a, b) and then linkPaths(b, c) doesn't link c to a.

#### Live Demo

In 1.0 the checkbox is unchecked:
http://jsbin.com/hocuwazada/edit?html,output

Same code, but with 2.0 the checkbox stays checked:
http://jsbin.com/xevedoyeme/edit?html,output

#### Steps to Reproduce

           this._foo2 = this._foo;
            this._foo3 = this._foo;
            this.linkPaths('_foo', '_foo2');
            this.linkPaths('_foo2', '_foo3')
            this.set('_foo3.checked', false);

#### Expected Results

A notification is triggered on _foo.

#### Actual Results

A notification is triggered on _foo2 but not _foo.

### Browsers Affected

- [ X] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions

- Polymer: v2.0-preview
- webcomponents: v2.0-preview

kevinpschaaf:
Fixes #4542

The `_setPendingPropertyOrPath` implementation that the public `set` leverages does not trigger `_propertiesChanged` if the root property being set has no accessor.  Generally this is correct, except that `linkPaths` can result in triggering linked paths with accessors, but where the root property being set has none.  As such, `set` should always result in `_propertiesChanged` running, to match the 1.x behavior and give `computeLinkedPaths` an opportunity to run.

Additionally, since `computeLinkedPaths` only did one loop over the links, if a newly generated link was subsequently linked to a path earlier in the list, it would never compute to that link.  As such, `computeLinkedPaths` was moved to `runComputedEffects` and takes advantage of the same while loop that runs until no more pending properties are generated.

### DIFF
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -371,12 +371,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    */
   function runComputedEffects(inst, changedProps, oldProps, hasPaths) {
     let computeEffects = inst.__computeEffects;
-    if (computeEffects) {
-      let inputProps = changedProps;
-      while (runEffects(inst, computeEffects, inputProps, oldProps, hasPaths)) {
+    let inputProps = changedProps;
+    while (inputProps) {
+      computeEffects && runEffects(inst, computeEffects, inputProps, oldProps, hasPaths);
+      computeLinkedPaths(inst, inputProps, hasPaths);
+      if ((inputProps = inst.__dataPending)) {
         Object.assign(oldProps, inst.__dataOld);
         Object.assign(changedProps, inst.__dataPending);
-        inputProps = inst.__dataPending;
         inst.__dataPending = null;
       }
     }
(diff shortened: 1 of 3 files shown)


lib/mixins/property-effects.html
```diff
@@ -1292,32 +1288,28 @@
...
+            // object has already been updated
+            let old = Polymer.Path.get(this, path);
+            path = /** @type {string} */ (Polymer.Path.set(this, path, value));
+            // Use property-accessor's simpler dirty check
+            if (!path || !super._shouldPropertyChange(path, value, old)) {
+              return false;
             }
-            this.__dataHasPaths = true;
           }
+          this.__dataHasPaths = true;
```
sorvell: This seems broken. If there's no accessor then the value is not applied because `_setPendingProperty` will not actually set the value.

sorvell: Consider setting all the linked paths here and avoiding doing it in `runComputedEffects`.

kevinpschaaf: For paths, the only time `Polymer.Path.set` is required to update the value is _iff_ it is `!isPathNotification`; if it is a path notification, the value is guaranteed to be set.

PR #4552, opened 2017-04-19
Copying bower -> package.json to allow for yarn installations

maxknee:


azakus:
I'm not sure how this would actually help with yarn installations. Wouldn't a user need to use a git dependency in their `package.json` for Polymer, which is rather discouraged?

kevinpschaaf:
Closing, no response and not intending to make this change otherwise.

Issue #4554, opened 2017-04-20
dom-repeat.itemsIndexAs documentation might be incorrect


mattpalermo:
### Description

Documentation for dom-repeat.itemsIndexAs is exactly the same as dom-repeat.indexAs. I suspect that the dom-repeat.itemsIndexAs documentation should say things which are "opposite" to the dom-repeat.indexAs documentation.

### Possible fix

```javascript
        /**
         * The name of the variable to add to the binding scope with the index
         * for the item. If `sort` or `filter` is provided, the item index will reflect the
         * original array order (rather than the sorted and filtered order).
         */
```

### Versions
Current master

arthurevans:
From what I can see, it looks like the fix proposed by @mattpalermo is correct ... @kevinpschaaf?

kevinpschaaf:
Thanks! Fixed in https://github.com/Polymer/polymer/pull/4557

PR #4557, opened 2017-04-21
Minor doc fixes.

### LINKED ISSUE #4553: Add example to API doc for Polymer.Debounce
The API is confusing, and it really needs a class-level comment with an example.

sorvell:
Fixes #4553, #4554

### DIFF
--- a/lib/utils/debounce.html
+++ b/lib/utils/debounce.html
@@ -68,19 +68,28 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
   /**
    * Creates a debouncer if no debouncer is passed as a parameter
-   * or it cancels an active debouncer otherwise. For example, the following
-   * code can be called multiple times within a microtask and the provided
-   * function will be debounced (called once at the next microtask checkpoint):
+   * or it cancels an active debouncer otherwise. The following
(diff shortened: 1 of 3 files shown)


lib/utils/debounce.html
```diff
@@ -68,7 +68,19 @@
...
   /**
    * Creates a debouncer if no debouncer is passed as a parameter
-   * or it cancels an active debouncer otherwise.
+   * or it cancels an active debouncer otherwise. For example, the following
+   * code can be called multiple times within a microtask and the provided
+   * function will be debounced (called once at the next microtask checkpoint):
+   *
+   * let job = Polymer.Debouncer.debounce(job, Polymer.Async.microTask, () => {
+   *   console.log('debounced!');
+   * });
```
arthurevans: Isn't it important for the value of `job` to be preserved across multiple calls here? If I copy this code into an event handler, will it debounce correctly?

That is, wouldn't you usually store the job like `this._job = Polymer.Debouncer.debounce(_job ...)`

sorvell: Indeed. Updated with hopefully a more clear example.

arthurevans:
👍  :shipit: 🎉

Issue #4543, opened 2017-04-18
Provide Alternatives to "Meta" Programming for observers and computed properties


ChadKillingsworth:
Observers and computed properties currently require string values. This hides them from most static analysis engines and refactoring tools. They are also difficult to use with Closure-compiler's advanced mode as properties get renamed.

Need to provide an alternative way to use these property effects that does not depend on strings.

## Simple Observers

This is the easiest fix. Allow a function to be specified directly instead of just a string.

```js
class FooElement extends Polymer.Element {
  static get is() { return 'foo-element'; }
  static get properties() {
    return {
      bar: {
        type: String,
        observer: FooElement.prototype.barChanged_
      }
    }
  }
  barChanged_(newValue, oldValue) {
    console.log(newValue, oldValue)
  }
}
```

## Complex Observers and Computed Properties

These are grouped together because they suffer from the same problem: change notification. A computed property is best defined as an ES6 getter and a complex observer is already an instance method. The only thing needed is change notifications from properties that these methods depend on. This isn't hard to wire up manually with a simple observer and calls to `notifyPath`, but it could be a bit easier.

Is there perhaps a way to express the dependency in reverse?

```js
class FooElement extends Polymer.Element {
  static get is() { return 'foo-element'; }
  static get properties() {
    return {
      bar: type: String
    }
  }
  static get observers() {
    return [{
      path: 'bar',
      observer: 'helloBar'
    }, {
      path: 'bar',
      observer: FooElement.prototype.barIsHello
    }];
  }
  static get helloBar() {
    return this.bar + ' hello';
  }

  /** @return {boolean} */
  barIsHello() {
   if (/ hello$/.test(this.bar)) {
     return true;
   }
   return false;
  }
}
```

ChadKillingsworth:
#4574 takes care of the simple observer issue.

#4518 allows ES6 getters to be used instead of computed properties. I'm going to run some tests and make sure, but I think that [linkPaths](https://www.polymer-project.org/2.0/docs/api/elements/Polymer.Element#method-linkPaths) should provide a way to indicate that changes on one path should recalculate the getter value. This solves the computed properties use case.

For complex observers, there is no obvious solution. However perhaps it is acceptable to leave these as-is so as to discourage their use.

PR #4574, opened 2017-04-29
Support property observers which are direct function references

ChadKillingsworth:
Partially addresses #4543

Provides better static analysis and refactoring support in multiple tools. Alleviates the need for property reflection with Closure-compiler renaming.

The test case included isn't actually that useful (but does test the change). Real benefits come with the class style syntax:

```js
class FooElement extends Polymer.Element {
  static get is() { return 'foo-element'; }
  static get properties() {
    return {
      bar: {
        type: String,
        observer: FooElement.prototype.barChanged_
      }
    }
  }
  barChanged_(newValue, oldValue) {
    console.log(newValue, oldValue)
  }
}
```

### DIFF
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -208,12 +208,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    * @private
    */
   function runObserverEffect(inst, property, props, oldProps, info) {
-    let fn = inst[info.methodName];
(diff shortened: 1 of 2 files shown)


ChadKillingsworth:
Sorry for the delay. Rebased to tip of tree.

aomarks:
Filed https://github.com/Polymer/docs/issues/2421 to add documentation.

aomarks:
> Filed Polymer/docs#2421 to add documentation.

Oops, ignore me, missed that there already was an issue.

PR #4628, opened 2017-05-24
fix ArraySplice @types to more closely match code

danbeam:
There's a bunch of `.length` checks on the `current` and `previous` array args passed to the public calculateSplices() method.

I've updated from `{Array}` to `{!Array}` in every case that an unprotected `.length` check occurs (which would fail if the parameter is `null`).

I also saw no way that `null` would ever be /returned/ from this method, so I made the return type `{!Array}` (non-nullable) as well.

### DIFF
--- a/lib/utils/array-splice.html
+++ b/lib/utils/array-splice.html
@@ -146,19 +146,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    *   l: The length of the current array
    *   p: The length of the old array
    *
-   * @param {Array} current The current "changed" array for which to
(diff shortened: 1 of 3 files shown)


danbeam:
yeah, i'll rebase today

danbeam:
@TimvdLippe done

TimvdLippe:
@danbeam Thanks! Did you regenerate the typescript types as well? Should be `gulp generate-types`.

danbeam:
> Should be `gulp generate-types`.

is actually `gulp generate-typescript`, btw :wink:

PR #4632, opened 2017-05-25
Vend mixins for listeners block and hostAttributes block

### LINKED ISSUE #4625: Feature request: vend missing mixins
In 1.0 we trained people that they could use declarative `listeners` and `hostAttributes` blocks. In 2.0 we took those away but also didn't vend a helpful mixin for them to use.
Note: I am **not** suggesting we put this in `Polymer.Element`, but rather as a standalone thing in the `mixins` folder.

@azakus @tjsavage @sorvell @kevinpschaaf

azakus:
Allow for easier migration to `Polymer.Element` for elements that use legacy `hostAttributes` and `listeners` blocks in `Polymer({})`

Fixes #4625

### DIFF
--- a/lib/mixins/host-attributes-mixin.html
+++ b/lib/mixins/host-attributes-mixin.html
@@ -1,0 +1,34 @@
+<!--
+@license
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<link rel="import" href="property-accessors.html">
+<link rel="import" href="../utils/mixin.html">
+<script>
+  (function() {
+    'use strict';
+
+    Polymer.HostAttributes = Polymer.dedupingMixin((base) => {
+      let attributeBase = Polymer.PropertyAccessorsMixin(base);
+
+      class HostAttributes extends attributeBase {
+        ready() {
+          let hostAttributes = this.constructor.hostAttributes;
+          if (hostAttributes) {
+            for (let name in hostAttributes) {
+              this._ensureAttribute(name, hostAttributes[name]);
+            }
+          }
+          super.ready();
+        }
+      }
+
+      return HostAttributes;
+    });
+  })();
+</script>
--- a/lib/mixins/host-attributes-mixin.html
+++ b/lib/mixins/host-attributes-mixin.html
@@ -14,7 +14,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     'use strict';
 
     Polymer.HostAttributes = Polymer.dedupingMixin((base) => {
-      let attributeBase = Polymer.PropertyAccessorsMixin(base);
+      let attributeBase = Polymer.PropertyAccessors(base);
 
       class HostAttributes extends attributeBase {
         ready() {
--- a/lib/mixins/listeners-mixin.html
+++ b/lib/mixins/listeners-mixin.html
@@ -1,0 +1,34 @@
+<!--
+@license
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<link rel="import" href="template-stamp.html">
+<link rel="import" href="../utils/mixin.html">
+<script>
+  (function() {
+    'use strict';
+
+    Polymer.Listeners = Polymer.dedupingMixin((base) => {
+      let listenerBase = Polymer.TemplateStamp(base);
+
+      class Listeners extends listenerBase {
+        ready() {
+          let listeners = this.constructor.listeners;
+          if (listeners) {
+            for (let l in listeners) {
+              this._addMethodEventListenerToNode(this, l, listeners[l]);
+            }
+          }
+          super.ready();
+        }
+      }
+
+      return Listeners;
+    });
+  })();
+</script>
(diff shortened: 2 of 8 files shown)


lib/mixins/host-attributes-mixin.html
```diff
@@ -0,0 +1,65 @@
...
+<script>
+  (function() {
+    'use strict';
+
+    /**
+     * Element class mixin to provide default attributes and values, similar to
+     * the "hostAttributes" configuration property used in Polymer 1.x.
+     *
+     * By returning a map from a static class property named "hostAttributes"
+     * , the custom element will ensure that the given attributes are set to given value if
```
notwaldorf: lol comma first eh

lib/mixins/host-attributes-mixin.html
```diff
@@ -0,0 +1,65 @@
...
+     *
+     * For example, if an element implements hostAttributes like so:
+     *
+     * ```
+     *   class Foo extends Polymer.HostAccessors(Polymer.Element) {
+     *     static get hostAttributes() { return {'foo': 'bar' }}
+     *   }
+     * ```
+     *
+     * An instance of that element will have an attribute named "foo"
```
notwaldorf: I forgot to read this sentence because it was too spaced out, and too close to the type (so i assumed it was a compiler thing.

How about "For example, if an element implements `hostAttributes` like below, then an instance of that element will boot up with an attribute named "foo"`

lib/mixins/listeners-mixin.html
```diff
@@ -0,0 +1,68 @@
...
+     * ```
+     * class Foo extends Polymer.Listeners(Polymer.Element) {
+     *   static get listeners() {
+     *     "foo": "fooHandler"
+     *   }
+     *   fooHandler(ev) {...}
+     * }
+     * ```
+     *
+     * An instance of that element will have a "foo" event handler that calls "fooHandler" on that element.
```
notwaldorf: same as before

test/unit/attributes-elements.html
```diff
@@ -201,3 +202,94 @@
     }
   });
 </script>
+
+<script>
+  class XBasicLite extends Polymer.mixinBehaviors([PropertyTypeBehavior], Polymer.HostAttributes(Polymer.Element)) {
+    static get hostAttributes() {
```
sorvell: Add a test for subclassing that shows inheriting/overriding.

test/unit/events-elements.html
```diff
@@ -93,3 +94,15 @@
     });
   </script>
 </dom-module>
+
+<script>
+  class XListenersLite extends Polymer.mixinBehaviors([EventLoggerImpl], Polymer.Listeners(Polymer.Element)) {
+    static get listeners() {
```
sorvell: Add a test for subclassing that shows inheriting/overriding.

lib/mixins/host-attributes-mixin.html
```diff
@@ -0,0 +1,65 @@
...
+       */
+      let attributeBase = Polymer.PropertyAccessors(base);
+
+      /**
+       * @polymerMixinClass
+       * @implements {Polymer_HostAttributes}
+       */
+      class HostAttributes extends attributeBase {
+        ready() {
+          let hostAttributes = this.constructor.hostAttributes;
```
sorvell: This duplicates code in [legacy-element-mixin](https://github.com/Polymer/polymer/blob/master/lib/legacy/legacy-element-mixin.html#L149). Ideally, it wouldn't and legacy-element-mixin would use this mixin. Alternatively, this could just be moved to another repo and then the implementation can diverge slightly.

lib/mixins/listeners-mixin.html
```diff
@@ -0,0 +1,68 @@
...
+      let listenerBase = Polymer.TemplateStamp(base);
+
+      /**
+       * @polymerMixinClass
+       * @extends {listenerBase}
+       * @implements {Polymer_Listeners}
+       */
+      class Listeners extends listenerBase {
+        ready() {
+          let listeners = this.constructor.listeners;
```
sorvell: Subclassing is tricky with this implementation. Ideally you can easily inherit/override these.

MajorBreakfast:
I hope that this lands. The new way where you use `bind()` in the constructor, `addEventListener()` in `connectedCallback` and `removeEventListener()` in `disconnectedCallback` is definitely not DRY.

sz332:
Any information about the status of this PR?

arthurevans:
@azakus Ping? Is this landing? If so, we could ref it from the docs.

MajorBreakfast:
I did some investigating and I don't think that add the `listeners` array is sensible anymore:

You only need `bind()` when calling `addEventListener()` on something else than the custom element itself (e.g. window). The `this` inside the callback will be correct as the custom element is the current target of the event (and the current target is automatically used as `this`). Take a look at the code example below.
The `listeners` array only ever worked for listening for events on the custom elements itself. The same can be easily accomplished by using `addEventListener()` in `ready()`. This also won't lead to a memory leak, because the element and callback only hold references to each other, there's no external reference. Same won't be true if you call `window.addEventListener()`, but, again, the `listeners` array was never able to do that either. Then, you need to do this in the `connectedCallback()` and `disconnectedCallback`.

```JS
ready() {
  super.ready();
  this.addEventListener('click', this._onClick);
}

_onClick(event) { /* ... */ }
```

I've already submitted a PR to improve this in the documentation https://github.com/Polymer/docs/pull/2343 Currently it's a bit misleading.

sz332:
@MajorBreakfast  but why ready? Why not connectedCallback ?

MajorBreakfast:
`connectedCallback()`, unlike `ready()`, may be called more than once. It's called every time you attach your element. You'd end up with multiple listeners if you used that.

sz332:
@MajorBreakfast thank you, great explanation. Probably this should be also added to the documentation. And just being curious: why not in constructor?

MajorBreakfast:
I think it works within the constructor just as well. Without Polymer this is probably the preferred place to add your listeners (because without Polymer there's no `ready()`). `ready()` runs a bit later (first time the element is added to the DOM). But, there's absolutely no difference if no events are fired that you listen to. Maybe setting them up in `ready()` has an advantage, but I don't see it. Works either way.

sz332:
@MajorBreakfast if your add the listener in connectedCallback, but remove it in disconnectedCallback, is this okay?

MajorBreakfast:
You can do that. But this approach is only necessary if you listen on something like `window`, because there you'd run into a memory leak if you used the simple way with `ready()`. Also, be sure to do it exactly as stated in the Polymer docs https://www.polymer-project.org/2.0/docs/devguide/events#imperative-listeners

sz332:
@MajorBreakfast would it be possible to extend the documentation with the information you shared in this PR? Because it was very enlightening, and now I understand how the event listener works, and probably others had also problem with this. Thanks.

MajorBreakfast:
@sz332 I've recently submitted this as a PR https://github.com/Polymer/docs/pull/2343/files It's not merged yet. Also, I've added the information to use `ready()` instead of `connectedCallback()` just now.

PR #4635, opened 2017-05-26
Not an RC anymore

arthurevans:
Subject line says it all.


README.md
```diff
@@ -238,7 +239,8 @@ The Polymer library uses a BSD-like license that is available [here](./LICENSE.t
 -----------
 
 <a name="release-candidate"></a>
```
TimvdLippe: Also delete this line probably?

arthurevans: I was thinking of leaving it in in case folks had it bookmarked, but I don't feel strongly about it ... 🍡

PR #4637, opened 2017-05-26
Upgrade babel-preset-babili

nazar-pc:
Latest release of babel-preset-babili fixes breaking complex regular expression with corresponding upstream bug report being https://github.com/babel/babili/issues/526 and PR that fixes it being https://github.com/babel/babili/pull/490 that also affects Polymer.


azakus:
LGTM

PR #4645, opened 2017-05-31
Fix some expressions end without a semicolon

baocang:


TimvdLippe:
Closing in favor of #4651. We will discuss there whether we want to enforce this rule or not.

PR #4649, opened 2017-05-31
Special-case undefined textarea.value same as input. Fixes #4630

### LINKED ISSUE #4630: <textarea> welcomes user with 'undefined' text value (only for object subproperties)
### Description
An `undefined` value, when binding to `<textarea>`,  is treated as 'undefined' text, but only if it appears in an object subproperty.
Interestingly `null` gets treated just fine. `undefined` in a flat property is also just fine.
`<input type="text">` doesn't have the issue.
 See the live demo.

#### Live Demo
http://jsbin.com/xukejamegi/edit?html,output

#### Expected Results
It should work like `<input type="text">`

#### Actual Results
The user value is an 'undefined' text.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 9
- [x] Safari 8
- [x] IE 11

### Versions

- Polymer: v1.9.1, v2.0.0 (we are using 1.9.1 - please **do** try to fix 1.X)
- webcomponents: 0.7.24

kevinpschaaf:
### Reference Issue
Fixes #4630

### DIFF
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -598,7 +598,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     if (binding.kind !== 'attribute') {
       // Some browsers serialize `undefined` to `"undefined"`
       if (binding.target === 'textContent' ||
-          (node.localName == 'input' && binding.target == 'value')) {
+          (binding.target === 'value' &&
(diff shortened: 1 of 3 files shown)


TimvdLippe:
@azakus shall we merge this one?

PR #4666, opened 2017-06-05
Add explicit imports across files inside of polymer

FredKSchott:
This PR adds back missing imports. These explicit references are needed for some automated tooling.

/cc @justinfagnani

### DIFF
--- a/lib/utils/style-gather.html
+++ b/lib/utils/style-gather.html
@@ -8,6 +8,7 @@ Code distributed by Google as part of the polymer project is also
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 -->
 <link rel="import" href="resolve-url.html">
+<link rel="import" href="../elements/dom-module.html">
 <script>
 (function() {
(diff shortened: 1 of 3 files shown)


lib/utils/style-gather.html
```diff
@@ -8,6 +8,7 @@
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 -->
 <link rel="import" href="resolve-url.html">
+<link rel="import" href="../elements/dom-module.html">
```
justinfagnani: This file shouldn't have an import of dom-module. The check below is to see if it's loaded and use it if so, but there are some uses of style-gather that don't necessitate loading dom-module, like custom-style.

FredKSchott: Agreed, removing

FredKSchott:
@justinfagnani @sorvell PTAL, PR is now a single line

justinfagnani:
LGTM from my end

PR #4668, opened 2017-06-06
guard all references to Polymer.DomModule

FredKSchott:
Per our discussion, in case it doesn't exist.

/cc @justinfagnani


FredKSchott:
Bump! also /cc @azakus

justinfagnani:
LGTM from me at least :)

PR #4675, opened 2017-06-07
Store original template content inside templateInfo

baszczewski:
In some scenarios it is important to have access to original template content even without 'preserve-content' attribute. 

For example in my application I have custom table element where users can insert many templates inside lighting dom for all columns. From this columns I have to create row template dynamically (only this one will be templatized). Unfortunately cell templates (without 'preserve-content') don't work correctly because of consumed property binding (content from _contentForTemplate is also not enough here). 
From other side I don't want use  'preserve-content' attribute because this hits API usability.

### DIFF
--- a/lib/mixins/template-stamp.html
+++ b/lib/mixins/template-stamp.html
@@ -306,7 +306,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        *   metadata to `nodeInfo`
        */
       static _parseTemplateNestedTemplate(node, outerTemplateInfo, nodeInfo) {
+        let originalContent = node.content.cloneNode(true);
         let templateInfo = this._parseTemplate(node, outerTemplateInfo);
+        templateInfo.originalContent = originalContent;
         let content = templateInfo.content =
           node.content.ownerDocument.createDocumentFragment();
         content.appendChild(node.content);


baszczewski:
I signed it!

kevinpschaaf:
We are maintaining the parsing API as protected API that users are free to override in their own subclasses (or class mixins) exactly for purposes like this.

e.g.

```js
class MyBaseClass extends Polymer.Element {

  static _parseTemplateNestedTemplate(node, outerTemplateInfo, nodeInfo) {
    let originalContent = node.content.cloneNode(true);
    let ret = super._parseTemplateNestedTemplate(node, outerTemplateInfo, nodeInfo);
    nodeInfo.templateInfo.originalContent = originalContent;
    return ret;
  }

}
```

Is there a reason this needs to be a general feature (that will tax every Polymer element with the cost of cloning nested template content) as opposed to something you do in the base class of elements that need this?

baszczewski:
Thank you for your answer and some details about Polymer internals. It is really helpful. 

Unfortunetly this not resolve my problem completly because my template is "broken" by dom-bind and not my custom element itself. So I have to change dom-bind implementantion or create my own version with originalContent. Probly this will be acceptable for me.

PR #4684, opened 2017-06-12
Closure warnings

azakus:
Full support for Closure Compiler


kevinpschaaf:
What exactly is the difference between the type of warnings that go away with `@unrestricted` vs `@suppress(missingProperties)`?

lib/legacy/legacy-element-mixin.html
```diff
@@ -433,8 +459,8 @@
        * that contains an insertion point (`<slot>`) inside its subtree.
        */
       distributeContent() {
-        if (window.ShadyDOM && this.shadowRoot) {
-          this.shadowRoot.forceRender();
+        if (window.ShadyDOM) {
+          window.ShadyDOM.flush();
```
kevinpschaaf: Curious why this is showing as a change?  Just wasn't rebased recently?

lib/mixins/property-effects.html
```diff
@@ -86,7 +89,7 @@
    * Runs all effects of a given type for the given set of property changes
    * on an instance.
    *
-   * @param {Object} inst The instance with effects to run
+   * @param {PropertyEffectsInstance | Polymer_PropertyEffects} inst The instance with effects to run
```
kevinpschaaf: What does this mean?  Seems like one or the other, no?  Why doesn't it need to be a union of those two things?

azakus:
`@unrestricted` allows for adding instance properties outside of `constructor()`.

`@suppress {missingProperties}` suppresses the warnings generated by accessing properties that the compiler does not know that the class has.
Good examples include anywhere a class is dynamically assigned a property, or a type interface doesn't include private properties.

lib/elements/array-selector.html
```diff
@@ -31,15 +31,26 @@
...
+   * master `items` array and a `selected` array
    */
   let ArraySelectorMixin = Polymer.dedupingMixin(superClass => {
 
     /**
+     * @constructor
+     * @extends {superClass}
+     * @implements {Polymer_PropertyEffects}
+     */
+    let propertyEffectsBase = Polymer.PropertyEffects(superClass);
```
azakus: Seems reasonable to me. That seems like a bug in master as well then.

lib/legacy/legacy-element-mixin.html
azakus: This must have been missed in a merge conflict. I'll pull the master branch version.

lib/legacy/legacy-element-mixin.html
```diff
@@ -188,7 +210,7 @@
        * @return {string} Serialized value
        */
       serialize(value) {
-        return this._serializeValue(value);
+        return this._serializeValue(value) || '';
```
azakus: Ok.

ChadKillingsworth:
Man I'm glad to see this PR.

> @suppress {missingProperties} suppresses the warnings generated by accessing properties that the compiler does not know that the class has.
Good examples include anywhere a class is dynamically assigned a property, or a type interface doesn't include private properties.

Be careful when you do this that you don't inadvertently lie to the compiler about types. The type based optimizations will create some really hard to debug issues with blatantly wrong type information. Missing type information is not a problem for this though.

lib/mixins/property-effects.html
azakus: Actually, PropertyEffectsInstance already extends Polymer_PropertyEffects so I can probably remove the type union here.

In general though, I would have to make a new type to do a union like you would want.

azakus:
@ChadKillingsworth good points. I've been trying to use suppressions sparingly, mostly around edges where the lazy instantiation of class properties, but this is a good time to maybe try to shape the classes with `null`s for those properties ahead of time.

ChadKillingsworth:
@azakus You can also just add a stub definition for the compiler:

```js
constructor() {
  /** @type {number} */
  this.myPropIThatIsNotInitialized;
}
```

lib/mixins/property-effects.html
azakus: Looks like I need a bunch of casts now because a lot of methods in `PropertyEffects` class call out to these functions and the type definition doesn't really exist

lib/mixins/property-effects.html
```diff
@@ -1614,7 +1671,7 @@
         if (this.__propagateEffects) {
           runEffects(this, this.__propagateEffects, changedProps, oldProps, hasPaths);
         }
-        let templateInfo = this.__templateInfo;
+        let templateInfo = /** @type {TemplateInfo} */(this.__templateInfo);
```
kevinpschaaf: Can remove cast since it's typed in ctor right now (to be moved to types file)

lib/mixins/property-effects.html
```diff
@@ -1681,10 +1738,9 @@
        *   reported in index order (raw results from `Object.observe` are not
        *   ordered and must be normalized/merged before notifying).
        * @public
-       * @suppress {missingProperties}
       */
       notifySplices(path, splices) {
-        let info = {};
+        let /** PathInfo */ info = {path: ''};
```
kevinpschaaf: Can eliminate the type info here since it should be structural.

lib/mixins/template-stamp.html
```diff
@@ -349,8 +346,8 @@
        * and `id` attributes.
        *
        * @param {Element} node Node to parse
-       * @param {Object} templateInfo Template metadata for current template
-       * @param {Object} nodeInfo Node metadata for current template.
+       * @param {!TemplateInfo} templateInfo Template metadata for current template
+       * @param {!NodeInfo} nodeInfo Node metadata for current template.
        * @param {string} name Attribute name
        * @param {*} value Attribute value
```
kevinpschaaf: value arg should always be string; can remove cast below

lib/mixins/property-effects.html
```diff
@@ -1089,12 +1102,58 @@
...
+     * @mixinClass
      * @implements {Polymer_PropertyEffects}
+     * @extends {propertyEffectsBase}
      * @unrestricted
      */
     class PropertyEffects extends propertyEffectsBase {
 
+      constructor() {
+        super();
+        /** @type {boolean} */
```
azakus: This is hung up by some sort of internal compiler error :/

azakus: https://github.com/google/closure-compiler/issues/2556

ChadKillingsworth:
I worked with this branch a bit over the weekend. A few notes:

 * Moving the interfaces from externs to source works great - except for the static properties. Since the compiler doesn't properly model static inheritance yet, leaving the static properties as externs seems appropriate to me.
 * Any extern or interface function that takes no arguments and returns nothing (it's only used for side effects) needs an `@return {undefined}` annotation. Otherwise, the compiler assumes it's of type `function(...?):?` - which is pretty horrible.
 * I recommend switching from `@record` to `@interface`. I had no problems doing that myself. `@record` doesn't get near the renaming that `@interface` does and you aren't currently using structural type checking.

Here's an example of the changes I made: https://github.com/ChadKillingsworth/polymer/commit/6f2776473f90009f463d604292a5c96617f78335

azakus:
@ChadKillingsworth thanks!

azakus:
Looks like the difference between putting all interfaces into externs and splitting statics out into externs is around 2kb in binary size.
Worthwhile, but not 100% necessary in my mind. 
I think this can merge to master with the last two modifications described in https://github.com/Polymer/polymer/pull/4684#issuecomment-312627603

azakus:
https://github.com/google/closure-compiler/pull/2338 has merged, so now we can go ahead and merge this branch!

ChadKillingsworth:
Yay!

PR #4697, opened 2017-06-19
Updating the license name to the SPDX identifier.

reiz:
By using the SPDX identifier other tools like [VersionEye](https://www.versioneye.com) can handle the license more easily.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -3,7 +3,7 @@
   "main": [
     "polymer.html"
   ],
-  "license": "http://polymer.github.io/LICENSE.txt",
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Sadly we can not accept such changes to the license without a legal procedure.

reiz:
@TimvdLippe Is the legal procedure triggered now through this pull request? Pretty much all other NPM & Bower modules are using a license name in the `license` attribute. The polymer project is an exception here. Would be nice if we could change that.

PR #4699, opened 2017-06-20
Adds `selectIndexes` and`selectAll` to `array-selector`

### LINKED ISSUE #3176: array-selector needs selectAll method
I'm using array-selector with an iron-list and many thousands of items.  I need to implement a `selectAll` method and have found two ways of doing so, neither optimal.  The first is to simply assign to the array-selector's `selected` property.  This is very fast and appears to work at first but bypasses the internal linkages such that nothing can be deselected.  The second technique is this:

```
var s = this.$.selector;
var items = s.items;
for (var i=0; i<items.length; ++i){
    s.select(items[i]);
}
```

This works but is very slow, and in my case causes Chrome to prompt me to kill the page due to unresponsiveness.

If the internals of array-selector allow for it, a `selectAll` would be extremely helpful.

### LINKED ISSUE #3827: array-selector to select all / deselect all
Looking at the documentation it seems we can't do select all / deselect all with array-selector. It would be great if this element can get that feature. Its very common that we have to use select all / deselect all. Anyone knows how we can implement a work-around for this with array-selector ? I tried to at onload to iterate the master array and run the code to select / deselect all but its not working for me. Thanks for your help !

Westbrook:
### Reference Issue
Fixes #3176
Fixes #3827

These additions are currently for `2.x`, while the above tickets are about `1.x`. I'm working on a similar PR for the `1.x` line of code, so that between the two I can fully address https://github.com/PolymerElements/iron-list/issues/436 and the hybrid nature of the component requires that this support functionality gets into both lines.

### DIFF
--- a/lib/elements/array-selector.html
+++ b/lib/elements/array-selector.html
@@ -328,10 +328,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           if (idxs.length) return this.selectIndex(idxs[0]);
         } else {
           this.__selectedMap.clear();
-          let items = idxs.map((idx) => {
+          let items = [];
+          idxs.forEach((idx) => {
             let item = this.items[idx];
             this.__selectedMap.set(item, idx);
-            return item;
+            items.push(item);
           });
           this.__updateLinks();
           this.set('selected', items);
--- a/lib/elements/array-selector.html
+++ b/lib/elements/array-selector.html
@@ -321,7 +321,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * @param {array} idxs Indexes from `items` array to select
        */
       selectIndexes(idxs) {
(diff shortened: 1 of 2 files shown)


Westbrook:
Any thoughts on this from the team? I'd be happy to make any changes to the code, comments, tests, or PR that would allow us to move forward with this. It seems like a good amount of people would also be happy for the addition, and it would it would be really helpful in the conversion to 2.0 that my team and I are working on.

TimvdLippe:
Sorry for the long delay. After a discussion we decided to not include this behavior in `<array-selector>`. We would like to minimize the API surface we need to support and ship and it is quite easy to implement this functionality yourself. Check out http://jsbin.com/sebohupovu/edit?html,console,output which extends `array-selector` and has the corresponding methods implemented.

PR #4705, opened 2017-06-22
selectItems, selectAll, and tests

### LINKED ISSUE #3176: array-selector needs selectAll method

### LINKED ISSUE #3827: array-selector to select all / deselect all

Westbrook:
### Reference Issue
Fixes #3176
Fixes #3827

This relates to the 2.x version of this PR #4699.

These two PRs will allow me to fully address the changes needed for PolymerElements/iron-list#436.

### DIFF
--- a/src/lib/template/array-selector.html
+++ b/src/lib/template/array-selector.html
@@ -212,6 +212,41 @@ is false, `selected` is a property representing the last selected item.  When
           this.linkPaths('selectedItem', 'items.' + key);
         }
       }
+    },
+
+    /**
+     * Selects the given items.
+     *
+     * @param {array} items Items from `items` array to select
+     */
+    selectItems: function(items) {
+      if (arguments.length > 1 || (typeof items !== 'undefined' && !Array.isArray(items))) {
+        items = Array.from(arguments);
+      }
+      var icol = Polymer.Collection.get(this.items);
+      var key;
+      var skey;
+      var item;
+      if (this.multi) {
+        this.set('selected', items);
+        this._selectedColl = Polymer.Collection.get(this.selected);
+        for (var i=0; i<items; i++) {
+          item = items[i];
+          key = icol.getKey(item);
+          skey = this._selectedColl.getKey(item);
+          this.linkPaths('selected.' + skey, 'items.' + key);
+        }
+      } else if (items.length){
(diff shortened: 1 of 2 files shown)


Westbrook:
Any thoughts on this from the team? I'd be happy to make any changes to the code, comments, tests, or PR that would allow us to move forward with this. It seems like a good amount of people would also be happy for the addition, and it would it would be really helpful in the conversion to 2.0 that my team and I are working on.

Westbrook:
I know that there is very little 1.x work being done, but I did see the recent patch that went out and am seeing the `<custom-style>` work being back ported to there, does anyone have any thoughts as to whether we could get this on the docket for the release that has to come together for that? Possibly @TimvdLippe, who's been making amazing waves on the issues/PR scene?

TimvdLippe:
The `<custom-style>` changes were required to prevent an upcoming Chrome change to break Polymer websites. This PR targets new functionality, so depending if #4699 gets merged, we might merge this one. The reason we have not reviewed the PRs is because we have to quantify the benefit of shipping and maintaing the extra API. E.g. maybe it is possible already with element extension to implement this functionality. Probably this afternoon we can take a look and give you some feedback :smile:

TimvdLippe:
Closing per https://github.com/Polymer/polymer/pull/4699#issuecomment-331311907 Check out the JSBin over there 😄

PR #4706, opened 2017-06-22
Resolves Issue with DeSerialization of Date: #4696

### LINKED ISSUE #4696: Issue with DeSerialization of Date
var x = "1497810600000" //string casted timestamp
then, new Date(X) --> Invalid date.

The constructor of Date Object allows string input for long form of date only and not timestamp.
Can we update the deserialize function in Polymer.Base (probably in _deserializeValue () method of property-accessors.html) with following:

```
case 'Date':
var serialized;
        if(value instanceof Date) {
            serialized = Polymer.Base.deserialize(value, 'Date').getTime();
        } else {
            value = isNaN(value) ? value : Number(value);
            serialized = new Date(value).getTime();
        }
break;

```

ronak1009:
Fixes #4696: Issue with DeSerialization of Date


sorvell:
Sorry for the delay. Would you mind adding a test?

ronak1009:
@sorvell I have added the test case and updated the pull request

test/unit/property-accessors.html
```diff
@@ -112,6 +112,22 @@
...
+    var dateArr = [];
+    dateArr[0] = new Date();
+    dateArr[1] = String(dateArr[0].getTime()); //string timestamp
+    dateArr[2] = dateArr[0].getTime(); //number timestamp
+    dateArr[3] = dateArr[0].toString(); //full date
+    setTimeout(function(){
+      let ser = [];
+      ser[0] = window.XFoo.prototype._deserializeValue(dateArr[1], Date);
+      ser[1] = window.XFoo.prototype._deserializeValue(dateArr[2], Date);
+      ser[2] = window.XFoo.prototype._deserializeValue(dateArr[3], Date);
```
TimvdLippe: It seems like there is no assertion for this line? Could you add one for it? Also the variable name `ser` seems cryptic to me. Would be better to use a longer name to improve readability

ronak1009: here I am just storing the return values of the deserializeValue function. I have used them in assert statements below.
I took this approach so that assert statements becomes more readable.

TimvdLippe: I don't disagree with the separation of the `serializedValue` array and the assertions, but I meant that I don't see any reference to `ser[2]` in the assertions on the lines below.

TimvdLippe: Ping @ronak1009 could you rebase and update the variable names per this comment 😄

TimvdLippe:
I manually updated the test suite in https://github.com/Polymer/polymer/commit/86a64b6c7e9b5a1d4dfdddc1f35e4a8efe76c0b5

PR #4710, opened 2017-06-24
Create Polymer.DomSwitch mixin.

### LINKED ISSUE #4709: Create Polymer.DomSwitch mixin
### Description
Create Polymer.DomSwitch mixin or base class which chooses between multiple templates based on the value of the `type` property.

#### Live Demo
https://jsbin.com/mecuhoxadi/edit?html,output

### Versions
- Polymer: master
- webcomponents: master

So far I have only tested this feature on Chrome.

coreyfarrell:
This mixin allows creation of components which choose a template to
stamp based on the value of the `type` property.

### Notes
* This is not ready for merge as it requires tests and the documentation could probably use some more love.
* I'm unsure the code to dispatch dom-change event is needed or if this will happen automatically due to stamping a template?

### Reference Issue
Fixes #4709

### DIFF
--- a/lib/elements/dom-switch.html
+++ b/lib/elements/dom-switch.html
@@ -1,0 +1,38 @@
+<!--
+@license
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<link rel="import" href="../../polymer-element.html">
+<link rel="import" href="../mixins/dom-switch.html">
+
+<dom-module id="dom-switch">
+  <script>
+    (function() {
+      'use strict';
+
+      /**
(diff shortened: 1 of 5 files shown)


coreyfarrell:
Basic tests have been added.

TimvdLippe:
The usage of the dom-switch seems very cumbersome and verbose to me, especially looking at https://github.com/Polymer/polymer/pull/4710/files#diff-add0058828e84802e42329abc2516d80R44 I wonder if this could be reduced in some way.

coreyfarrell:
@TimvdLippe it's worth noting you linked to `<dom-module id="x-two-of-each">` which tests that if multiple template's of the same type are provided only the first is used.  https://github.com/Polymer/polymer/pull/4710/files#diff-add0058828e84802e42329abc2516d80R24 `x-one-of-each` is a more reasonable test case (without duplicate/unused templates).

It's important to note this isn't meant to be a replacement for `iron-pages` (which is easier to use but less efficient).  In my own use case I have dozens of templates an element which extends `DomSwitch`, and that element is used inside a `dom-repeat` bound to an array with dozens of items.  With DomSwitch the shadowRoot for each instance contains only the active type, vs `iron-pages` which would stamp DOM for everything with most of it hidden.

TimvdLippe:
We had a quick discussion about this PR. We would like to not expand the different dom-* elements in Polymer, as maintaining dom-if and dom-repeat is already a high-cost maintenance burden. If you would like to ship this element as is and publish on webcomponents.org, that would be okay. However, we will not incorporate this in the Polymer core library.

PR #4721, opened 2017-07-05
support limit property (post filter/sort)

### LINKED ISSUE #4720: Feature Request: dom-repeat limit
### Description
I have been trying to establish a method of limiting the items using a filter callback.  This method works fine, unless you are also using a sort callback.

`dom-repeat` currently filters elements and then sorts them, presumably this is a less intensive process for larger datasets than sorting first.

It would be a very useful feature to limit the number of instances generated after both filtering and sorting.

### Example
`<template is="dom-repeat" items="[[items]]" sort="_sort" filter="_filter" limit="5">`

TomK:
- refactor __limit to __chunkLimit
 - add limit property and observer
 - constrain instance generation by __chunkLimit AND limit
 - clone limit tests to ensure chunk and limit both work independently

### Reference Issue
Fixes #4720

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -203,6 +203,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         },
 
         /**
+         * Define the total number of items to render after filtering and sorting.
+         * Setting limit to zero or negative equates to no limit.
+         */
+        limit: {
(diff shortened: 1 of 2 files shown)


TomK:
Please let me know if i'm barking up the wrong tree with this solution.

TomK:
Closing this, pending discussion.

PR #4724, opened 2017-07-05
Add support for passive event listeners

### LINKED ISSUE #3604: Add support for passive event listeners
Echoing @RByers on pushing for support of passive event listeners which could see fairly significant scrolling performance improvements across the board.

Request on jquery: https://github.com/jquery/jquery/issues/2871
Explainer: https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md
Chrome status page: https://www.chromestatus.com/features/5745543795965952

This hits Chrome Stable (51) in about a month.

### LINKED ISSUE #3787: Perf: Do not add document-level blocking touch listeners
All polymer apps get a touch listener added to `document` by default?

screen shot 2016-07-19 at 5 46 31 pm
screen shot 2016-07-19 at 5 47 01 pm
screen shot 2016-07-19 at 5 47 29 pm

Doing so is typically frowned upon b/c it can have negative effects on page load and scrolling perf, especially on mobile. What are these events for? Can they be passive events by default...or can we have users opt-into passive events somehow if they know they're not going to call `preventDefault()`?

https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md

cc @RByers @azakus

### LINKED ISSUE #4667: gestures.html adds non-passive event listener to a scroll-blocking 'touchstart' event
This causes chrome to log Violation warnings
screen shot 2017-06-05 at 5 04 35 pm

azakus:
Provide a way for elements to have passive gesture listeners with
`this._passiveGestures = true`

Fixes #4667
Fixes #3787
Fixes #3604

TODO:
- [ ] add tests


dtapuska:
@TimvdLippe  Tim you should be able to test this via calling preventDefault() and checking whether defaultPrevented was true on the Event.

azakus:
After some thought, I don't think this approach is really workable. It requires individual elements to agree on setting `passive`, but at an application level, it may not be possible to know which element set the listeners that may affect scrolling.

Instead #4829 has a more limited version that is 100% app controlled.

PR #4730, opened 2017-07-11
Passing template argument to render function

anuragpathak:
When Mutation observer listens to addition of Template element and then it passes it again to render function.
Earlier , Again this.querySelector was getting called, to prevent it if argument is passed, use it else go for querySelector


anuragpathak:
JS DOC Error

TimvdLippe:
Closing per above comment. Please reopen if you can show your described usecase.

PR #4745, opened 2017-07-23
Do not assign to a readonly property on window

rictic:
If `window.customElements` exists there's a good chance that it's a readonly property as per spec, and assigning to readonly properties throws an exception in strict mode.


rictic:
Reviewed and approved in person by @sorvell

PR #4748, opened 2017-07-24
 Use customElements.get rather than referring to the global for Polymer.DomModule

rictic:


rictic:
Reviewed and approved in person by @sorvell

PR #4762, opened 2017-08-01
Better distinguish param name from namespaced name

rictic:


lib/utils/settings.html
```diff
@@ -90,8 +90,8 @@
    * @memberof Polymer
    * @param {(function(*,string,string,Node):*)|undefined} sanitizeDOMValue the global sanitizeDOMValue callback
```
azakus: Shouldn't this change to `newSanitizeDOMValue` as well?

rictic: +1, done

rictic:
This change is just to make things simpler for tooling

PR #4768, opened 2017-08-03
Rewrite expression parser to a state machine in optional mixin

### LINKED ISSUE #1291: Curlies in styles break the preprocessor
Things like:

``` css
.foo:after {
  content: '{';
}
```

### LINKED ISSUE #1734: Consider annotator differentiating between `[[3, 4]]` and `[[binding]]`

### LINKED ISSUE #1978: consider if `true` and `false` are supported literal inputs to inline computations

### LINKED ISSUE #3870: Binding to properties having '/' in the key doesn't work

### LINKED ISSUE #4239: Computed bindings can't handle boolean literals true or false
Demo: http://plnkr.co/edit/r5FwQDDiSkwA4E7smAzx

In the code below, the 2nd and 3rd binding doesn't call the function at all. The last one works, but is not what I want, since "1" doesn't capture the meaning as much as "true". Looks like a bug to me.

```html
<link rel="import"  href="https://rawgit.com/Polymer/polymer/v1.7.1/polymer.html">

<dom-module id="dom-element">
  <template>
    <p>foo(field): {{foo(field)}}</p><br>
    <p>foo(field, true): {{foo(field, true)}}</p><br>
    <p>foo(field, false): {{foo(field, false)}}</p><br>
    <p>foo(field, 1): {{foo(field, 1)}}</p><br>
  </template>
</dom-module>

<script>
  Polymer({
    is: "dom-element",
    
    properties: {
      field: {
        type: Number
      },
    },
    
    foo: function(val, twice=false) {
      return twice ? val*2 : val
    },
  });
</script>
```

### LINKED ISSUE #4723: Data binding doesn't parse Japanese unicode characters
Version 1.6.1:

`value="{{data.商品名}}"`
evaluates to the string "{{data.商品名}}" instead of dereferencing data.商品名

TimvdLippe:
Implement a new binding expression parser as statement machine instead of a regex. This solves the problem of maintaining a whitelist of characters that are allowed in bindings, e.g. it allows Japanese unicode characters, slashes, etc... It can better deal with unbalanced brackets, e.g. `{{asdf]]` and `[[asdf}}`, since we can maintain state and escaping in strings is more intuitive.

Regarding performance: I ran the supplied test on current master and on this branch. The new binding expression parsers took 950 ms for 1000 templates while current master took 700 ms. Have to take a deep dive in what is taking a lot of time, but would like to get some initial feedback first.

TODO:
- [x] Documentation
- [x] Performance improvements
- [x] Check compliance with Closure Compiler
- [ ] Add mechanism to check if parsing succeeded, to improve error handling of incorrect bindings

### Reference Issue
Fixes #1734
Fixes #1978
Fixes #3870
Fixes #4239 
Fixes #4723
Supersedes and closes #3871


TimvdLippe:
Performance is now on-par with the old regex.

sorvell:
It would be great to see if this PR could be done as an optional mixin. Hopefully there are enough hooks in the binding system to make this possible.

The concern is the size and speed of the additional code here. This plus the fact that these issues are mostly pretty corner case makes me think that perhaps an optional mixin might be a better fit.

azakus:
Just a few eslint issues: https://travis-ci.org/Polymer/polymer/builds/265406074

TimvdLippe:
This is now extracted to a mixin and tested separately. We might still want to take a closer look at the performance.

TimvdLippe:
@sorvell renamed!

lib/utils/strict-binding-parser.html
```diff
@@ -0,0 +1,405 @@
+<!--
```
sorvell: Let's move this file to `lib/mixins`since it's.... a mixin.

TimvdLippe: Done

lib/mixins/strict-binding-parser.html
```diff
@@ -0,0 +1,405 @@
...
+              }
+              break;
+            }
+            case STATE.METHOD: {
+              switch (char) {
+                case ')': {
+                  storeMethodVariable(bindingData, text, i);
+                  storeMethod(bindingData, templateInfo);
+                  bindingData.startChar = i + 1;
+                  storeMethod(bindingData, templateInfo);
```
sorvell: Is this a typo to call `storeMethod` a second time here? If so, let's fix and add a test to verify.

sorvell: @TimvdLippe acknowledged issue and will fix.

TimvdLippe: I no longer have write access to this repository, so I submitted #4866 on this branch to fix the issue

sorvell:
We need to fix the test issues before merging.

justinfagnani:
Hey @TimvdLippe, we talked about this a bit already, but there's a lot of overlap here with the polymer-expressions 2.0 branch: https://github.com/Polymer/polymer-expressions/tree/2.0-preview

I think that recursive descent parsers are generally easier to reason about and maintain than hand-written state machines. It'd be interesting to see if polymer-expressions were suitable to use as the more strict parser either if 1) were stripped down to just the syntax and features that Polymer currently supports, or 2) we offered the full syntax as an opt-in for more powerful expressions.

@kevinpschaaf mentions that this PR handles finding and matching the expression delimiters too, and that would need to be done for polymer-expressions. He and I have both not been able to put together the time to finish the push to integrating it is a mixin. I wonder if you could investigate that and try to get it actually working to compare against this option.

TimvdLippe:
@justinfagnani Yes I did consider using `polymer-expressions` instead. However, the original bindings were detected by a regex, which is quite performant, but also quite unmaintainable. I did multiple iterations of my overall idea (on how to remove the whitelist of accepted characters and take into account delimiters). Most solutions were simply not performant enough to be comparable with the regex.

The pure state machine was the solution that did get very close (not 100% same speed), while it is more maintainable than the original regex (imo). I fear that a full-blown generalized parser (like in `polymer-expressions`) is too big and slow for such a small syntax, that's why I did not go down that route.

Nonetheless, I will try to port the expressions in a similar manner to this PR and confirm/deny my suspicion.

TimvdLippe:
@sorvell Build is passing now and docs + types are updated.

TimvdLippe:
For me: update the docs to state performance vs correctness trade-off, fix conflicts and then merge. Verbal LGTM by @sorvell

PR #4769, opened 2017-08-03
Test cleanup for tooling

rictic:
A bunch of minor noop changes that make it easier for our tooling to work with the tests.

### DIFF
--- a/lib/utils/gestures.html
+++ b/lib/utils/gestures.html
@@ -337,10 +337,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (handled.skip) {
         return;
       }
-      let recognizers = Gestures.recognizers;
       // reset recognizer state
-      for (let i = 0, r; i < recognizers.length; i++) {
-        r = recognizers[i];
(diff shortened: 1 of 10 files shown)


TimvdLippe:
Last commit seems to have broken the linter :cry:

FredKSchott:
FWIW: Any tooling that needs to analyze user-code should know how to handle many of these edge cases. Not -1 on anything in this PR, but make sure there are separate issues to track any edge-cases removed from core.

Ex: We've known `recognizers` was harder to analyze/tool than `Gestures.recognizers` but didn't change it at the time because we knew we needed to handle our users doing this anyway.

rictic:
Yeah. I've filed bugs for the main issues. The `Gestures.recognizer` name issue is a big one that we need to handle correctly.

Requiring that a project run under strict mode seems like a smaller deal to me. We can just document that when it's a limitation and normal linters can catch that stuff. (here the linter was explicitly configured to treat `stamped` as a global).

The rest of the changes are pretty obscure edge cases, IMO, and we can just document them as known limitations.

Issue #4137, opened 2016-11-09
Implement build time optimizations


sorvell:
In addition to polymer [build](https://www.polymer-project.org/1.0/docs/tools/polymer-cli) and the [polymer css build](https://github.com/PolymerLabs/polymer-css-build), there are a number of other possible build time optimizations that are worth exploring. Our current timeline is to invest in these features shortly after shipping version 2.0 of Polymer.

* [strip whitespace](https://github.com/Polymer/vulcanize/issues/396)
* pre-compile parsing of templates for bindings
* pre-compile polymer classes to include behaviors and property effects

TimvdLippe:
Pre-building has been researched and a prototype is available at https://github.com/polymerlabs/loose-it Any further issues/improvements can be filed in that repository.

PR #4782, opened 2017-08-09
Small refactoring in effect metadata

TimvdLippe:
Refactor effect metadata system with a couple of modifications:

* Add fallback for when data does not exist. This is primarily the start for #4137 and potential later changes to choose better defaults to reduce memory
* Factor out the effect functions from `property-effects`. This is primarily a readibility improvement.
* Rewrite `findTemplateNode` to operate on an array instead of nested objects. This stores significantly less information in a compressed format.

### Reference Issue
These are preparations #4137

### DIFF
--- a/lib/legacy/polymer-fn.html
+++ b/lib/legacy/polymer-fn.html
@@ -12,6 +12,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 <script>
 
+  window.elements = {};
+
   (function() {
     'use strict';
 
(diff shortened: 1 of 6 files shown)


TimvdLippe:
As this PR grew big over time, I am factoring out some of the sub-changes. This makes reviewing easier and make sure that the changes are not all-or-nothing, but pick those that we like.

I extracted the `findTemplateNode` change to #4853

sorvell:
Since there are potentially breaking changes in this PR, we are likely to consider it for Polymer 3.x.

TimvdLippe:
Closing the PR for now, but keeping the branch alive. We will revisit this for Polymer 3.

PR #4783, opened 2017-08-09
Fix compiling with Polymer({}) calls

azakus:
Add types for instance properties of ElementMixin and LegacyElementMixin


TimvdLippe:
How can I test that this solves the closure warnings?

azakus:
@TimvdLippe with https://github.com/polymerlabs/polymer-closure-sample and a compiled-from-head closure compiler :P

TimvdLippe:
Ah that explains why my closure gulp task was not working yesterday and I couldn't find an update on npm. Thanks!

azakus:
The next release of closure-compiler should have the `---polymer_version=2` flag supported.

PR #4786, opened 2017-08-10
Update closure annotation for all paths

beckysiegel:
Polymer.Path.get accepts both a string path or an Array path, so
functions that call this should allow for either as well.  Already
changed for Polymer.prototype.push here:

https://github.com/Polymer/polymer/pull/4776


azakus:
@beckysiegel Oh sorry, closure-types.js is a generated file from `gulp generate-closure-externs`!
I've already updated it with the changes from your PR (#4776) in c20b6574776cfb5e0036df6346d8fe8a2020dc03

azakus:
Oops, my mistake, this affects more than that file 😨

beckysiegel:
@azakus done!

PR #4789, opened 2017-08-11
Update closure compiler to support polymer pass v2

azakus:
Looser typing for Polymer({}) to support polymer pass transform

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
   },
   "scripts": {
     "build": "gulp",
-    "test": "gulp lint && wct"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
     "del": "^3.0.0",
     "dom5": "^2.3.0",
     "eslint-plugin-html": "^2.0.1",
-    "google-closure-compiler": "^20170626.0.0",
+    "google-closure-compiler": "^20170806.0.0",
```
azakus: No more building from HEAD!

Issue #4792, opened 2017-08-13
Should we consider to use !== rather then != 


AlphaWong:
https://github.com/Polymer/polymer/blob/ca1ce19682b0003359090491d7e81a58d26b9a4a/lib/elements/dom-if.html#L160

Should we consider the following condition that 
```js
// true
'abc' != ['abc']

//false
'abc' !== ['abc']
```
when we would like to have a comparison.

Best,
Alpha

kevinpschaaf:
The `if` property is boolean typed, and we accept truthy/falsy values.  So can you explicate why you think `'abc'` (truthy) and `['abc']` (truthy) should be considered a change for the `if` value and thus e.g. fire `dom-changed`?

AlphaWong:
Deat Kevin,

Thanks for your review. I originally idea is using strict comparison to limit developer expectation.
live demo https://jsfiddle.net/dnfwypoa/2/
```js
const showDom = function showDom(flag) {
  if (flag == true) console.log('Dom show.');
  else console.log('Dom not show.');
}

const showDomStrict = function showDomStrict(flag) {
  if (flag === true) console.log('Dom show');
  else console.log('Dom not show.');
}

showDom(1);
// output : Dom show.
showDomStrict(1);
// output : Dom not show
```
However, thx for @web-padawan in https://github.com/Polymer/polymer/pull/4802#issuecomment-323744111 mentions that it may change the <dom-if> behavior.

My standpoint is that some developer may put 0 / 1 to <dom-if> as they think 0 / 1 and true / false are same.

This change can force them to use Boolean rather then 0 / 1.

sorvell:
Yes, we cannot really consider this since it would be a breaking change, but I think the opinion is reasonable and there's a trade off here.

PR #4795, opened 2017-08-15
UPT: Use strict compare & remove an else statment

AlphaWong:
Dear community,
 
Plz review my code

Best,
Alpha

### Reference Issue
https://github.com/Polymer/polymer/issues/4792

### DIFF
--- a/lib/elements/dom-if.html
+++ b/lib/elements/dom-if.html
@@ -80,7 +80,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           observer: '__debounceRender'
         }
 
-      }
+      };
 
     }
 
--- a/lib/elements/dom-if.html
+++ b/lib/elements/dom-if.html
@@ -157,7 +157,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (!this.restamp && this.__instance) {
         this._showHideChildren();
       }
-      if (this.if != this._lastIf) {
+      if (this.if !== this._lastIf) {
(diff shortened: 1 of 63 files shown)


lib/elements/dom-if.html
```diff
@@ -175,12 +175,13 @@
           if (!template) {
             // Wait until childList changes and template should be there by then
             let observer = new MutationObserver(() => {
-              if (this.querySelector('template')) {
-                observer.disconnect();
-                this.__render();
-              } else {
+              // If template tag not found, it will throw
```
azakus: Why change this `if` statement?

AlphaWong: As an early throw will be done if the first condition has been hit.
I do not think it is necessary to create an else block.

Thanks for your review.

TimvdLippe: I would say to revert this change. It makes the code harder to read imo.

AlphaWong:
okay, I will revert that change.

Thanks

AlphaWong:
I just rebase to latest remote polymer and revert the change of the error throw.
Please review my code.

Thanks
Alpha

TimvdLippe:
I think your rebase went wrong as there a couple of extra commits in this PR. Could you remove those commits and force push?

AlphaWong:
I am not sure as I rebase to polymer/polymer rather than my personal branch.
That my commits should behind your commit (change semi quotes error level ).

I will change it tmr. 
Thanks a lot !!

AlphaWong:
PR continue:
https://github.com/Polymer/polymer/pull/4802

Sorry for my lack experience action.

PR #4802, opened 2017-08-21
UPT: Use strict compare

AlphaWong:
Dear community,

Sorry for previous PR mixes with a lot of useless commits.
According to your suggestions that I clear up those commits.

Thanks,
Alpha

### Previous PR:
https://github.com/Polymer/polymer/pull/4795

### Reference Issue
#4792


AlphaWong:
Good point.
my change will cause a behavior change.
may be it need other to review my change.

Thanks

PR #4821, opened 2017-08-31
Explicitly set display none on dom-* elements

### LINKED ISSUE #4773: <dom-if> interferes with justify-content: space-between
### Description

Using `<template is='dom-if'>` in source code causes `<dom-if>` element to be put in resulting HTML (regardless of whether the condition is true or false).
If the parent element has `justify-content: space-between; display: flex;` then `<dom-if>` participates in element count for justification sake and the result looks broken.

This can be fixed by styling `<dom-if>` with `display: none`. Since this is a regression from Polymer 1.9, maybe the library can apply this automatically? I can't think of a use case where one wouldn't want `display: none` on the `<dom-if>` 

#### Live Demo

http://jsbin.com/megukipana/1/edit?html,output

#### Steps to Reproduce
Example:
```html
    <template>
      <style>
        .parent {
          border: 1px solid red;
          justify-content: space-between;
          width: 300px;
          display: flex;
        }
        .child {
          border: 1px solid green;
          width: 100px;
        }
      </style>
      <div class='parent'>
        <div class='child'>Child1</div>        
        <template is='dom-if' if='1'>
          <div class='child'>Child2</div>
        </template>
      </div>
    </template>
```

#### Expected Results
Child1 is to the left, Child2 is to the right.

#### Actual Results
Space to the right if Child2 caused by empty `<dom-if>`

### Browsers Affected

- [x] Chrome
- [x] Firefox

### Versions

- Polymer: v2.0
- webcomponents: vX.X.X

TimvdLippe:
### Reference Issue
Fixes #4773

### DIFF
--- a/lib/elements/dom-bind.html
+++ b/lib/elements/dom-bind.html
@@ -67,6 +67,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
 
       connectedCallback() {
+        this.style.display = 'none';
         this.render();
       }
 
--- a/lib/elements/dom-if.html
+++ b/lib/elements/dom-if.html
@@ -128,6 +128,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     connectedCallback() {
       super.connectedCallback();
+      this.style.display = 'none';
       if (this.if) {
         this.__debounceRender();
       }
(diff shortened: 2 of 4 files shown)


TimvdLippe:
Sauce seems to be having issues. We can rerun the build tomorrow.

PR #4829, opened 2017-09-07
App-level support for passive gesture touch listeners

### LINKED ISSUE #3604: Add support for passive event listeners

### LINKED ISSUE #3787: Perf: Do not add document-level blocking touch listeners

### LINKED ISSUE #4667: gestures.html adds non-passive event listener to a scroll-blocking 'touchstart' event

azakus:
Apps can set `Polymer.passiveTouchGestures` to `true` to force all touch listeners for gestures to use passive event handlers.
This may improve scrolling performance, at the cost of being unable to control scrolling with `gestureEvent.preventDefault()`.
Instead, elements must use `Polymer.Gestures.setTouchAction(node, 'direction')`

This PR overrules #4724

Fixes #4667
Fixes #3787
Fixes #3604

### DIFF
--- a/lib/utils/gestures.html
+++ b/lib/utils/gestures.html
@@ -40,22 +40,34 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
   })();
 
+  /**
+   * @param {string} name Possible mouse event name
+   * @return {boolean} true if mouse event, false if not
+   */
+  function isMouseEvent(name) {
+    return MOUSE_EVENTS.indexOf(name) > -1;
+  }
+
   /* eslint no-empty: ["error", { "allowEmptyCatch": true }] */
   // check for passive event listeners
   let SUPPORTS_PASSIVE = false;
   (function() {
     try {
-      let opts = Object.defineProperty({}, 'passive', {get: function() {SUPPORTS_PASSIVE = true;}});
+      let opts = Object.defineProperty({}, 'passive', {get() {SUPPORTS_PASSIVE = true;}});
       window.addEventListener('test', null, opts);
       window.removeEventListener('test', null, opts);
     } catch(e) {}
   })();
 
+  // decide whether to use {passive: true} for gestures listening for touch events
+  let PASSIVE_TOUCH = Boolean(HAS_NATIVE_TA && SUPPORTS_PASSIVE && Polymer.passiveTouchGestures);
+
   // Check for touch-only devices
   let IS_TOUCH_ONLY = navigator.userAgent.match(/iP(?:[oa]d|hone)|Android/);
 
   let GestureRecognizer = function(){}; // eslint-disable-line no-unused-vars
-  GestureRecognizer.prototype.reset = function(){};
+  /** @type {function()} */
+  GestureRecognizer.prototype.reset;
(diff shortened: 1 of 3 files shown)


azakus:
Hmm, looks like closure linting is broken on travis now :(

lib/utils/settings.html
```diff
@@ -93,5 +93,15 @@
...
+
+  /**
+   * Globally settable property to make Polymer Gestures use passive TouchEvent listeners when recognizing gestures.
+   * When set to `true`, gestures made from touch will not be able to prevent scrolling, allowing for smoother
+   * scrolling performance.
+   * Defaults to `false` for backwards compatibility.
+   *
+   * @memberof Polymer
+   */
+  Polymer.passiveTouchGestures = Polymer.passiveTouchGestures || false;
```
sorvell: Should be a function for forward compatibility

lib/utils/gestures.html
```diff
@@ -40,22 +40,34 @@
...
     try {
-      let opts = Object.defineProperty({}, 'passive', {get: function() {SUPPORTS_PASSIVE = true;}});
+      let opts = Object.defineProperty({}, 'passive', {get() {SUPPORTS_PASSIVE = true;}});
       window.addEventListener('test', null, opts);
       window.removeEventListener('test', null, opts);
     } catch(e) {}
   })();
 
+  // decide whether to use {passive: true} for gestures listening for touch events
+  let PASSIVE_TOUCH = Boolean(HAS_NATIVE_TA && SUPPORTS_PASSIVE && Polymer.passiveTouchGestures);
```
sorvell: `passiveTouchGestures` should be a function for forward compat.

azakus:
This will ship as Polymer 2.1

matthew-dean:
This is labelled as being related to touch gestures, but will this also allow `this.listen` to support passive scroll event listeners? Since this closes #3604 it suggests this is the case, but is that true? This seems related to just toggling an app-wide default for gestures, and not supporting component-level listeners to individually set passive event listeners, correct?

matthew-dean:
Just to re-iterate, I think this incorrectly closes #3604 and #3604 should be re-opened.

PR #4839, opened 2017-09-13
Fix #4818

### LINKED ISSUE #4818: Array changes don't propagate into dom-if=false
### Description

When array property is bound to sub-element under `dom-if`, then changes in the array don't propagate to sub-element if `dom-if`condition is false. Then, when `dom-if` condition becomes true the binding becomes out of sync.

#### Live Demo

http://jsbin.com/pasuwapoxo/1/edit?html,console,output

#### Steps to Reproduce
Example:

1. See jsbin and repeatedly click on the red rectangle
2. The value propagates only when `dom-if` is true and the visual value is out of sync with the real value printed to the log.

### Browsers Affected

- [x] Chrome
Didn't check others.

### Versions

- Polymer: v2.0.2

sorvell:
WIP: needs tests and heavy review.

Adds boolean argument to `_flushProperties` which "forces" any changes to objects to propagate. This is then used by `dom-if` to ensure when `if` becomes true, data propagates as expected.

Fixes #4818

### DIFF
--- a/lib/elements/dom-if.html
+++ b/lib/elements/dom-if.html
@@ -238,7 +238,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           this.__instance._setPendingProperty(prop, this.__dataHost[prop]);
         }
         this.__invalidProps = null;
-        this.__instance._flushProperties();
+        this.__instance._flushProperties(true);
       }
     }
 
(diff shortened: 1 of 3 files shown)


MajorBreakfast:
@sorvell @TimvdLippe 
#5094 has some nice tests for this. Shall I provide a PR to this branch?
Am I allowed to use async functions in tests? (Makes them easier to read)

sorvell:
After consideration, we've determined that there's too much risk of breaking user code if we change the core functionality around how data is passed between elements.  Therefore we're not going to merge this PR.

Instead, we recommend using immutable data patterns to have more control over how data propagates between elements. We plan to provide some helpers around working with immutable data, see: https://github.com/Polymer/polymer/issues/5126.

PR #4840, opened 2017-09-15
Basic element

sorvell:
Adds `Polymer.BasicElement`. This is a minimal starting point for building an element that uses `PropertyAccessors` which requires a bit of boiler plate out of the box.

Provides a `properties` objects similar to `Polymer.Element` for configuring accessors with a few differences:
* only `type` is supported and may be `foo: String` or `foo: {type: String}`.
* properties do not automatically compose to subclasses, so users should do `Object.assign({...}, super.properties});`

This element has no opinion about how or when any dom should be stamped This should make it a good choice to use with libraries like [lit-html](https://github.com/polymerlabs/lit-html). Any side effects from setting properties should be handled in `_propertiesChanged`.


cdata:
Discussed in person w/ @sorvell . Summary of my feedback: on a design level I would expect `BasicElement` to be a suitable base class for ultra-fast custom elements. This leads me to consider `PropertyAccessors` to be overkill for `BasicElement`'s base implementation. In general, separation of `PropertiesChanged` and `PropertyAccessors` seems good, although a purist factoring might ask `PropertyAccessors` to be split into yet more mixins. For perf reasons, it seems fine to group the way it has been grouped.

justinfagnani:
@sorvell Do subclasses really need to manually compose properties blocks? If `properties` induces accessors on the prototype, then the prototype chain should take care of the rest, no?

justinfagnani:
I'd like to consider adding a more generic method than `_propertiesChanged` as the override point for things like a lit-html mixin. `renderCallback` is something that I've used for this, and SkateJS does to. IMO, it nicely corresponds to the native reaction methods. This way we could write a lit-html or other render mixin that works again any base class that provides a `renderCallback`.

lib/mixins/properties-changed.html
```diff
@@ -0,0 +1,410 @@
...
+       */
+      _initializeProperties() {
+        // Capture instance properties; these will be set into accessors
+        // during first flush. Don't set them here, since we want
+        // these to overwrite defaults/constructor assignments
+        for (let p in this.__dataHasAccessor) {
+          if (this.hasOwnProperty(p)) {
+            this.__dataInstanceProps = this.__dataInstanceProps || {};
+            this.__dataInstanceProps[p] = this[p];
+            delete this[p];
```
cdata: Is it necessary to delete each property when they are just going to be re-configured with accessors anyway?

lib/mixins/properties-changed.html
```diff
@@ -0,0 +1,410 @@
...
+   */
+   Polymer.PropertiesChanged = Polymer.dedupingMixin(superClass => {
+
+    return class PropertiesChanged extends superClass {
+
+      /**
+       * Creates property accessors for the given property names.
+       * @param {array} props Array of names of accessors.
+       */
+       static createProperties(props) {
```
cdata: Weird indentation

lib/mixins/properties-changed.html
```diff
@@ -0,0 +1,410 @@
...
+       * Users may override this function to implement behavior that is
+       * dependent on the element having its property data initialized, e.g.
+       * from defaults (initialized from `constructor`, `_initializeProperties`),
+       * `attributeChangedCallback`, or values propagated from host e.g. via
+       * bindings.  `super.ready()` must be called to ensure the data system
+       * becomes enabled.
+       *
+       * @public
+       */
+       ready() {
```
cdata: Weird indentation

lib/mixins/properties-changed.html
```diff
@@ -0,0 +1,410 @@
...
+
+      /**
+       * Returns a property name that corresponds to the given attribute.
+       * By default, converts dash to camel case, e.g. `foo-bar` to `fooBar`.
+       * @param {string} attribute Attribute to convert
+       * @return {string} Property name corresponding to the given attribute.
+       *
+       * @protected
+       */
+       _propertyForAttribute(attribute) {
```
cdata: Weird indentation

lib/mixins/properties-changed.html
```diff
@@ -0,0 +1,410 @@
...
+      }
+
+      /**
+       * Override point to provide a type to which to deserialize a value to
+       * a given property.
+       * @param {string} name Name of property
+       *
+       * @protected
+       */
+       _typeForProperty(name) {} //eslint-disable-line no-unused-vars
```
cdata: Weird indentation

cdata:
> renderCallback is something that I've used for this, and SkateJS does to. IMO, it nicely corresponds to the native reaction methods.

Isn't there a catch-22 with naming things like `renderCallback` _because_ they correspond with platform-native names? Something something once it catches on in author code land it can't be standardized?

Otherwise, no argument.

> If properties induces accessors on the prototype, then the prototype chain should take care of the rest, no?

I like that a subclass can be explicit about _how_ it inherits properties from its superclass. To me this seems less magical, and I as an author can be deliberate in configuring what properties my class gets.

OTOH, it seems like there is at least one scenario where the proposed implementation suffers compared to relying on prototype chain inheritance. Say I have an base class `x-widget` and two inheriting implementations. If I create instances of all three custom elements, and the two inheriting implementations use 100% of the properties of their parent, won't I be configuring the same set of accessors three times (instead of potentially just once for `x-widget`)?

arthurevans:
@justinfagnani How is `renderCallback` more generic than `_propertiesChanged`? 

It seems to me that `propertiesChanged` is a generic callback that describes exactly what happened—some properties changed. The idea of `renderCallback` injects a React-flavored opinion about how those properties manifest themselves. I know it's enshrined in the React world and other libraries that take their start from it, but `render` is a weird name for "create or update a DOM tree." `createOrUpdateDOMSubtreeThatThePlatformMayEventuallyRender`, maybe.

justinfagnani:
@arthurevans `renderCallback` is more generic because it doesn't say _why_ the subclass should do something, but _what_ the subclass should do. It maybe be that the subclass is being asked to render because properties changed, or because it's the initial render, or because some global state, or a timer fired, etc., that the element has declared a dependency on. `_propertiesChanged` is exposing that a specific thing happened, and the name doesn't offer guidance on what to do then.

This doesn't have anything to do with React, I don't think. I view it as an additional, optional, non-standard Custom Element reaction. Render is a concept in many frameworks and how people talk about rendering DOM.

Already, on top of BasicElement's `_propertiesChanged`, PolymerLitElement is going to define a `render()` callback that developers implement. What I'm proposing here could mean that that mixin works against any base class that calls `renderCallback` as opposed only BasicElement/PropertyAccessors that defined `_propertiesChanged`, unless we try to advocate `_propertiesChanged` as an ad-hoc community standard.

cdata:
> renderCallback is more generic because it doesn't say why the subclass should do something, but what the subclass should do

The original contention is that `renderCallback` is good because it is closer to the platform callbacks. Those callbacks don't describe what should happen; they describe what did happen (element was connected to the document, element's attributes changed). They are more akin to event handlers than hooks intended to implement specific behavior.

lib/elements/basic-element.html
```diff
@@ -0,0 +1,128 @@
...
+   * @polymer
+   * @memberof Polymer
+   * @constructor
+   * @implements {Polymer_PropertiesChanged}
+   * @extends HTMLElement
+   * @appliesMixin Polymer.PropertiesChanged
+   * @summary Custom element base class that provides minimal starting point
+   * using PropertiesChanged to create properties
+   */
+   class BasicElement extends Polymer.PropertiesChanged(HTMLElement) {
```
azakus: For closure, mixins have to be of the form:
```js
const basicElementBase = Polymer.PropertiesChanged(HTMLElement);

class BasicElement extends basicElementBase {
```

lib/elements/basic-element.html
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Implements standard custom elements getter to observes the attributes
+     * listed in `properties`.
```
azakus: `@return {!Array<string>}`

lib/elements/basic-element.html
```diff
@@ -0,0 +1,128 @@
...
+  'use strict';
+
+  /**
+   * Custom element base class that provides minimal starting point
+   * using PropertyAccessors to create properties.
+   *
+   * @customElement
+   * @polymer
+   * @memberof Polymer
+   * @constructor
```
azakus: These couple bits (`@constructor`, `@implements`, `@extends`) have to be moved down to `basicElementBase` as I described below.

lib/utils/async.html
```diff
@@ -73,24 +73,31 @@
       after(delay) {
         return  {
           run(fn) { return setTimeout(fn, delay); },
-          cancel: window.clearTimeout.bind(window)
+          cancel(handle) {
```
azakus: This file's changes seem like they could be handled in a separate PR

lib/mixins/property-accessors.html
```diff
@@ -95,13 +93,21 @@
   Polymer.PropertyAccessors = Polymer.dedupingMixin(superClass => {
 
     /**
+     * @constructor
+     * @extends {superClass}
+     * @implements {Polymer_PropertiesChanged}
+     * @unrestricted
```
azakus: I don't know if `@unrestricted` is needed here.

daKmoR:
This is an fantastic idea :) love it :)

I was just wondering if we could also have a simple way of setting default values?
e.g.
```
		static get properties() {
			return {
				header: {
					type: String,
					value: 'Init name'
				},
````

should be just some if right?
```
// in some init function? 
if (prop.value) {
  this._setProperty(prop.value);
}
````

but I can't seem to find the correct place to add it... if I could I would create a merge request...

**Comment by cdata** [comment]
@daKmoR If I understood @sorvell correctly, the user is intended to set default values in the constructor of your element. This is a departure from traditional Polymer style, but much closer to standard ES6 class style.

Also, regarding extra features like `observers`, this element base class is intended to be absolutely minimal. So, you can write mixins to support such features, but they probably aren't suitable for `BasicElement`'s core implementation.

**Comment by arthurevans** [comment]
@daKmoR Yep, if you set a property default in `constructor`, it's set before the element can be configured using attributes—by spec, `constructor` is called before any attribute values are set. I _think_ you're always going to get a `_propertiesChanged` callback, though, even if it's only setting the default values.

Setting defaults in `constructor` works in `Polymer.Element` elements, too. Since it is idiomatic ES6, we'll make that clearer in the docs going forward.

**Comment by daKmoR** [comment]
hmm that's what I did before... ok so maybe it's easier to copy (or reimplement) it if needed... most of the lit-elements so far seems to got that route... see https://github.com/PolymerLabs/lit-html/issues/88

thx for the info.

does anybody know how much performance is gained by using a "complex" properties-changed.html with caching instead of "plain" getters and setters?

File: lib/mixins/property-accessors.html [hunk 11]
```diff
@@ -269,7 +233,7 @@
         this.__serializing = true;
         value = (arguments.length < 3) ? this[property] : value;
         this._valueToNodeAttribute(this, value,
-          attribute || caseMap.camelToDashCase(property));
+          attribute || this._attributeNameForProperty(property));
```
azakus: This is not defined anywhere. Should it be?

closure.log
```diff
@@ -0,0 +1,76 @@
+gulp-google-closure-compiler: externs/closure-types.js:941: WARNING - property disconnectedCallback not defined on any supertype of Polymer_LegacyElementMixin
```
abdonrd: I guess the `.log` should be removed. :)

lib/mixins/properties-changed.html
```diff
@@ -0,0 +1,486 @@
...
+         * Returns an attribute name that corresponds to the given property.
+         * The attribute name is the lowercased property name. Override to
+         * customize this mapping.
+         * @param {string} property Property to convert
+         * @return {string} Attribute name corresponding to the given property.
+         *
+         * @protected
+         */
+        static attributeNameForProperty(property) {
+          return property.toLowerCase();
```
TimvdLippe: This is different to the previous mapping where it was camelcase to kebabcase right?

lib/mixins/properties-mixin.html
```diff
@@ -0,0 +1,212 @@
...
+       * `PropertiesChanged`.
+       */
+      connectedCallback() {
+        this._enableProperties();
+      }
+
+      /**
+       * Called when the element is removed from a document
+       */
+      disconnectedCallback() {}
```
TimvdLippe: Is this required to be here?

lib/utils/async.html
```diff
@@ -67,13 +64,15 @@
...
-       * @param {number} delay Time to wait before calling callbacks in ms
+       * @param {number=} delay Time to wait before calling callbacks in ms
        * @return {AsyncInterface} An async timeout interface
        */
       after(delay) {
-        return  {
+        return {
           run(fn) { return setTimeout(fn, delay); },
-          cancel: window.clearTimeout.bind(window)
+          cancel(handle) {
```
TimvdLippe: This recently changed on master to add `@function` to the doc. Have to recheck if the analyzer can still understand it.

test/unit/polymer.properties-mixin-with-property-accessors.html
```diff
@@ -0,0 +1,357 @@
...
+  <test-fixture id="sub-mixin-element-attr">
+    <template>
+      <sub-mixin-element prop="attr" prop2="attr" prop3="attr" mixin="5" camel-case="camelCase"></sub-mixin-element>
+    </template>
+  </test-fixture>
+
+  <script>
+  suite('class extends Polymer.PropertiesMixin', function() {
+
+    var el;
```
TimvdLippe: This testsuite still uses vars. Since you updated the vars to consts in the previous file, do it here as well :)

test/unit/polymer.properties-mixin.html
```diff
@@ -0,0 +1,396 @@
...
+  <test-fixture id="sub-mixin-element-attr">
+    <template>
+      <sub-mixin-element prop="attr" prop2="attr" prop3="attr" mixin="5" custom-observed-attr="custom"></sub-mixin-element>
+    </template>
+  </test-fixture>
+
+  <script>
+  suite('class extends Polymer.PropertiesMixin', function() {
+
+    var el;
```
TimvdLippe: Same here ;)

lib/mixins/properties-mixin.html
```diff
@@ -0,0 +1,212 @@
...
+
+      /**
+       * Finalize an element class. This includes ensuring property
+       * accessors exist on the element prototype. This method is called by
+       * `finalize` and finalizes the class constructor.
+       *
+       * @protected
+       */
+      static _finalizeClass() {
+        const props = ownProperties(this);
```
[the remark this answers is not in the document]
kevinpschaaf: Looked at this with Justin, the answer is because `observedAttributes` needs access to properties, and want to keep that as minimal as possible.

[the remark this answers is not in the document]
kevinpschaaf: But agree `ownProperties` function & memoizer could be removed and just have own properties added in `get _properties()` and use `Object.keys()` to get "own properties".

lib/mixins/properties-mixin.html
```diff
@@ -0,0 +1,212 @@
...
+       * Finalizes an element definition, including ensuring any super classes
+       * are also finalized. This includes ensuring property
+       * accessors exist on the element prototype. This method calls
+       * `_finalizeClass` to finalize each constructor in the prototype chain.
+       */
+       static finalize() {
+        if (!this.hasOwnProperty(JSCompiler_renameProperty('__finalized', this))) {
+          const superCtor = superForClass(this);
+          if (superCtor) {
+            superCtor.finalize();
```
kevinpschaaf: Justin idea: eliminate `superForClass` and instead:
```js
if (super.prototype instanceof PropertiesMixin) {
   Object.getPrototypeOf(this).finalize();
}
```
Or, can we not do this because closure doesn't support property access off of super?

lib/mixins/properties-mixin.html
sorvell: It's there as a minor convenience so users don't have to check before calling `super.disconnectedCallback`

test/unit/polymer.properties-mixin-with-property-accessors.html
sorvell: Yes!

test/unit/polymer.properties-mixin.html
sorvell: Yup, thx.

lib/mixins/properties-mixin.html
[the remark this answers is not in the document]
sorvell: We need to pass only "own properties" to `createProperties`. To do that with only `_properties`, we'd have to copy the top level properties into another object which is more code and slower. While we could change `createProperties` to only create properties for `ownPropertyNames` that seems like an odd choice.

Also, we have `ownProperties` to avoid hitting the `properties` getter more than once (to avoid running potentially $ user code). This getter can be called in one of two cases, (1) define is called and `_properties` is accessed via `observedAttributes` being inspected; (2) the class is "finalized" without being defined by virtue of being a superclass.

This is definitely biased for avoiding runtime work v. minimizing code size.

lib/mixins/properties-mixin.html
```diff
@@ -0,0 +1,212 @@
...
+
+      /**
+       * Overrides PropertiesChanged method to return type specified in the
+       * static `properties` object for the given property.
+       * @param {string} name Name of property
+       * @return {*} Type to which to deserialize attribute
+       *
+       * @protected
+       */
+      static typeForProperty(name) {
```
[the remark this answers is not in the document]
sorvell: Yes, this `info` argument here is a "normalized" property: https://github.com/Polymer/polymer/blob/basic-element/lib/mixins/element-mixin.html#L203

lib/mixins/properties-mixin.html
sorvell: That's right and there's another spot we'd need the same check. Will rename to `superPropertiesClass`

lib/mixins/properties-mixin.html
```diff
@@ -0,0 +1,208 @@
...
+        super._initializeProperties();
+      }
+
+      /**
+       * Called when the element is added to a document.
+       * Calls `_enableProperties` to turn on property system from
+       * `PropertiesChanged`.
+       */
+      connectedCallback() {
+        this._enableProperties();
```
justinfagnani: This should check for and call `super.connectedCallback`

lib/mixins/properties-mixin.html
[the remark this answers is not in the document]
justinfagnani: I don't think it's _that_ odd to define the `_properties` object that this generates and `createProperties` accepts as have a distinction between own and inherited properties :) The behavior would be that the _normalized_ properties object has a prototype chain that mirrors the prototype chain of the constructor, just like the proto chain of the constructors mirrors the proto chain of their `.prototype` properties. Makes sense to me, but I admit it might not for everyone.

Arguing against myself out-loud: the difference between `Object.keys(this._properties)` and `for (let p in this._properties) {}` might be way too subtle. It's also admittedly a very minor benefit, because it's per-class...

...but it is there: slightly less code, properties are only looped through once to build `__properties`, less duplication of key/values between `__ownProperties` and `__properties`, one less static property and shape change.

Anyway, super, super small nit. I'm hyper-zoomed in here, apologies...

lib/mixins/properties-mixin.html
justinfagnani: The problem with an empty stub is like with my above comment: if the mixin is applied to a superclass with a `disconnectedCallback`, the superclasses method won't be called.

Also, if we do this for `disconnectedCallback` we might want to for `adoptedCallback`?

lib/mixins/properties-mixin.html
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * Mixes `moreProps` into `props` but upgrades shorthand type
+     * syntax to { type: Type}.
+     *
+     * @param {Object} props Properties to normalize
+     * @return {Object} Copy of input `props` with normalized properties that
+     * are in the form {type: Type}
+     * @private
+     */
+    function normalizeProperties(props) {
```
justinfagnani: This can move out of the mixin function to avoid creating a new function object per mixin application.

lib/mixins/properties-mixin.html
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * Returns the super class constructor for the given class, if it is an
+     * instance of the PropertiesMixin.
+     *
+     * @param {PropertiesMixinConstructor} constructor PropertiesMixin constructor
+     * @return {PropertiesMixinConstructor} Super class constructor
+     */
+    function superPropertiesClass(constructor) {
+      const superCtor = Object.getPrototypeOf(constructor);
+      if (superCtor.prototype instanceof PropertiesMixin) {
```
justinfagnani: This code is fine, just want to point out a subtle thing with mixins, that might warrant a comment, for any reader not familiar with class expressions or this mixin pattern: `PropertiesMixin` here is a new variable per call to the mixin function. If this mixin could be applied multiple times to a proto chain, the check could unexpectedly fail, but it's fine here because of the deduping wrapper.

lib/mixins/properties-mixin.html
```diff
@@ -0,0 +1,208 @@
...
+
+    /**
+     * Returns a memoized version of the `properties` object for the
+     * given class. Properties not in object format are converted to at
+     * least {type}.
+     *
+     * @param {PropertiesMixinConstructor} constructor PropertiesMixin constructor
+     * @return {Object} Memoized properties object
+     */
+    function ownProperties(constructor) {
```
justinfagnani: This can also move out of the mixin function. There maybe other cases in the other mixins.

lib/mixins/properties-mixin.html
kevinpschaaf: Good point, will do.

lib/mixins/properties-mixin.html
kevinpschaaf: Makes sense; need to check with @azakus, it might need to stay in the closure so jscompiler has access to `PropertiesMixinConstructor` assuming it's supposed to be `PropertiesMixin` (since that symbol no longer exists) but not entirely sure if it needs to be some outside type)... Will confirum with Dan on Mon.

lib/mixins/properties-mixin.html
kevinpschaaf: Yep, will do.

lib/mixins/properties-mixin.html
justinfagnani: oh yeah... I forgot the comment might need to be in scope :(

lib/mixins/properties-mixin.html
kevinpschaaf: Confirming whether closure supports the `if (super.connectedCallback)` incantation and will add if so (if not, that will be unfortunate but will document that this needs to be the first mixin applied that implements `connectedCallback`).

lib/mixins/properties-mixin.html
kevinpschaaf: I'm inclined to remove `disconnectedCallback` from this mixin altogether as YAGNI (this mixin is still likely to be used in a higher level base element that handles rendering where that author can make a choice of whether to add a superable stub for convenience).

kevinpschaaf: Eh, I can go both ways I guess.  This used to be implemented in `ElementMixin` and just moved up the layering to here (assuming the same reason we had it for `Polymer.Element` will want to be carried forth to future concrete base classes as well).  Having this pattern where users are encouraged to always `super.disconnectedCallback` makes it safer for existing mixins to add a `disconnectedCallback` override at a later point.

Adopted callback is so obscure, I think it would only a waste of bytes. :P

PR #4842, opened 2017-09-17
Fix inline code highlight in readme doc

tschwartz:


TimvdLippe:
Hm, not sure if we can accept PRs to 3.0-preview right now, as it is auto-generated and might be overwritten later.

PR #4852, opened 2017-09-21
Move @externs before @license because Closure likes that.

aomarks:


aomarks:
Note to self: closure-types.js is generated by https://github.com/Polymer/gen-closure-declarations. Fix it there too.

PR #4853, opened 2017-09-22
Change findTemplateNode to work on array of indices

TimvdLippe:
Originally part of the pre-building of effects, but that PR got too big to review. As such, I broke out this particular change which is completely stand-alone.

Previously, nested objects were used to represent how to retrieve a node in a template. It used a recursive pattern to find a parent. This solution only stores 1 integer per node and is iterative. This is especially a benefit when serializing this format, as it takes only 1 character to represent a node lookup instead of >20.

### DIFF
--- a/lib/mixins/template-stamp.html
+++ b/lib/mixins/template-stamp.html
@@ -44,20 +44,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   }
 
   function findTemplateNode(root, nodeInfo) {
-    // recursively ascend tree until we hit root
-    let parent = nodeInfo.parentInfo && findTemplateNode(root, nodeInfo.parentInfo);
-    // unwind the stack, returning the indexed node at each level
-    if (parent) {
-      // note: marginally faster than indexing via childNodes
-      // (http://jsperf.com/childnodes-lookup)
-      for (let n=parent.firstChild, i=0; n; n=n.nextSibling) {
-        if (nodeInfo.parentIndex === i++) {
-          return n;
+    let parent = root;
+    for (let i = 0, l = nodeInfo.length; i < l; i++) {
+      for (let n = parent.firstChild, j = 0; n; n = n.nextSibling) {
+        if (nodeInfo[i] === j++) {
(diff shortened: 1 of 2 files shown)


lib/mixins/template-stamp.html
```diff
@@ -154,8 +150,7 @@
        *         templateInfo: {object}, // nested template metadata
        *         // Metadata to allow efficient retrieval of instanced node
        *         // corresponding to this metadata
-       *         parentInfo: {number},   // reference to parent nodeInfo>
-       *         parentIndex: {number},  // index in parent's `childNodes` collection
+       *         parentInfo: {Array},   // indices to traverse to this node from the template root
```
sorvell: Unfortunately, this is a breaking change. I don't think we can accept this until Polymer 3.0.

sorvell:
This is a good change, but it is (minorly) breaking. We'll leave it open for consideration for Polymer 3.x.

lib/mixins/template-stamp.html
kevinpschaaf: It may be possible to factor `findTemplateNode` such that it is still recursive, but able to operate either on a `parentInfo` tree (for backward compatibility) or a `parentIndex` array (for efficient serialization).  We can consider this if we move forward with https://github.com/Polymer/polymer/pull/4782 (or a non-breaking refactoring of it).  In that way, the runtime code would still generate backward-compatible `parentInfo` trees (i.e. https://github.com/Polymer/polymer/pull/4853/files#diff-84d426316fb9e8502af8f5eed8675b40L286 would be reverted), and the build-time tool would transform that into an array of indices.

TimvdLippe: It could certainly work by being able to parse both recursive notation as well as the array notation, but I don't think it would be good to land that before Polymer 3. Since the build-time tool is likely not ready before 3 is ready, landing it right now increases the Polymer code size and adds more conditions.

TimvdLippe:
We have not incorporated this in 3.x and I think we have no plans of pursueing this any further. Therefore I am closing this PR.

PR #4870, opened 2017-10-06
Declare Polymer.Templatizer directly, for Closure.

aomarks:
Closure cannot follow the indirect `@polymerBehavior` annotation, so define Polymer.Templatizer directly.


lib/legacy/templatizer-behavior.html
```diff
@@ -87,9 +87,8 @@
      * between versions 1.x and 2.x.
      *
      * @polymerBehavior
-     * @memberof Polymer
```
rictic: I don't think we need to get rid of this `@memberof`

rictic: nm, this is better.

PR #4875, opened 2017-10-09
fix(dom-api-shady): getOwnerRoot issue with window

AliMd:
fix _ownerShadyRootForNode process _ownerShadyRootForNode for window element!

When we use some jquery plugin we have error on `document.documentElement.contains(node)` becouse node is window!
I don't know when and why node is window but i have always this error when tap on some jquery modules! 
I will try to find the exactly what happened but bytheway `getOwnerRoot` must skip process on window


TimvdLippe:
I am closing this pull request for now, as it appears to be stale. Please reopen once you have a reproduction case so that we can review and understand the problem 😄

Issue #4860, opened 2017-10-03
:dir support for 1.x and native ShadowDOM 2.x


azakus:
ShadyCSS supports `:dir()` to provide different styling for RTL and LTR layouts.
This should be backported to 1.x, and also polyfilled to be supported in Native ShadowDOM for 2.x.

The approach for 2.x is somewhat simple: With a mixin, we can detect the direction of the page and set an attribute on all elements implementing the mixin. Then, we can transform `:dir()` into a simple attribute selector.

azakus:
Fixed with #4880 and #4881

PR #4880, opened 2017-10-11
:dir() support

azakus:
### Reference Issue
Polymer 2.x fix for #4860

### Goal
Allows for native and polyfilled ShadowDOM elements to use `:dir` CSS selector to provide text directionality based layouts.

### Details
- Modifies `Polymer.ElementMixin` to provide override points for stylesheet modification.
- Elements can opt out of the global page text direction by setting the `dir` attribute directly in `ready()`.
- Automatic left-to-right or right-to-left styling is sync'd with the `<html>` element
- Applications must set `<html dir="ltr">` or `<html dir="rtl">` to sync direction
- Opting out of the global direction styling is permanent


lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+
+    let observed = false;
+
+    /**
+     * @param {!Polymer_DirMixin} instance Instance to set RTL status on
+     */
+    function setRTL(instance) {
+      if (!instance.__origRTLStatus) {
+        const el = /** @type {!HTMLElement} */(instance);
+        if (document.documentElement.getAttribute('dir') === 'rtl') {
```
sorvell: Let's capture this state so we can make this application function faster.

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+          el.removeAttribute('dir');
+        }
+      }
+    }
+
+    /**
+     * Element class mixin that allows elements to use the `:dir` CSS Selector to have
+     * text direction specific styling.
+     *
+     * With this mixin, any stylesheet provided in the template will be transform `:dir` into
```
sorvell: will transform

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+     * `:host([dir])` and sync direction with the page via the element's `dir` attribute.
+     *
+     * Elements can opt out of the global page text direction by setting the `dir` attribute
+     * directly in `ready()`.
+     *
+     * Caveats:
+     * - Automatic left-to-right or right-to-left styling is sync'd with the `<html>` element
```
sorvell: `<html>` element *only*, setting the dir attribute on an element further up the DOM tree will have no effect.

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+     * - Applications must set `<html dir="ltr">` or `<html dir="rtl">` to sync direction
```
sorvell: Let's put this as the first caveat and then say that changing `dir` on html is supported.

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Element class mixin that allows elements to use the `:dir` CSS Selector to have
+     * text direction specific styling.
+     *
+     * With this mixin, any stylesheet provided in the template will be transform `:dir` into
+     * `:host([dir])` and sync direction with the page via the element's `dir` attribute.
+     *
+     * Elements can opt out of the global page text direction by setting the `dir` attribute
+     * directly in `ready()`.
```
sorvell: or in HTML?

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+          DIR_INSTANCES.forEach(i => setRTL(i));
+        }).observe(document.documentElement, {attributes: true, attributeFilter: ['dir']});
+      }
+
+      /**
+       * @constructor
+       * @extends {base}
+       * @implements {Polymer_ElementMixin}
+       */
+      const elementBase = Polymer.ElementMixin(base);
```
sorvell: What about removing this dependency and requiring users to at least also use PropertyAccessors?

lib/mixins/element-mixin.html
```diff
@@ -577,13 +547,8 @@
...
         // note: finalize template when we have access to `localName` to
         // avoid dependence on `is` for polyfilling styling.
-        if (this._template && !this._template.__polymerFinalized) {
-          this._template.__polymerFinalized = true;
-          const baseURI =
-            importPath ? Polymer.ResolveUrl.resolveUrl(importPath) : '';
-          finalizeTemplate(/** @type {!PolymerElement} */(this.__proto__), this._template, baseURI,
-            /**@type {!HTMLElement}*/(this).localName);
-        }
+        this.constructor._finalizeTemplate(/** @type {!PolymerElement} */(this.__proto__), this._template,
```
sorvell: Since this is a static now, let's not pass in the prototype.

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+       * @mixinClass
+       * @implements {Polymer_DirMixin}
+       */
+      class Dir extends elementBase {
+
+        /**
+         * @override
+         * @suppress {missingProperties} Interfaces in closure do not inherit statics, but classes do
+         */
+        static _processStyleText(is, template, baseURI) {
```
sorvell: Let's only enable the runtime element support if we find dir in the css.

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+        constructor() {
+          super();
+          /** @type {boolean} */
+          this.__origRTLStatus = false;
+        }
+
+        /**
+         * @suppress {invalidCasts} Closure doesn't understand that `this` is an HTMLElement
+         */
+        ready() {
```
sorvell: Let's make these runtime features opt-in via a flag.

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,130 @@
...
+     * - Automatic left-to-right or right-to-left styling is sync'd with the `<html>` element
+     * - Applications must set `<html dir="ltr">` or `<html dir="rtl">` to sync direction
+     * - Opting out of the global direction styling is permanent
+     *
+     * @mixinFunction
+     * @polymer
+     * @appliesMixin Polymer.ElementMixin
+     * @memberof Polymer
+     */
+    Polymer.DirMixin = Polymer.dedupingMixin((base) => {
```
sorvell: Let's mix this into LegacyElement so we have hybrid support since 1.x has dir support now.

lib/mixins/dir-mixin.html
azakus: Done.

lib/mixins/dir-mixin.html
azakus: Not sure if this is still needed if we use the presence of `:dir` to activate.

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,171 @@
...
+    const EL_DIR = /([\s\w#\.\[\]\*]*):dir\((ltr|rtl)\)/g;
+    const EL_DIR_REPLACMENT = ':host([dir="$2"]) $1';
+
+    /**
+     * @type {!Array<!Polymer_DirMixin>}
+     */
+    const DIR_INSTANCES = [];
+
+    /** @type {MutationObserver} */
+    let OBSERVER = null;
```
sorvell: Let's not all caps for variables?

lib/mixins/dir-mixin.html
sorvell: Seems ok. They are different and an element could have either or both.

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,171 @@
...
+         */
+        ready() {
+          super.ready();
+          this.__autoDirOptOut = /** @type {!HTMLElement} */(this).hasAttribute('dir');
+        }
+
+        /** @suppress {missingProperties} Closure's understanding of prototypes is bad */
+        connectedCallback() {
+          if (elementBase.prototype.connectedCallback) {
+            /** @type {!Polymer_ElementMixin} */(elementBase).prototype.connectedCallback.call(this);
```
sorvell: Why not `super.connectedCallback()` (and same for disconnected...)

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,171 @@
...
+          }
+          if (this.constructor.__activateDir) {
+            takeRecords();
```
sorvell: Don't think this will fly (same in disconnected). We cannot update *all* elements as takeRecords does when one is connected

lib/mixins/dir-mixin.html
```diff
@@ -0,0 +1,171 @@
...
+@license
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<link rel="import" href="../utils/style-gather.html">
+<link rel="import" href="../utils/resolve-url.html">
```
sorvell: Doesn't seem like the imports above are used, remove?

lib/mixins/dir-mixin.html
azakus: PropertyAccessors doesn't have connectedCallback or disconnectedCallback

lib/mixins/dir-mixin.html
azakus: Ok, just means I'll have to pull document direction on connected.

lib/mixins/dir-mixin.html
azakus: sounds good

lib/mixins/dir-mixin.html
sorvell: Specifically if you find it, call `super.` rather than `.call`.

azakus: oh sure

PR #4891, opened 2017-10-19
[1.x] Adds support for `shady-unscoped` attribute to style elements.

### LINKED ISSUE #4856: Need efficient way to do reset styles across large component sets
Polymer needs a cheap way of applying shared styles in shadow and shady dom contexts.

Ideally, this would apply to hybrid elements and therefore work in 1.x and 2.x as its an important piece of the migration process.

The specific use case we are trying to address is shared styles that are used across all elements (e.g. a reset-style.css file). Since /deep/ is being deprecated, the current solution is to use a <style include> in every single component.

This works great in shadow dom, but is inefficient in shady dom. Since the current shady dom implementation would create a lot of duplicate code.

From discussing with @sorvell it seems like we can add a solution where a <style include> gets appended to the root without going through the style scoping. That seems like a solution that would work well, especially if it can continue to work with css variables (although this is not a blocker for this feature).

(aside: not a fan of the amount of boilerplate we're adding, but that's a separate discussion).

sorvell:
Add support for styles with a `shady-unscoped` attribute
Since `/deep/` has been removed from Shady DOM v0, users must share styles between elements via style modules (e.g. `<style include="shared-style">`).

Under ShadyDOM, these styles are scoped and duplicated between elements. If these styles are intended to replace global styling done with `html /deep/ ...` this is inefficient since the styles are needlessly copied to each element style.

Therefore, this adds support for a `shady-unscoped` attribute. If this attribute is placed on a style and native Shadow DOM is not in use, then the style is copied once to the document and are left unscoped.

Please note, css custom properties are not supported in `shady-unscoped` styles.

Fixes #4856

### DIFF
--- a/test/unit/shady-unscoped-style-import-css.html
+++ b/test/unit/shady-unscoped-style-import-css.html
@@ -1,0 +1,3 @@
+.import {
+  border: 2px solid yellow;
+}
--- a/test/unit/shady-unscoped-style-import-css.html
+++ b/test/unit/shady-unscoped-style-import-css.html
@@ -1,3 +1,0 @@
-.import {
-  border: 2px solid yellow;
-}
--- a/test/unit/shady-unscoped-style-import.html
+++ b/test/unit/shady-unscoped-style-import.html
@@ -1,5 +1,14 @@
+<!--
+@license
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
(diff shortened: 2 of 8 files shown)


test/unit/shady-unscoped-style-import.html
```diff
@@ -0,0 +1,26 @@
+<dom-module id="global-shared1">
```
azakus: license header

test/unit/shady-unscoped-style.html
```diff
@@ -0,0 +1,121 @@
+<!doctype html>
```
azakus: license header

test/unit/shady-unscoped-style-import-css.html
```diff
@@ -0,0 +1,3 @@
+.import {
```
azakus: license header

PR #4892, opened 2017-10-19
Add support for `shady-unscoped` attribute

sorvell:
For hybrid mode compatibility, adds similar support to https://github.com/Polymer/polymer/pull/4891.

This PR also prevents multiple styles from being collapsed into a single style in the element template. This will primarily affect style sharing and is being done to allow browsers to recognize shared styles as similar so that they can perform optimizations.

Note, currently the `ShadyCSS` polyfill always collapses multiple styles into a single element, but this may be changed in a future version of the polyfill. This means that users of either the @apply shim or ShadyCSS scoping shim will still have only a single style element in the element template.

Related #4856


test/unit/multi-style.html
```diff
@@ -0,0 +1,115 @@
+<!doctype html>
+<html>
```
azakus: License header

test/unit/shady-unscoped-style-import-css.html
```diff
@@ -0,0 +1,3 @@
+.import {
```
azakus: license header

test/unit/shady-unscoped-style-import.html
```diff
@@ -0,0 +1,26 @@
+<dom-module id="global-shared1">
```
azakus: license header

test/unit/shady-unscoped-style.html
```diff
@@ -0,0 +1,121 @@
+<!doctype html>
+<html>
```
azakus: license header

azakus:
Depends on ShadyCSS support for `shady-unscoped` attribute on `<style>` elements.

azakus:
webcomponents/shadycss#141

azakus:
webcomponents 1.0.19 has `shady-unscoped` support

PR #4902, opened 2017-10-24
Polymer 1: Allow arbitrary whitespace in CSS imports 

freshp86:
This is a port of https://github.com/Polymer/polymer/pull/4780 for Polymer 1.

### DIFF
--- a/src/lib/style-util.html
+++ b/src/lib/style-util.html
@@ -165,7 +165,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       },
 
       cssFromModules: function(moduleIds, warnIfNotFound) {
-        var modules = moduleIds.trim().split(' ');
+        var modules = moduleIds.trim().split(/\s+/);
         var cssText = '';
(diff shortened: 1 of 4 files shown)


freshp86:
/cc @sorvell @kevinpschaaf 

It would be nice if this fix can be ported to Polymer 1, so that we (in Chromium) can avoid errors like [this one](https://chromium-review.googlesource.com/c/chromium/src/+/734861) as well as having to add multiple comments in the code, like [these](https://cs.chromium.org/search/?q=polymer/pull/3668&type=cs).

freshp86:
Ping. Can someone take a look?

freshp86:
Thanks for merging. Can we spawn a release for v1.11.1 so that this fix can eventually make it in Chromium?

freshp86:
@azakus Is this included in a released version of Polymer 1 yet? If not, are there any plans to make one? Otherwise this fix can't be leveraged in our Chromium code.

azakus:
No sorry, its been too long since a release was cut. I'll try to make one today!

PR #4913, opened 2017-11-01
Clarify API docs for PropertyAccessors mixin

### LINKED ISSUE #4906: PropertyAccessors API doc needs clarification
Doc says:

For basic usage of this mixin, simply declare attributes to observe via the standard static get observedAttributes(), implement _propertiesChanged on the class, and then call MyClass.createPropertiesForAttributes() once on the class to generate property accessors for each observed attribute prior to instancing. Last, call this._flushProperties() once to enable the accessors.

Questions:
1) What does "prior to instancing" mean? Should this be done once, at a class level, before creating an instance of a component? Or once per instance? 
2) _flushProperties--in the constructor? Or does this need to be called after the constructor?
3) API doc for _flushProperties says "Generally, this method should not be called in user code," but the comment at the top of the file tells us to do just that. Maybe, "Generally this method should not be called in user code, _unless you are using the `PropertyAccessor` mixin directly."??

The following code appears to work for me, but I arrived at it by a process of guesswork and experimentation, and it may not be optimal.

```
  class CustomElement extends Polymer.PropertyAccessors(HTMLElement) {
    
    static get observedAttributes() { return ['one', 'two']; }
    
    constructor() {
      super();
      CustomElement.createPropertiesForAttributes();
    }
    
    connectedCallback() {
      this._flushProperties();
    }
    _propertiesChanged() {
      this.textContent = this.one + this.two;
    }
  }

  // Register the new element with the browser
  customElements.define('custom-element', CustomElement);

```

/cc @sorvell @kevinpschaaf

arthurevans:
Fixes #4906.


TimvdLippe:
Ping @arthurevans to either process above feedback or merge :smile:

PR #4919, opened 2017-11-02
instance.$.foo should only give Elements

rictic:
You need to be at least an `Element` to have an id attribute, and you need to have an id attribute to be in the `$` map.


azakus:
Concept is LGTM, but that file is generated, so the change to make is probably in `lib/mixins-element-mixin.html`

PR #4924, opened 2017-11-07
Fix the manually implemented "flatMap" ;-)

### LINKED ISSUE #4923: Uncaught (in promise) TypeError: undefined is not a function (legacy-element-mixin.html:907)
### Description
[obj] was wrongly transformed to obj and caused error on line 907

### Versions

- Polymer: v2.2.0
- webcomponents: v1.0.17

dkokic:
[obj] was wrongly transformed to obj and caused error on line 907

### Reference Issue

Fixes #4923

### DIFF
--- a/lib/legacy/legacy-element-mixin.html
+++ b/lib/legacy/legacy-element-mixin.html
@@ -897,7 +897,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        */
       _logger(level, args) {
         // accept ['foo', 'bar'] and [['foo', 'bar']]
-        if (Array.isArray(args) && args.length === 1) {
+        if (Array.isArray(args) && args.length === 1 && Array.isArray(args[0])) {
           args = args[0];
         }
         switch(level) {


TimvdLippe:
Ping @dkokic to add a regression test for this pull request.

dkokic:
I would like to add the required regression test, but would need some guidance where to place the test file for legacy-element-mixin.html file.

The error (in my case) is being produced within iron-ajax error handler (iron-ajax.html:526):
```
if (this.verbose) {
  Polymer.Base._error(error);
}
```
This invokes the method in legacy-element-mixin.html:934
```
_error(...args) {
  this._logger('error', args);
}
```
which packs all the arguments into an array *args* and invokes the _logger method:
```
/**
 * Facades `console.log`/`warn`/`error` as override point.
 *
 * @param {string} level One of 'log', 'warn', 'error'
 * @param {Array} args Array of strings or objects to log
 */
 _logger(level, args) {
  // accept ['foo', 'bar'] and [['foo', 'bar']]
  if (Array.isArray(args) && args.length === 1) {
    args = args[0];
  }
  switch(level) {
    case 'log':
    case 'warn':
    case 'error':
      console[level](...args);
  }
}
```

The problem is clearly in the method implementation as it fails when args is an array with the single element.

TimvdLippe:
@dkokic https://github.com/Polymer/polymer/blob/master/test/unit/logging.html is probably the best spot.

Issue #4603, opened 2017-05-16
importHref() could use better docs


justinfagnani:
See https://www.polymer-project.org/2.0/docs/api/#function-Polymer.importHref

Current docs:

>Polymer.importHref(href: string, onload: Function=, onerror: Function=, optAsync: boolean=): HTMLLinkElement
>
> Convenience method for importing an HTML document imperatively.
>
>This method creates a new <link rel="import"> element with the provided URL and appends it to the document to start loading. In the onload callback, the import property of the link element will contain the imported document contents.

The signature of the callbacks isn't specified, so the phrase "In the onload callback, the import property of the link element" doesn't make a lot of sense. What link element?

Also, the docs don't specify what happens if I load a URL twice - does it always create a new link element (and thus always fire a new load or error event), or does it reuse existing link elements? ie, should a caller remember if it already imported a URL?

JamesCoyle:
I'd like to know the behavior on loading an existing resource too. Will it load the resource again and duplicate the link element, will it realize it exists and just run the callback function, or will it detect it exists and not run the callback function? 

A lot of the polymer 2.0 still seems to be somewhat lacking in detail and a lot of things have been bunched into the Upgrade Guide rather than having separate sections like they did in the 1.0 docs. It seems like it assumes you are already familiar with Polymer 1.0 which not everyone is.

hariprasadiit:
@paulcuth thanks for info

`<link rel="import" href="/bower_components/polymer/lib/utils/import-href.html">` couldn't find this anywhere. Polymer docs should mention about this.

xelra:
I've been trying to make sense of the behavior of `ImportHref()` for more than a day. The documentation is so sparse and I simply don't understand what the exact conditions are for `onerror` to be called.

I've written a small element that checks auth and is supposed to lazy-load elements using `ImportHref()`. The code works, but every time `authenticated` changes, it triggers the console log message from `onerror`. And it does that despite `importHref` clearly importing successfully.

Does anyone know why?

```
<!-- my-check.html -->

<link rel="import" href="../bower_components/polymer/polymer-element.html">
<link rel="import" href="../bower_components/polymer/lib/utils/import-href.html">

<link rel="lazy-import" href="my-app.html">
<link rel="lazy-import" href="my-login.html">

<dom-module id="my-check">
	<template>
		<style>
			:host {
				display: block;
				height: 100%;
			}
		</style>

		<dom-if if="[[!authenticated]]" restamp>
			<template>
[...]

PR #4928, opened 2017-11-10
Add TypeScript declarations.

### LINKED ISSUE #4025: polymer 2.0 typings
Please update the typings for polymer and webcomponents first. It opens up the door for easy polymer 2.0 typescript experimenting.

```
npm i -D @types/polymer
```

image

image

image

PS `iron-component-page` is completely broken but the `/demo` code that only needs polymer 2 is working.
https://github.com/gertcuykens/hello-world

`web-component-tester` is also working except how do you do `getContentChildren();` in polymer 2?

```
      test('distributed children', function() {
        var els = myEl.getContentChildren();
        assert.equal(els.length, 0, 'no distributed nodes');
      });
```

aomarks:
Generated by https://github.com/Polymer/gen-typescript-declarations. Regenerate with `gulp generate-typescript`. Types are output to a `types/` directory.

Fixes #4025


gulpfile.js
```diff
@@ -267,8 +267,18 @@ gulp.task('generate-externs', ['clean'], () => {
   });
 });
 
+gulp.task('generate-typescript', () => {
+  let genTs = require('@polymer/gen-typescript-declarations').generateDeclarations;
+  del.sync(['**/*.d.ts']);
```
azakus: Just use promise api with `del`

justinfagnani:
I think siblings makes sense, especially with the move to modules coming up.

/cc @FredKSchott and @usergenic because we'll need modulizer to re-generate the types.

gulpfile.js
FredKSchott: (async/await ftw)

aomarks:
@justinfagnani I'm fine with that, but note that once we're on modules, we'll also be on NPM, so we can and should add types as part of `prepack`.

I think the objection to sibling files from @azakus is that it clutters up the source repo. Siblings for 2.0 is simpler from my perspective, but I understand not wanting to clutter the repo.

Can you elaborate on how this will work with modulizer? I would have thought we'd just blow away the types before we run modulizer, and re-run the generator afterwards, once it has support for modules?

TimvdLippe:
I have updated the description to reference #4025 which this PR fixes as well.

gulpfile.js
aomarks: Done.

aomarks:
Hi all. This is ready for another look. Highly recommend reviewing the 3 commits separately.

- Types are now output to a `types/` directory instead of as sibling files. We made this decision to avoid cluttering the repo. For Polymer 3 on NPM, we will instead output as siblings as part of prepublish, instead of committing to the repo at all.

- Integrates all of the fixes listed at https://github.com/Polymer/gen-typescript-declarations/issues/23. There are still a few missing APIs, but I don't think we should block merging on that. All of the missing APIs are things that Analyzer doesn't know about at all, so they have also never been in our documentation.

- Added a commit with some minor JSDoc tweaks.

aomarks:
Also, this now includes the `extra-types.d.ts` file, which are TypeScript versions of the hand-written types we needed from the Closure externs.

lib/utils/mixin.html
```diff
@@ -38,6 +38,7 @@
    * @memberof Polymer
    * @template T
    * @param {T} mixin ES6 class expression mixin to wrap
+   * @return {T}
```
azakus: From what I remember, the magic voodoo that Closure Compiler uses for mixins doesn't like a return type here. Can you double check that `gulp lint-closure` doesn't produce any warnings?

aomarks: > Generally LGTM, minus that nit and need to double-check closure with return type on deduping mixin.

Added some items to the CHANGELOG. Where should I mention the Node version in the README? It doesn't currently mention the gulp script. (async/await was actually 7.6 btw).

aomarks: > From what I remember, the magic voodoo that Closure Compiler uses for mixins doesn't like a return type here. Can you double check that gulp lint-closure doesn't produce any warnings?

I ran `gulp lint`, and it produces a lot of warnings. Nothing about `dedupingMixin` that I can see, though. Here's the log, does it look normal?: https://gist.github.com/aomarks/3adb586df000a36e682cc4f87137606b

types/lib/legacy/polymer.dom.d.ts
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Legacy DOM and Event manipulation API wrapper factory used to abstract
+   * differences between native Shadow DOM and "Shady DOM" when polyfilling on
+   * older browsers.
+   *
+   * Note that in Polymer 2.x use of `Polymer.dom` is no longer required and
+   * in the majority of cases simply facades directly to the standard native
+   * API.
+   */
+  function dom(obj: Node|Event|null): DomApi|EventApi|null;
```
TimvdLippe: I would specify this as the following:

```ts
function dom(obj: Node): DomApi;
function dom(obj: Event): EventApi;
function dom(obj: null): DomApi;
```

types/lib/utils/async.d.ts
```diff
@@ -0,0 +1,109 @@
...
+       * Returns a sub-module with the async interface providing the provided
+       * delay.
+       */
+      function after(delay: number): AsyncInterface|null;
+
+
+      /**
+       * Enqueues a function called in the next task.
+       */
+      function run(fn: Function|null): number;
```
TimvdLippe: Can not be `|null`:

```js
> Polymer.Async.timeOut.run()
< VM1220:1 Uncaught TypeError: Failed to execute 'setTimeout' on 'Window': 1 argument required, but only 0 present.
    at <anonymous>:1:23
```

types/lib/utils/async.d.ts
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Async interface wrapper around `requestAnimationFrame`.
+     */
+    namespace animationFrame {
+
+
+      /**
+       * Enqueues a function called at `requestAnimationFrame` timing.
+       */
+      function run(fn: Function|null): number;
```
TimvdLippe: Can not be `|null`:

```js
> Polymer.Async.animationFrame.run()
< Uncaught TypeError: Failed to execute 'requestAnimationFrame' on 'Window': 1 argument required, but only 0 present.
    at <anonymous>:1:30
```

types/lib/utils/async.d.ts
```diff
@@ -0,0 +1,109 @@
...
+     * Async interface wrapper around `requestIdleCallback`.  Falls back to
+     * `setTimeout` on browsers that do not support `requestIdleCallback`.
+     */
+    namespace idlePeriod {
+
+
+      /**
+       * Enqueues a function called at `requestIdleCallback` timing.
+       */
+      function run(fn: (p0: IdleDeadline|null) => any): number;
```
TimvdLippe: `=> any` can be `=> void`

types/lib/utils/async.d.ts
```diff
@@ -0,0 +1,109 @@
...
+     * Promises are avoided as an implementation choice for the time being
+     * due to Safari bugs that cause Promises to lack microtask guarantees.
+     */
+    namespace microTask {
+
+
+      /**
+       * Enqueues a function called at microtask timing.
+       */
+      function run(callback: Function|null): number;
```
TimvdLippe: Should be `callback?: Function` as invoking `Polymer.Async.microTask.run()` is okay.

types/lib/utils/flattened-nodes-observer.d.ts
```diff
@@ -0,0 +1,90 @@
...
+   *     this._observer.disconnect();
+   *   }
+   * }
+   * customElements.define(TestSelfObserve.is, TestSelfObserve);
+   * ```
+   */
+  class FlattenedNodesObserver {
+    _shadyChildrenObserver: MutationObserver|null;
+    _nativeChildrenObserver: MutationObserver|null;
+    _boundSchedule: () => any;
```
TimvdLippe: Hm are these actually supposed to be part of the public API @kevinpschaaf @sorvell ? I do not see them mentioned on https://www.polymer-project.org/2.0/docs/api/classes/Polymer.FlattenedNodesObserver

types/lib/utils/flush.d.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/// <reference path="boot.d.ts" />
+
+declare namespace Polymer {
+
+
+  /**
+   * Adds a `Polymer.Debouncer` to a list of globally flushable tasks.
+   */
+  function enqueueDebouncer(debouncer: Polymer.Debouncer|null): void|null;
```
TimvdLippe: Can not be `null`. That would throw an error later in `flushDebouncers()`

types/lib/utils/gestures.d.ts
```diff
@@ -0,0 +1,80 @@
...
+     * Similar to `document.elementFromPoint`, but pierces through
+     * shadow roots.
+     */
+    function deepTargetFind(x: number, y: number): Element|null;
+
+
+    /**
+     * Adds an event listener to a node for the given gesture type.
+     */
+    function addListener(node: Node|null, evType: string, handler: Function|null): boolean;
```
TimvdLippe: All of these (and in the later methods) can not be `|null`.

types/lib/utils/import-href.d.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+   * Convenience method for importing an HTML document imperatively.
+   *
+   * This method creates a new `<link rel="import">` element with
+   * the provided URL and appends it to the document to start loading.
+   * In the `onload` callback, the `import` property of the `link`
+   * element will contain the imported document contents.
+   */
+  function importHref(href: string, onload?: Function|null, onerror?: Function|null, optAsync?: boolean): HTMLLinkElement|null;
```
TimvdLippe: For some reason this type is different than the one specified in `legacy-element-mixin.d.ts`. Duplication FTW?

types/lib/utils/settings.d.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/// <reference path="boot.d.ts" />
+/// <reference path="resolve-url.d.ts" />
+
+declare namespace Polymer {
+
+  /**
+   * Legacy settings.
+   */
+  namespace settings {
```
TimvdLippe: We should probably add the declarations for the following:

```ts
namespace settings {
  useNativeCSSProperties: boolean;
  useNativeCustomElements: boolean;
  useShadow: boolean;
}
```

types/lib/utils/unresolved.d.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * DO NOT EDIT
+ *
+ * This file was automatically generated by
+ *   https://github.com/Polymer/gen-typescript-declarations
+ *
+ * To modify these typings, edit the source file(s):
+ *   lib/utils/unresolved.html
+ */
+
```
TimvdLippe: Maybe do not generate the file if there were no types in it?

types/lib/elements/array-selector.d.ts
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * When `true`, calling `select` on an item that is already selected
+     * will deselect the item.
+     */
+    toggle: boolean;
+
+    /**
+     * Clears the selection state.
+     */
+    clearSelection(): any;
```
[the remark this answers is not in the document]
aomarks: Done.

types/lib/elements/dom-repeat.d.ts
```diff
@@ -0,0 +1,268 @@
...
+    itemsIndexAs: string;
+
+    /**
+     * A function that should determine the sort order of the items.  This
+     * property should either be provided as a string, indicating a method
+     * name on the element's host, or else be an actual function.  The
+     * function should match the sort function passed to `Array.sort`.
+     * Using a sort function has no effect on the underlying `items` array.
+     */
+    sort: Function|null;
```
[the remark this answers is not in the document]
aomarks: TODO Update how Analyzer emits typings for Polymer properties. Talked to @rictic, we're thinking:

- If there is a `@type` annotation, emit it (already the case).
- If there is a default `value`, emit such that the type is non-nullable and non-undefined.
- Otherwise, assume the type could be null or undefined.

aomarks:
Addressed most of them comments, still a few more to look into. Pushed some changes to the generator, and a bunch of tweaks to the JSDoc annotations in Polymer.

types/lib/utils/unresolved.d.ts
aomarks: Currently we mirror the HTML import graph to the reference import graph, so that complicates that a bit, since dependents need to know whether their dependency has anything. See also https://github.com/Polymer/gen-typescript-declarations/blob/master/src/gen-ts.ts#L114

types/lib/utils/import-href.d.ts
aomarks: Done. Made them the same.

types/lib/utils/gestures.d.ts
aomarks: Done. Updated `@param` annotations.

types/lib/utils/flush.d.ts
aomarks: Done. Updated `@param` annotation.

types/lib/utils/async.d.ts
aomarks: Done. Updated `@param` annotation.

types/lib/utils/async.d.ts
aomarks: Done. Updated `@param` annotation.

types/lib/utils/async.d.ts
aomarks: Done. Updated `@param` annotation.

types/lib/utils/async.d.ts
aomarks: Done. Updated `@param` annotation.

types/lib/legacy/polymer.dom.d.ts
```diff
@@ -0,0 +1,145 @@
...
+     *   import
+     * @returns Clone of given node imported to this owner document
+     */
+    importNode(node: Node|null, deep: boolean): Node|null;
+
+    /**
+     * @returns Returns a flattened list of all child nodes and nodes assigned
+     * to child slots.
+     */
+    getEffectiveChildNodes(): any[]|null;
```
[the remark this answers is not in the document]
aomarks: `Node` or `HTMLElement`? Did `Node` for now. Updated `@param` annotation.

types/lib/legacy/polymer-fn.d.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Legacy class factory and registration helper for defining Polymer
+ * elements.
+ *
+ * This method is equivalent to
+ * `customElements.define(info.is, Polymer.Class(info));`
+ *
+ * See `Polymer.Class` for details on valid legacy metadata format for `info`.
+ */
+declare function Polymer(info: PolymerInit): HTMLElement;
```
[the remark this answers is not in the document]
aomarks: > Too bad TypeScript does not allow "dynamic" specification of the interface HTMLElementTagNameMap such that new Polymer({is: 'my-element'}) and document.createElement('my-element').

Yeah, that would be cool, but don't think it's possible. I'm sure you saw that we generate an extension to that interface when we find a custom element, but I think if users want it they'll just need to write it manually.

> However, I think the return type is wrong, as this functions returns a class and not an instance of HTMLElement.

Done. Good catch! Updated `@param` annotation.

types/lib/legacy/legacy-element-mixin.d.ts
```diff
@@ -0,0 +1,620 @@
...
+
+    /**
+     * Polyfill for Element.prototype.matches, which is sometimes still
+     * prefixed.
+     *
+     * @param selector Selector to test.
+     * @param node Element to test the selector against.
+     * @returns Whether the element matches the selector.
+     */
+    elementMatches(selector: string, node?: Element|null): boolean;
```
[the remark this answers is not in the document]
aomarks: Done. Updated `@param` annotation.

FYI actually `node?` won't take `null`, it will only take `undefined` (either explicitly or by omission), but I think in this case it's ok to constrain that?

types/lib/legacy/legacy-element-mixin.d.ts
```diff
@@ -0,0 +1,620 @@
...
+     * @param href URL to document to load.
+     * @param onload Callback to notify when an import successfully
+     *   loaded.
+     * @param onerror Callback to notify when an import
+     *   unsuccessfully loaded.
+     * @param optAsync True if the import should be loaded `async`.
+     *   Defaults to `false`.
+     * @returns The link element for the URL to be loaded.
+     */
+    importHref(href: string, onload: Function|null, onerror: Function|null, optAsync: boolean): HTMLLinkElement|null;
```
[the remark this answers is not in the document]
aomarks: Done. Updated `@param` annotations. Also made the return type non-nullable.

types/lib/legacy/legacy-element-mixin.d.ts
```diff
@@ -0,0 +1,620 @@
...
+     * By default (if no waitTime is specified), async callbacks are run at
+     * microtask timing, which will occur before paint.
+     *
+     * @param callback The callback function to run, bound to `this`.
+     * @param waitTime Time to wait before calling the
+     *   `callback`.  If unspecified or 0, the callback will be run at microtask
+     *   timing (before paint).
+     * @returns Handle that may be used to cancel the async job.
+     */
+    async(callback: Function|null, waitTime?: number): number;
```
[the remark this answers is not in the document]
aomarks: Done. Updated `@param` annotation.

types/lib/legacy/legacy-element-mixin.d.ts
```diff
@@ -0,0 +1,620 @@
...
+     *   context) when the wait time elapses.
+     * @param wait Optional wait time in milliseconds (ms) after the
+     *   last signal that must elapse before invoking `callback`
+     * @returns Returns a debouncer object on which exists the
+     * following methods: `isActive()` returns true if the debouncer is
+     * active; `cancel()` cancels the debouncer if it is active;
+     * `flush()` immediately invokes the debounced callback if the debouncer
+     * is active.
+     */
+    debounce(jobName: string, callback: () => any, wait: number): Object|null;
```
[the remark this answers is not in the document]
aomarks: Done. Updated `@param` annotation.

types/lib/legacy/legacy-element-mixin.d.ts
```diff
@@ -0,0 +1,620 @@
...
+
+    /**
+     * Returns a list of nodes distributed within this element that match
+     * `selector`. These can be dom children or elements distributed to
+     * children that are insertion points.
+     *
+     * @param selector Selector to run.
+     * @returns List of distributed elements that match selector.
+     */
+    queryDistributedElements(selector: string): Array<Node|null>|null;
```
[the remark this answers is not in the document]
aomarks: Done. Updated `@param` annotation to `!Array<!Node>`. In Closure, all types except primitives are nullable by default.

types/lib/legacy/legacy-element-mixin.d.ts
```diff
@@ -0,0 +1,620 @@
...
+     * @param type Name of event type.
+     * @param detail Detail value containing event-specific
+     *   payload.
+     * @param options Object specifying options.  These may include:
+     *  `bubbles` (boolean, defaults to `true`),
+     *  `cancelable` (boolean, defaults to false), and
+     *  `node` on which to fire the event (HTMLElement, defaults to `this`).
+     * @returns The new event that was fired.
+     */
+    fire(type: string, detail?: any, options?: any): Event|null;
```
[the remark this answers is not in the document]
aomarks: Done. Added support for record types in https://github.com/Polymer/gen-typescript-declarations/pull/41.

types/lib/legacy/legacy-element-mixin.d.ts
```diff
@@ -0,0 +1,620 @@
...
+
+    /**
+     * Provides an override implementation of `attributeChangedCallback`
+     * which adds the Polymer legacy API's `attributeChanged` method.
+     *
+     * @param name Name of attribute.
+     * @param old Old value of attribute.
+     * @param value Current value of attribute.
+     */
+    attributeChangedCallback(name: string, old: string|null, value: string|null): any;
```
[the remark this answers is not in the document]
aomarks: I think when a method is overridden, Analyzer doesn't lift up type annotations from its ancestor, so each implementation has to declare the types separately.

types/lib/legacy/class.d.ts
```diff
@@ -0,0 +1,105 @@
...
+   * In addition, the following Polymer-specific callbacks may be provided:
+   * - `registered`: called after first instance of this element,
+   * - `created`: called during `constructor`
+   * - `attached`: called during `connectedCallback`
+   * - `detached`: called during `disconnectedCallback`
+   * - `ready`: called before first `attached`, after all properties of
+   *   this element have been propagated to its template and all observers
+   *   have run
+   */
+  function Class(info: PolymerInit): () => any;
```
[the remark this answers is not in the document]
aomarks: Done. Added support for constructors in https://github.com/Polymer/gen-typescript-declarations/pull/41.

types/lib/elements/dom-module.d.ts
```diff
@@ -0,0 +1,51 @@
...
+   *     <dom-module id="foo">
+   *       <img src="stuff.png">
+   *     </dom-module>
+   *
+   * Then in code in some other location that cannot access the dom-module above
+   *
+   *     let img = customElements.get('dom-module').import('foo', 'img');
+   */
+  class DomModule extends HTMLElement {
+    attributeChangedCallback(name: any, old: any, value: any): any;
```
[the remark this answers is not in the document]
aomarks: Done. Added `@param` annotation.

types/lib/elements/dom-bind.d.ts
```diff
@@ -0,0 +1,51 @@
...
+    Polymer.OptionalMutableData(
+    Polymer.GestureEventListeners(
+    Polymer.Element))) {
+
+    /**
+     * assumes only one observed attribute
+     */
+    attributeChangedCallback(): any;
+    connectedCallback(): any;
+    disconnectedCallback(): any;
```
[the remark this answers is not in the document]
aomarks: Done. Added `@returns` annotation.

types/lib/elements/array-selector.d.ts
```diff
@@ -0,0 +1,187 @@
...
+     * @returns Whether the item is selected
+     */
+    isIndexSelected(idx: number): boolean;
+
+    /**
+     * Deselects the given item if it is already selected.
+     *
+     * @param item Item from `items` array to deselect
+     */
+    deselect(item: any): void|null;
```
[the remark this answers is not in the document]
aomarks: Done. Fixed nullable void bug in https://github.com/Polymer/gen-typescript-declarations/pull/40.

types/lib/elements/array-selector.d.ts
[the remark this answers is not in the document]
aomarks: Done. Added `@returns` annotation.

types/lib/elements/array-selector.d.ts
```diff
@@ -0,0 +1,187 @@
...
+   * `select(item)` and `deselect(item)` API, updates the `selected` property
+   * which may be bound to other parts of the application, and any changes to
+   * sub-fields of `selected` item(s) will be kept in sync with items in the
+   * `items` array.  When `multi` is false, `selected` is a property
+   * representing the last selected item.  When `multi` is true, `selected`
+   * is an array of multiply selected items.
+   */
+  function ArraySelectorMixin<T extends new(...args: any[]) => {}>(base: T): {
+    new(...args: any[]): ArraySelectorMixin & Polymer.ElementMixin
+  } & T
```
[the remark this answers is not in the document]
aomarks: :)

types/lib/utils/settings.d.ts
aomarks: I moved the `@namespace` annotation from `settings` to `Settings` so at least that's correct now. But actually there is no way in Analyzer to model this object and its interface. Going to punt on the full interface for now. The main way people use this is with a `<script>` tag in their `index.html` anyway, so it's unlikely it would need to be references from TypeScript.

types/lib/legacy/legacy-element-mixin.d.ts
```diff
@@ -0,0 +1,620 @@
...
+
+    /**
+     * Convenience method to run `querySelector` on this local DOM scope.
+     *
+     * This function calls `Polymer.dom(this.root).querySelector(slctr)`.
+     *
+     * @param slctr Selector to run on this local DOM scope
+     * @returns Element found by the selector, or null if not found.
+     */
+    $$(slctr: string): Element|null;
```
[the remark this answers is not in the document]
aomarks: That would be cool. I'd need a way to directly override the TS definition though, because there'd be no way to define that with a JSDoc annotation. Will probably punt for now. Might end up writing an override mechanism later.

types/lib/elements/dom-repeat.d.ts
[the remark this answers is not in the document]
aomarks: Doing this right would probably be a change to Analyzer. Draft PR out at https://github.com/Polymer/polymer-analyzer/pull/791. Needs a bit more discussion. Proposal is that all Polymer properties should be emitted as possibly null or undefined, unless they have a `@type` annotation to constrain it further.

types/lib/utils/flattened-nodes-observer.d.ts
azakus: Yeah these are probably meant to be private or protected, as denoted by the `_` prefix

kevinpschaaf: Basically, these are intended to be "private".

On element classes/mixins, we have followed a convention of `_` single underscore is "protected" (intended for subclassers not instance users, but we will maintain the API per semver) and `__` as "private" (implementation detail, only the author should use, can change at any time, no semver guarantee).

We were a little sloppier on non-element classes, since we basically have way less expectation that anyone will subclass them, but that's not really defensible.  At a minimum we should `@private` annotate these, if not follow the same `__` convention.

justinfagnani: `@private` seems like the way to go, IMO

aomarks: Sounds good.

I just pulled all of the JSDoc changes, including these new `@private` annotations, into a separate PR at https://github.com/Polymer/polymer/pull/4997/files.

types/lib/elements/dom-repeat.d.ts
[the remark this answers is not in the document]
aomarks: Done via https://github.com/Polymer/polymer-analyzer/pull/791. Properties are null always `|null|undefined` unless a `@type` annotation says otherwise.

aomarks:
PTAL.

This PR is up to date with the latest changes to Polymer, Analyzer, and the type generator.

gulpfile.js
```diff
@@ -260,15 +260,24 @@ gulp.task('lint', (done) => {
...
+  await del(['types/**/*.d.ts', '!types/extra-types.d.ts']);
+  const config = await fs.readJson('gen-tsd.json');
+  const files = await genTs('.', config);
+  for (const [filePath, contents] of files) {
+    await fs.outputFile(path.join('types', filePath), contents);
+  }
 });
 
 gulp.task('update-version', () => {
   gulp.src('lib/utils/boot.html')
```
justinfagnani: fyi, this task looks like it's missing a return statement. It should return the stream pipeline.

types/extra-types.d.ts
```diff
@@ -0,0 +1,163 @@
...
+ * automatically generated.
+ */
+
+/// <reference path="lib/mixins/property-effects.d.ts" />
+
+/**
+ * Types from "externs/closure-types.js"
+ */
+
+type Polymer_PropertyEffects = Polymer.PropertyEffects;
```
justinfagnani: Why is this necessary? Add a comment?

types/extra-types.d.ts
```diff
@@ -0,0 +1,163 @@
+/**
+ * This file contains the types that are required for compilation of the
+ * Polymer generated type declarations, but which could not themselves be
+ * automatically generated.
+ */
+
+/// <reference path="lib/mixins/property-effects.d.ts" />
+
+/**
```
justinfagnani: Shouldn't this be a `/*` comment, since it's not a doc comment on the subsequent line?

types/extra-types.d.ts
```diff
@@ -0,0 +1,163 @@
...
+  notify?: boolean;
+  observer?: string;
+}
+
+type PolymerElementProperties = {
+  [key: string]: PolymerElementPropertiesMeta
+};
+
+interface PolymerInit {
+  is: string;
```
justinfagnani: Are there any comments to copy here? TODO maybe?

types/extra-types.d.ts
```diff
@@ -0,0 +1,163 @@
...
+  target: string;
+  parts: BindingPart[];
+  literal?: string;
+  isCompound: boolean;
+  listenerEvent?: string;
+  listenerNegate?: boolean;
+}
+
+/**
+ * Types from "lib/utils/async.html"
```
justinfagnani: Why do we have types from regular source files? Is it cause we don't have interfaces in there? Might be worth a comment.

types/lib/elements/array-selector.d.ts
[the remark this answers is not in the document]
justinfagnani: I think these would be easier to read if you define the `Constructor<T>` type.

types/lib/elements/array-selector.d.ts
```diff
@@ -0,0 +1,187 @@
...
+   *   },
+   *   toggleSelection(e) {
+   *     let item = this.$.employeeList.itemForElement(e.target);
+   *     this.$.selector.select(item);
+   *   }
+   * });
+   * ```
+   */
+  class ArraySelector extends
+    Polymer.ArraySelectorMixin(
```
justinfagnani: I wasn't sure that .d.ts files could have expressions in the extends clause. Does this really build the type up as expected?

types/lib/elements/dom-repeat.d.ts
```diff
@@ -0,0 +1,268 @@
...
+     * it tries to allow to actually fire in a given second. It does this
+     * by measuring the time between `rAF`s and continuously adjusting the
+     * number of items created each `rAF` to maintain the target framerate.
+     * Setting this to a higher number allows lower latency and higher
+     * throughput for event handlers and other tasks, but results in a
+     * longer time for the remaining items to complete rendering.
+     */
+    targetFramerate: number|null|undefined;
+    _targetFrameTime: number|null|undefined;
+    disconnectedCallback(): any;
```
justinfagnani: I think these should be typed as returning `void`

types/lib/legacy/class.d.ts
```diff
@@ -0,0 +1,110 @@
...
+
+declare class PolymerGenerated {
+  created(): any;
+  _registered(): any;
+  _applyListeners(): any;
+
+  /**
+   * only apply if not already set.
+   */
+  _ensureAttributes(): any;
```
justinfagnani: I think all of these should return `void` too

types/lib/legacy/polymer.dom.d.ts
```diff
@@ -0,0 +1,149 @@
...
+     */
+    getOwnerRoot(): Node|null;
+
+    /**
+     * For slot elements, returns the nodes assigned to the slot; otherwise
+     * an empty array. It is equivalent to `<slot>.addignedNodes({flatten:true})`.
+     *
+     * @returns Array of assigned nodes
+     */
+    getDistributedNodes(): Array<Node|null>|null;
```
justinfagnani: Some places you have `Foo[]` and here `Array<Foo>`. TODO to parse Closure and convert?

types/lib/mixins/properties-changed.d.ts
```diff
@@ -0,0 +1,257 @@
...
+    _initializeInstanceProperties(props: object|null): any;
+
+    /**
+     * Updates the local storage for a property (via `_setPendingProperty`)
+     * and enqueues a `_proeprtiesChanged` callback.
+     *
+     * @param property Name of the property
+     * @param value Value to set
+     */
+    _setProperty(property: string, value: any): any;
```
justinfagnani: should return `void`

types/lib/utils/html-tag.d.ts
```diff
@@ -0,0 +1,41 @@
...
+   *       <style>:host{ content:"..." }</style>
+   *       <div class="shadowed">${this.partialTemplate}</div>
+   *       ${super.template}
+   *     `;
+   *   }
+   *   static get partialTemplate() { return Polymer.html`<span>Partial!</span>`; }
+   *
+   * @returns Constructed HTMLTemplateElement
+   */
+  function html(strings: ITemplateArray|null, ...values: any[]): HTMLTemplateElement;
```
justinfagnani: `null` shouldn't be allowed here

types/lib/utils/import-href.d.ts
```diff
@@ -0,0 +1,27 @@
...
+   * Convenience method for importing an HTML document imperatively.
+   *
+   * This method creates a new `<link rel="import">` element with
+   * the provided URL and appends it to the document to start loading.
+   * In the `onload` callback, the `import` property of the `link`
+   * element will contain the imported document contents.
+   *
+   * @returns The link element for the URL to be loaded.
+   */
+  function importHref(href: string, onload?: Function, onerror?: Function, optAsync?: boolean): HTMLLinkElement;
```
justinfagnani: It'd be nice to have better types for the callbacks. Issue in Polymer core?

types/extra-types.d.ts
```diff
@@ -0,0 +1,163 @@
...
+ */
+interface IdleDeadline {
+  didTimeout: boolean;
+  timeRemaining(): number;
+}
+
+/**
+ * Closure type equivalence for tagged template literal function argument.
+ */
+type ITemplateArray = TemplateStringsArray;
```
justinfagnani: Ah, I didn't see this when I looked at `Polymer.html`. Can't you just add this to the Closure -> TS type converter functions?

TimvdLippe:
I have a couple more comments:

1. It seems that all `static` methods are not typed. For example https://www.polymer-project.org/2.0/docs/api/elements/Polymer.Element#staticmethod-createPropertiesForAttributes is not found in the types.
2. Running the tasks outputs some warnings, but I have no clue where they are coming from: 
```
$ yarn run build generate-typescript
yarn run v1.3.2
$ gulp generate-typescript
[14:11:58] Using gulpfile ~/Projects/polymer/polymer/gulpfile.js
[14:11:58] Starting 'generate-typescript'...
Unknown syntax.
Unknown syntax.
Unknown syntax.
Could not find a name for class.
[14:12:04] Finished 'generate-typescript' after 5.85 s
Done in 6.75s.
```

I have opened #4998 to tighten almost all of the types to the correct value. This removes all `: any`, types methods that were completely missing types and fixes up some incorrect types.

types/lib/legacy/polymer.dom.d.ts
rictic: This is deliberate. If it's an array of a named type, we do `Foo[]`, if it's anything more complicated we do `Array<Foo | Bar | Baz>`, following the google typescript style guide.

types/lib/elements/array-selector.d.ts
aomarks: Yep, you can do this and it works.

aomarks:
> I think these would be easier to read if you define the Constructor<T> type.

I didn't want to pollute the global namespace with this type. I would have to somehow ensure it's only defined once when multiple repos' generated types are compiled together.

aomarks:
> It seems that all static methods are not typed. For example https://www.polymer-project.org/2.0/docs/api/elements/Polymer.Element#staticmethod-createPropertiesForAttributes is not found in the types.

Oops, yeah quite an omission. I've added static method support in https://github.com/Polymer/gen-typescript-declarations/pull/46. Turned out representing static methods for mixins required significant refactoring of the way mixins are emitted.

aomarks:
> One thing that would be nice is if we could run the generated types through the type checker to make sure they're at least well formed and all referenced built-in types are defined

I've been testing manually, the current typings compile cleanly. It would be nice to automate it, maybe fail on Travis if it's not compiling. Maybe we should do the same for Closure compile too. Don't think I'll do that in this PR though.

types/lib/utils/html-tag.d.ts
aomarks: Done (updated annotation in Polymer).

types/lib/utils/import-href.d.ts
aomarks: Yep, would just need to update the `@params` in `import-href.html`. Punting for now.

types/extra-types.d.ts
aomarks: Fixed via https://github.com/Polymer/gen-typescript-declarations/pull/47.

types/extra-types.d.ts
aomarks: Some interfaces are defined directly in the source if they are only used in that one place. Not sure I need to comment that here? Did end up removing one case, since `AsyncModule` turned out to be a re-definition of `AsyncInterface`.

types/extra-types.d.ts
aomarks: Added TODO. Nothing in the externs to copy so I'd be writing something new.

types/extra-types.d.ts
aomarks: No longer necessary. Fixed in https://github.com/Polymer/gen-typescript-declarations/pull/47.

types/extra-types.d.ts
aomarks: Done.

gulpfile.js
aomarks: Wasn't changed in this PR, but added a return.

aomarks:
> Running the tasks outputs some warnings, but I have no clue where they are coming from:

I would like to improve the warnings output by integrating with the very nice Analyzer warnings printer. For another time though.

> Unknown syntax.
> Unknown syntax.
> Unknown syntax.

These were from not supporting `Object<foo, bar>` syntax. Fixed in https://github.com/Polymer/gen-typescript-declarations/pull/48.

> Could not find a name for class.

This is from `https://github.com/Polymer/polymer/blob/master/lib/utils/templatize.html#L61`. Safe to ignore.

types/lib/mixins/properties-changed.d.ts
aomarks: Fixed by Tim in https://github.com/Polymer/polymer/pull/4998

types/lib/elements/dom-repeat.d.ts
aomarks: Fixed in https://github.com/Polymer/polymer/pull/4998

types/lib/legacy/class.d.ts
aomarks: Fixed in https://github.com/Polymer/polymer/pull/4998

aomarks:
I think this is ready to merge. All comments addressed.

lib/utils/mixin.html
aomarks: Closure is compiling cleanly now.

types/lib/legacy/polymer.dom.d.ts
aomarks: That would be really nice, but no way to represent this with Closure so we'd need to manually override this. Don't have that mechanism yet so will think a bit and punt for a later PR.

types/lib/utils/import-href.d.ts
justinfagnani: Funny, I filed an issue a ways back and it got comments recently: https://github.com/Polymer/polymer/issues/4603

justinfagnani:
LGTM still

43081j:
Just a couple of things but they could be sorted in a future PR:

Annotations:

- dom-repeat's `indexForElement` has an annotation of `{*}` for its return type, but should probably be `number`
- properties-changed mixin should have `createdProperties` as `void` return type
- properties-mixin should have `finalize` as `void` return type
- async `animationFrame.run` should have `function(number):void` parameter
- render-status `beforeNextRender` and `afterNextRender` parameters should be `function():void`

Misc:

- does anyone know of progress or a PR to add `connectedCallback` and what not to typescript's shipped definitions? there's a few places we have to define these but they should really be inherited
- `animationFrame.run` and `animationFrame.cancel` are duplicated

Great work @aomarks, will be testing this out more through the week.

PR #4934, opened 2017-11-14
add return jsdoc to void functions

43081j:
This is just a quick find and replace to add `@return {void}` to any functions/methods which never (and shouldn't) return.

This'll help with the typescript generator and until polymer/polymer-analyzer#753 is solved (if it does get solved).

I may have missed some as this was mostly scripted but it looks like i got most of the public methods. I left any overridden methods alone and such.


TimvdLippe:
@rictic Thanks for the review! @43081j could you rebase, then we can merge this PR 😄

43081j:
here you go. just waiting for the ci build to pass but should be fine.

aomarks:
Nice! Thanks for this @43081j.

PR #4936, opened 2017-11-15
dom-repeat: add filteredItemCount to display total count of items with filter applied

### LINKED ISSUE #4935: Feature request: dom-repeat: property for total count of filtered items
This is a feature request for `dom-repeat`.

Currently, if you want to get the count of items used by a `dom-repeat` after applying a `filter` function, you can use [`renderedItemCount`](https://www.polymer-project.org/2.0/docs/api/elements/Polymer.DomRepeat#property-renderedItemCount). This works well for many scenarios, but is not ideal if you are using "chunked" rendering and want to know the total filtered count upfront.

As an example, if I have a list of 1000 items that is being filtered down to 500 items and I set `initialCount` to 10, then if I am using `renderedItemCount` to let the user know how many items are in the list, the value will have to update many times before it finally settles. What I really want is some way to know the value of 500 and not have it change every time a "chunk" is finished rendering.

I propose adding a new property, `filteredItemCount` that updates when the `filter` changes and reflects the total count of filtered items regardless of if they have finished rendering or not.

trentmwillis:
Fixes #4935. Open to feedback on the solution/implementation 😄

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -229,9 +229,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
          * Count of currently rendered items after `filter` (if any) has been applied.
          * If "chunking mode" is enabled, `renderedItemCount` is updated each time a
          * set of template instances is rendered.
-         *
          */
         renderedItemCount: {
+          type: Number,
+          notify: true,
+          readOnly: true
+        },
+
+        /**
+         * Total count of items after `filter` (if any) has been applied. If
+         * "chunking mode" is enabled, this number will reflect the eventual end
+         * count and not necessarily the number of items currently rendered.
+         */
+         filteredItemCount: {
           type: Number,
           notify: true,
           readOnly: true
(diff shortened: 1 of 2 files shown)


trentmwillis:
Friendly ping :) any chance this could get a review?

MajorBreakfast:
I took a look at the code and all seems fine to me. I think this is a valid improvement.

I like the naming of the new `filteredItemCount` property because it is consistent with the existing `renderedItemCount` property.

lib/elements/dom-repeat.html
```diff
@@ -554,6 +564,7 @@
       for (let i=this.__instances.length-1; i>=instIdx; i--) {
         this.__detachAndRemoveInstance(i);
       }
+      this._setFilteredItemCount(isntIdxToItemsIdx.length);
```
kevinpschaaf: Would it make sense to gate setting the filtered item count on actually having a filter applied (and documenting it that way)?   If you don't have a filter, the total eventual count is just `items.length`, and this would avoid the perf impact of firing a notifying event every render for the (majority) case where no filter is applied.  Or is that too confusing?

trentmwillis: Is the perf impact likely to be noticeable?

By only doing setting `filteredItemCount` when a `filter` is applied, we risk pushing some logic into user-space. For example, if I have a repeat that conditionally has a `filter` function, then I'll need my component to conditionally choose whether to use `filteredItemCount` or `items.length` based on whether or not `filter` is set.

trentmwillis:
I ran into a scenario again today where this would be useful. Any chance of getting it merged? I can rebase onto master if desired.

trentmwillis:
Closing this as there seems to be no interest from the maintainers.

PR #4941, opened 2017-11-17
Clarify docs on target-framerate. Fixes #4897

### LINKED ISSUE #4897: [docs] targetFramerate does not make it clear what units.
The name of the property implies 'Hz', and the words in the comment also imply 'Hz', but it isn't explicit.

Please confirm the units of the target framerate are Hz...ie the intention is to `initial-count` items are rendered every `target-framerate` frames.

If I understand correctly, if I set 'initial-count' to '1', and 'target-framerate' to '1', then dom-repeat will attempt to render one additional item per second. Is that correct?

kevinpschaaf:
### Reference Issue
Fixes #4897

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -254,12 +254,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
          * fps) to target by throttling the number of instances rendered each
          * frame to not exceed the budget for the target frame rate.  The
          * framerate is effectively the number of `requestAnimationFrame`s that
-         * it tries to allow to actually fire in a given second, and it targets
-         * this by empirically measuring the time between rAF's and throttling
-         * up/down the number of items created each `rAF`. Setting this to a 
-         * higher number will allow lower latency and higher throughput for
-         * event handlers and other tasks, but will result in a longer time for
-         * the remaining items to complete rendering.
+         * it tries to allow to actually fire in a given second. It does this
+         * by measuring the time between `rAF`s and continuously adjusting the
+         * number of items created each `rAF` to maintain the target framerate.
+         * Setting this to a higher number allows lower latency and higher
+         * throughput for event handlers and other tasks, but results in a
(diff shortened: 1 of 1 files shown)


TimvdLippe:
@davidmaxwaterman could you review this PR as well?

lib/elements/dom-repeat.html
```diff
@@ -250,12 +250,16 @@
...
 
         /**
-         * When `initialCount` is used, this property defines a frame rate to
-         * target by throttling the number of instances rendered each frame to
-         * not exceed the budget for the target frame rate.  Setting this to a
+         * When `initialCount` is used, this property defines a frame rate (in
+         * fps) to target by throttling the number of instances rendered each
+         * frame to not exceed the budget for the target frame rate.  The
+         * framerate is effectively the number of `requestAnimationFrame`s that
+         * it tries to allow to actually fire in a given second, and it targets
```
arthurevans: This sentence is kind of long. Maybe break sentence here: "... a given second. It does this by measuring the time between `rAF`s and continuously adjusting the number of items created each `rAF` to maintain the target framerate."

I also tried an optional rewording here to emphasize the fact that the process is ongoing rather than one-time. Totally optional if you prefer the current text.

lib/elements/dom-repeat.html
```diff
@@ -250,12 +250,16 @@
...
+         * this by empirically measuring the time between rAF's and throttling
+         * up/down the number of items created each `rAF`. Setting this to a 
          * higher number will allow lower latency and higher throughput for
```
arthurevans: will allow => allows

lib/elements/dom-repeat.html
```diff
@@ -250,12 +250,16 @@
...
-         * things like event handlers, but will result in a longer time for the
-         * remaining items to complete rendering.
+         * event handlers and other tasks, but will result in a longer time for
```
arthurevans: will result => results

PR #4942, opened 2017-11-17
Ensure re-sort/filter always happens after array item set. Fixes #3626

### LINKED ISSUE #3247: dom-repeat doesn't sort properly when updated with a larger array
When using `dom-repeat` with an items array and a `sort` attribute, it will properly sort the items.  However, if I update the items array from a callback handler to a new array that is longer than the previous one, then it will not be sorted properly.  It appears that the items are actually sorted in the correct order, however the items start somewhere in the middle of the list and then wrap around.  It appears that the number of items wrapped around at the end is equal to the same number of items that was in the previous items array.  So, I would suspect this bug is due to some DOM node reuse perhaps?

To clarify the array ordering, if the previous array had 3 items and the new array has 10 items, then the new sort order is: `3, 4, 5, 6, 7, 8, 9, 0, 1, 2`

As a workaround I was able to get the correct sort order by delaying setting of the new array in the callback as follows.  (`index` is the property bound to `items` in the `dom-repeat`)  I'm not sure why this helps.

```
event_callback(e:CustomEvent) {
    setTimeout(()=> { this.index = e.detail.response; }, 0);
}
```

### LINKED ISSUE #3254: dom-repeat filter not updating when entire Array item is set
Summary:

```
 <template is="dom-repeat" filter="aFilter" observe="show" items="{{items}}">
        <div>{{item.name}}</div>
    </template>

  properties: {  items: { type: Array, value: function() { return [ 
        {show: false, name: 'A'}, 
        {show: true, name: 'B'}, 
      ]}
  }
```

This updates the dom-repeat as expected:

```
this.set('items.0.show', true);
```

This however does not update the dom-repeat, although I would expect it to:

```
  var newItem = { show: true, name: 'A' };
  this.set('items.0', newItem);
```

As a workaround I am using splice, which works as expected:

```
this.splice('items', 0, 1, newItem);
```

Fiddle: https://jsfiddle.net/stevenjb_chromium/nbndnsk7/

### LINKED ISSUE #3626: dom-repeat doesn't call `_render` when an item changes

kevinpschaaf:
Previously, `this.set('items.1', {...})` did not cause a re-render when sorted/filtered, even if paths were observed.  This fixes the bug by ensuring that changing the entire item always causes a re-sort/filter.

### Reference Issue
Fixes #3626
Fixes #3254 
Fixes #3247

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -456,13 +456,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
 
     __handleObservedPaths(path) {
-      if (this.__observePaths) {
-        path = path.substring(path.indexOf('.') + 1);
-        let paths = this.__observePaths;
-        for (let i=0; i<paths.length; i++) {
-          if (path.indexOf(paths[i]) === 0) {
-            this.__debounceRender(this.__render, this.delay);
-            return true;
+      // Handle cases where path changes should cause a re-sort/filter
+      if (this.__sortFn || this.__filterFn) {
+        if (!path) {
+          // Always re-render if the item iteself changed
+          this.__debounceRender(this.__render, this.delay);
+        } else if (this.__observePaths) {
+          // Otherwise, re-render if the path changed matches an observed path
+          path = path.substring(path.indexOf('.') + 1);
+          let paths = this.__observePaths;
+          for (let i=0; i<paths.length; i++) {
+            if (path.indexOf(paths[i]) === 0) {
+              this.__debounceRender(this.__render, this.delay);
+            }
           }
         }
       }
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -459,7 +459,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // Handle cases where path changes should cause a re-sort/filter
       if (this.__sortFn || this.__filterFn) {
         if (!path) {
-          // Always re-render if the item iteself changed
+          // Always re-render if the item itself changed
           this.__debounceRender(this.__render, this.delay);
         } else if (this.__observePaths) {
           // Otherwise, re-render if the path changed matches an observed path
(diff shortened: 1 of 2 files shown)


lib/elements/dom-repeat.html
```diff
@@ -456,13 +456,19 @@
     }
 
     __handleObservedPaths(path) {
-      if (this.__observePaths) {
```
TimvdLippe: Previously the condition was only on `this.__observePaths`. Is there a case where `this.__sortFn` and `this.__filterFn` are both falsy, but `this.__observePaths` is truthy? In that case, this would regress, as the condition is no longer hit.

lib/elements/dom-repeat.html
```diff
@@ -456,13 +456,19 @@
...
-        path = path.substring(path.indexOf('.') + 1);
-        let paths = this.__observePaths;
-        for (let i=0; i<paths.length; i++) {
-          if (path.indexOf(paths[i]) === 0) {
-            this.__debounceRender(this.__render, this.delay);
-            return true;
+      // Handle cases where path changes should cause a re-sort/filter
+      if (this.__sortFn || this.__filterFn) {
+        if (!path) {
+          // Always re-render if the item iteself changed
```
TimvdLippe: s/iteself/itself

lib/elements/dom-repeat.html
sorvell: According to the [docs](https://github.com/Polymer/polymer/blob/master/lib/elements/dom-repeat.html#L208), `observe` (which sets `__observePaths`) is only relevant when using `sort` and/or `filter` so this seems ok.

PR #4944, opened 2017-11-18
Fix issue when bindings were updated when dom-if was false

### LINKED ISSUE #4877: Changing binding property while dom-if's if is false throws an error
### Description
Recently I found a problem when I use dom-if and lazy-pages:

```html
<lazy-pages app="{{app}}">
  <template is="dom-if">
    <my-page app="{{app}}"></my-page>
  </template>
</lazy-pages>
```
When I try to change the 'if' attribute of dom-if following changing the property of `app`, an error occurs:
image

And another problem is when dom-if's if property is false, changing a data-binding will not actually reflect on the elements which are inside the dom-if element. For example, look at the following code:

```html
<dom-module id="lazy-pages">
  <template>
    [[app.name]]
    <slot></slot>
  </template>
</dom-module>

<dom-module id="my-page">
  <template>
    [[app.name]]
  </template>
</dom-module>

<lazy-pages id="outer' app="{{app}}">
  <template id="inner" is="dom-if">
    <my-page app="{{app}}"></my-page>
  </template>
</lazy-pages>
```

Run the following code:
```
inner.if = false;
outer.set('app.name', 'a new name');

// will result in:
<lazy-pages>
  #shadow-root
    <text>a new name</text>
    <my-page>
      #shadow-root
        <text>the old name</text>    <--------- value not updated
    </my-page>
    ...
```

Is it intended to be like this?

But if you set it's `if` property back to `true`, those data still remain outdated...

#### Live Demo

http://plnkr.co/edit/uQ5OOiJKWe0nBpd7QKqM?p=preview

#### Steps to Reproduce

Open dev-tools and try to click the DOM 2 times and you'll see the error shown in dev-tools.
Strangely, if I give an id to that dom-if element, everything will work fine...

Second issue's live demo: http://plnkr.co/edit/T6GRQGChLe8F13lklTWR?p=preview

#### Expected Results
No error is throw.

#### Actual Results
Error is thrown.

### Versions

- Polymer: v2.1.1

TimvdLippe:
I was able to extract the issue without the original usage of `<iron-lazy-pages>` based on the plunkr posted in #4877. It appears that the line was using the wrong datahost store. Per https://github.com/Polymer/polymer/issues/4877#issuecomment-337265625 changing the line using the same pattern as used in `Templatize` seemed to fix the issue.
### Reference Issue

Fixes #4877

### DIFF
--- a/lib/elements/dom-if.html
+++ b/lib/elements/dom-if.html
@@ -235,7 +235,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       let props = this.__invalidProps;
       if (props) {
         for (let prop in props) {
-          this.__instance._setPendingProperty(prop, this.__dataHost[prop]);
+          this.__instance._setPendingProperty(prop, this.__instance.__dataHost.__dataHost[prop]);
         }
         this.__invalidProps = null;
         this.__instance._flushProperties();
--- a/test/unit/dom-if-elements.html
+++ b/test/unit/dom-if-elements.html
@@ -344,3 +344,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   });
   </script>
 </dom-module>
+
+<dom-module id="x-container">
+  <template>
+    <slot></slot>
+  </template>
+  <script>
+  Polymer({
+    is: 'x-container',
+    properties: {
+      prop: Object
+    }
+  });
+  </script>
+</dom-module>
--- a/test/unit/dom-if.html
+++ b/test/unit/dom-if.html
@@ -62,6 +62,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     </dom-bind>
   </div>
 
+  <dom-bind id="data-binding-false">
+    <template>
+      <x-container id="data-binding-false-container" prop="{{prop}}">
+        <template is="dom-if">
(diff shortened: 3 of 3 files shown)


TimvdLippe:
@sorvell I can't seem to get this build passing on Sauce. It is failing for various reasons all the time, but it does not seem to be related in any regard to this PR

lib/elements/dom-if.html
```diff
@@ -235,7 +235,7 @@
       let props = this.__invalidProps;
       if (props) {
         for (let prop in props) {
-          this.__instance._setPendingProperty(prop, this.__dataHost[prop]);
+          this.__instance._setPendingProperty(prop, this.__instance.__dataHost.__dataHost[prop]);
```
sorvell: I believe this is correct, but it's awkward.

If the `dom-if` has a binding then its `dataHost` will be set and equivalent the the template `dataHost` but it not then it's unset. The template dataHost is always [set](https://github.com/Polymer/polymer/blob/master/lib/utils/templatize.html#L324) if there are any host properties.

I think we need to add an API for this `templateDataHost` or something to the `TemplateInstance` class since this is so confusing. Wdyt think @kevinpschaaf ?

TimvdLippe:
Blocked on #4965

TimvdLippe:
I am no longer pursuing this PR.

PR #4956, opened 2017-11-27
Allow non-JSON literals when property type is "Object".

### LINKED ISSUE #4464: Inconsistent attribute deserialization (eg: string to object vs number to object).
Upgrading to Polymer 2.0 (2.0-preview) and I've noticed an apparent change with attribute deserialization. Specifically if a property of non-String type is assigned a string value via its corresponding attribute then it is NOT deserialized (successfully), however other values (eg numbers or booleans, etc) will deserialize perfectly fine even when the property's type does not match.

Eg:
```
properties: {
  myobj: { type: Object }
}
...
<my-thing myobj="hello"></my-thing> <!-- doesn't work -->
<my-thing myobj="234"></my-thing> <!-- works! -->
```
This issue seems to affect string values specifically. If a number, bool, etc is specified then it sets the property successfully but if it's just some text string then it fails.

JSBin: https://jsbin.com/paqazel/edit?html,output

**Expected results** would be either that string values deserialize for non-string types (as they seemed to do before 2.0) or that they don't deserialize but neither do other values that don't match the declared type.

**Actual results** are that string values behave special and the inconsistency makes me feel like something is wrong here, and it made me confused and spend a bunch of time trying to figure out what I was doing wrong when upgrading to 2.0.

Or maybe I've misunderstood something?
Thanks :)

Chrome 58
Polymer 2.0-preview

pkaske:
It seems there are in general no tests if properties are correctly deserialized or I didn't see or get it???
If someone can give me a hint I would try to implement them.

### Reference Issue
Fixes #4464

### DIFF
--- a/lib/mixins/property-accessors.html
+++ b/lib/mixins/property-accessors.html
@@ -362,6 +362,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
               outValue = JSON.parse(/** @type {string} */(value));
             } catch(x) {
               // allow non-JSON literals like Strings and Numbers
+              outValue = value;
             }
             break;
 
--- a/test/unit/attributes-elements.html
+++ b/test/unit/attributes-elements.html
@@ -90,7 +90,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         observer: 'attr1Changed'
       },
       shorthandNumber: Number,
-      shorthandBool: Boolean
+      shorthandBool: Boolean,
+      foreignTypeObjectProp: {
+        type: Object,
+        value: 'none'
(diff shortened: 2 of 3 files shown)


TimvdLippe:
@pkaske You can add a test in this suite: https://github.com/Polymer/polymer/blob/4b0def6d04f2e552eef66dc39b7dd03561fbeb2c/test/unit/attributes.html#L90

pkaske:
@TimvdLippe I added some test. Are they sufficient? I'm jet not very experienced with testing web components, leave alone polymers internals. I'm more then open and thankful for any suggestions, tips, critique and hint to used anti patterns :)

PR #4957, opened 2017-11-28
Adds `restamp` mode to dom-repeat

### LINKED ISSUE #1713: dom-repeat is not properly updating elements on filter/sort

dfreedm:
(this is a "take two" of https://github.com/Polymer/polymer/pull/4363)

Fixes #1713

When restamp is true, template instances are never reused with different items. Instances mapping to current items are moved to their new location, new instances are created for any new items, and instances for removed items are discarded. This mode is generally more expensive than restamp: false as it results in more instances to be created and discarded during updates and disconnection/reconnection churn. By default, object identity is used for mapping instances to items. Set restampKey to provide key based identity.

When restamp: true is used, restampKey can be used to provide a path into the item object that serves as a unique key for the item rather than using object identity, for purposes of mapping instances to items. Instances for items with the same key will be reused, while all others will be either restamped or discarded.

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -269,6 +269,32 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         _targetFrameTime: {
           type: Number,
           computed: '__computeFrameTime(targetFramerate)'
+        },
+
+        /**
+         * When `restamp` is true, template instances are never reused with
+         * different items.  Instances mapping to current items are moved
+         * to their new location, new instances are created for any new
+         * items, and instances for removed items are discarded.  This mode is
+         * generally more expensive than `restamp: false` as it results in
+         * more instances to be created and discarded during updates and
+         * disconnection/reconnection churn.  By default, object identity is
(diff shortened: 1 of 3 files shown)


jogibear9988:
@azakus will this be fixed any time?

PR #4960, opened 2017-11-29
Add some @function annotations to APIs that are defined by assignment.

aomarks:
This helps analyzer understand when something is a function despite not having any function-looking syntax on the right-hand-side.

Part of https://github.com/Polymer/gen-typescript-declarations/issues/23


aomarks:
LMK if you can think of any other APIs that might be affected by this.

TimvdLippe:
I have restarted the builds, as they were failing. Let's see if the issues are related to your PR or general flakiness.

azakus:
The only way this would break tests is with `eslint` failing, but that would fail before the actual test runner started. Rerunning the tests again, and if they fail the same way I'm just going to merge.

PR #4962, opened 2017-11-29
Do not set touchend listeners to passive

### LINKED ISSUE #4961: passiveTouchGestures probably shouldn't apply to touchend
Touch scrolling is designed to be fast when `touchstart` and `touchmove` listeners are passive, but it's fine for `touchend` to be non-passive (since it doesn't block scrolling).  See [here](https://developers.google.com/web/updates/2017/01/scrolling-intervention#breakage_and_guidance) for some details including:

> Although it's still necessary to do so for mobile Safari, websites should not rely on calling preventDefault() inside of touchstart and touchmove listeners as this is no longer guaranteed to be honored in Chrome. Developers should apply the touch-action CSS property on elements where scrolling and zooming should be disabled to notify the browser before any touch events occur. To suppress the default behavior of a tap (such as the generation of a click event), call preventDefault() inside of a touchend listener.

From [code inspection](https://github.com/Polymer/polymer/blob/21ff47184edf0fd247776dedbf5b764c1a68e8a9/lib/utils/gestures.html#L484) it looks to me like Polymer's `passiveTouchGestures` feature (#3604) sets all touch listeners to passive, including `touchend`.  This will make adoption harder because it disables the ability to suppress `click` events.  See for example b/66423965 inside Google.

Perhaps `passiveTouchGestures` can simply be updated to apply only to `touchstart` and `touchmove` listeners?

/cc @azakus @dtapuska

azakus:
`touchend` listeners do not need to be passive to enable more performant
scrolling.

With this change, most of the tradeoffs with enabling
`passiveTouchGestures` disappear, leaving only the inability to control
scrolling from `track`, `down`, and `move` gestures.

Fixes #4961

### DIFF
--- a/lib/utils/gestures.html
+++ b/lib/utils/gestures.html
@@ -62,9 +62,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   /**
    * Generate settings for event listeners, dependant on `Polymer.passiveTouchGestures`
    *
+   * @param {string} eventName Event name to determine if `{passive}` option is needed
    * @return {{passive: boolean} | undefined} Options to use for addEventListener and removeEventListener
    */
-  function PASSIVE_TOUCH() {
+  function PASSIVE_TOUCH(eventName) {
+    if (isMouseEvent(eventName) || eventName === 'touchend') {
+      return;
+    }
     if (HAS_NATIVE_TA && SUPPORTS_PASSIVE && Polymer.passiveTouchGestures) {
       return {passive: true};
     } else {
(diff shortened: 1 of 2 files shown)


lib/utils/gestures.html
```diff
@@ -481,7 +481,7 @@
           gobj[dep] = gd = {_count: 0};
         }
         if (gd._count === 0) {
-          let options = !isMouseEvent(dep) && PASSIVE_TOUCH();
+          let options = !isMouseEvent(dep) && dep !== 'touchend' && PASSIVE_TOUCH();
```
kevinpschaaf: Move `isMouseEvent` and `touchend` check into `PASSIVE_TOUCH()`

TimvdLippe:
@azakus Since #4963 has been merged, I suppose this one can as well?

PR #4963, opened 2017-11-29
[1.x] Do not set touchend listeners to passive

azakus:
`touchend` listeners do not need to be passive to enable more performant
scrolling.

With this change, most of the tradeoffs with enabling
`passiveTouchGestures` disappear, leaving only the inability to control
scrolling from `track`, `down`, and `move` gestures.

1.x port of #4962

Related to #4961

### DIFF
--- a/src/standard/gestures.html
+++ b/src/standard/gestures.html
@@ -58,7 +58,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   })();
 
   // decide whether to use {passive: true} for gestures listening for touch events
-  function PASSIVE_TOUCH() {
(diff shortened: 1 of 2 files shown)


src/standard/gestures.html
```diff
@@ -349,7 +349,7 @@
           gobj[dep] = gd = {_count: 0};
         }
         if (gd._count === 0) {
-          var options = !isMouseEvent(dep) && PASSIVE_TOUCH();
+          var options = !isMouseEvent(dep) && dep !== 'touchend' && PASSIVE_TOUCH();
```
kevinpschaaf: Move `isMouseEvent` and `touchend` check into `PASSIVE_TOUCH()`

PR #4964, opened 2017-11-29
[1.x] Make `:dir` more robust

### LINKED ISSUE #4952: :dir selector not working on polymer 1.11.0
Styling with `:dir` selector doesn't work properly in Safari with polymer 1.x

E.g. consider the following element template:
```html
  <style>
    div {
      color: blue;
    }
    :host(:dir(rtl)) div  {
      color: red;
    }
  </style>
  <div>am i red?</div>
```
The div element will be wrongfully blue when we set `<html dir="rtl">`. This works as expected with polymer 2.x, but not with polymer 1.x
http://jsbin.com/pujiden/1/edit?html,output

azakus:
Fixes #4952

### DIFF
--- a/test/unit/dir.html
+++ b/test/unit/dir.html
@@ -109,6 +109,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     </script>
   </dom-module>
 
+  <dom-module id="other-element">
+    <script>
+      addEventListener('WebComponentsReady', function() {
+        Polymer({is: 'other-element'});
+      });
+    </script>
+  </dom-module>
+
   <dom-module id="x-complicated">
     <template>
       <style>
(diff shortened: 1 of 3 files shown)


test/unit/dir.html
```diff
@@ -109,6 +109,28 @@
...
 
+  <dom-module id="x-complicated">
+    <template>
+      <style>
+        #target {
+          padding: 20px;
+          color: white;
+          background: rgb(0, 0, 255);
+        }
+        :dir(rtl) #target {
```
kevinpschaaf: Add test for `#target:dir(rtl)`, since this is actually the correct selector for what people will typically want.

Need to make sure users are writing `:dir()` as a compound selector, not a descendant selector.  cc @arthurevans @sorvell @bicknellr @valdrinkoshi

valdrinkoshi: FYI did some tests with the compound selector, it works fine with most cases except when applied to another custom element, e.g. `my-element:dir(rtl)`: http://jsbin.com/cewijoy/1/edit?html,output

FWIW, testing descendant selectors on FF (where `:dir` is implemented natively) works as advertised.

valdrinkoshi: Filed the compound selector issue here https://github.com/Polymer/polymer/issues/4966

azakus: @kevinpschaaf Test added

@valdrinkoshi Can you file a bug with that example separately?

valdrinkoshi: done already, #4966

PR #4967, opened 2017-11-30
Annotate Polymer function with @global.

aomarks:
This will let Analyzer discover it thanks to
https://github.com/Polymer/polymer-analyzer/pull/774.

Part of https://github.com/Polymer/gen-typescript-declarations/issues/23

### DIFF
--- a/lib/legacy/polymer-fn.html
+++ b/lib/legacy/polymer-fn.html
@@ -24,6 +24,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      *
      * See `Polymer.Class` for details on valid legacy metadata format for `info`.
      *
+     * @global
      * @override
      * @function Polymer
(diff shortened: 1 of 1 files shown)


TimvdLippe:
I am okay with an extra annotation, but it leaves me wondering why detecting an assignment starting with `window.` wouldn't work. Is that because the `polymerFn` is used instead of `Polymer`?

aomarks:
Our approach in analyzer has so far generally been to just follow the `@memberof` annotations, rather than trying to understand the LHS of the assignment expression. If a function doesn't have `@memberof`, it would be totally ignored. So I just extended this to also respect `@global`, since `@memberof window` was a bit weird (see https://github.com/Polymer/polymer-analyzer/pull/774). I guess we could have had special handling for `window` instead, this was just simpler to do. Note it is a JSDoc thing: http://usejsdoc.org/tags-global.html.

PR #4970, opened 2017-12-04
Fix :dir selectors with nested custom elements

### LINKED ISSUE #4966: [polymer 2.x] :dir compound selector not working when applied on a custom element
The compound selector works fine with all cases except when applied to another custom element, e.g. `my-element:dir(rtl)` http://jsbin.com/cewijoy/1/edit?html,output

azakus:
Fixes #4966


azakus:
@kevinpschaaf fixes case where `other-element:dir(rtl){}` was being rewritten to `other-:host([dir="rtl"]) element{}`

PR #4971, opened 2017-12-04
`html` tag function for generating templates

### LINKED ISSUE #4937: Supply `html` tag function for generating template in modules
### Description
Currently `static get template()` can return an `HTMLTemplateElement` _or_ a string which is internally parsed into a template.  When moving to modules in Polymer 3, all templates from `dom-modules` are automatically moved into strings returned from the static template getter by `modulizer`, and all newly authored elements would be expected to just return strings.  This will be problematic for any Polymer 2.0 elements that extended an element and used `super.template` to wrap/modify its super class template _assuming super.template returned an `HTMLTemplateElement`_.

As such, we propose to deprecate `Polymer.Element`'s support for `static get template()` returning strings, such that `super.template` is guaranteed to return an `HTMLTemplateElement`, and never a string.  In its place, we'll provide a `html` tagged template literal tag function that does the string -> template conversion, and modulizer will be changed to wrap template strings with `html` automatically.

This approach solves both the backwards-compatibility problem mentioned above, eliminates the ambiguously-typed nature of the current `template` API, and also has an extra benefit of providing a hint to code editors to syntax-highlight the template string as HTML (Sublime, Atom, and Github apparently already do this automatically, and VS Code [has a plugin](https://marketplace.visualstudio.com/items?itemName=bierner.lit-html) to do so).

In addition, the `html` tag function should have a small amount of sugar to allow creating holes in the template, so that subclassers can override those with `html`-tagged strings (which return templates) and have the hole values interpolated as strings (via `innerHTML`) before being converted to a template, e.g.:

Super class (provides a template with header and footer "holes"):
```js
class SuperClass extends Polymer.Element {
  static get template() {
    return html`
      <div>${this.headerTemplate}</div>
      [[myProp.stuffThatGoesHere]]
      <div>${this.footerTemplate}</div>
    `;
  }
  static get headerTemplate() { return html`<h1>Header</h1>` }
  static get footerTemplate() { return html`<h1>Footer</h1>` }
}
```

Sub-class (both wraps the super template, and overrides the header/footer "holes"):
```js
class SubClass extends SuperClass {
  static get template() {
    return html`
      <div class="frame">${super.template}</div>
    `;
  }
  static get headerTemplate() { return html`<h2>Sub-header</h2>` }
  static get footerTemplate() { return html`<h2>Sub-footer</h2>` }
}
```

To make the examples above work, the `html` tag function must take the _`innerHTML`_ of any interpolated values that are `HTMLElement`s; this is the only sugar proposed, other than taking the final interpolated string and parsing it into a new `HTMLTemplateElement`.

_Note, the `html` tag function described here to be shipped with Polymer (which simply returns a vanilla `HTMLTemplateElement`) should not be confused with the `html` tag function provided with the [`lit-html`](https://github.com/PolymerLabs/lit-html) library, which is similar in concept but separate in implementation.  I think the benefits of de-facto standard HTML syntax highlighting in editors outweighs the potential confusion._ 

### Versions
- Polymer: v2

azakus:
Fixes #4937


lib/utils/html-fn.html
```diff
@@ -0,0 +1,61 @@
...
+     *   static get partialTemplate() { return Polymer.html`<span>Partial!</span>`; }
+     *
+     * @memberof Polymer
+     * @param {!Array<string>} strings Constant parts of tagged template literal
+     * @param {!Array<*>} values Variable parts of tagged template literal
+     * @return {!HTMLTemplateElement} Constructed HTMLTemplateElement
+     */
+    Polymer.html = function html(strings, ...values) {
+      /** @type {!Array<string>} */
+      const parts = [strings[0]];
```
justinfagnani: 🏌️: you can get the interpolation with:

```js
const htmlString = values.reduce((r, v, i) => r + htmlValue(v) + strings[i+1], strings[0]);
```

lib/utils/html-fn.html
```diff
@@ -0,0 +1,61 @@
...
+     * @param {*} value Object to stringify into HTML
+     * @return {string} HTML stringified form of `obj`
+     */
+    function htmlValue(value) {
+      if (value instanceof HTMLElement) {
+        return /** @type {!HTMLElement} */(value).innerHTML;
+      } else if (value instanceof Node) {
+        return /** @type {!Node} */(value).textContent;
+      } else {
+        return value.toString();
```
justinfagnani: I think this should be `String(value)` - that'll handle `null`, `undefined`, and matches the template literal spec, afaik

lib/utils/html-fn.html
```diff
@@ -0,0 +1,61 @@
...
+<script>
+  (function() {
+    'use strict';
+
+    /**
+     * @param {*} value Object to stringify into HTML
+     * @return {string} HTML stringified form of `obj`
+     */
+    function htmlValue(value) {
+      if (value instanceof HTMLElement) {
```
justinfagnani: `instanceof Element`? `Element` has `innerHTML` I think.

lib/utils/html-fn.html
```diff
@@ -0,0 +1,61 @@
...
+        return /** @type {!HTMLElement} */(value).innerHTML;
```
justinfagnani: Should this really be `innerHTML`? Seems like we want `innerHTML` for template for sure, but for other elements I'd think we'd want `outerHTML`. That might be a case for _only_ supporting templates.

lib/utils/html-fn.html
azakus: wouldn't it be more efficient to do the concatenation at the end?

lib/utils/html-fn.html
kevinpschaaf: I think we should only support `HTMLTemplateElement`'s `innerHTML` and String interpolation.

lib/utils/html-fn.html
```diff
@@ -0,0 +1,63 @@
...
+<script>
+  (function() {
+    'use strict';
+
+    /**
+     * @param {*} value Object to stringify into HTML
+     * @return {string} HTML stringified form of `obj`
+     */
+    function htmlValue(value) {
+      if (value === null || value === undefined) {
```
kevinpschaaf: Justin convinced me we shouldn't do this; let's go with just HTMLTemplateElement's innerHTML or cast to string.

lib/utils/html-fn.html
justinfagnani: Apparently string concatenation is so optimized that it's faster than join. There's a rope version of a string in V8, so that kind of makes sense.

lib/utils/html-fn.html
azakus: Ok

lib/utils/html-fn.html
```diff
@@ -0,0 +1,61 @@
...
+    function htmlValue(value) {
+      if (value instanceof HTMLTemplateElement) {
+        return /** @type {!HTMLTemplateElement} */(value).innerHTML;
+      } else {
+        return String(value);
+      }
+    }
+
+    /**
+     * Small tagged template string function to facilitate building a `<template>` element.
```
justinfagnani: I don't think the public documentation needs "small". I'd go with something like:

A template literal tag[1] that creates an HTML `<template>` element from the contents of the string. This allows you to write a Polymer Template in JavaScript.

Interpolated values are converted to strings when the template is created, they are not intended as a replacement for Polymer data-binding. There is special support for HTMLTemplateElement values, allowing for easy composition of superclass or partial templates.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_template_literals

azakus: I'll just copy that in there, thank you.

azakus:
@justinfagnani No, I generate the Changelog when we make a release
A docs issue would be great!

lib/utils/html-fn.html
azakus: SGTM

lib/utils/html-fn.html
```diff
@@ -0,0 +1,63 @@
...
+     *   }
+     *   static get partialTemplate() { return Polymer.html`<span>Partial!</span>`; }
+     *
+     * @memberof Polymer
+     * @param {!Array<string>} strings Constant parts of tagged template literal
+     * @param {...*} values Variable parts of tagged template literal
+     * @return {!HTMLTemplateElement} Constructed HTMLTemplateElement
+     */
+    Polymer.html = function html(strings, ...values) {
+      // use raw strings to preserve literal escapes in strings
```
TimvdLippe: Could we add a regression test for this to make sure we do not break this contract?

test/smoke/html-fn.html
```diff
@@ -0,0 +1,64 @@
...
+-->
+<!DOCTYPE html>
+<html>
+<head>
+  <script src="../../../webcomponentsjs/webcomponents-lite.js"></script>
+  <link rel="import" href="../../polymer-element.html">
+</head>
+<body>
+  <script>
+    const html = Polymer.html;
```
TimvdLippe: Is there a specific reason you have to do this assignment? I suspect it has to do with syntax highlighting not working on it?
```js
Polymer.html`template`
```
If this is the reason, I would add it in the documentation of the function to encourage this pattern instead. (and also point to Polymer 3 having a simple import for this, making the assignment useless)

justinfagnani:
One comment on file names: I'd consider naming the file `html-tag.html`, as user might not see this as a normal function.

lib/mixins/element-mixin.html
```diff
@@ -270,6 +271,8 @@
       let template = /** @type {PolymerElementConstructor} */ (klass).template;
       if (template) {
         if (typeof template === 'string') {
+          console.warn('String templates are deprecated. ' +
+            `Please return a <template>, or use Polymer.html\`\` for element <${klass.is}>`);
```
justinfagnani: TODO to link to docs for `Polymer.html`

lib/utils/html-fn.html
```diff
@@ -0,0 +1,63 @@
...
+    function htmlValue(value) {
+      if (value instanceof HTMLTemplateElement) {
+        return /** @type {!HTMLTemplateElement} */(value).innerHTML;
+      } else {
+        return String(value);
+      }
+    }
+
+    /**
+     * A template literal tag1 that creates an HTML <template> element from the contents of the string.
```
justinfagnani: "tag1" -> "tag"

lib/mixins/element-mixin.html
```diff
@@ -14,6 +14,7 @@
 <link rel="import" href="../utils/case-map.html">
 <link rel="import" href="../utils/style-gather.html">
 <link rel="import" href="../utils/resolve-url.html">
+<link rel="import" href="../utils/html-fn.html">
```
kevinpschaaf: I think we decided to bundle the `html` tag function with the entry points, so that in modules a user can do e.g. `import {Element, html} from 'polymer-element.js';` as a convenience.

So, let's remove the import here, and add it to the two main end-user entry points (`polymer.html` and `polymer-element.html`).  In those, add `Polymer.html = Polymer.html;` so that modulizer adds exports for the `html` function from those.

lib/utils/html-fn.html
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * @param {*} value Object to stringify into HTML
+     * @return {string} HTML stringified form of `obj`
+     */
+    function htmlValue(value) {
+      if (value instanceof HTMLTemplateElement) {
+        return /** @type {!HTMLTemplateElement} */(value).innerHTML;
+      } else {
+        return String(value);
```
kevinpschaaf: Actually, the String cast is unnecessary since you'll get the same behavior during concatenation.  So it's really just `value instanceof HTMLTemplateElement ? value.innerHTML : value;`

PR #4976, opened 2017-12-06
Select <style> more strictly

### LINKED ISSUE #4975: Style node management causes blocking error in ShadowDOM application.
The changes found here (https://github.com/Polymer/polymer/commit/819652eb#diff-41695d1bc0c64b86f0e5a7878f58df0f) throw the following error:

```Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.```

When there is an `<svg>` element with a `<style>` tag inside of the template of an element that has styles included from a shared source.

#### Live Demo

http://jsbin.com/lumapiquda/1/edit?html,console,output

#### Steps to Reproduce
1. Create `my-element`
2. Add a styled SVG element to that component.
3. Import external styles into that element.
4. Run element

#### Expected Results
No errors throw.

#### Actual Results
```Uncaught DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.```

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [x] Safari 11
- [ ] Safari 8
- [ ] IE 11

### Versions

- Polymer: v2.3.0
- webcomponents: master

### PR DESCRIPTION
**Pull request by Westbrook**
Select only direct children because you need to be able to user "insertBefore" later

Fixes #4975

ToDo:
- [ ] confirm x-browser (`:scope` might not be the answer because of that)
- [ ] add tests

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -417,7 +417,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     function processElementStyles(klass, template, is, baseURI) {
       const styles = Polymer.StyleGather.stylesFromModuleImports(is).concat(
         Polymer.StyleGather.stylesFromTemplate(template));
-      let templateStyles = template.content.querySelectorAll('style');
+      let templateStyles = template.content.querySelectorAll(':scope > style');
       let lastStyle = templateStyles[templateStyles.length-1];
       // ensure all gathered styles are actually in this template.
       for (let i=0; i < styles.length; i++) {

### CODE REVIEW

**Comment by Westbrook** [comment]
Reading up on the very clever and interesting `:scope > style` selector, it looks like there may be some issue with x-browser support, I need to confirm on a system with IE, but in that case would something like the following be overkill:
```
function processElementStyles(klass, template, is, baseURI) {
      const styles = Polymer.StyleGather.stylesFromModuleImports(is).concat(
        Polymer.StyleGather.stylesFromTemplate(template));
      let templateStyles = template.content.querySelectorAll('style');
      let lastDirectChildStyleIndex = templateStyles.length-1
      let lastDirectChildStyle = templateStyles[lastDirectChildStyleIndex];
      // a direct child of the tempalte is needed for application of
      // "included" styles below via `insertBefore`
      while (!lastDirectChildStyle.parentNode.isSameNode(template.content)) {
        lastDirectChildStyleIndex--;
        if (lastDirectChildStyleIndex < 0) {
          // if no direct child style elements, choose the first element child
          lastDirectChildStyle = template.content.firstElementChild;
        } else {
          lastDirectChildStyle = templateStyles[lastDirectChildStyleIndex];
        }
      }
      // ensure all gathered styles are actually in this template.
      for (let i=0; i < styles.length; i++) {
        let s = styles[i];
        // if the style is not in this template, it's been "included" and
        // we put a clone of it in the template.
        if (s.getRootNode() != template.content) {
          s = s.cloneNode(true);
          template.content.insertBefore(s, lastDirectChildStyle);
        }
        s.textContent = klass._processStyleText(s.textContent, baseURI);
      }
      if (window.ShadyCSS) {
        window.ShadyCSS.prepareTemplate(template, is);
      }
    }
```
Reading it back here, should it always be `template.content.firstElementChild`, which allows to a removal of all the back checking?

PR #4979, opened 2017-12-06
Handle styles that are not direct children of templates correctly

### LINKED ISSUE #4975: Style node management causes blocking error in ShadowDOM application.

azakus:
Find the correct spot to add "included" styles in the template by
keeping track of which "concrete" style in the template we have
encountered and inserting the "included" styles before them.

Fixes #4975

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -417,18 +417,18 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     function processElementStyles(klass, template, is, baseURI) {
       const styles = Polymer.StyleGather.stylesFromModuleImports(is).concat(
         Polymer.StyleGather.stylesFromTemplate(template));
-      const templateStyles = Array.from(template.content.querySelectorAll('style'));
+      const templateStyles = template.content.querySelectorAll('style');
       // keep track of the last "concrete" style in the template we have encountered
       let templateStyleIndex = 0;
       // ensure all gathered styles are actually in this template.
       for (let i=0; i < styles.length; i++) {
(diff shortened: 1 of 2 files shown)


azakus:
Still investigating a few busted cases in Safari 10 and 9

Westbrook:
I also like how this adheres more closely to the cascade I'd expect to find when applying something like:
```
<style include="first-shared-style">
/* styles that override some of the shared styles */
</style>
<style include="second-shared-style-with-more-overrides">
/* a fourth override */
</style>
```
Thanks for the quick attention to this!

PR #4981, opened 2017-12-09
Fix import path for html-tag.

aomarks:


kevinpschaaf:
LGTM

Would want to add a test but it's actually non-trivial since `polymer.html` imports `polymer-element.html`, which also imports the (correct) html tag function.

PR #4988, opened 2017-12-12
FlattenedNodesObserver: do not fail on node without children

### LINKED ISSUE #4987: IE11 - Unable to get property 'length' of undefined or null reference
### Description

image

#### Live Demo
Reproducible with [px-data-grid](https://github.com/vaadin/px-data-grid) repo.

#### Steps to Reproduce
1. Clone the https://github.com/vaadin/px-data-grid repo
2. Install deps
3. `polymer serve`
4. Open the served document in IE11 

#### Expected Results
No error is throw

#### Actual Results
Exception `Unable to get property 'length' of undefined or null reference` 

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [x] IE 11

### Versions
- Polymer: v2.3.1
- webcomponents: v1.0.20

limonte:
### Reference Issue
Fixes #4987

### Description
Document doesn't have children in IE:

For comparison, Chrome:

### DIFF
--- a/lib/utils/flattened-nodes-observer.html
+++ b/lib/utils/flattened-nodes-observer.html
@@ -130,7 +130,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     connect() {
       if (isSlot(this._target)) {
         this._listenSlots([this._target]);
-      } else {
+      } else if (this._target.children) {
         this._listenSlots(this._target.children);
         if (window.ShadyDOM) {
           this._shadyChildrenObserver =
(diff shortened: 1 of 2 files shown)


limonte:
> Please add a test case that we can run to verify the original issue and make sure we do not regress.

Added test which fails in IE without the `if (this._target.children)` check and passes with it. Please take a look @TimvdLippe

limonte:
Tests are failing in Firefox not because of this PR, there's something else broken

TimvdLippe:
Yes I started a discussion about the test failures, will get back to that. In the mean time, @azakus could you run this branch on IE to confirm the issue is fixed? Sadly I don't have a Windows development machine to my disposal.

azakus:
Got stuck on https://github.com/webcomponents/template/issues/29 for a bit, reviewing now

limonte:
Thanks @azakus! It would be great if we can have the patch release, our deadline is near and the IE11 support is important for our customer.

PR #4991, opened 2017-12-12
Fix regression with imported css

### LINKED ISSUE #4986: Polymer 2.3.1: Uncaught TypeError: Cannot read property 'parentNode' of undefined
With Polymer 2.3.1, I get the following error message. Downgrading to Polymer 2.3.0 resolves the issue.

```
Uncaught TypeError: Cannot read property 'parentNode' of undefined
    at processElementStyles (element-mixin.html:431)
    at Function._finalizeTemplate (element-mixin.html:650)
    at HTMLElement._initializeProperties (element-mixin.html:591)
    at HTMLElement._initializeProperties (legacy-element-mixin.html:163)
    at new PropertyAccessors (property-accessors.html:147)
    at new TemplateStamp (template-stamp.html:121)
    at new PropertyEffects (property-effects.html:1134)
    at new PolymerElement (element-mixin.html:559)
    at new GestureEventListeners (gesture-event-listeners.html:46)
    at new Dir (dir-mixin.html:130)
```

### Description
After updating my project to Polymer 2.3.1, the project stopped working with the above error message. In the mentioned location `element-mixin.html:431`, the variable `templateStyle` is undefined, because `templateStyles` is an empty node list.

#### Expected Results
No JS error.

#### Actual Results
JS error.

### Browsers Affected
- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions
- Polymer: v2.3.1
- webcomponents: v1.0.20

azakus:
Fix case where `<link rel="import" type="css">` is the only styling for
an element

Fixes #4986

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -255,7 +255,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // ensure all gathered styles are actually in this template.
       for (let i=0; i < styles.length; i++) {
         let s = styles[i];
-        let templateStyle = templateStyles[templateStyleIndex];
+        // if no styles in the template, fall back to first element child of template
+        let templateStyle = templateStyles.length ?
+          templateStyles[templateStyleIndex] : template.content.firstElementChild;
         // if the style is not in this template, it's been "included" and
         // we put a clone of it in the template before the style that included it
         if (templateStyle !== s) {
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -247,17 +247,22 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @private
      */
     function processElementStyles(klass, template, is, baseURI) {
-      const styles = Polymer.StyleGather.stylesFromModuleImports(is).concat(
-        Polymer.StyleGather.stylesFromTemplate(template));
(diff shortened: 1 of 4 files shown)


lib/mixins/element-mixin.html
```diff
@@ -255,7 +255,9 @@
       // ensure all gathered styles are actually in this template.
       for (let i=0; i < styles.length; i++) {
         let s = styles[i];
-        let templateStyle = templateStyles[templateStyleIndex];
+        // if no styles in the template, fall back to first element child of template
+        let templateStyle = templateStyles.length ?
+          templateStyles[templateStyleIndex] : template.content.firstElementChild;
```
TimvdLippe: Lacking the knowledge, but I assume the `firstElementChild` represents the `style` tag represents the `.css` file?

sorvell: This just says put any non-local styles either before the closest local template style or at the top of the template

azakus: No, the `<link>` element won't make it into the template.

sorvell: This will still throw if there's no `firstElementChild`.

azakus: Yup, good point. Let me make the smallest possible repro.

PR #4993, opened 2017-12-15
Templatize: remove slots when hiding children

### LINKED ISSUE #4977: dom-if that contains a slot not working (without restamp)
### Description
The following code is not correctly working:
```
<template is="dom-if" if="[[toggle]]">
  <slot></slot>
</template>
```
When the `toggle` becomes `true` and the template is stamped for the first time, the slotted content appears. When the `toggle` becomes false again, the slotted content does not disappear. When the `restamp` attribute is applied to `dom-if`, it works. But without `restamp`, it doesn't.

#### Live Demo
http://jsbin.com/fazewarake/edit?html,console,output

#### Steps to Reproduce
1. Create `test-slotted` element
1. Create `test-main` element and put `<test-slotted>` in the content
1. Create a `dom-if` in `test-main` with a toggle that toggles a `<slot>`
1. Observe how the slotted element does not toggle

#### Expected Results
Toggling the slotted element using `dom-if` works.

#### Actual Results
The slotted element appears, but then doesn't disappear when the `dom-if` becomes false. Note that when applying the `restamp` attribute to `dom-if`, it works.

### Versions
- Polymer: v2.3.0
- webcomponents: v1.0.20

sorvell:
Fixes #4977. When templatized content is hidden (via `_showHideChildren`) as in `dom-if` top level `<slot>` elements are now removed instead of being hidden. This ensures that assigned nodes are not displayed in false-y dom-if elements.


TimvdLippe:
While it works, it does feel weird to remove these slots. Would this solution still work if you change the light dom nodes?

lib/utils/templatize.html
```diff
@@ -165,6 +165,18 @@
               } else {
                 n.textContent = n.__polymerTextContent__;
               }
+            // remove and replace slot
+            } else if (n.localName === 'slot') {
+              if (hide) {
+                n.__polymerParent__ = n.parentNode;
+                n.__polymerSibling__ = n.nextSibling;
```
kevinpschaaf: If the sibling was removed (e.g. due to `dom-if` going false) then the slot would never come back.

May be better to replaceChild with a non-slot?

lib/utils/templatize.html
```diff
@@ -172,7 +172,22 @@
...
+            } else if (n.localName === 'slot') {
+              if (hide) {
+                n.__polymerParent__ = n.parentNode;
+                n.__polymerReplaced__ = new Comment('hidden-slot');
+                n.parentNode.replaceChild(n.__polymerReplaced__, n);
+              } else {
+                const parent = n.__polymerParent__;
+                const replace = n.__polymerReplaced__;
+                if (parent && replace) {
+                  parent.replaceChild(n, n.__polymerReplaced__);
```
TimvdLippe: Nit: can be `parent.replaceChild(n, replace)`

PR #4994, opened 2017-12-15
Adds `Polymer.DisableUpgradeMixin`

### LINKED ISSUE #4909: Provide a disable-upgrade option for Polymer 2
In Polymer 1.x users could add a `disable-upgrade` attribute to elements to defer upgrade work until the attribute was removed. This feature is not possible in Polymer 2 due to removal of bespoke API (e.g. users implement `connectedCallback` rather than `attached`); however, we should be able to add a mixin that users can opt-in to using.

sorvell:
Fixes #4909. Elements using this mixin will not "enable" themselves when they have a `disable-upgrade` attribute. This mirrors a feature of Polymer 1.x but is now opt-in.


lib/mixins/disable-upgrade-mixin.html
```diff
@@ -0,0 +1,89 @@
...
+          if (!this.__dataEnabled && value == null && this.isConnected) {
+            super.connectedCallback();
+          }
+        } else {
+          super.attributeChangedCallback(name, old, value);
+        }
+      }
+
+      __shouldEnable() {
+        return this.__dataEnabled || !this.hasAttribute(DISABLED_ATTR);
```
TimvdLippe: Where is `this.__dataEnabled` coming from? I do not see it referenced anywhere.

PR #4996, opened 2017-12-15
[ProperitesChanged] Fix deserialization

### LINKED ISSUE #4995: Recent merge `basic-element` branch breaks deserialization
A change in the `basic-element` branch made the `type` given to `PropertiesChanged._deserializeValue` callable. This means the type given in an element's `properties` object became callable. 

This breaks the case where a user might pass a class constructor as the property type.

### LINKED ISSUE #5002: Docs for PropertiesChanged mixin are incorrect
Docs comment for [PropertiesChanged mixin](https://github.com/Polymer/polymer/blob/master/lib/mixins/properties-changed.html#L30) reference API that doesn't exist on this class.
This looks like a copy-paste error from PropertyAccessors mixin.

sorvell:
Fixes #4995.
Fixes #5002.

A change merged with the `basic-element` caused the `type` argument to `_deserializeValue` to be called as a function to determine the deserialized output. This was an unintended breaking change from master and is reverted here. For example, the `basic-element` change regressed using a class constructor as a property type.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,9 @@ before_script:
 - bower install
 - gulp lint-eslint
 script:
-- xvfb-run wct --verbose -l firefox
+- xvfb-run wct -l chrome
+- xvfb-run wct -l firefox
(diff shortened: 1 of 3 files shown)


TimvdLippe:
After restarting tests are still timing out in Firefox, so some more updates are necessary apparently.

azakus:
Back down to the still bad, but consistent Safari timeouts.

PR #4997, opened 2017-12-15
Tighten Closure type annotations.

aomarks:
I pulled these changes out of https://github.com/Polymer/polymer/pull/4928. These are various changes to the Polymer core JSDoc annotations that improve the generated TypeScript declarations.

- Adds various `!` non-nullable operators. In Closure, all non-primitive types are non-nullable (we should try to use non-nullable types by default going forward if possible).
- Marked some things `@private` that are not part of our public API.
- Added `@return {void}` to some functions that don't return anything.
- A few other signature changes.

`gulp closure` produces no warnings

### DIFF
--- a/lib/elements/array-selector.html
+++ b/lib/elements/array-selector.html
@@ -211,7 +211,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       /**
        * Clears the selection state.
-       *
+       * @returns {void}
        */
       clearSelection() {
         // Unbind previous selection
--- a/lib/elements/array-selector.html
+++ b/lib/elements/array-selector.html
@@ -211,7 +211,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       /**
        * Clears the selection state.
-       * @returns {void}
+       * @return {void}
        */
       clearSelection() {
         // Unbind previous selection
--- a/lib/elements/dom-bind.html
+++ b/lib/elements/dom-bind.html
@@ -61,16 +61,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         this.__children = null;
       }
 
-      // assumes only one observed attribute
+      /** @returns {void} */
       attributeChangedCallback() {
+        // assumes only one observed attribute
         this.mutableData = true;
       }
 
+      /** @returns {void} */
       connectedCallback() {
         this.style.display = 'none';
         this.render();
       }
 
+      /** @returns {void} */
       disconnectedCallback() {
         this.__removeChildren();
       }
--- a/lib/elements/dom-bind.html
+++ b/lib/elements/dom-bind.html
@@ -61,19 +61,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         this.__children = null;
       }
 
-      /** @returns {void} */
+      /** @return {void} */
       attributeChangedCallback() {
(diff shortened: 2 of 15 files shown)


lib/elements/array-selector.html
```diff
@@ -211,7 +211,7 @@
 
       /**
        * Clears the selection state.
-       *
+       * @returns {void}
```
azakus: `@return`

lib/elements/dom-bind.html
```diff
@@ -61,16 +61,19 @@
         this.__children = null;
       }
 
-      // assumes only one observed attribute
+      /** @returns {void} */
```
azakus: `@return`

lib/elements/dom-bind.html
```diff
@@ -61,16 +61,19 @@
...
       attributeChangedCallback() {
+        // assumes only one observed attribute
         this.mutableData = true;
       }
 
+      /** @returns {void} */
```
azakus: `@return` is only necessary if the function takes arguments

lib/elements/dom-bind.html
```diff
@@ -61,16 +61,19 @@
...
       connectedCallback() {
         this.style.display = 'none';
         this.render();
       }
 
+      /** @returns {void} */
```
azakus: `@return` is only necessary if the function takes arguments

lib/elements/dom-module.html
```diff
@@ -74,6 +74,10 @@
       return null;
     }
 
+    /**
+     * @param {string} name
+     * @returns {void}
```
azakus: `@return`

lib/elements/dom-module.html
```diff
@@ -74,6 +74,10 @@
       return null;
     }
 
+    /**
+     * @param {string} name
```
azakus: `@param {?string} old`
`@param {?string} value`

lib/legacy/legacy-element-mixin.html
```diff
@@ -134,6 +134,7 @@
        * @param {string} name Name of attribute.
        * @param {?string} old Old value of attribute.
        * @param {?string} value Current value of attribute.
+       * @returns {void}
```
azakus: `@return`

lib/elements/dom-bind.html
aomarks: I added it because otherwise in TypeScript we have to output returns `any` in this case, because we have no idea what the return type is.

lib/elements/array-selector.html
aomarks: Done.

lib/elements/dom-bind.html
aomarks: Done.

lib/elements/dom-bind.html
aomarks: See below. Think we should keep these.

lib/elements/dom-module.html
aomarks: Done. Also had to add descriptions according to linter.

lib/elements/dom-module.html
aomarks: Done.

lib/legacy/legacy-element-mixin.html
aomarks: Done.

PR #4998, opened 2017-12-16
Tighten more types for TypeScript and Closure

TimvdLippe:
This PR is part of the work being done in #4928

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -113,7 +113,7 @@ class AddClosureTypeImport extends Transform {
 
 gulp.task('clean', () => del([DIST_DIR, 'closure.log']));
 
-gulp.task('closure', ['clean'], () => {
+gulp.task('closure', ['generate-externs'], () => {
 
   let entry, splitRx, joinRx, addClosureTypes;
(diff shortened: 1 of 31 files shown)


aomarks:
Does running `gulp closure` produce any warnings?

TimvdLippe:
> Does running gulp closure produce any warnings?

Nope

gulpfile.js
```diff
@@ -113,7 +113,7 @@ class AddClosureTypeImport extends Transform {
 
 gulp.task('clean', () => del([DIST_DIR, 'closure.log']));
 
-gulp.task('closure', ['clean'], () => {
+gulp.task('closure', ['generate-externs'], () => {
```
aomarks: Is this change ok? @azakus?

TimvdLippe: Reasoning for this change is that `closure-lint` would not properly pick up the new changes. Whenever we run `closure`, the externs should be up-to-date to generate the correct output. `generate-externs` in turn invokes `clean`, so that should still work.

PR #5000, opened 2017-12-19
add void return type annotations

### LINKED ISSUE #5010: TypeScript typings missing forwarded DomApi methods/properties
A number of methods and properties from `Node` are forwarded using a utility method, and Analyzer is not able to discover that they are present on the `DomApi` class: https://github.com/Polymer/polymer/blob/master/lib/legacy/polymer.dom.html#L238

### LINKED ISSUE #5012: mixinBehaviors has incorrect return type
Great to see Polymer 2 typings are now on the master branch.

I noticed that mixinBehaviors has an incorrect return type:
` function mixinBehaviors(behaviors: object|any[]|null, klass: HTMLElement|{new(): HTMLElement}): {new(): HTMLElement};`

 It returns an HTMLElement, ignoring the super class and the behaviors that were passed to it. This is due to how the JSDoc is set up: https://github.com/Polymer/polymer/blob/master/lib/legacy/class.html#L38

Adding the behavior types is a bit more complicated, but at least it could return the type of the class passed as the second argument. I think something like this:

`function mixinBehaviors<T extends HTMLElement>(behaviors: object|any[]|null, klass: T): T;`

43081j:
This is just to add a few annotations for improving the types we just merged in #4928.

I wasn't too sure about dom-repeat's `indexForElement` but if you follow it through the source, it seems it can only ever return a number.

---

* Fixes #5010
* Fixes #5012

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -674,7 +674,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * `modelForElement(el).set('item.<sub-prop>', value)`
      * should be used.
      *
-     * @param {HTMLElement} el Element for which to return the item.
+     * @param {!HTMLElement} el Element for which to return the item.
      * @return {*} Item associated with the element.
      */
     itemForElement(el) {
@@ -687,8 +687,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * If `sort` is provided, the index will reflect the sorted order (rather
      * than the original array order).
      *
-     * @param {HTMLElement} el Element for which to return the index.
-     * @return {*} Row index associated with the element (note this may
+     * @param {!HTMLElement} el Element for which to return the index.
+     * @return {?number} Row index associated with the element (note this may
      *   not correspond to the array index if a user `sort` is applied).
      */
     indexForElement(el) {
@@ -709,7 +709,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      *     model.set('item.checked', true);
      *   }
      *
-     * @param {HTMLElement} el Element for which to return a template model.
+     * @param {!HTMLElement} el Element for which to return a template model.
      * @return {TemplateInstanceBase} Model representing the binding scope for
      *   the element.
      */
--- a/lib/legacy/class.html
+++ b/lib/legacy/class.html
@@ -33,9 +33,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * to ensure that any legacy behaviors can rely on legacy Polymer API on
      * the underlying element.
      *
-     * @param {!(Object|Array)} behaviors Behavior object or array of behaviors.
-     * @param {!HTMLElement|function(new:HTMLElement)} klass Element class.
-     * @return {function(new:HTMLElement)} Returns a new Element class extended by the
+     * @template T
+     * @param {!(PolymerInit|Array<!PolymerInit>)} behaviors Behavior object or array of behaviors.
+     * @param {function(new:T)} klass Element class.
+     * @return {function(new:T)} Returns a new Element class extended by the
      * passed in `behaviors` and also by `Polymer.LegacyElementMixin`.
      * @memberof Polymer
      * @suppress {invalidCasts, checkTypes}
--- a/lib/legacy/legacy-element-mixin.html
+++ b/lib/legacy/legacy-element-mixin.html
@@ -316,9 +316,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * properties.  To ensure only `ownProperties` are copied from source
        * to target and that accessor implementations are copied, use `extend`.
        *
-       * @param {Object} target Target object to copy properties to.
-       * @param {Object} source Source object to copy properties from.
-       * @return {Object} Target object that was passed as first argument.
+       * @param {!Object} target Target object to copy properties to.
+       * @param {!Object} source Source object to copy properties from.
+       * @return {!Object} Target object that was passed as first argument.
        */
       mixin(target, source) {
         for (let i in source) {
@@ -352,12 +352,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * returns a document fragment containing the imported content.
        *
        * @param {HTMLTemplateElement} template HTML template element to instance.
-       * @return {DocumentFragment} Document fragment containing the imported
+       * @return {!DocumentFragment} Document fragment containing the imported
        *   template content.
       */
       instanceTemplate(template) {
         let content = this.constructor._contentForTemplate(template);
-        let dom = /** @type {DocumentFragment} */
+        let dom = /** @type {!DocumentFragment} */
           (document.importNode(content, true));
         return dom;
       }
@@ -377,7 +377,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        *  `bubbles` (boolean, defaults to `true`),
        *  `cancelable` (boolean, defaults to false), and
        *  `node` on which to fire the event (HTMLElement, defaults to `this`).
-       * @return {Event} The new event that was fired.
+       * @return {!Event} The new event that was fired.
        */
       fire(type, detail, options) {
         options = options || {};
@@ -631,7 +631,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       /**
        * Checks whether an element is in this element's local DOM tree.
        *
-       * @param {Element=} node The element to be checked.
+       * @param {!Element} node The element to be checked.
        * @return {boolean} true if node is in this element's local DOM tree.
        */
       isLocalDescendant(node) {
@@ -679,7 +679,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        *   context) when the wait time elapses.
        * @param {number} wait Optional wait time in milliseconds (ms) after the
        *   last signal that must elapse before invoking `callback`
-       * @return {Object} Returns a debouncer object on which exists the
+       * @return {!Object} Returns a debouncer object on which exists the
        * following methods: `isActive()` returns true if the debouncer is
        * active; `cancel()` cancels the debouncer if it is active;
        * `flush()` immediately invokes the debounced callback if the debouncer
@@ -768,9 +768,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * Convenience method for creating an element and configuring it.
        *
        * @param {string} tag HTML element tag to create.
-       * @param {Object} props Object of properties to configure on the
+       * @param {Object=} props Object of properties to configure on the
        *    instance.
-       * @return {Element} Newly created and configured element.
+       * @return {!Element} Newly created and configured element.
        */
       create(tag, props) {
         let elt = document.createElement(tag);
@@ -795,9 +795,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * element will contain the imported document contents.
        *
        * @param {string} href URL to document to load.
-       * @param {!Function=} onload Callback to notify when an import successfully
+       * @param {?function(!Event):void=} onload Callback to notify when an import successfully
        *   loaded.
-       * @param {!Function=} onerror Callback to notify when an import
+       * @param {?function(!ErrorEvent):void=} onerror Callback to notify when an import
        *   unsuccessfully loaded.
        * @param {boolean=} optAsync True if the import should be loaded `async`.
        *   Defaults to `false`.
--- a/lib/legacy/polymer.dom.html
+++ b/lib/legacy/polymer.dom.html
@@ -235,23 +235,62 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
   }
 
-  /** @type {function(boolean=):Node} */
+  /**
+   * @function
+   * @param {boolean=} deep
+   * @return {!Node}
+   */
   DomApi.prototype.cloneNode;
-  /** @type {function(Node):!Node} */
+  /**
+   * @function
+   * @param {Node} node
+   * @return {!Node}
+   */
   DomApi.prototype.appendChild;
-  /** @type {function(Node, Node):!Node} */
+  /**
+   * @function
+   * @param {Node} newChild
+   * @param {Node} refChild
+   * @return {!Node}
+   */
   DomApi.prototype.insertBefore;
-  /** @type {function(Node):!Node} */
+  /**
+   * @function
+   * @param {Node} node
+   * @return {!Node}
+   */
   DomApi.prototype.removeChild;
-  /** @type {function(Node, Node):!Node} */
+  /**
+   * @function
+   * @param {Node} oldChild
+   * @param {Node} newChild
+   * @return {!Node}
+   */
   DomApi.prototype.replaceChild;
-  /** @type {function(string, (string|number|boolean)):void} */
+  /**
+   * @function
+   * @param {string} name
+   * @param {string} value
+   * @return {void}
+   */
   DomApi.prototype.setAttribute;
-  /** @type {function(string):void} */
+  /**
+   * @function
+   * @param {string} name
+   * @return {void}
+   */
   DomApi.prototype.removeAttribute;
-  /** @type {function(string):?Element} */
+  /**
+   * @function
+   * @param {string} selector
+   * @return {?Element}
+   */
   DomApi.prototype.querySelector;
-  /** @type {function(string):!NodeList<!Element>} */
+  /**
+   * @function
+   * @param {string} selector
+   * @return {!NodeList<!Element>}
+   */
   DomApi.prototype.querySelectorAll;
 
   forwardMethods(DomApi.prototype, [
--- a/lib/legacy/polymer.dom.html
+++ b/lib/legacy/polymer.dom.html
@@ -235,6 +235,62 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     }
   }
 
+  forwardMethods(DomApi.prototype, [
+    'cloneNode', 'appendChild', 'insertBefore', 'removeChild',
+    'replaceChild', 'setAttribute', 'removeAttribute',
+    'querySelector', 'querySelectorAll'
+  ]);
+
+  forwardReadOnlyProperties(DomApi.prototype, [
+    'parentNode', 'firstChild', 'lastChild',
+    'nextSibling', 'previousSibling', 'firstElementChild',
+    'lastElementChild', 'nextElementSibling', 'previousElementSibling',
+    'childNodes', 'children', 'classList'
+  ]);
+
+  forwardProperties(DomApi.prototype, [
+    'textContent', 'innerHTML'
+  ]);
+
+
+  /**
+   * Event API wrapper class returned from `Polymer.dom.(target)` when
+   * `target` is an `Event`.
+   */
+  class EventApi {
+    constructor(event) {
+      this.event = event;
+    }
+
+    /**
+     * Returns the first node on the `composedPath` of this event.
+     *
+     * @return {!EventTarget} The node this event was dispatched to
+     */
+    get rootTarget() {
+      return this.event.composedPath()[0];
+    }
+
+    /**
+     * Returns the local (re-targeted) target for this event.
+     *
+     * @return {!EventTarget} The local (re-targeted) target for this event.
+     */
+    get localTarget() {
+      return this.event.target;
+    }
+
+    /**
+     * Returns the `composedPath` for this event.
+     * @return {!Array<EventTarget>} The nodes this event propagated through
+     */
+    get path() {
+      return this.event.composedPath();
+    }
+  }
+
+  Polymer.DomApi = DomApi;
+
   /**
    * @function
    * @param {boolean=} deep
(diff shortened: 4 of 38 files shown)


43081j:
yup was just finishing off some annotations, will do shortly

types/lib/utils/gestures.d.ts
```diff
@@ -42,15 +42,15 @@ declare namespace Polymer {
      *
      * @returns Returns true if a gesture event listener was added.
      */
-    function addListener(node: Node, evType: string, handler: Function): boolean;
+    function addListener(node: Node, evType: string, handler: (p0: Event|null) => void): boolean;
```
43081j: i need help here. `handler` should have an optional parameter, rather than just null.. `(p0?: Event|null)` (assuming strict null checks was on). my annotation was `function(?Event=):void` before, but that seems to produce `(p0: any)`...

43081j:
cc @aomarks 

this should be sorted now. needs some careful code reviewing just to be sure we haven't lost any type information though.

lib/elements/dom-repeat.html
```diff
@@ -687,8 +687,8 @@
      * If `sort` is provided, the index will reflect the sorted order (rather
      * than the original array order).
      *
-     * @param {HTMLElement} el Element for which to return the index.
-     * @return {*} Row index associated with the element (note this may
+     * @param {!HTMLElement} el Element for which to return the index.
+     * @return {number} Row index associated with the element (note this may
```
aomarks: It looks like this could also return `null`, see the short circuit evaluation where `modelForElement` can return `null`.

types/lib/utils/gestures.d.ts
aomarks: The converter was missing the ability to parse optional and rest parameters in type expressions. so it was reverting to `any`. Will be fixed in https://github.com/Polymer/gen-typescript-declarations/pull/51. When that's merged, I'll do a patch release and you should be able to use the correct syntax.

lib/legacy/legacy-element-mixin.html
```diff
@@ -768,9 +768,9 @@
        * Convenience method for creating an element and configuring it.
        *
        * @param {string} tag HTML element tag to create.
-       * @param {Object} props Object of properties to configure on the
+       * @param {?Object=} props Object of properties to configure on the
```
aomarks: `Object` is nullable by default, so don't need the `?`.

lib/legacy/polymer.dom.html
```diff
@@ -303,8 +303,8 @@
    * @summary Legacy DOM and Event manipulation API wrapper factory used to
    * abstract differences between native Shadow DOM and "Shady DOM."
    * @memberof Polymer
-   * @param {!Node|Event} obj Node or event to operate on
-   * @return {DomApi|EventApi} Wrapper providing either node API or event API
+   * @param {?(Node|Event)=} obj Node or event to operate on
```
aomarks: `Node` and `Event` are nullable by default, so don't need the `?`.

lib/utils/async.html
```diff
@@ -78,7 +78,6 @@
       /**
        * Enqueues a function called in the next task.
        *
-       * @function
```
aomarks: I wonder why we had these `@function` annotations... @justinfagnani do you know if these are important?

types/lib/utils/async.d.ts
```diff
@@ -43,20 +43,6 @@ declare namespace Polymer {
       function run(fn: Function, delay?: number): number;
 
 
-      /**
-       * Enqueues a function called in the next task.
-       *
-       * @returns Handle used for canceling task
-       */
-      function run(fn: Function, delay?: number): number;
```
aomarks: Oh, I guess those `@function` annotations were important!

lib/utils/async.html
43081j: some seemed to have it and some didn't. those which did would produce duplicate type declarations for some reason... so either we shouldn't have `@function`, or the generator is having trouble with it.

types/lib/utils/async.d.ts
43081j: these were duplicates, i think we still have them elsewhere in the file. `@function` caused the generator to output `run` twice...

43081j:
addressed most comments.

im having trouble running `gulp closure`. it seems to throw errors around base classes in mixins not existing...

TimvdLippe:
@43081j These warnings are outputted from the analyzer, so they are expected. If the task generates a `closure.log`, then there are warnings/errors that have to be solved.

types/lib/utils/gestures.d.ts
aomarks: This PR is released in `0.3.4` so bump the dependency and you should be able to get the right output for `function(?Event=):void`.

lib/utils/async.html
aomarks: Seems like a bug in polymer-analyzer.

lib/legacy/polymer.dom.html
```diff
@@ -303,8 +303,8 @@
    * @summary Legacy DOM and Event manipulation API wrapper factory used to
    * abstract differences between native Shadow DOM and "Shady DOM."
    * @memberof Polymer
-   * @param {!Node|Event} obj Node or event to operate on
-   * @return {DomApi|EventApi} Wrapper providing either node API or event API
+   * @param {(Node|Event)=} obj Node or event to operate on
+   * @return {!DomApi|!EventApi} Wrapper providing either node API or event API
```
justinfagnani: Is there a way to type this in Closure to simulate TypeScript overrides? Something like: `@type {(function(Node=):!DomApi)|(function(Event=):!EventApi)}`

lib/utils/async.html
justinfagnani: Not sure why it was annotated that way. I wonder if it had to do with the container being a namespace. It'd be good to run these annotation-changing PRs through doc generation.

lib/utils/gestures.html
```diff
@@ -429,7 +429,7 @@
      * @memberof Polymer.Gestures
      * @param {!Node} node Node to add listener on
      * @param {string} evType Gesture type: `down`, `up`, `track`, or `tap`
-     * @param {!Function} handler Event listener function to call
+     * @param {!function(?Event):void} handler Event listener function to call
```
justinfagnani: Should this be `function(!Event):void`?

lib/utils/gestures.html
```diff
@@ -447,7 +447,7 @@
      * @memberof Polymer.Gestures
      * @param {!Node} node Node to remove listener from
      * @param {string} evType Gesture type: `down`, `up`, `track`, or `tap`
-     * @param {!Function} handler Event listener function previously passed to
+     * @param {!function(?Event):void} handler Event listener function previously passed to
```
justinfagnani: same here

lib/utils/gestures.html
43081j: yup i would think so. i set it as this because its what `_add` took

lib/legacy/legacy-element-mixin.html
```diff
@@ -795,9 +795,9 @@
        * element will contain the imported document contents.
        *
        * @param {string} href URL to document to load.
-       * @param {!Function=} onload Callback to notify when an import successfully
+       * @param {!function(!Event):void=} onload Callback to notify when an import successfully
        *   loaded.
-       * @param {!Function=} onerror Callback to notify when an import
+       * @param {!function(!ErrorEvent):void=} onerror Callback to notify when an import
```
TimvdLippe: I think this can be just `:void`

lib/legacy/polymer.dom.html
```diff
@@ -55,9 +55,9 @@
      * Returns an instance of `Polymer.FlattenedNodesObserver` that
      * listens for node changes on this element.
      *
-     * @param {Function} callback Called when direct or distributed children
+     * @param {function(!Element, { target: !Element, addedNodes: !Array<!Element>, removedNodes: !Array<!Element> }):void} callback Called when direct or distributed children
```
TimvdLippe: I would extract this object type to an internal type. You are using it later in `FlattenedNodesObserver` as well, which is a little bit of duplication.

lib/utils/import-href.html
```diff
@@ -36,9 +36,9 @@
    *
    * @memberof Polymer
    * @param {string} href URL to document to load.
-   * @param {!Function=} onload Callback to notify when an import successfully
+   * @param {!function(!Event):void=} onload Callback to notify when an import successfully
```
TimvdLippe: Can be `:void`

lib/legacy/legacy-element-mixin.html
43081j: It cannot. the callback is optional.

lib/utils/import-href.html
43081j: again the callback is optional, it cannot.

43081j:
i managed to run `gulp closure` and committed the changes. but i get an error when it tries to access `compiler.jar` so maybe it didn't work exactly as it should.

aomarks:
> i managed to run gulp closure and committed the changes. but i get an error when it tries to access compiler.jar so maybe it didn't work exactly as it should.

Sounds like you have an installation problem. What OS are you on? I assume you have Java installed?

I ran the compiler from your branch and I do get these new warnings:
```
polymer.html_script_27.js:342: WARNING - Returned type does not match declared return type.
Expected : DocumentFragment
Found    : (DocumentFragment|null)
More details:
The found type is a union that includes an unexpected type: null
        return dom;
        ^^^^^^^^^^^

polymer.html_script_27.js:789: WARNING - Invalid type for parameter 2 of function Polymer.importHref.
Expected : (function(Event): undefined|undefined)
Found    : (function(Event): undefined|null)
More details:
The found type is a union that includes an unexpected type: null
        return Polymer.importHref(href, loadFn, errorFn, optAsync);
                                        ^^^^^^

polymer.html_script_27.js:789: WARNING - Invalid type for parameter 3 of function Polymer.importHref.
Expected : (function(ErrorEvent): undefined|undefined)
Found    : (function(ErrorEvent): undefined|null)
More details:
The found type is a union that includes an unexpected type: null
        return Polymer.importHref(href, loadFn, errorFn, optAsync);
                                                ^^^^^^^
```

BTW there shouldn't be anything to commit from running `gulp closure`, we don't check in the compiled output, we just make sure it will compile cleanly.

43081j:
ah i just committed what was updated (it updated `externs` which is already in the repo).

do you want me to reset that to whats in master?

im on windows btw, but i was using bash-on-windows so i think ill just give it a go in linux or something...

lib/utils/gestures.html
```diff
@@ -466,7 +466,7 @@
      * @private
      * @param {!HTMLElement} node Node on which to add the event.
      * @param {string} evType Event type to add.
-     * @param {function(Event?)} handler Event handler function.
+     * @param {function(!Event)} handler Event handler function.
```
43081j: @aomarks instead of changing `addListener` to take `!Event=`, i just changed `_add` to not be optional. this internally ends up doing `addEventListener` and i dont think theres an event which would fire without an event...

aomarks:
> ah i just committed what was updated (it updated externs which is already in the repo).
> 
> do you want me to reset that to whats in master?

Oh, actually we just changed the `gulp closure` script to update the externs, so I was wrong -- any changes to those files should be checked in.

43081j:
i've committed the externs changes.

also fixed any remaining issues i think.

lib/legacy/polymer.dom.html
```diff
@@ -282,6 +282,7 @@
 
     /**
      * Returns the `composedPath` for this event.
+     * @return {!Array<Node>}
```
43081j: @aomarks do you know why these don't turn up in the types? the getters. also, this should be `EventTarget` in typescript but closure doesn't seem to know what `EventTarget` is, so i set it as `Node` for now...

lib/legacy/polymer.dom.html
```diff
@@ -235,6 +235,25 @@
...
+  /** @type {function(Node, Node):!Node} */
+  DomApi.prototype.replaceChild;
+  /** @type {function(string, (string|number|boolean)):void} */
+  DomApi.prototype.setAttribute;
+  /** @type {function(string):void} */
+  DomApi.prototype.removeAttribute;
+  /** @type {function(string):?Element} */
+  DomApi.prototype.querySelector;
+  /** @type {function(string):!NodeList<!Element>} */
+  DomApi.prototype.querySelectorAll;
```
43081j: how do we want to handle these? this seems the right way to define them but the generator doesn't seem to pick them up.

lib/legacy/polymer.dom.html
aomarks: Only downside to that is that we'd also have to also manually write it for typescript in our `extra-types.d.ts`, because analyzer doesn't pick up on type declarations.

lib/legacy/polymer.dom.html
aomarks: I don't think we currently have a way to tell Analyzer about methods added to a class after declaration. @rictic is there any trick here?

lib/legacy/polymer.dom.html
```diff
@@ -274,14 +293,15 @@
...
-     * @return {Node} The local (re-targeted) target for this event.
+     * @return {!EventTarget} The local (re-targeted) target for this event.
      */
     get localTarget() {
       return this.event.target;
     }
 
     /**
      * Returns the `composedPath` for this event.
+     * @return {!Array<EventTarget>} The nodes this event propagated through
```
43081j: @aomarks it turns out closure does know what `EventTarget` is but i got confused because of the whole getters-are-not-yet-recognised issue. so these don't produce any types, but they pass the closure checks.

i did change some more types here from `Node` to `EventTarget` because that matches the dom spec.

types/extra-types.d.ts
```diff
@@ -13,7 +13,7 @@ interface PolymerElementPropertiesMeta {
   computed?: string;
   reflectToAttribute?: boolean;
   notify?: boolean;
-  observer?: string;
+  observer?: string|((val: any, old: any) => void);
```
43081j: are we ok doing this here? for polymer/polymer-decorators#23. though the externs js has `function(*, *)` or some such thing.

aomarks: Yes, I think this is correct.

43081j:
polymer/polymer-analyzer#816 is merged now so once there's a tagged release, we should be able to update here and detect the getters in the `DomApi` and what not.

aomarks:
> polymer/polymer-analyzer#816 is merged now so once there's a tagged release, we should be able to update here and detect the getters in the DomApi and what not.

We'll also need to wait for https://github.com/Polymer/gen-typescript-declarations/pull/56 to be merged, and then bump the analyzer version again in the deps. We're pinning the gen-typescript-declarations analyzer dependency, because the `3.0-pre` releases have a lot of breaking change potential.

43081j:
@aomarks fair enough, if you could let me know when thats ready ill bump it.

i've also just changed the dom api method annotations so they are defined as `@function`. assuming polymer/polymer-analyzer#821 gets accepted, we need to define them as a function so they're not detected as properties.

lib/legacy/polymer.dom.html
```diff
@@ -235,6 +235,64 @@
...
+   * @return {!Node}
+   */
+  DomApi.prototype.replaceChild;
+  /**
+   * @function
+   * @param {string} name
+   * @param {string} value
+   * @return {void}
+   */
+  DomApi.prototype.setAttribute;
```
43081j: FYI closure's types define `setAttribute` as taking `(string|number|boolean)`, whereas typescript's lib defines it as taking only `string`. Here i took the typescript version, but let me know if you disagree.

43081j:
@rictic @aomarks  do you know why the static method in dom-module doesn't get picked up? the analyser picks it up in master, and the generator has logic to read it from the analysis result in master. are we just behind? or is it something else?

lib/legacy/class.html
```diff
@@ -33,9 +33,10 @@
      * to ensure that any legacy behaviors can rely on legacy Polymer API on
      * the underlying element.
      *
-     * @param {!(Object|Array)} behaviors Behavior object or array of behaviors.
-     * @param {!HTMLElement|function(new:HTMLElement)} klass Element class.
-     * @return {function(new:HTMLElement)} Returns a new Element class extended by the
+     * @template T
+     * @param {!(PolymerInit|Array<!PolymerInit>)} behaviors Behavior object or array of behaviors.
+     * @param {function(new:T)} klass Element class.
+     * @return {function(new:T)} Returns a new Element class extended by the
```
43081j: can i get some review here? two things changed:

* we no longer accept an instance of an element, it didn't seem to make any sense unless im missing something (further down it ends up extending what you passed)
* i added a `T` so we accept and return the same type

Unfortunately, closure compiler isn't able to have something like... `T extends HTMLElement`. so this is still kinda wrong but better typed than before i suppose?

cc @rictic @aomarks

justinfagnani: FYI, Tuesday's when most people will start coming back to work.

43081j: haha no worries. who knows why im on here on NYE :tada: think its time to celebrate instead

justinfagnani: Definitely :)

Thanks for your recent contributions, looking forward to more in 2018 😄

FredKSchott:
@43081j this is awesome! Lets make sure we confirm https://github.com/Polymer/polymer-modulizer/ doesn't break with this PR, I'll run our fixture tests later today to confirm.

43081j:
ah that would be great if you wouldn't mind @FredKSchott. 

do remember too that we're waiting on a new analyzer and gen-ts-declarations release before we can merge.

btw as a small summary here's what we're waiting for to make this mergeable:

* polymer/gen-typescript-declarations#56
* polymer/polymer-analyzer#821

FredKSchott:
@43081j ah got it, thanks for the summary. I'll actually hold off then for now to prioritize some other work until this is closer to merging. 

*(For anyone reading this who has time to test them before I do, you can sync polymer-modulizer fixture with `npm run build:fixtures` and run the tests with `npm run test:full`)*

aomarks:
@43081j I just released 0.3.5 of the generator, which has your 3 Analyzer PRs, and support for emitting `readonly`. Want to bump this PR?

43081j:
thanks for that :)

i've bumped the version and re-generated the types but it doesn't seem to have picked up on the `DomApi` prototype methods. so either we've found a bug or that PR isn't in 0.3.5 🤔 

don't quite have time to look into it yet but will do soon

aomarks:
Just a high level note: I think we should try to get this merged soon, to make sure as many of these improvements as possible get into the 2.4 release. If there are any things we're not quite sure about, I'd suggest removing them from this PR for now and tackling them incrementally with separate PRs.

Is `gulp closure` compiling cleanly with the latest changes?

aomarks:
> i've bumped the version and re-generated the types but it doesn't seem to have picked up on the DomApi prototype methods. so either we've found a bug or that PR isn't in 0.3.5 🤔

Hmm, https://github.com/Polymer/polymer-analyzer/pull/821 is definitely in Analyzer `3.0.0-pre.7`.

BTW one way you can test this end-to-end when you're working on analyzer is to `npm link` your local analyzer into your local polymer.

43081j:
i would just like to get these dom api methods appearing and then i think it should be merged as is.

ill try get them sorted tomorrow. i had a little dig into it and it is detecting them but they just don't end up in the types.

43081j:
does it maybe make sense to add `@memberof` to the prototype methods and create the namespaced name from that?

```ts
/**
 * @function
 * @memberof Polymer
 * @return {void}
 */
DomApi.prototype.Foo;
```

would then become `Polymer.DomApi` as the class name, and `Foo` as the method name.

aomarks:
> i would just like to get these dom api methods appearing and then i think it should be merged as is.
> 
> ill try get them sorted tomorrow. i had a little dig into it and it is detecting them but they just don't end up in the types.
> 
> edit:
> 
> this is because the class name is set as Polymer.DomApi but, at the point of attaching methods to the prototype, the identifier is DomApi. this is because of @memberof

I think the simplest solution is to fully qualify the references on these declarations, instead of trying to re-assign via annotations. This seems to work for me without further changes to Analyzer:

```js
  /**
   * @function
   * @param {boolean=} deep
   * @return {!Node}
   */
  Polymer.DomApi.prototype.cloneNode;
```

produced:

```ts
class DomApi {
    ...
    cloneNode(deep?: boolean): Node;
}
```

justinfagnani:
@aomarks I agree. Also agree that this was confusing, thanks for digging in!

43081j:
updated the domapi types

lib/legacy/polymer.dom.html
```diff
@@ -235,6 +235,64 @@
...
+   * @param {string} selector
+   * @return {?Element}
+   */
+  Polymer.DomApi.prototype.querySelector;
+  /**
+   * @function
+   * @param {string} selector
+   * @return {!Array<!Element>}
+   */
+  Polymer.DomApi.prototype.querySelectorAll;
```
43081j: this should be a `NodeList` like defined [here](https://github.com/google/closure-compiler/blob/032a257f7ec86effe9171736b138dae0f3d9d86a/externs/browser/w3c_dom1.js#L436). what should we do about this? it shouldn't be an `Array`

lib/legacy/polymer.dom.html
```diff
@@ -237,70 +237,61 @@
 
   /**
    * @function
-   * @memberof Polymer
    * @param {boolean=} deep
    * @return {!Node}
    */
-  DomApi.prototype.cloneNode;
+  Polymer.DomApi.prototype.cloneNode;
```
aomarks: I think these now all need to move below the `Polymer.DomApi = DomApi;` line because it's still undefined at this point.

lib/legacy/polymer.dom.html
aomarks: I see `NodeList` in the TypeScript standard lib: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L12654 (warning: huge file)

lib/legacy/polymer.dom.html
43081j: whoops! good eyes, i should've seen that 🙈

lib/legacy/polymer.dom.html
43081j: im having trouble with this because it should be (in TS) `NodeListOf<Element>`, but in closure they seem to use `NodeList<Element>`. either way, i just end up with `any` unless i pass _just_ `NodeList` (without a `T`) :/

aomarks: Ah, I see. `gen-typescript-declarations` doesn't currently handle the general case of a type applications `Foo<Bar>`, it only handles the special cases `Array<Foo>` and `Object<Foo, Bar>`. I just filed https://github.com/Polymer/gen-typescript-declarations/issues/58 to track that.

We'll also need some special logic for converting `NodeList<Foo>` to `NodeListOf<Foo>`, since TypeScript has two different types for when it is parameterized vs not. Filed https://github.com/Polymer/gen-typescript-declarations/issues/59 for that. Although, I do wonder if the TypeScript types could be updated so that `NodeList` has a default generic type parameter of `Node` (which is possible as of TypeScript 2.3: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults).

Let's not block merging this PR on fixing those, though. It looks like TypeScript's `NodeList` effectively has `Node` as the default type, so if Closure is ok with a non-parameterized `NodeList`, then that seems ok, even if we're sure that this can only return `Elements` -- definitely better than `Array`.

aomarks:
> @43081j this is awesome! Lets make sure we confirm https://github.com/Polymer/polymer-modulizer/ doesn't break with this PR, I'll run our fixture tests later today to confirm.

We ran this PR through modulizer, and the changes look good, so I think we're clear to merge from that perspective.

43081j:
i've updated to use `NodeList` but we should remember in the associated issues to update it once we can parameterise it i think.

looks ready to merge to me now. anything extra can be done in a future PR so we dont delay this anymore

also thanks very much for all the code reviewing and what not, its taken time but seems ready finally 👍

FredKSchott:
@43081j awesome work, the new docs look great 👍 👍 👍

types/lib/utils/templatize.d.ts
```diff
@@ -16,14 +16,22 @@ declare class TemplateInstanceBase extends
   Polymer.PropertyEffects(
   Polymer.Element) {
 
+  /**
+   * Find the parent model of this template instance.  The parent model
+   * is either another templatize instance that had option `parentModel: true`,
+   * or else the host element.
+   */
+  readonly parentModel: Polymer.PropertyEffects;
+  _methodHost: Polymer.PropertyEffects;
```
TimvdLippe: It seems that we need to update our fields to make them `@private`. Extracted issue for that: https://github.com/Polymer/polymer/issues/5028

PR #5003, opened 2017-12-21
Fail travis on non updated files

TimvdLippe:
In #4928 we added support for TypeScript by auto-generating types based on the Closure types defined in the library. However, we have to remember that whenever we update our types, that we also regenerate these TypeScript types.

I browsed the man pages of `git diff` and came across `git diff --exit-code`: https://git-scm.com/docs/git-diff#git-diff---exit-code This makes `git diff` fail whenever there are changes.

As such, this PR introduces the `git diff` in the `before_script` stage on Travis. This task now runs the generation script of TypeScript types and then verifies that there are no uncommitted changes. Whenever someone updates any Closure types (or adds them with new features), the Travis build will now fail if the developer forgot to update the TypeScript types.

In the commit history of this PR, you can see that the 2nd commit intentionally failed, by updating 1 of our types and not committing the corresponding TypeScript `d.ts` updates.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_script:
 - npm install -g bower gulp-cli@1
 - bower install
 - gulp lint-eslint
+- gulp generate-typescript
 - git diff --exit-code
 script:
 - xvfb-run wct -l chrome
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_script:
 - npm install -g bower gulp-cli@1
 - bower install
 - gulp lint-eslint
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -12,6 +12,8 @@ before_script:
 - npm install -g bower gulp-cli@1
 - bower install
 - gulp lint-eslint
+- gulp generate-typescript
```
azakus: Also add `gulp generate-externs`

TimvdLippe: Done 👍 Edited on GitHub, so let's see if Travis complains

PR #5007, opened 2017-12-27
fix(legacy-element-mixin): _log with single parameter #5006

### LINKED ISSUE #5006: [bug] _log with single parameter does not show properly
### Description
When `_log` call with single string parameter (like `this._log('test')`) it does not show properly on the console!
Because _logger convert `log('test')` to `log('t', 'e', 's', 't')`!

#### Live Demo

http://jsbin.com/luhaxab/1/edit?js,console

AliMD:
### Reference Issue

Fixes #5006
Supersedes and closes #4924

### DIFF
--- a/test/unit/logging.html
+++ b/test/unit/logging.html
@@ -78,6 +78,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       assert.deepEqual(fnEl.stream, [{level: 'log', args: ['fn']}, {level: 'warn', args: ['fn']}, {level: 'error', args: ['fn']}]);
       assert.deepEqual(classEl.stream, [{level: 'log', args: ['class']}, {level: 'warn', args: ['class']}, {level: 'error', args: ['class']}]);
     });
+    test('_log with single parameter', () => {
+      const msg = 'log test';
+      let logedMsg;
+      const orgConsole = console;      
+      window.console = {
+        log: msg => {
+          logedMsg = msg
+        }
+      };
+      Polymer.Base._log(msg);
+      window.console = orgConsole;
+      assert.equal(msg, logedMsg);
+    });
     test('_logf', () => {
       let args = ['hi', 'there'];
       let output = fnEl._logf(...args);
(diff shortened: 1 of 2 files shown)


AliMD:
@TimvdLippe If I want to make a test cant use `https://github.com/Polymer/polymer/blob/master/test/unit/logging.html`
Because it's about `_logger` itself and this fn replace with another fn for test
I force to make another test unit and I don't think its necessary 
I make a simple test with overwrite console.log for make you happy ;)

AliMD:
@kevinpschaaf @azakus Could you please change the reviewer 😄

AliMD:
@TimvdLippe Any order sir?

test/unit/logging.html
```diff
@@ -85,6 +85,22 @@
       assert.includeMembers(output, args, `${args} should be in ${output}`);
     });
   });
+  suite('_logger', () => {
+    test('_log with single parameter', () => {
+      const msg = 'log test';
+      const logHistory = [];
+      const orgConsole = console;      
+      window.console = {
```
TimvdLippe: You can do the following with `sinon` (http://sinonjs.org/releases/v4.1.3/spies/):

```js
test('_log with single parameter', () => {
  const msg = 'log test';
  const spy = sinon.spy();
  const oldConsole = window.console;

  window.console = spy;
  Polymer.Base._log(msg);
  window.console = oldConsole;

  assert.isTrue(spy.calledOnce);
  assert.isTrue(spy.calledWith(msg));
});
```

AliMD: @TimvdLippe Any pros for this case?

TimvdLippe: For consistency, as we use it throughout our testing suite: https://github.com/Polymer/polymer/search?utf8=%E2%9C%93&q=sinon.spy&type=

AliMD:
@TimvdLippe Should I close this PR?
Any hope to merge?

TimvdLippe:
Please process the feedback of https://github.com/Polymer/polymer/pull/5007#discussion_r159123572 and then this is good to be merged :smile:

PR #5019, opened 2018-01-05
[1.x] Allow adding type="custom" to avoid deprecation warning. Fixes #5017

### LINKED ISSUE #5017: Allow users to avoid deprecation warning re: styles in HTML imports
Per https://github.com/Polymer/polymer/issues/4679 (changes landed in https://github.com/Polymer/polymer/pull/4849 and https://github.com/Polymer/polymer/pull/4850), we updated `<custom-style>` to automatically move itself to the main document to avoid breakage due to the [planned deprecation](https://bugs.chromium.org/p/chromium/issues/detail?id=523952) in Chrome of `<styles>` in HTML Imports applying to the main document.

However, since the Chrome deprecation warning is logged as soon as the HTML parser sees the style in the import, the deprecation warning is misleading since the style will immediately be moved to the main document when the `custom-style` upgrades, avoiding any actual problems.  This has confused users and also leads to muddied statistics about the actual impact of the deprecation.

This issue proposes to allow users to add a [`type="custom"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style#attr-type) attribute to `<style>` tags to prevent them from being parsed as `text/css`, which avoids the deprecation warning.  When `custom-style` then moves the style to the main document, it can remove the `type` attribute, allowing the style to be parsed and apply as normal.

kevinpschaaf:
Remove `type="custom"` when moving `custom-style` to main document.  Allows users to add `type="custom"` to avoid the deprecation warning.

Fixes #5017

### DIFF
--- a/src/lib/custom-style.html
+++ b/src/lib/custom-style.html
@@ -134,6 +134,11 @@ Note, all features of `custom-style` are available when defining styles as part 
       cases of ordering w.r.t the main document styles.
       */
       if (this.ownerDocument !== window.document && this.__appliedElement === this) {
+        // Allow users to add `type="custom"` to <style> tags to avoid deprecation warning
+        var type = this.getAttribute('type');
+        if (type == 'custom') {
+          this.removeAttribute('type');
+        }
         document.head.appendChild(this);
       }
       // needed becuase elements in imports do not get 'attached'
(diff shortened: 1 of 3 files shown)


src/lib/custom-style.html
```diff
@@ -134,6 +134,11 @@
       cases of ordering w.r.t the main document styles.
       */
       if (this.ownerDocument !== window.document && this.__appliedElement === this) {
+        // Allow users to add `type="custom"` to <style> tags to avoid deprecation warning
+        var type = this.getAttribute('type');
```
azakus: Let's make this unconditional to support `type="custom"` in the main page. Less footgunny.

PR #5023, opened 2018-01-07
Disallow non-templates as interpolations in Polymer.html

justinfagnani:
This is to prevent XSS vulnerabilities from accidentally interpolating unsafe user-provided values into templates that would be interpreted as HTML.

With this change, the only values that can be used to construct templates are JS template literals and other HTML templates. We're not checking that HTML templates are created from Polymer.html, but if an app is letting user-controlled values into an HTML template as HTML, it already has an XSS.

cc @arthurevans and @katejeffreys for documenting this change.


lib/utils/html-tag.html
```diff
@@ -20,20 +20,22 @@
...
+     * A template literal tag that creates an HTML <template> element from the
+     * contents of the string.
      *
      * This allows you to write a Polymer Template in JavaScript.
      *
-     * Interpolated values are converted to strings when the template is created,
-     * they are not intended as a replacement for Polymer data-binding.
+     * Templates can be composed by interpolating `HTMLTemplateElement`s in
+     * expressions in the JavaScript template literal. The nested template's
+     * `innerHTML` is inluced in the containing template.
```
graynorton: 'inluced' => 'included'

justinfagnani:
@sorvell sounds good.

I would want to offer this as another template tag so that it can only be used on literals, not on string variables, and it should disallow interpretations of plain strings.

For a function that can consume any string, I would want `unsafe` in the name, ie `Polymer.unsafeHTML()`.

justinfagnani:
Given that 2.4 got tagged already, I think we should merge this as-is for a quick 2.4.1. We can always add more supported data types without breaking changes.

lib/utils/perf-tester.js
```diff
@@ -0,0 +1,483 @@
+(function() {
+
+  'use strict';
+
+  class PerfTester extends HTMLElement {
```
TimvdLippe: @sorvell Was this class intended to be committed in this PR? If so, we need some docs on what it actually does and how we can use it (assuming you used it to measure the performance of the changes in this PR).

lib/utils/html-tag.html
```diff
@@ -12,13 +12,31 @@
   (function() {
     'use strict';
 
+    class HTMLLiteral {
```
justinfagnani: docs

lib/utils/html-tag.html
```diff
@@ -12,13 +12,31 @@
...
+      constructor(string) {
+        this.innerHTML = string;
```
justinfagnani: I wouldn't name this `innerHTML`, this isn't an element. I'd go with `value`

lib/utils/html-tag.html
```diff
@@ -59,5 +77,33 @@
...
+     *   static get styleTemplate() { return Polymer.htmlLiteral`.shadowed { background: gray; }`; }
+     *
+     * @memberof Polymer
+     * @param {!ITemplateArray} strings Constant parts of tagged template literal
+     * @param {...*} values Variable parts of tagged template literal
+     * @return {!HTMLLiteral} Constructed literal string
+     */
+    Polymer.htmlLiteral = function(strings, ...values) {
+      return new HTMLLiteral(values.reduce((acc, v, idx) =>
+        acc + literalValue(v) + strings[idx + 1], strings[0]));
```
justinfagnani: style nit: 4 space indent for wrapped expressions

lib/utils/html-tag.html
```diff
@@ -12,13 +12,31 @@
...
      * @param {*} value Object to stringify into HTML
      * @return {string} HTML stringified form of `obj`
      */
     function htmlValue(value) {
-      if (value instanceof HTMLTemplateElement) {
-        return /** @type {!HTMLTemplateElement} */(value).innerHTML;
+      if (value instanceof HTMLTemplateElement || value instanceof HTMLLiteral) {
+        return /** @type {!HTMLTemplateElement | !HTMLLiteral} */(value).innerHTML;
       } else {
         throw new Error(`non-template value passed to Polymer.html: ${value}`);
```
justinfagnani: this message should probably be updated to allow for htmlLiteral too

lib/utils/html-tag.html
```diff
@@ -12,13 +12,31 @@
   (function() {
     'use strict';
 
+    class HTMLLiteral {
+      constructor(string) {
+        this.innerHTML = string;
+      }
```
justinfagnani: Consider adding a `toString()` implementation with the literal value

lib/utils/html-tag.html
```diff
@@ -12,28 +12,48 @@
...
+     * A template literal tag that creates an HTML <template> element from the
+     * contents of the string.
      *
      * This allows you to write a Polymer Template in JavaScript.
      *
-     * Interpolated values are converted to strings when the template is created,
-     * they are not intended as a replacement for Polymer data-binding.
+     * Templates can be composed by interpolating `HTMLTemplateElement`s in
+     * expressions in the JavaScript template literal. The nested template's
+     * `innerHTML` is included in the containing template.
```
azakus: This should also include using `htmlLiteral`

lib/utils/perf-tester.js
sorvell: argh, mistakenly committed; removing.

lib/utils/html-tag.html
kevinpschaaf: It was an attempt to write less code, since both types can return their string value via `innerHTML`; fair that it's a bit short-cutty.

lib/utils/html-tag.html
azakus: Done!

alchmy:
It seems like a pretty major problem to not be able to separate the HTML templates into separate files. This is absolutely key to being able to write modular code. I would naturally think it would be a developers choice if they wanted to add these arbitrary files. It sort of seems like it's just a way to keep code within a Polymer environment. Perhaps I haven't fully understood what's involved, but if I can't use arbitrary HTML templates, it's actually a deal breaker for me in terms of Polymer vs Vanilla WC.

miztroh:
@alchmy: You may want to check out the HTML Modules discussion.

https://github.com/w3c/webcomponents/issues/645

justinfagnani:
@alchmy this isn't about not being able to break up templates, this is about not opening XSS vulnerabilities by blindly allowing user input to flow into privileged templates. You can compose templates, they just have be composed of values produced by the `html` tag.

PR #5024, opened 2018-01-08
Move absolute url logic to element-mixin

### LINKED ISSUE #4788: `resolveUrl` ignores baseURI when given absolute URL
### Description
`Polymer.ResolveUrl.resolveUrl` when called with an absolute path, will return _just_ the path without the `baseURI`. This means you can't construct paths for URLs with a different domain, or if you wanted to share around the full path URL.

I'm assuming it's just [this line](https://github.com/Polymer/polymer/blob/master/lib/utils/resolve-url.html#L30). Happy to make a PR, just want to confirm the expected behaviour.

#### Live Demo
http://jsbin.com/defave/edit?js,console

#### Steps to Reproduce
Call `Polymer.ResolveUrl.resolveUrl` with an absolute path, and any baseURI e.g.

```js
Polymer.ResolveUrl.resolveUrl('/foo', 'http://localhost');
```

#### Expected Results
Expect the absolute path appended to the baseURI e.g. `http://localhost/foo`

#### Actual Results
Just the absolute path e.g. `/foo`

### Browsers Affected
Just tested in these:

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [x] Safari 10
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions
Polymer v2.0.2 - could be more, this is just the one I'm testing it on

TimvdLippe:
This is backwards-compatible with the previous URL resolution.
Now, `Polymer.ResolveUrl.resolveUrl` correctly appends the url
to the basePath.

### Reference Issue
Fixes #4788

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -638,6 +638,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * @return {string} Rewritten URL relative to base
        */
       resolveUrl(url, base) {
+        // Preserve backward compatibility with `this.resolveUrl('/foo')` resolving
+        // against the main document per #2448
         if (url && ABS_URL.test(url)) {
           return url;
         }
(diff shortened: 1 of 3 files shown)


TimvdLippe:
Ah yes, thanks!

PR #5029, opened 2018-01-10
Gestures: fall back to event.target when composedPath is empty.

### LINKED ISSUE #5030: Gestures crashes when composedPath is empty
### Description
Found via https://github.com/Polymer/polymer-decorators/issues/27.

In Safari, with a Gestures `tap` listener registered, dispatching a `click` event on `window` will cause the gestures library to crash, because the `composedPath()` array is empty. Seems like a Safari bug.

#### Live Demo

https://jsfiddle.net/qj04909d/2/

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [x] Safari 11
- [ ] IE 11

### Versions

- Polymer: v2.3.1
- webcomponents: vX.X.X

aomarks:
This came up in Safari, where a click event on window ended up crashing gestures. No original target could be found, because `event.composedPath()` returned an empty array, which we assumed couldn't happen.

Found via https://github.com/Polymer/polymer-decorators/issues/27

Fixes #5030

### DIFF
--- a/lib/utils/gestures.html
+++ b/lib/utils/gestures.html
@@ -316,8 +316,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     _findOriginalTarget: function(ev) {
       // shadowdom
       if (ev.composedPath) {
-        const target = /** @type {EventTarget} */(ev.composedPath()[0]);
-        return target;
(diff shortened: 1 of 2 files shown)


aomarks:
PTAL, added regression test.

PR #5034, opened 2018-01-10
add correct return type for `querySelectorAll`

43081j:
Just a quick fix now that we have a new version of the generator released.

Changing `NodeList` to the correct parameterised type, `NodeList<Node>`.

### DIFF
--- a/lib/legacy/polymer.dom.html
+++ b/lib/legacy/polymer.dom.html
@@ -345,7 +345,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   /**
    * @function
    * @param {string} selector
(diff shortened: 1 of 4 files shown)


lib/legacy/polymer.dom.html
```diff
@@ -345,7 +345,7 @@
   /**
    * @function
    * @param {string} selector
-   * @return {!NodeList}
+   * @return {!NodeList<!Node>}
```
aomarks: Should be `!NodeList<!Element>` no?

43081j: that was a close one! 🙈

PR #5035, opened 2018-01-10
Allow templatizer to be used without owner or host prop forwarding. Fixes #4458

### LINKED ISSUE #3422: Error when calling set or notifyPath on a two-way bound property from within dom-template
Reproducible test case can be found here: http://jsbin.com/dunikitije/edit?html,console,output

Possibly related to issue #3307

### LINKED ISSUE #4458: Polymer.Templatize issues.
I'm upgrading to Polymer 2.0 and in that process decided to change a component from using Templatizer behaviour to using Poymer.Templatize, which seems to be the new thing to use... but I am having issues getting it to work as expected.

1) The example in the docs fails outright due to the options object being undefined.
https://www.polymer-project.org/2.0/docs/api/namespaces/Polymer.Templatize

`Polymer.Templatize.templatize(template)` errors due to no options object existing,
instead if I pass an empty options object (and owner/this) then it works. Eg: `Polymer.Templatize.templatize(template, this, {})`.
Should the docs be fixed? or should Templatize deal with undefined options?

2) The initial data values are not reflected in the generated template. Am I misunderstanding how this is supposed to work? Setting the values on the instance directly works, but not the initial values!?
For example this won't output 'initial' as expected:
```
let instance = new TemplateClass({myProp: 'initial'});
this.shadowRoot.appendChild(instance.root);
```
but then if you do the following it will output 'new value' correctly:
`instance.myProp = 'new value';`

JSBin example: https://jsbin.com/zimewuf/edit?html,output

PS. Templatize in Polymer.Temp**al**tize.templatize is misspelled in the docs.

Chrome 58
Polymer: 2.0-preview

kevinpschaaf:
Makes templatizer safe to be used without an owner (only needed when forwarding props to/from host via `options.forwardHostProp` and automatically hiding/showing children if the owner is hidden).  If no owner is used, default template processing and binding evaluation from `PropertyEffects` will be used (normally it delegates to any overrides in the owner).  Also no-ops `dispatchEvent` so that path notifications when not forwarding to a host does not throw.

In this change, filtering the user `props` bag on `options.instanceProps` was removed since it seems not strictly required.

### Reference Issue
Fixes #4458
Fixes #3422

### DIFF
--- a/lib/utils/templatize.html
+++ b/lib/utils/templatize.html
@@ -102,6 +102,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             this._setPendingProperty(hprop, this.__dataHost['_host_' + hprop]);
           }
         }
+        // Any instance props passed in the constructor will overwrite host props;
+        // normally this would be a user error but we don't specifically filter them
         for (let iprop in props) {
           this._setPendingProperty(iprop, props[iprop]);
         }
@@ -391,11 +393,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      *     // Customize property forwarding and event model decoration
      *     let TemplateClass = Polymer.Templatize.templatize(template, this, {
      *       parentModel: true,
-     *       instanceProps: {...},
      *       forwardHostProp(property, value) {...},
+     *       instanceProps: {...},
      *       notifyInstanceProp(instance, property, value) {...},
      *     });
-     *
      *
      * @namespace
      * @memberof Polymer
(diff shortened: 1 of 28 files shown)


kevinpschaaf:
@TimvdLippe See https://github.com/Polymer/polymer/issues/4458#issuecomment-356697713, basically this PR makes it work with only one argument, the other two are now actually optional where they were not before.  Whoops, I didn't mark the 2nd `owner` argument optional; making that change now.

lib/utils/templatize.html
```diff
@@ -96,15 +97,13 @@
...
-              this._setPendingProperty(iprop, props[iprop]);
-            }
+        if (options.forwardHostProp) {
+          for (let hprop in this.__hostProps) {
+            this._setPendingProperty(hprop, this.__dataHost['_host_' + hprop]);
           }
         }
-        for (let hprop in this.__hostProps) {
-          this._setPendingProperty(hprop, this.__dataHost['_host_' + hprop]);
+        for (let iprop in props) {
```
sorvell: Let's add a comment saying that passing in a prop in `props` that has the same name as a bound host prop may shadow it. This is considered a user error (warn?). We should add this to the docs too.

TimvdLippe:
Could we check if this PR also fixes #5063? Maybe that can be fixed at the same time. If not, we have to separate the work.

PR #5036, opened 2018-01-10
Update mixinBehaviors annotation. Behaviors don't satisfy PolymerInit.

aomarks:
Behaviors don't have an `is`, so this type annotation was slightly off. I think what we really want is something like `PolymerInit` minus the `is`, but for now this at least makes things compile.

### DIFF
--- a/lib/legacy/class.html
+++ b/lib/legacy/class.html
@@ -34,7 +34,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * the underlying element.
      *
      * @template T
-     * @param {!(PolymerInit|Array<!PolymerInit>)} behaviors Behavior object or array of behaviors.
+     * @param {!Object|!Array<!Object>} behaviors Behavior object or array of behaviors.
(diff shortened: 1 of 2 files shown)


aomarks:
> Would the intersection of PolymerInit and {is?: undefined} work maybe?

We don't have a way to express that in Closure, and it doesn't seem to work in TypeScript anyway (I think the stricter `is` wins in the intersection). I think we'd need to just write two interfaces or something.

justinfagnani:
> it doesn't seem to work in TypeScript anyway (I think the stricter is wins in the intersection)

An intersection of types has unions of shared members, since the type has to satisfy both interfaces.

justinfagnani:
Can you make `is` optional on `PolymerInit`?

aomarks:
> Can you make is optional on PolymerInit?

It's maybe not optional in some cases, and there might be code that would have to be updated to check for undefined to make Closure happy. So I think we just need two interfaces.

For now just going to revert this back to `object`, which is what we had before.

PR #5038, opened 2018-01-11
Port keyframe name fix from ShadyCSS

### LINKED ISSUE #3475: CSS keyframes reference incorrectly renamed when prefixed by another keyframes set
### Description

If keyframes `slide` and `slide-back` are defined in a element, the `slide-back` keyframes declaration are renamed to `@keyframes slide-back-x-el-0 {...}`, but the animation property is incorrectly renamed to `animation: slide-x-el-0-back`.
#### Live Demo

https://jsbin.com/vidaxokeho/edit?html,output
### Browsers Affected

- [x] Chrome
### Versions

- Polymer: v1.3.0

azakus:
Fixes #3475

### DIFF
--- a/src/lib/style-properties.html
+++ b/src/lib/style-properties.html
@@ -419,7 +419,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       // Transforms `@keyframes` names to be unique for the current host.
       // Example: @keyframes foo-anim -> @keyframes foo-anim-x-foo-0
       _scopeKeyframes: function(rule, scopeId) {
-        rule.keyframesNameRx = new RegExp(rule.keyframesName, 'g');
+        // Animation names are of the form [\w-], so ensure that the name regex does not partially apply
+        // to similarly named keyframe names by checking for a word boundary at the beginning and
+        // a non-word boundary or `-` at the end.
+        rule.keyframesNameRx = new RegExp('\\b' + rule.keyframesName + '(?!\\B|-)', 'g');
         rule.transformedKeyframesName = rule.keyframesName + '-' + scopeId;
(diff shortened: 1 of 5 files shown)


src/lib/style-properties.html
```diff
@@ -419,7 +419,10 @@
       // Transforms `@keyframes` names to be unique for the current host.
       // Example: @keyframes foo-anim -> @keyframes foo-anim-x-foo-0
       _scopeKeyframes: function(rule, scopeId) {
-        rule.keyframesNameRx = new RegExp(rule.keyframesName, 'g');
+        // Animation names are of the form [\w-], so ensure that the name regex does not partially apply
+        // to similarly named keyframe names by checking for a word boundary at the beginning and
+        // a non-word boundary or `-` at the end.
+        rule.keyframesNameRx = new RegExp('\\b' + rule.keyframesName + '(?!\\B|-)', 'g');
```
sorvell: It looks like `foo` will match in `a-foo`

azakus: For the prefix issue, looks like we'll need something more complicated like `(?:animation-name\s*:|@(?:-moz|-webkit|-ms)?-?keyframes)\s*a-foo(?!\B|-)`

azakus: After reviewing this, it turns out not to matter if `foo` matches `a-foo`. We're only looking to add a suffix scoping name to the keyframe, and the regex fix makes sure that that suffix scoping is only added to animations already without a suffix.

The bug here was specifically if the match was found as a prefix for an animation, such that `a` would match `a-foo` and incorrectly add the scoping suffix as an infix modification, breaking the naming scheme.

Therefore, if `a-foo` is matched for `foo`, it will no longer match for `a-foo`, and all animations will have the scoping suffix added correctly.

PR #5039, opened 2018-01-12
Fix label tap by checking matched label pairs

### LINKED ISSUE #4717: iOS native label -> checkbox tap doesn't always toggle check status
### Description
Having a LABEL linked with an INPUT CHECKBOX via native FOR attribute, including Polymer 1.9.1/1.9.2, tapping on label doesn't always toggle checkboxes status.

#### Live Demo
https://jsfiddle.net/t2wr2cob/

#### Steps to Reproduce
Visit the above on an iOS device, I tested on iPad 10.3, iPhone 10.3
Tap on the 'Native checkbox' label a few times, it often works the first few times, then you'll see the label highlight as it received the tap, but the checkbox doesn't toggle.
Occasionally it worked every time, I re-ran the fiddle and it would break again.

If you disable polymer 1.9.2 from being loaded, but leave webcomponents-lite loading, problem doesn't exist. It's only when you include polymer.html#v1.9.2 that it breaks.

#### Expected Results
Checkbox toggles on tap

#### Actual Results
Checkbox often does not toggle on tap

### Browsers Affected
- [x] Mobile Safari 10

### Versions
- Polymer: v1.9.2
- webcomponents: v1.9.2

azakus:
Fixes #4717

### DIFF
--- a/lib/utils/gestures.html
+++ b/lib/utils/gestures.html
@@ -97,9 +97,58 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   /** @type {(function(MouseEvent): void | undefined)} */
   GestureRecognizer.prototype.click;
 
-  // keep track of any labels hit by tghe mouseCanceller
+  // keep track of any labels hit by the mouseCanceller
   /** @type {!Array<!HTMLLabelElement>} */
   const clickedLabels = [];
+
+  /** @type {!Map<string, boolean>} */
+  const labellable = {
+    'button': true,
+    'input': true,
+    'keygen': true,
+    'meter': true,
+    'output': true,
+    'textarea': true,
+    'progress': true,
+    'select': true
+  };
+
+  /**
+   * @param {HTMLElement} el Element to check labelling status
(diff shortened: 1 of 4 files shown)


TimvdLippe:
@azakus Is this PR mergeable with just the smoke test or would you like some unit tests for this one as well?

azakus:
Added tests

lib/utils/gestures.html
```diff
@@ -97,6 +97,55 @@
...
+  /**
+   * @param {HTMLElement} el Element that may be labelled.
+   * @return {!Array<!HTMLLabelElement>} Relevant label for `el`
+   */
+  function matchingLabels(el) {
+    /** @type {!Array<!HTMLLabelElement>} */
+    let labels = el.labels;
+    // IE doesn't have `labels` and Safari doesn't populate `labels`
+    // if element is in a shadowroot.
+    // In this instance, finding the non-ancestor labels is enough,
```
sorvell: Let's test that case.

azakus: Done

PR #5041, opened 2018-01-13
No need of this line

awesomehimanshu:
let x = y;
y=x ;
Why to use next line there will be no difference .

### DIFF
--- a/lib/utils/settings.html
+++ b/lib/utils/settings.html
@@ -78,6 +78,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    */
 
   /**
+   * Sets the global sanitizeDOMValue available via `Polymer.sanitizeDOMValue`.
    *
    * @memberof Polymer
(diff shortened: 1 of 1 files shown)


TimvdLippe:
As stated in the comment: `This is needed for tooling`. I don't think we can remove this, can we @FredKSchott ?

awesomehimanshu:
It doesn't make any sense to do that .

TimvdLippe:
From a JavaScript point of view, it is not necessary. However, our tooling (primarily `polymer-analyzer`) can not understand all usages. Therefore we have to standardize how we treat declarations of variables, like this one, such that the tooling can understand it. These assignments are not needed for runtime, but are required for static analysis.

kevinpschaaf:
Please see https://github.com/Polymer/polymer-modulizer#overview.

> It converts assignments to global namespace objects to module exports, and references to namespaces as imported names.

Assignments to global namespace objects become exports when converting to ES modules.  When Polymer 3.0 is released, module source will become the source of truth, so code like this is facilitating a migration path to ES modules while the source of truth is still in HTML imports.

PR #5043, opened 2018-01-15
Remove unnecessary limit check

MajorBreakfast:
https://github.com/Polymer/polymer/blob/master/lib/elements/dom-repeat.html#L555-L556

```JS
const limit = Math.min(isntIdxToItemsIdx.length, this.__limit);
      for (; instIdx<limit; instIdx++) {
```

So, for each interation of the loop this holds:
```JS
instIdx < Math.min(isntIdxToItemsIdx.length, this.__limit)
```

Which implies:
```JS
instIdx < this.__limit
```

Consequently `&& instIdx < this.__limit` is essentially the same as `&& true` and can be left out.

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -558,7 +558,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         let itemIdx = isntIdxToItemsIdx[instIdx];
         let item = items[itemIdx];
         itemsIdxToInstIdx[itemIdx] = instIdx;
-        if (inst && instIdx < this.__limit) {
+        if (inst) {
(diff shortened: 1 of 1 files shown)


MajorBreakfast:
I stumbled over this because the `else` branch creates new instances. It makes no sense to create new instances when the limit is exceeded. But, the check is redundant and thus this never happens.

MajorBreakfast:
@TimvdLippe I (and, it seems, Google also) have never heard "SFTM" before :) What does it mean? :D 

I think the CI is currently a bit broken: "Error response status: 6 Selenium error: no such session"

MajorBreakfast:
Ah, I figured that it must mean something along those lines :)

PR #5044, opened 2018-01-15
dom-repeat accepts template provided by slot

MajorBreakfast:
Makes this work:

```HTML
<my-repeater items="{{items}}">
  <template>[[item]]</template>
</my-repeater>
```

```HTML
<!-- Template for `<my-repeater>` uses a `<slot>`-->
<dom-repeat items="{{items}}">
  <slot></slot>
</dom-repeat>
```

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -323,16 +323,29 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
     }
 
+    __getTemplate () {
+      for (let el = this.firstElementChild; el; el = el.nextElementSibling) {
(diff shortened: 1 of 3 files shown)


MajorBreakfast:
I've now changed it so that it uses a `for...of` loop to iterate the `assignedNodes` array. That's a lot nicer in code, but am I allowed to use it?

sorvell:
We're not inclined to accept this PR for the reasons you suggest. The difficulty styling the elements and the lack of dynamism makes this somewhat forced. Instead, we agree that using templatize to handle a specific use case is probably a better call. Nonetheless, thanks for the effort and discussion.

MajorBreakfast:
@sorvell 👍

PR #5048, opened 2018-01-16
Fix incorrect path modification in dom-repeat __handleObservedPaths()

### LINKED ISSUE #4983: Bug in observe property handling in dom-repeat (__handleObservedPaths)
### Description

Hey,

I think I found a bug in handling of the dom-repeat template's observe property.

If the observe property should observe a path to an item's subproperty, such as this:
```
[
    {
        mainObject: { 
            subObject: {
                value: null
            }
        }
    }
]
```
The following doesn't work (leaving out nonessential attributes), even though it should:
`<template is="dom-repeat" observe="mainObject.subObject.value">`
Omiting the first-level subproperty does the trick, but I don't think this is the expected behavior:
`<template is="dom-repeat" observe="subObject.value">`

I believe the issue lies here:

**polymer/lib/elements/dom-repeat.html:**
```
    __handleObservedPaths(path) {
      // Handle cases where path changes should cause a re-sort/filter
      if (this.__sortFn || this.__filterFn) {
        if (!path) {
          // Always re-render if the item itself changed
          this.__debounceRender(this.__render, this.delay);
        } else if (this.__observePaths) {
          // Otherwise, re-render if the path changed matches an observed path
          path = path.substring(path.indexOf('.') + 1);
          let paths = this.__observePaths;
          for (let i=0; i<paths.length; i++) {
            if (paths[i].indexOf(path) === 0) {
              this.__debounceRender(this.__render, this.delay);
            }
          }
        }
      }
    }
```

Specifically here:
`path = path.substring(path.indexOf('.') + 1);`
This line removes the first object from the notified item path, changing `mainObject.subObject.value` from the example above to `subObject.value`, which is then never found in the for loop, as it compares the shortened path to paths from `observe="mainObject.subObject.value"` property.

I hope the live demo helps:

#### Live Demo
http://jsbin.com/haqonocufo/1/edit?html,output

#### Steps to Reproduce
Create array of complex objects, bind the array to dom-repeat template, create a filter function, set a subproperty to observe path of the subproperty and change the subproperty after initial render.

#### Expected Results
Template gets re-rendered because of the observed subproperty change.

#### Actual Results
Template doesn't get re-rendered.

### Versions
- Polymer: v2.3.1
- webcomponents: v1.0.20

MajorBreakfast:
Fixes #4983

The path modification is wrong. The path is already modified in `__handleItemPath()`
https://github.com/Polymer/polymer/blob/master/lib/elements/dom-repeat.html#L649

This causes the sort and filter to fail if a nested property is modified, e.g. `'items.1.prop.nestedProp'`. The current code gets the already modified `'prop.nestedProp'` and incorrectly modifies it to `nestedProp`.

@kevinpschaaf

### DIFF
--- a/test/unit/dom-repeat-elements.html
+++ b/test/unit/dom-repeat-elements.html
@@ -367,6 +367,32 @@ window.getData = () => [
   });
 </script>
 
+<dom-module id="x-repeat-filter-and-sort-by-nested-property">
+    <template>
+      <template id="repeater" is="dom-repeat" items="{{items}}">
+        <x-foo itema-prop="{{item.prop.nestedProp}}"></x-foo>
+      </template>
+    </template>
+    <script>
+    Polymer({
+      is: 'x-repeat-filter-and-sort-by-nested-property',
+      properties: {
+        items: {
+          value () {
+            return [
+              { prop: { nestedProp: 0 } },
+              { prop: { nestedProp: 1 } },
+              { prop: { nestedProp: 2 } },
+              { prop: { nestedProp: 3 } },
+              { prop: { nestedProp: 4 } },
+            ];
+          }
+        }
+      }
+    });
+    </script>
+  </dom-module>
+
 
 <dom-module id="x-simple-repeat">
   <template>
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -473,7 +473,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           this.__debounceRender(this.__render, this.delay);
         } else if (this.__observePaths) {
           // Otherwise, re-render if the path changed matches an observed path
-          path = path.substring(path.indexOf('.') + 1);
           let paths = this.__observePaths;
           for (let i=0; i<paths.length; i++) {
             if (path.indexOf(paths[i]) === 0) {
--- a/test/unit/dom-repeat.html
+++ b/test/unit/dom-repeat.html
@@ -43,6 +43,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     </template>
   </test-fixture>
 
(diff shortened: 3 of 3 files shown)


TimvdLippe:
Please add a regression test so we can verify the original issue remains fixed.

MajorBreakfast:
There you go.

BTW the test suite for dom-repeat is over 4000 lines and of code. Would it be alright if I broke it up into multiple files in another PR? (not today, but sometime in the future) I'm thinking of one file per custom component and one file per suite. I'm not a big fan of long files because it leads me to waste a lot of time trying to find where I just left off.

test/unit/dom-repeat-elements.html
```diff
@@ -367,6 +367,30 @@
...
+      <template id="repeater" is="dom-repeat" items="{{items}}">
+        <x-foo itema-prop="{{item.prop.nestedProp}}"></x-foo>
+      </template>
+    </template>
+    <script>
+    Polymer({
+      is: 'x-repeat-filter-and-sort-by-nested-property',
+      properties: {
+        items: {
+          value: [
```
TimvdLippe: This should be `function() { return [...]; }`

MajorBreakfast: May I use ES2015 style `value () { return [] }`?

MajorBreakfast:
@TimvdLippe I was thinking about a folder ;)

MajorBreakfast:
```
safari 11.0.2            Tests passed
firefox 57               Tests passed
chrome 62                Tests passed
```

MajorBreakfast:
Ready to merge: It fixes a bug in Polymer that the tests previously didn't cover

MajorBreakfast:
@TimvdLippe Ah okay. Thx for the update :)

PR #5049, opened 2018-01-16
Use sudo to fix chrome testing

azakus:
There is a bug with the chrome sandbox in the Travis CI container
system, which is fixed by using `sudo: required`.

Related to travis-ci/travis-ci#8836


azakus:
Closing in favor of #5050

PR #5051, opened 2018-01-18
Fix issue where el.splice could not clear full array

### LINKED ISSUE #5032: Omitting deleteCount, this.splice() acts different from array.splice()
### Description

Should this.splice('thisArray', 0) remove all entries? _"If deleteCount is omitted ... then all of the elements from start through the end of the array will be deleted."_ https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

#### Live Demo

http://jsbin.com/wogogaseya/1/edit?html,output

#### Steps to Reproduce

        let arrgh = [1,2,3];
        this.arrgh = [1,2,3];
        arrgh.splice(0);
        this.splice('arrgh',0);
        console.log('Lengths:',arrgh.length,this.arrgh.length);

// Lengths: 0 3

https://github.com/Polymer/polymer/blob/master/lib/mixins/property-effects.html#L1942

#### Expected Results
Lengths: 0 0

#### Actual Results
Lengths: 0 3

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Safari 9
- [ ] Safari 8
- [ ] IE 11

### Versions

- Polymer: vX.X.X
- webcomponents: vX.X.X

TimvdLippe:
The gist of the issue is that, depending on the number of arguments passed on to `array.splice`, splice does different things. We were eagerly setting `start` to 0 and passing all 3 arguments. Instead, check which case we actually have and fix the case when only `start` is passed in.
### Reference Issue
Fixes #5032

### DIFF
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -1946,7 +1946,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         //
         let ret;
         // Omit any additional arguments if they were not passed in
-        if (start !== undefined && !deleteCount && !items.length) {
+        if (arguments.length === 2) {
           ret = array.splice(start);
         // Either start was undefined and the others were defined, but in this
         // case we can safely pass on all arguments
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -1933,13 +1933,33 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // Normalize fancy native splice handling of crazy start values
         if (start < 0) {
           start = array.length - Math.floor(-start);
-        } else {
+        } else if (start) {
           start = Math.floor(start);
         }
-        if (!start) {
-          start = 0;
+        // array.splice does different things based on the number of arguments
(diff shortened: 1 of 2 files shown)


lib/mixins/property-effects.html
```diff
@@ -1933,13 +1933,33 @@
...
+        // you pass in. Therefore, array.splice(0) and array.splice(0, undefined)
+        // do different things. In the former, the whole array is cleared. In the
+        // latter, no items are removed.
+        // This means that we need to detect whether 1. one of the arguments
+        // is actually passed in and then 2. determine how many arguments
+        // we should pass on to the native array.splice
+        //
+        let ret;
+        // Omit any additional arguments if they were not passed in
+        if (start !== undefined && !deleteCount && !items.length) {
```
sorvell: For simplicity, let's just check if `arguments.length ==2`

TimvdLippe: Fixed :+1:

TimvdLippe:
@sorvell I have updated the if-condition. Could you approve this PR?

PR #5058, opened 2018-01-22
Delete package-lock.json

limonte:
`package-lock.json` is a joke:

- https://stackoverflow.com/q/45022048/1331425
- https://github.com/npm/npm/issues/17979

`yarn.lock` works just perfectly if you need a lock file. Using `yarn.lock` would make sense since v3 will be based on Yarn, not on npm.


azakus:
Let's add the package-lock.json to gitignore as well.

web-devel:
https://github.com/npm/npm/issues/17979#issuecomment-332701215

limonte:
> Let's add the package-lock.json to gitignore as well.

Done.

PR #5059, opened 2018-01-24
Simplify code for <dom-repeat>'s `sort` and `filter` properties

MajorBreakfast:
This change removes two ternary operators that spanned two lines and replaces them with less complicated code that does the same thing.

### DIFF
--- a/lib/elements/dom-repeat.html
+++ b/lib/elements/dom-repeat.html
@@ -392,22 +392,24 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       return this.__dataHost._methodHost || this.__dataHost;
     }
 
-    __sortChanged(sort) {
-      let methodHost = this.__getMethodHost();
-      this.__sortFn = sort && (typeof sort == 'function' ? sort :
-        function() { return methodHost[sort].apply(methodHost, arguments); });
-      if (this.items) {
-        this.__debounceRender(this.__render);
+    __functionFromPropertyValue(functionOrMethodName) {
+      if (typeof functionOrMethodName === 'string') {
+        let methodName = functionOrMethodName;
+        let obj = this.__getMethodHost();
+        return function() { return obj[methodName].apply(obj, arguments); };
       }
+
+      return functionOrMethodName;
+    }
+
+    __sortChanged(sort) {
+      this.__sortFn = this.__functionFromPropertyValue(sort);
+      if (this.items) { this.__debounceRender(this.__render); }
     }
 
     __filterChanged(filter) {
-      let methodHost = this.__getMethodHost();
-      this.__filterFn = filter && (typeof filter == 'function' ? filter :
-        function() { return methodHost[filter].apply(methodHost, arguments); });
-      if (this.items) {
-        this.__debounceRender(this.__render);
-      }
+      this.__filterFn = this.__functionFromPropertyValue(filter);
+      if (this.items) { this.__debounceRender(this.__render); }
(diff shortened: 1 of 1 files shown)


MajorBreakfast:
Yay, the CI works again 😀

lib/elements/dom-repeat.html
```diff
@@ -392,22 +392,26 @@
...
 
-    __sortChanged(sort) {
-      let methodHost = this.__getMethodHost();
-      this.__sortFn = sort && (typeof sort == 'function' ? sort :
-        function() { return methodHost[sort].apply(methodHost, arguments); });
-      if (this.items) {
-        this.__debounceRender(this.__render);
+    __sortChanged(fn) {
+      if (typeof fn === 'string') {
+        let methodName = fn, obj = this.__getMethodHost();
```
merlinnot: `const`?

lib/elements/dom-repeat.html
```diff
@@ -392,22 +392,26 @@
...
 
-    __filterChanged(filter) {
-      let methodHost = this.__getMethodHost();
-      this.__filterFn = filter && (typeof filter == 'function' ? filter :
-        function() { return methodHost[filter].apply(methodHost, arguments); });
-      if (this.items) {
-        this.__debounceRender(this.__render);
+    __filterChanged(fn) {
+      if (typeof fn === 'string') {
+        let methodName = fn, obj = this.__getMethodHost();
```
merlinnot: `const`?

MajorBreakfast: I totally agree. This should be const. However, `const` makes the closure compiler scream:
```
polymer.html_script_36.js:381: WARNING - All constants must be typed. The compiler could not infer the type of constant obj. Please use an explicit type annotation. For more information, see:
https://github.com/google/closure-compiler/wiki/Using-NTI-(new-type-inference)#warnings-about-uninferred-constants
        const methodName = fn, obj = this.__getMethodHost();
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
The same warning of course also for `methodName`

MajorBreakfast:
Now I've extracted the shared logic into a helper method. I've decided to call it `__functionFromPropertyValue()` because that was the most descriptive name I came up with.

merlinnot:
@TimvdLippe We don't have to assign anything to a property, we just have to use a property to identify property on `this`, like so (pseudocode):
```js
    __functionFromPropertyValue(functionOrMethodName) {
      if (typeof functionOrMethodName === 'string') {
        const methodName = functionOrMethodName;
        const obj = this.__getMethodHost();
        return function() { return obj[methodName].apply(obj, arguments); };
      }
      return functionOrMethodName;
    }
    __propertyChanged(which, to) {
      this[which] = this.__functionFromPropertyValue(to);
      if (this.items) { this.__debounceRender(this.__render); }
    }
    __sortChanged(sort) {
      this__propertyChanged('__sortFn', sort);
    }
    __filterChanged(filter) {
      this__propertyChanged('__filterFn', filter);
    }
```

What do you think guys? I've also changed `let`s to `const`s ❤️

MajorBreakfast:
@merlinnot I considered that, IMO makes it more complicated to read and understand again.

MajorBreakfast:
About the ˋconstˋ thing: Let‘s hear what the others say. One solution would be to turn the warning off. Depends on the reason why it was put in.

MajorBreakfast:
👍

PR #5062, opened 2018-01-25
Use javascript string escaping in Polymer.html

### LINKED ISSUE #5060: html-tag.js is using the .raw version of strings; y tho?
### Description
Consider the template string literal assignment from a recently modulized element iron-a11y-announcer:
```js
html`<div aria-live\$="[[mode]]">[[_text]]</div>`
```

Note the bound attribute `aria-live$=` is represented here with an escaped `\$` instead of just the `$`.  This is because of a naive, but valid, escaping that takes place when the original html file was ported into the javascript templated string literal in polymer-modulizer.

Polymer `html-tag.js`, however, uses the `.raw` value of this literal, resulting in its injection of the escaped character into the DOM, giving everyone binding to attributes the following experience with the current modulized code: https://github.com/Polymer/polymer-modulizer/issues/292

Here's the html-tag code: https://github.com/Polymer/polymer/blob/ee1d71f5f896ee57a0ee6d1f377ae6caa76e556b/lib/utils/html-tag.html#L57

Can we be clear there's a reason for the `html()` function to use `.raw`?  Because I don't see a case where you'd want javascript-escape sequences to be literally interpreted when injecting HTML.  Is this for convenience of embedding JS source in the template?  I think it just makes things even harder to escape inside the embedded JS source then too as a consequence...

azakus:
There's a lot of pitfalls

Fixes #5060

### DIFF
--- a/lib/utils/html-tag.html
+++ b/lib/utils/html-tag.html
@@ -52,12 +52,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @return {!HTMLTemplateElement} Constructed HTMLTemplateElement
      */
     Polymer.html = function html(strings, ...values) {
-      // use raw strings to preserve literal escapes in strings
-      /** @type {!Array<string>} */
-      const rawStrings = strings.raw;
       const template = /** @type {!HTMLTemplateElement} */(document.createElement('template'));
       template.innerHTML = values.reduce((acc, v, idx) =>
-        acc + htmlValue(v) + rawStrings[idx + 1], rawStrings[0]);
+        acc + htmlValue(v) + strings[idx + 1], strings[0]);
       return template;
     };
   })();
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
Went over the tradeoffs with a few people in the bullpen. There are a few considerations.

Pros for using `raw`:
* Can mostly use HTML escaping rules, not JS escaping rules (e.g., you can safely write e.g. `<div>Are you happy (yes\no)?</div>` in the HTML without escaping the `\` since the `\` means nothing special in HTML)

Cons for using `raw`:
* You can't get around needing to escape a couple of JS-specific characters in the HTML, even when using `raw`: e.g.  `` ` `` and `$` when preceeding a `{` still needs to be escaped when used in HTML to prevent closing the string or starting a part.  For example, a common footgun would be `<div>The price is ${{price}}</div>`; the `$` needs to be escaped to `\${{price}}` per JS rules.  
* The `\` required to escape the backtick and dollar signs per the above would need to be post-processed out of the `raw` text, since they still show up in the `raw` output, making the processing in the `html` tag function more complex
* Since most users are familiar with backticked-strings conforming to JS escaping rules, having mostly non-JS escaping rules, but needing to remember a couple of them just sounds confusing

usergenic:
👏

PR #5068, opened 2018-01-27
Fix issue with observers being called twice

### LINKED ISSUE #5067: Observers called multiple times in inheritance chain
### Description
I have two classes, `base-element` defines an observer and `my-element` extends `base-element` with no own properties. Both are placed in the dom with the property set. The observer of just `base-element` is called once. The observer of `my-element` (which was inherited from `base-element`) is called twice.

http://jsbin.com/wimozeravo/1/edit?html,console,output

Adding an empty properties set to `my-element` makes the observer fire only once:

http://jsbin.com/giniqeweco/1/edit?html,console,output

It seems that when a class in an inheritance chain defines a property observer, for each class in the inheritance chain that does not define a properties getter, the property observer is called once extra. 

This can be illustrated with mixins (which are just classes).

Here there are two mixins and a class. The first mixin defines an observer and the others don't define any properties. The observer is fired thrice: 

http://jsbin.com/hilularuce/1/edit?html,console,output

With empty properties it's fired once:

http://jsbin.com/domakayufu/1/edit?html,console,output

TimvdLippe:
Previously, we were checking whether the class actually had the property (https://github.com/Polymer/polymer/pull/4840/files#diff-41695d1bc0c64b86f0e5a7878f58df0fL118), however, in the changed version we were not. This change makes sure that if a class does not define a `properties` getter, it does not traverse up the prototype chain and retrieve the `properties` block from the superClass (thus processing all properties twice).
### Reference Issue
Fixes #5067


lib/mixins/properties-mixin.html
```diff
@@ -87,10 +88,14 @@
...
      */
     function ownProperties(constructor) {
-      if (!constructor.hasOwnProperty(
-        JSCompiler_renameProperty('__ownProperties', constructor))) {
-        const props = constructor.properties;
-        constructor.__ownProperties = props ? normalizeProperties(props) : null;
+      if (!constructor.hasOwnProperty(JSCompiler_renameProperty('__ownProperties', constructor))) {
+        let props = null;
+
+        if (constructor.hasOwnProperty('properties') && constructor.properties) {
```
sorvell: I imagine we probably need `JSCompiler_renameProperty` for `properties`, don't we?

TimvdLippe: The weird part is, with or without produces the same output. Nonetheless, I have added the line.

Westbrook:
Is this the sort of fix that we could get a 2.4.1 release out of? Possibly soon?

PR #5070, opened 2018-01-29
Use containers for testing again

azakus:
Disable chrome security sandbox due to travis-ci/travis-ci#8836

Remove node module caching with the removal of package lock

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ node_js: stable
 addons:
   firefox: latest
   chrome: stable
+cache:
+  directories:
+  - node_modules
 before_script:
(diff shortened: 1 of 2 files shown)


azakus:
If we're not using package-lock, I'm not sure having the npm cache is a great idea. The only way the installed modules will update now is if we push updates to package.json.

TimvdLippe:
Before npm introduced the package lock, we were using the cache as well. I don't recall any problems back then. Per https://docs.travis-ci.com/user/caching/#Things-not-to-cache Travis advises to use the cache especially for npm.

azakus:
SGTM

Issue #5069, opened 2018-01-29
Uncaught TypeError: this.keyEventTarget.addEventListener is not a function


lpellegr:
Since an update to Polymer 2.4.0 I am getting the following error on a page that makes use of paper elements and iron-a11y-keys.

Reverting to Polymer 2.3.1 solves the issue.

Unfortunately, I am not able to reproduce in a minimal use case for now.

Here is the error if it can help:

> Uncaught TypeError: this.keyEventTarget.addEventListener is not a function
>     at HTMLElement.<anonymous> (iron-a11y-keys-behavior.html:441)
>     at Array.forEach (<anonymous>)
>     at HTMLElement._listenKeyEventListeners (iron-a11y-keys-behavior.html:435)
>     at HTMLElement.attached (iron-a11y-keys-behavior.html:329)
>     at HTMLElement.attached (class.html:244)
>     at HTMLElement.attached (class.html:242)
>     at HTMLElement.connectedCallback (legacy-element-mixin.html:107)
>     at HTMLElement._attachDom (element-mixin.html:593)
>     at HTMLElement._readyClients (element-mixin.html:566)
>     at HTMLElement._flushClients (property-effects.html:1565)
> (anonymous) @ iron-a11y-keys-behavior.html:441
> _listenKeyEventListeners @ iron-a11y-keys-behavior.html:435
> attached @ iron-a11y-keys-behavior.html:329
> attached @ class.html:244
> attached @ class.html:242
> connectedCallback @ legacy-element-mixin.html:107
> _attachDom @ element-mixin.html:593
> _readyClients @ element-mixin.html:566
> _flushClients @ property-effects.html:1565
> _propertiesChanged @ property-effects.html:1699
> _flushProperties @ properties-changed.html:325
> _flushProperties @ property-effects.html:1551
> ready @ property-effects.html:1656
> ready @ element-mixin.html:551
> _enableProperties @ properties-changed.html:308
> connectedCallback @ properties-mixin.html:203
> connectedCallback @ element-mixin.html:537
> _attachDom @ element-mixin.html:593
> _readyClients @ element-mixin.html:566
> _flushClients @ property-effects.html:1565
> _propertiesChanged @ property-effects.html:1699
> _flushProperties @ properties-changed.html:325
> _flushProperties @ property-effects.html:1551
> ready @ property-effects.html:1656
> ready @ element-mixin.html:551
> ready @ noticeable-project-posts-upsert.html:282
[...]

PR #5075, opened 2018-01-30
[element-mixin] Do not create property accessors unless a property ef…

### LINKED ISSUE #5073: Using mixin in a base class triggers TypeError in descendants in Polymer 2.4.0
_From @dpikhulya on January 30, 2018 17:41_

Adding Polymer.PaperButtonBehavior to a base class, and extending that base class in two different child elements makes the following exception to be thrown by the second subclass instance:

```
properties-changed.html:133 Uncaught TypeError: Cannot read property '_imperativeKeyBindings' of undefined
    at HTMLElement.get (properties-changed.html:133)
    at HTMLElement._prepKeyBindings (iron-a11y-keys-behavior.html:397)
    at HTMLElement.registered (iron-a11y-keys-behavior.html:325)
    at HTMLElement._registered (class.html:197)
    at HTMLElement._registered (class.html:185)
    at HTMLElement._registered (class.html:185)
    at HTMLElement._registered (class.html:185)
    at HTMLElement._registered (class.html:185)
    at HTMLElement._initializeProperties (legacy-element-mixin.html:174)
    at new PropertiesChanged (properties-changed.html:153)
    ...
```
In this example we have a hierarchy like this:
```
AbstractButton <- Polymer.PaperButtonBehavior
CustomButton1 <- AbstractButton
CustomButton2 <- AbstractButton
```
If we place BOTH `<custom-button-1>` and `<custom-button-2>` on the same page, only one of them (the second one) will fail with this exception. Commenting either of subclass instances doesn't trigger an error. 

The same example works fine with Polymer 2.3.1.

See [this example](http://jsbin.com/dovujaquji/1/edit?html,output), which fails to display "custom-button-2" text as a button.

This might not be specific to PaperButtonBehavior, but I didn't delve deeper.

### Expected outcome

No exception is thrown, and both AbstractButton descendants are displayed as buttons.

### Actual outcome

TypeError exception is thrown, the second subclass instance is not displayed properly.

### Live Demo

http://jsbin.com/dovujaquji/1/edit?html,output

### Steps to reproduce

Just open that page and see the exception in the console:
image

### Browsers Affected

Checked this to be reproducible at least in Chrome 64 and Firefox 58.

_Copied from original issue: PolymerElements/paper-behaviors#86_

sorvell:
…fect exists

Reverts an unintentional breaking change. Previously (2.3.x) property accessors were created only for properties in the properties object which had explicit property effects. In 2.4.0, property accessors were created for any property name returned in the properties object. This was a breaking change in rare cases (for example in https://github.com/PolymerElements/iron-a11y-keys-behavior).

This change reverts the 2.4.0 behavior. Accessors are now again only added for properties which have explicit effects. In addition, `_addPropetyToAttributeMap` was added to `PropertiesChanged` to record a mapping from the attribute to the property name (needed for deserialization).

### Reference Issue
Fixes #5073

### DIFF
--- a/externs/closure-types.js
+++ b/externs/closure-types.js
@@ -24,6 +24,10 @@ function Polymer_PropertiesChanged(){}
 Polymer_PropertiesChanged.prototype._createPropertyAccessor = function(property, readOnly){};
 /**
 * @param {string} property Name of the property
+*/
+Polymer_PropertiesChanged.prototype._addPropertyToAttributeMap = function(property){};
+/**
+* @param {string} property Name of the property
 * @param {boolean=} readOnly When true, no setter is created
 * @return {void}
 */
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -231,11 +231,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (info.observer) {
         proto._createPropertyObserver(name, info.observer, allProps[info.observer]);
       }
-      // always ensure an accessor is made for properties but don't stomp
-      // on existing values.
-      if (!info.readOnly && !(name in proto)) {
-        proto._createPropertyAccessor(name);
-      }
+      // always create the mapping from attribute back to property for deserialization.
+      proto._addPropertyToAttributeMap(name);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


TimvdLippe:
Does this also fix #5069?

kevinpschaaf:
@TimvdLippe We root-caused #5069 to a user error which surfaced due to a attribute-deserialization fix in 2.4, comments are updated over there.

PR #5079, opened 2018-01-31
Added Polymer.version to polymer-externs

mikesamuel:


TimvdLippe:
I think this file is generated so we need to annotate the type in the source code instead.

aomarks:
> I think this file is generated so we need to annotate the type in the source code instead.

This one is manually maintained actually, `closure-types.js` is the generated one.

PR #5083, opened 2018-02-01
Use class syntax in <dom-repeat> documentation (#5077)

MajorBreakfast:
Fixes https://github.com/Polymer/polymer/issues/5077

PR #5085, opened 2018-02-01
Update JSDocs to use <dom-repeat> tags

### LINKED ISSUE #5077: Outdated documentation of dom-repeat
### Description
The documentation for dom-repeat is using Polymer 1.x syntax. This documentation is present as a JSDoc in the source file and being displayed at webcomponents.org. Happy to create a PR but as this touches public doc the team is probably better placed to review.

#### Live Demo
https://www.webcomponents.org/element/polymer/polymer/v2.4.0/elements/dom-repeat (or any 2.x.x version)

#### Expected Results
Usage of `<dom-repeat>` and Polymer 2 syntax to create a new component with `class EmployeeList extends Polymer.Element`

#### Actual Results
Currently displaying `<template is="dom-repeat">` and Polymer 1 syntax to create a new component using `Polymer({is: 'employee-list'})`

### Versions
- Polymer: v2.x

Note that dom-if would benefit of a similar documentation (missing example)

floribon:
Follow-up of #5083, Fixes #5077 

The [documentation page for v2 dom-repeat](https://www.polymer-project.org/2.0/docs/api/elements/Polymer.DomRepeat) still uses Polymer 1 element creation syntax and the `<template is="dom-repeat">` shorthand.

My understanding from https://www.polymer-project.org/2.0/docs/about_20#type-extension is that the wraped element syntax `<dom-repeat>` should now be promoted.

### DIFF
--- a/lib/elements/array-selector.html
+++ b/lib/elements/array-selector.html
@@ -373,19 +373,23 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    *   <template>
    *
    *     <div> Employee list: </div>
-   *     <template is="dom-repeat" id="employeeList" items="{{employees}}">
+   *     <dom-repeat id="employeeList" items="{{employees}}">
+   *       <template>
    *         <div>First name: <span>{{item.first}}</span></div>
-   *         <div>Last name: <span>{{item.last}}</span></div>
-   *         <button on-click="toggleSelection">Select</button>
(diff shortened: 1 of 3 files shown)


TimvdLippe:
Manually merged, as the types of `array-selector` had to be updated. Thanks @floribon for the follow-up!

PR #5089, opened 2018-02-02
Update docs: templatize() cannot be called multiple times

MajorBreakfast:
I noticed that the docs didn't reflect what the code is actually doing. You **cannot** call `templatize()` twice on the same template.

### DIFF
--- a/lib/utils/templatize.html
+++ b/lib/utils/templatize.html
@@ -462,6 +462,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        *   from `instance.parentModel` in cases where template instance nesting
        *   causes an inner model to shadow an outer model.
        *
+       * All callbacks are called bound to the `owner`. Any context
+       * needed for the callbacks (such as references to `instances` stamped)
+       * should be stored on the `owner` such that they can be retrieved via
+       * `this`.
+       *
        * When `options.forwardHostProp` is declared as an option, any properties
        * referenced in the template will be automatically forwarded from the host of
        * the `<template>` to instances, with the exception of any properties listed in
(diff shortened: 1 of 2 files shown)


lib/utils/templatize.html
```diff
@@ -462,6 +462,10 @@
        *   from `instance.parentModel` in cases where template instance nesting
        *   causes an inner model to shadow an outer model.
        *
+       * All callbacks are called bound to the `owner`. So, context
+       * needed for the callbacks (such as references to `instances` stamped)
+       * should be stored on the `owner` such that they can be retrieved via `this`
```
kevinpschaaf: This is good info.

MajorBreakfast: Jep. I just moved that :D

lib/utils/templatize.html
```diff
@@ -471,14 +475,8 @@
...
        * be forwarded from the host).
        *
-       * Note that the class returned from `templatize` is generated only once
-       * for a given `<template>` using `options` from the first call for that
-       * template, and the cached class is returned for all subsequent calls to
-       * `templatize` for that template.  As such, `options` callbacks should not
-       * close over owner-specific properties since only the first `options` is
-       * used; rather, callbacks are called bound to the `owner`, and so context
-       * needed from the callbacks (such as references to `instances` stamped)
-       * should be stored on the `owner` such that they can be retrieved via `this`.
```
kevinpschaaf: In addition to the limitation you added here, we should clarify the point here, not delete it.  It is nuanced:

Consider the following:

```html
<dom-repeat items="[[items]]" id="outerRepeat">
  <template id="outerTemplate">
    <dom-repeat items="[[item.items]]" id="innerRepeat">
      <template id="innerTemplate">
        <div>[[item.text]]</div>
      </template>
    </dom-repeat>
  </template>
</dom-repeat>
```
The `innerRepeat` will call `templatize` to retrieve a constructor to generate instances of `innerTemplate` once for each outer item (and then stamp instances from the constructor returned for each inner item).  Technically it is a new instance of a `<template id="innerTemplate">` that is being templatized for each outer item, and so does not violate the rule you replaced it with ("can only be run once for a given `<template>`").

However there's an additional limitation that the note you deleted was trying to capture.  Templatizer caches the class generated from the first call to `templatize` for `innerTemplate`, and returns that generated class for every subsequent call to `templatize` on `innerTemplate`.  Without this caching, for every instance of `outerRepeat`, the `innerRepeat` would re-templatize the inner template, which is wasteful since we can track that the template is identical.  However, this caching means that the `options` passed into `templatize` are only respected for the first call of `templatize` for a given _nested template_.

Perhaps emphasizing this note applies to _*nested*_ templates, while adding additional note you provided, would be best.

MajorBreakfast: Thanks for the clarification! I'll update this PR with a revised explanation. There's no doubt that this should be documented because there's a lot going on.

I'll update this PR tomorrow. (It's midnight now here in Germany)

MajorBreakfast:
@kevinpschaaf I've rephrased the paragraph. I think it's now clear to the reader what happens. What do you think?

MajorBreakfast:
@kevinpschaaf I've refined it further. I've also added that one should not use arrow functions for the callbacks. Hehe. I just did that and it took me half a day to find the bug xD Unless you've anything to add, I think it's now good to merge.

MajorBreakfast:
👍

PR #5095, opened 2018-02-06
Throw an error if Polymer is loaded twice

### LINKED ISSUE #4134: Registration failed for type 'dom-module'/DuplicateDefinitionError: a type with name 'dom-module' is already registered
### Description
First of all it took about 6 hours to realize what causes this error. So thanks to me. 

**After build** error thrown when using any import other than `[entrypoint].html` in main `index.html` file.

In **Firefox 49.0.2** error looks like this:

> Error: DuplicateDefinitionError: a type with name 'dom-module' is already registered
> TypeError: Polymer.Base._getExtendedPrototype is not a function
> TypeError: this._desugarBehaviors is not a function

In **Chrome Version 54.0.2840.71 (64-bit)** looks like this:

>Uncaught DOMException: Failed to execute 'registerElement' on 'Document': Registration failed for type 'dom-module'. A type with that name is already registered.
>Uncaught TypeError: Polymer.Base._getExtendedPrototype is not a function
>Uncaught TypeError: this._desugarBehaviors is not a function

#### Steps to Reproduce
1. Create a new application/[or clone this repo and skip next 2 steps](https://github.com/iSuslov/polymer-import-error-in-index.html)
2. Add import in main `index.html` like it says in the [flexbox docs](https://elements.polymer-project.org/guides/flex-layout) in **Example: using classes in the main document** section. Like that:
```html
<head>

  ...

  <link rel="import" href="bower_components/iron-flex-layout/iron-flex-layout-classes.html">

  ...

</head>
```
3. Add `iron-flex-layout-classes` in `polymer.json` `includeDependencies` property. Otherwise it won't pick this dependency up. 
4. Build with `polymer build`
5. Serve with `polymer serve` bundled/unbundled version of the app.
5. Get a lot of pain with the error above.

#### Expected Results
No error

#### Actual Results
Error. See above.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 9
- [ ] Safari 8 most probably yes
- [ ] IE 11  most probably yes

### Versions

- Polymer-cli:  v0.16.0, v0.17.0 
- Polymer:  v1.7.0 
- webcomponents: v0.7.23

TimvdLippe:
Whenever Polymer is somehow loaded twice, there is a cryptic error shown in the console:
```
Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry
```
This happens when two urls resolve to different versions of Polymer (for example relative and absolute url). The following is now shown in the console when running the new smoke test:
image

Feel free to bikeshed on the actual message. Maybe @arthurevans has suggestions for this?
### Reference Issue
Fixes #4134

### DIFF
--- a/lib/utils/boot.html
+++ b/lib/utils/boot.html
@@ -14,12 +14,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   const userPolymer = window.Polymer;
 
   if (userPolymer && 'version' in userPolymer) {
-    throw new Error(`
-      Polymer was already loaded. This is likely an issue with an incorrect
-      link loading a fallback page or if a link to polymer.html redirects
-      to the wrong directory. E.g. "../polymer.html" and "../../polymer.html"
-      could load Polymer twice. Make sure that all urls pointing to Polymer
-      or dependencies that rely on Polymer are correctly deduped.`);
+    throw new Error(`Error: Polymer loaded twice.`);
   }
 
   /**
--- a/lib/utils/boot.html
+++ b/lib/utils/boot.html
@@ -13,6 +13,15 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   const userPolymer = window.Polymer;
 
+  if (userPolymer && 'version' in userPolymer) {
+    throw new Error(`
+      Polymer was already loaded. This is likely an issue with an incorrect
+      link loading a fallback page or if a link to polymer.html redirects
+      to the wrong directory. E.g. "../polymer.html" and "../../polymer.html"
+      could load Polymer twice. Make sure that all urls pointing to Polymer
+      or dependencies that rely on Polymer are correctly deduped.`);
+  }
+
   /**
    * @namespace Polymer
    * @summary Polymer is a lightweight library built on top of the web
--- a/test/smoke/polymer-loaded-twice.html
+++ b/test/smoke/polymer-loaded-twice.html
@@ -1,0 +1,11 @@
+<html>
+<head>
+  <link rel="import" href="../../polymer.html" />
+  <link rel="import" href="/polymer.html" />
+</head>
+<body>
+  <div>
+    Passes if there is an error in the console that explains the duplicate loading of Polymer
+  </div>
+</body>
+</html>
--- a/test/smoke/polymer-loaded-twice.html
+++ b/test/smoke/polymer-loaded-twice.html
@@ -1,7 +1,7 @@
 <html>
 <head>
-  <link rel="import" href="../../polymer.html" />
-  <link rel="import" href="/polymer.html" />
+  <link rel="import" href="../../lib/utils/boot.html" />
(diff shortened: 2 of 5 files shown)


lib/utils/boot.html
```diff
@@ -13,6 +13,15 @@
 
   const userPolymer = window.Polymer;
 
+  if (userPolymer && 'version' in userPolymer) {
+    throw new Error(`
+      Polymer was already loaded. This is likely an issue with an incorrect
```
sorvell: Error, Polymer loaded twice.

TimvdLippe: Done 👍

kevinpschaaf:
@TimvdLippe Are you able to add a unit test for this? Not sure how easy it is to get a second version of polymer actually loaded...

TimvdLippe:
I am not sure if we can. You can run the smoke test to see how it works. It is really easy to make such mistakes (in this case just two link elements)

TimvdLippe:
@kevinpschaaf Managed to write a unit test using an iframe.

test/smoke/polymer-loaded-twice.html
```diff
@@ -0,0 +1,11 @@
+<html>
+<head>
+  <link rel="import" href="../../polymer.html" />
+  <link rel="import" href="/polymer.html" />
```
TimvdLippe: Use query parameters instead absolute url to make sure you can serve this source from a different endpoint as well.

test/unit/loading-twice-iframe.html
```diff
@@ -0,0 +1,22 @@
...
+<body>
+  <script>
+  window.addEventListener('message', () => {
+    const relativeLink = document.createElement('link');
+    relativeLink.href = '../../polymer.html';
+    relativeLink.rel = 'import';
+    document.body.appendChild(relativeLink);
+
+    const absoluteLink = document.createElement('link');
+    absoluteLink.href = '/polymer.html';
```
sorvell: Let's load this with a query string so we can avoid an abs url since that puts constraints on serving.

TimvdLippe: It seems that, while query parameters can be used for cache busting, the browser will not load the same link twice even if the query parameters are different :sob:

test/smoke/polymer-loaded-twice.html
TimvdLippe: Load boot twice, verify error

TimvdLippe:
I fixed the unit test, however it appears it is failing on Safari:

```
AssertionError: Error message should include "Polymer loaded twice": expected false to be true (http://localhost:8081/components/chai/chai.js:210)
```

Sadly I do not have a Safari machine, so I can not debug this issue. It is probably an incorrect error message that is posted :cry:

TimvdLippe:
@sorvell Tests have now been fixed. PTAL

TimvdLippe:
I am no longer pursuing this PR.

PR #5097, opened 2018-02-06
[PropertiesChanged]: adds _shouldPropertiesChange

sorvell:
Fixes https://github.com/Polymer/polymer/issues/5098

Adds `_shouldPropertiesChange` method which can be overridden to customize when the `_propertiesChanged` method is called. This method gets the current, pending, and old property values and by default returns truthy...

### DIFF
--- a/lib/mixins/properties-changed.html
+++ b/lib/mixins/properties-changed.html
@@ -353,7 +353,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
          *   in `changedProps`
          * @return {boolean} true if changedProps is truthy
          */
-        _shouldPropertiesChange(props, changedProps, old) {
+        _shouldPropertiesChange(currentProps, changedProps, oldProps) { // eslint-disable-line no-unused-vars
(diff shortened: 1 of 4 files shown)


lib/mixins/properties-changed.html
```diff
@@ -332,13 +332,31 @@
          * @protected
          */
         _flushProperties() {
-          if (this.__dataPending && this.__dataOld) {
-            let changedProps = this.__dataPending;
+          const props = this.__data;
+          const changedProps = this.__dataPending;
+          const old = this.__dataOld;
+          if (this._shouldPropertiesChange(props, changedProps, old)) {
             this.__dataPending = null;
```
kevinpschaaf: Just a thought re: memory: should we be zeroing out `__dataOld` here, to so that the previous turn's worth of changed data can be gc'ed?  It currently gets reset on first change in new turn [here](https://github.com/Polymer/polymer/pull/5097/files#diff-980fcce29657756129c5008e85181523R271).

PR #5099, opened 2018-02-07
Tighten a type. Bump TS generator version. Format Travis type error.

aomarks:
- Make `EventApi.path`'s `EventTarget` type non-nullable.
- Bump TypeScript generator version.
- Make Travis update-types failure style the same as the elements.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,10 @@ before_script:
 - npm install -g bower gulp-cli@1
 - bower install
 - gulp lint-eslint
-- gulp update-types
-- git diff --exit-code
+- >-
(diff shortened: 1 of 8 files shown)


aomarks:
Hm, Travis is failing because `package-lock.json` changed. I guess `npm install` can do that? Not sure why this never came up before though. I'm doing it through `npm run update-types` instead of `gulp update-types` now, but it should be the exact same thing.

TimvdLippe:
@aomarks I have regenerated the lock file with npm 5.6.0. Not sure why your original install did not fully "lock" the lockfile.

PR #5103, opened 2018-02-08
Fix license links

abdonrd:
Edit: Oh, the auto request review is nice!


abdonrd:
Btw, would be nice to use `https` instead of `http`.

PR #5108, opened 2018-02-08
Use stronger check for PropertyEffects clients. Fixes #5107

### LINKED ISSUE #5017: Allow users to avoid deprecation warning re: styles in HTML imports

### LINKED ISSUE #5107: Calling this.set() on an element bound to a PropertiesMixin element throws
### Description
When a `Polymer.Element` element is bound to a `Polymer.PropertiesMixin` element (e.g lit-element) and `this.set('path.to.something', value)` is called on the `Polymer.Element`, an error is thrown.

The error is thrown because path notification in code `PropertyEffects` uses too naive a check that the bound element is another `PropertyEffects` element with path notification API.  Ideally we'd use `instanceof PropertyEffects` but this is not trivial with mixins.

#### Live Demo
https://jsbin.com/zujixol/1/edit?html,console,output

#### Steps to Reproduce
1. Open [jsBin](https://jsbin.com/zujixol/1/edit?html,console,output)
2. Click button
3. Note error thrown in console

#### Expected Results
No error is thrown

#### Actual Results
Error is thrown:
```
"TypeError: node._setPendingPropertyOrPath is not a function
    at Object.runBindingEffect [as fn] (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/property-effects.html:573:16)
    at runEffectsForProperty (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/property-effects.html:159:14)
    at runEffects (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/property-effects.html:125:13)
    at HTMLElement._propagatePropertyChanges (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/property-effects.html:1733:11)
    at HTMLElement._propertiesChanged (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/property-effects.html:1697:14)
    at HTMLElement._flushProperties (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/properties-changed.html:341:18)
    at HTMLElement._flushProperties (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/property-effects.html:1551:15)
    at HTMLElement._invalidateProperties (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/property-effects.html:1523:16)
    at HTMLElement.set (https://polygit.org/polymer+:master/webcomponents+:master/shadycss+webcomponents+:master/components/polymer/lib/mixins/property-effects.html:1858:20)
    at HTMLElement.change (https://null.jsbin.com/runner:25:14)"
```

### Browsers Affected
All

### Versions
- Polymer: v2.5.0

kevinpschaaf:
Check for private variable `__isPropertiesEffectsClient` to ensure a client is opting into synchronous batching and path notifications.

### Reference Issue
Fixes #5107


TimvdLippe:
I think you need to update your local version of the gen-typescript-declarations. It is pulling in changes that shouldnt be there post 1.0.0

PR #5110, opened 2018-02-09
Remove unnecessary mutableData property from MutableData mixin

MajorBreakfast:
Why is this set to `false`? I don't get it :)
If this is actually supposed to be set to `false`, then I'd like to add a comment why in this PR. But, as it's the MutableData mixin I'm assuming that the value should be `true`. I don't think that it's used anywhere. I think it was just added to make the Closure compiler happy.

### DIFF
--- a/lib/mixins/mutable-data.html
+++ b/lib/mixins/mutable-data.html
@@ -104,8 +104,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
 
     }
-    /** @type {boolean} */
-    MutableData.prototype.mutableData = false;
 
     return MutableData;
 


lib/mixins/mutable-data.html
```diff
@@ -105,7 +105,7 @@
 
     }
     /** @type {boolean} */
-    MutableData.prototype.mutableData = false;
+    MutableData.prototype.mutableData = true;
```
kevinpschaaf: So, `this.mutableData` is never used in `MutableData`, only `OptionalMutableData` (`MutableData` doesn't have a switch and always calls `mutablePropertyChange` with its `mutableData` arg set to `true`).

So it looks like this line is altogether unnecessary.  The blame shows it being added to help closure understand property types, but it must have been made in error; I don't see the closure warning count change my removing it, so I'd suggest just updating the PR to remove it altogether.

PR #5115, opened 2018-02-13
Fix issue with not genering the Templatizer docs

### LINKED ISSUE #5037: Polymer Templatize docs are not generated
### Description
The Polymer Templatize docs are no longer available on https://www.polymer-project.org/2.0/docs/api/ and its corresponding namespace url now 404s: https://www.polymer-project.org/2.0/docs/api/namespaces/Polymer.Templatize

This might be a problem in `polymer-analyzer` or a mistyping in one of our recent Closure/TypeScript changes. However, I did a `git blame` on the file and checked the recent changes, but I could not spot any obvious issues. @aomarks do you maybe have a clue what might be broken here?

TimvdLippe:
The docs were not properly generated, for two reasons:

1. The file exports to `Polymer`, but because of variable referencing, the `Polymer.Templatize` had no docs. Simply moving the `const` declaration to `Polymer.Templatize` fixes that issue.
1. There was a new-line between the declaration and the export. This meant that the JSDoc wasn't "attached" to the declaration. Removing the line fixes the generation of the namespace documentation.
### Reference Issue
Fixes #5037

### DIFF
--- a/lib/utils/templatize.html
+++ b/lib/utils/templatize.html
@@ -416,8 +416,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @summary Module for preparing and stamping instances of templates
      *   utilizing Polymer templating features.
      */
-
-    const Templatize = {
+    Polymer.Templatize = {
 
       /**
        * Returns an anonymous `Polymer.PropertyEffects` class bound to the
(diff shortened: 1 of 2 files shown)


TimvdLippe:
@kevinpschaaf Could you run it? Not on a laptop atm 😢

PR #5117, opened 2018-02-15
dispatchEvent returns boolean

aomarks:
This came up in the TypeScript declarations. Since we extend both `Element` and `TemplateInstanceBase`, the signatures on these two interfaces need to be compatible. `Element.dispatchEvent` returns a `boolean`, but `TemplateInstanceBase.dispatchEvent` currently returns `void`.

I don't really understand how this implementation of `dispatchEvent` is supposed to work and when it is called, so is it ok to always return `true` (meaning `preventDefault()` was not called) as I have done here?

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent

### DIFF
--- a/lib/utils/templatize.html
+++ b/lib/utils/templatize.html
@@ -246,7 +246,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * dispatch events safely no-op.
        *
        * @param {Event} event Event to dispatch
-       * @return {boolean}
+       * @return {boolean} Always true.
        */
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Hm, previously it would return undefined, which is falsy. Maybe the intention was to block events from bubbling up? Kevin or Steve probably know best.

sorvell:
My understanding is this was only added as a stub to prevent errors so I think this is fine.

PR #5118, opened 2018-02-15
Fix closure for forwardHostProp in Polymer.Templatize.templatize

### LINKED ISSUE #5063: Polymer.Templatize.templatize() is preserving forwardHostProp() function
Please see the reproduction here:
https://jsfiddle.net/exxe08ch/

Polymer.Templatize.templatize() is preserving forwardHostProp() function that is set on the first call.

As I can see, templatize() creates propagate effects only once for the first template's instance, then caches templatizeTemplateClass within the template element. (templatize.html:275)
As a result propagate effects are not recreated for new template's instances, forwardHostProp() function reference is preserved, and host property changes are not propagated to them.

Polymer: v2.3.1

TimvdLippe:
When creating multiple constructors on the same template with `Polymer.Templatize.templatize`, the caching in `templateInfo.templatizeTemplateClass` would not correctly use the closure of `forwardHostProp` if this was different between calls. Therefore, do a dirty check on `klass.__cachedUserForwardHostProp`, which will be different if the function references are different as well.
### Reference Issue
Fixes #5063

### DIFF
--- a/lib/utils/templatize.html
+++ b/lib/utils/templatize.html
@@ -306,12 +306,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       let userForwardHostProp = options.forwardHostProp;
       if (userForwardHostProp) {
         // Provide data API and property effects on memoized template class
-        let klass = templateInfo.templatizeTemplateClass;
-        if (!klass || klass.__cachedUserForwardHostProp !== userForwardHostProp) {
+        let klass = userForwardHostProp.templatizeTemplateClass;
+        if (!klass) {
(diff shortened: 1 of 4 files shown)


TimvdLippe:
@kevinpschaaf I have updated the code now to memoize on the function instead of the `templateInfo`. This should address the potential clash of overwriting the wrong class while also work per function.

kevinpschaaf:
This change fixes the "bug" by basically eliminating the performance optimization of caching the generated class, because there is never a case that you would actually used the cached class, since the `forwardHostProp` method is new for every call to `templatize`.

PR #5123, opened 2018-02-18
Simplify condition checking in stylesFromModule function

duykhoa:
This PR rearranges code and combines similar conditions into one.


duykhoa:
Thanks @TimvdLippe for your quick review.

PR #5125, opened 2018-02-21
Automatically detect capitalized HTML attributes

### LINKED ISSUE #1831: viewBox svg attribute
When using the data binding with $ attributes are converted to lower case.
Browsers expect svg viewBox attribute to be camel case "viewBox" but it is converted to lower case "viewbox". The data binding does not work.

The only solution for this is to use the dom api to change this attribute dynamically with setAttribute("viewBox", "0 0 100 100") ie in the "ready:" function.

Selecting the element by id seems not a good thing. If several elements have the same id in the same dom.

There should be a cleaner way to solve this.

### LINKED ISSUE #5114: Cannot set case sensitive attributes with $=
### Description
Setting an attribute with `$=` strips case information.

#### Steps to Reproduce
1. Create a new Polymer element
1. Add an attribute that relies on case, for instance `<svg viewBox$="[[calculateViewBox()]]">...` 

#### Expected Results
The result template should the attribute with case preserved, e.g. `<svg viewBox="0 0 50 50">...` 

#### Actual Results
Case information is lost, e.g. `<svg viewbox="0 0 50 50">...` 

Note that this specific example (`viewBox`) has already been raised in https://github.com/Polymer/polymer/issues/1831, but nothing has happened on that issue since 2016 - this is a more serious issue and `viewBox` is just an easily reproduced case.

Attributes in HTML are case sensitive, and Polyment needs to support that.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 9
- [x] Safari 8
- [x] IE 11

### Versions
- Polymer: ^2.5.0

TimvdLippe:
Support capitalized HTML attributes by providing a mixin that uses an heuristic based on capitalization.
### Reference Issue
Fixes #5114 
Fixes #1831


lib/utils/attribute-capitalization-mixin.html
```diff
@@ -0,0 +1,80 @@
...
+       * @override
+       * @param {Node} node Node to parse
+       * @param {TemplateInfo} templateInfo Template metadata for current template
+       * @param {NodeInfo} nodeInfo Node metadata for current template node
+       * @return {boolean} `true` if the visited node added node-specific
+       *   metadata to `nodeInfo`
+       * @protected
+       * @suppress {missingProperties} Interfaces in closure do not inherit statics, but classes do
+       */
+      static _parseTemplateNodeAttribute(node, templateInfo, nodeInfo, name, value) {
```
sorvell: Let's go ahead and just put this into `PropertyEffects`. It seems like it's only going to help and does no harm.

TimvdLippe: Done :+1:

lib/mixins/property-effects.html
```diff
@@ -2530,7 +2533,9 @@
           // Attribute or property
           let origName = name;
           let kind = 'property';
-          if (name[name.length-1] == '$') {
+          if (capitalAttributeRegex.test(name)) {
```
sorvell: Let's add a comment in the code here explaining why we're doing this. The only way we see a capital letter here is if the attr has a capital letter in it per spec. In this case, to make sure this binding works, we go ahead and make the binding to the attribute.

We should also add something simplified about this to the doc.

TimvdLippe: @sorvell Done :+1:

TimvdLippe:
Docs issue opened at https://github.com/Polymer/docs/issues/2490

PR #5130, opened 2018-02-25
Update Google Closure Compiler version and fix cast

TimvdLippe:
Update Closure to the latest released version. It uncovered 1 assignment as violating. `LegacyElementMixin` implicitly requires a `HTMLElement` for `base`. Had to cast to this type and suppress "invalid casts" to make closure happy.


kevinpschaaf:
@azakus You ok bumping the closure version here?

PR #5131, opened 2018-02-25
Update dom5 to 3.0.0

TimvdLippe:
There are no breaking changes for Polymer core :tada: 

CC @rictic to make sure we are updating correctly. I have tested `npm run build closure` which still appeared to work properly.


rictic:
Yep, polymer's gulp file doesn't use any of the changed APIs, so this looks good to me.

rictic:
One thing to note, dom5, like the other polymer tools, now requires node v6 or greater.

PR #5134, opened 2018-02-27
Add note for npm package

### LINKED ISSUE #4611: package.json missing shadycss dependency
I was trying using polymer 2.0 from npm, but it seems that some imports are failing because of a missing dependency in the package.json:  

```
ERROR in ./~/@polymer/polymer/lib/legacy/legacy-element-mixin.html
Error: Can't resolve '../../../shadycss/apply-shim.html' 

ERROR in ./~/@polymer/polymer/lib/elements/custom-style.html
Error: Can't resolve '../../../shadycss/custom-style-interface.html' 
```

shadycss is included as a `devDependency`, it probably should be a `dependency`.

Another problem is that npm, unlike bower, installs shadycss under `@webcomponents/shadycss`, which will make the import fail even if it is added as a `dependency`.

And yet another problem is that, given that npm installs polymer under `@polymer`, one more `../` should be added.

I fixed it by installing `@webcomponents/shadycss` and changing the imports to  `<link rel="import" href="../../../../@webcomponents/shadycss/apply-shim.html">` and `<link rel="import" href="../../../../@webcomponents/shadycss/custom-style-interface.html">`. But I guess this will break bower installation.

Is this an issue or am I doing something wrong?

azakus:
For people wanting to try Polymer with npm, they should use the v3 prereleases.

### Reference Issue
Fixes #4611

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ Polymer 2.0 is released, and will be the future focus of Polymer development goi
 
 👀 **Looking for Polymer v1.x?** Please see the [the v1 branch](https://github.com/Polymer/polymer/tree/1.x).
 
+⁉️ **Looking to use Polymer with npm?** Polymer 1.x and 2.x both use `bower` for version management, but v3 and on will use `npm`. Please see our [v3 announcement](https://www.polymer-project.org/blog/2017-08-22-npm-modules.html) and [v3 hand-on](https://www.polymer-project.org/blog/2017-08-23-hands-on-30-preview.html) to use a preview version of v3.
+
 ## Overview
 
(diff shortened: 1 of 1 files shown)


TimvdLippe:
I think we can close https://github.com/Polymer/polymer/issues/4611 with this by explaining Polymer 2 on NPM is not supported?

justinfagnani:
@TimvdLippe yep

README.md
```diff
@@ -27,6 +27,8 @@ Polymer 2.0 is released, and will be the future focus of Polymer development goi
 
 👀 **Looking for Polymer v1.x?** Please see the [the v1 branch](https://github.com/Polymer/polymer/tree/1.x).
 
+⁉️ **Looking to use Polymer with npm?** Polymer 1.x and 2.x both use `bower` for version management, but v3 and on will use `npm`. Please see our [v3 announcement](https://www.polymer-project.org/blog/2017-08-22-npm-modules.html) and [v3 hand-on](https://www.polymer-project.org/blog/2017-08-23-hands-on-30-preview.html) to use a preview version of v3.
```
justinfagnani: I'd add a note about why _this_ version (2.x) is on npm and what it's support is:

"Versions before 3.0 are also published to npm "as-is" and are generally unsupported. These packages are for advanced users who configure their own tooling to work around the differences between Bower and npm packages, like package folder layout."

azakus: Done!

Issue #1951, opened 2015-06-24
polymer element inside svg tags


forrejam:
Is there any plans to make SVG elements work with polymer?

``` html
    <template>
            <circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" />
            <my-custom-svg-element></my-custom-svg-element>
    </template>
```

I cant seem to get the `<foreignObject>` [work around](http://stackoverflow.com/questions/22393760/does-polymer-js-support-inner-svg-elements) working either. I'm using latest polymer from bower (1.0.4).

kevinpschaaf:
We can support this eventually.

Proof of concept here: http://jsbin.com/beqibe/edit?html,output

jirkafrancek:
Hi Kevin,

do you know if this has been fixed already? We'd like to use polymer in our future projects, but this issue is a blocker for us.

Many thanks!

HeSe:
Hi Kevin,

I have the same interest in this. 
Is this workaround still the way to go for now?
In which release can one expect a this currently P2 to be fixed?

Thank you!

christophe-g:
This is a first go :
https://github.com/Polymer/polymer/pull/3372
C.

fooloomanzii:
@kevinpschaaf Will it be supported? Although @christophe-g didn't find the time for the standard tests, #3372  looks like a promising solution. For full support of webContent there should be a solution for implementing webComponents of different namespaces, and you have by using SVG possibilities to reach high-precision graphic content

renfeng:
For people looking for a working solution,
https://github.com/Polymer/polymer/issues/1976#issuecomment-225454535

PR #5135, opened 2018-02-28
Make dom-elements work inside svgs

### LINKED ISSUE #1976: template doesn't work inside svg elements

TimvdLippe:
Use the `template` and `svg` innerHTML trick to let the HTML parser create the elements in the correct namespace. Then swap back the elements in a new template and continue iterating on that.

If you run the smoke test, you should see the following output (with the red square pulsing):

image

*Note*: This does not address #1951 yet, as I am not certain what the expected behavior is there. Since custom elements can nest svg elements, they would all need to stamp in the lightdom of the parent (e.g. the `svg`), but that seems like the wrong approach.

### Reference Issue
Fixes #1976

### DIFF
--- a/lib/mixins/template-stamp.html
+++ b/lib/mixins/template-stamp.html
@@ -16,6 +16,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   'use strict';
 
+  const SVG_NAMESPACE = 'http://www.w3.org/2000/svg';
+
   // 1.x backwards-compatible auto-wrapper for template type extensions
   // This is a clear layering violation and gives favored-nation status to
   // dom-if and dom-repeat templates.  This is a conceit we're choosing to keep
@@ -313,6 +315,32 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        *   metadata to `nodeInfo`
        */
       static _parseTemplateNestedTemplate(node, outerTemplateInfo, nodeInfo) {
+        // If a `<dom-repeat>` or a `<dom-if>` is inside of an svg, we need
+        // to create all elements in a different namespace. To do so, make an svg,
+        // let the HTML parser parse in that namespace, and then put back all
+        // created children in a fresh template.
+        if (node.parentElement.localName && templateExtensions[node.parentElement.localName]
+              && node.parentElement.parentElement && node.parentElement.parentElement.namespaceURI === SVG_NAMESPACE) {
+          const svg = document.createElementNS(SVG_NAMESPACE, 'svg');
+          svg.innerHTML = node.innerHTML;
+
+          const template = /** @type {HTMLTemplateElement} */ (document.createElement('template'));
+
+          while (svg.firstChild) {
+            template.content.appendChild(svg.removeChild(svg.firstChild));
+          }
+
+          node.parentNode.replaceChild(template, node);
+
+          // Clean up, such that `_parseTemplateChildNodes` does not run again
+          // on the old template children
+          while (node.firstChild) {
+            node.removeChild(node.firstChild);
+          }
+
+          node = template;
+        }
+
         let templateInfo = this._parseTemplate(node, outerTemplateInfo);
         let content = templateInfo.content =
           node.content.ownerDocument.createDocumentFragment();
--- a/lib/mixins/template-stamp.html
+++ b/lib/mixins/template-stamp.html
@@ -319,8 +319,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // to create all elements in a different namespace. To do so, make an svg,
         // let the HTML parser parse in that namespace, and then put back all
         // created children in a fresh template.
-        if (node.parentElement.localName && templateExtensions[node.parentElement.localName]
-              && node.parentElement.parentElement && node.parentElement.parentElement.namespaceURI === SVG_NAMESPACE) {
+        if (node.parentNode.localName && templateExtensions[node.parentNode.localName]
+              && node.parentNode.parentNode && node.parentNode.parentNode.namespaceURI === SVG_NAMESPACE) {
           const svg = document.createElementNS(SVG_NAMESPACE, 'svg');
           svg.innerHTML = node.innerHTML;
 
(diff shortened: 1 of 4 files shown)


renfeng:
Uncaught TypeError: Cannot read property 'localName' of undefined

        <template is="dom-repeat" items="[[positions]]" as="cx">
          <template is="dom-repeat" items="[[positions]]" as="cy">
            <ellipse rx="25" ry="25" cx$="[[cx]]" cy$="[[cy]]" stroke="black" />
          </template>
        </template>

lib/mixins/template-stamp.html
```diff
@@ -313,6 +315,32 @@
...
+        // If a `<dom-repeat>` or a `<dom-if>` is inside of an svg, we need
+        // to create all elements in a different namespace. To do so, make an svg,
+        // let the HTML parser parse in that namespace, and then put back all
+        // created children in a fresh template.
+        if (node.parentNode.localName && templateExtensions[node.parentNode.localName]
+              && node.parentNode.parentNode && node.parentNode.parentNode.namespaceURI === SVG_NAMESPACE) {
+          const svg = document.createElementNS(SVG_NAMESPACE, 'svg');
+          svg.innerHTML = node.innerHTML;
+
+          const template = /** @type {HTMLTemplateElement} */ (document.createElement('template'));
```
kevinpschaaf: Rather than creating a new template and replacing it, I think it's best to replace the contents of the existing template:

```js
node.textContent = '';
while (svg.firstChild) {
  node.content.appendChild(svg.firstChild);
}
```

lib/mixins/template-stamp.html
```diff
@@ -313,6 +315,32 @@
...
+
+          while (svg.firstChild) {
+            template.content.appendChild(svg.removeChild(svg.firstChild));
```
kevinpschaaf: Tip: no need to remove a child when moving it; append to new spot removes it from the previous, since a node can't exist in two trees at the same time.

lib/mixins/template-stamp.html
```diff
@@ -313,6 +315,32 @@
...
+          }
+
+          node.parentNode.replaceChild(template, node);
+
+          // Clean up, such that `_parseTemplateChildNodes` does not run again
+          // on the old template children
```
kevinpschaaf: Why would this have been the case?

lib/mixins/template-stamp.html
```diff
@@ -313,6 +315,32 @@
        *   metadata to `nodeInfo`
        */
       static _parseTemplateNestedTemplate(node, outerTemplateInfo, nodeInfo) {
+        // If a `<dom-repeat>` or a `<dom-if>` is inside of an svg, we need
+        // to create all elements in a different namespace. To do so, make an svg,
+        // let the HTML parser parse in that namespace, and then put back all
+        // created children in a fresh template.
+        if (node.parentNode.localName && templateExtensions[node.parentNode.localName]
+              && node.parentNode.parentNode && node.parentNode.parentNode.namespaceURI === SVG_NAMESPACE) {
+          const svg = document.createElementNS(SVG_NAMESPACE, 'svg');
```
kevinpschaaf: I was picturing this fixup being done in `_parseTemplate` so it applies to any template.  But I see now that the `dom-repeat` case requires checking up two levels to find the SVG namespace (rather than just one), which requires the whitelist of template wrappers. I also just realized `dom-repeat` can't be used in the main document to repeat SVG content because the `dom-repeat` must be parsed using the `<template is="...">` syntax otherwise it'll be kicked out of the SVG during parsing.  Conclusion: I think this is fine for now to use the whitelist and only do in nested templates, as you've done.

lib/mixins/template-stamp.html
TimvdLippe: I tried this at first, but this node does not have content. In other words, it is not actually a template. I think this happens because it parses template in the svg namespace, and this element does not exist there. Therefore I needed to create a new template to have the correct documentfragment to work with.

kevinpschaaf: Oh yeah, but we want to make a `Polymer.svg` helper that returns a template and use that here.  So it would be:

```js
const svgTemplate = Polymer.svg([node.innerHTML]);
node.textContent = '';
node.appendChild(svgTemplate.content);
```
(and we throw away the `svgTemplate`; if this was a hot path we'd care about the extra template node but it's not really)

lib/mixins/template-stamp.html
TimvdLippe: The children of template are not placed in a documentfragment (see previous reply) and are therefore direct children. Since parsetemplatenode still holds a reference to the original "template", it will iterate through the "real" children. But thats wrong, because we expect it to be in content instead.

lib/mixins/template-stamp.html
TimvdLippe: But `node.content` is undefined in this context. You cant assign it or use it here.

test/unit/svg-compatibility.html
```diff
@@ -0,0 +1,115 @@
+<!doctype html>
```
kevinpschaaf: Also note, need to add new unit tests to `runner.html` to get them into CI.

renfeng:
It works.

kevinpschaaf:
Note a change to shadycss (https://github.com/webcomponents/shadycss/pull/167) is required to get the tests here fully passing (ref: https://github.com/webcomponents/shadycss/issues/166)

TimvdLippe:
PR tentatively blocked on the resolution of https://github.com/w3c/webcomponents/issues/744

TimvdLippe:
Discussion has now moved to https://github.com/whatwg/html/issues/3553

TimvdLippe:
Closing per https://github.com/whatwg/html/issues/3563

Issue #4864, opened 2017-10-04
Improve internal API call-side by passing objects instead of literals


TimvdLippe:
The current effect system passes on a large number of arguments down into various arguments. This introduces the need for modifying a large number of methods to add a single argument (e.g. #4839). This boolean is on the call-side ambigous. In other words, the developer has to remember what the true and false in this position actually mean.

Another artifact of this approach is that when an argument is not needed, it can lead to call-sides such as `fn(prop, value, undefined, false, true)`, where the last 3 arguments are very ambigous and the caller only meant to set the last argument to true. What do they mean? why is `undefined` there? why should the other be false? Moreover, the order of arguments is quite arbitrary in the sense that it depends on the time they were added.

As such, I propose these "configurable" APIs are modified to pass an object instead. Since most of the APIs are internal, this should be a backwards-compatible change.

(Inspired by and for more information see https://github.com/yarnpkg/yarn/issues/3472)

PR #5145, opened 2018-03-10
Evaluated computed property dependencies first. Fixes #5143

### LINKED ISSUE #5143: [Feature request] Run computed properties in dependency order
### Description
Currently computed properties with invalidated dependencies are re-computed based on a combination of order they exist in the `properties` block and the order that dependencies are set, and will continue re-computing if subsequent dependent properties change.  This can result in computed properties running more than once in a turn, which is both inefficient if the computed property has expensive computation/side-effects, and may violate user expectation.

#### Live Demo
https://jsbin.com/lofapu/1/edit?html,console

#### Expected Results
Note that `computeZ` is run twice for the same change to `x`

#### Actual Results
Ideally `computeZ` would only be run once

### Browsers Affected
All

### Versions
- Polymer: v2.x

kevinpschaaf:
Updates the implementation of `runComputedEffect` to recursively walk the graph of dependencies that are themselves computed and re-compute them if needed, which ensures computed properties are not computed out of order.

The initial walk is still kicked off by the normal `runEffects` loop over changed properties, which ensures that no work related to computed properties occurs unless at least one dependency to a computed property has changed.  An alternate implementation of running every computed effect each flush was considered, but it seems worth keeping the "pay-for-play" model intact.  Thus, the effect de-duping mechanism still exists, which prevents a property later in the batch of changes from causing any computed effect that has already run during this flush from running again, and also prevents iloops for circular dependencies.

When walking the graph of computed dependencies to a computed property, we track when a computed effect is being run transitively (vs. directly due to a dependency change in `changedProps`) by passing a sentinel `TRANSITIVE_DEPENDENCY` property name to `runComputedEffect`'s `property` argument (which normally is the _input_ dependency which triggered the change).  This tracking is useful because transitively run computed effects need to dirty-check their dependencies, while effects that were directly run by `runEffects` can skip the expensive argument dirty check to occur since they are already guaranteed to have at least one argument dirty.

### Reference Issue
Fixes #5143

### DIFF
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -43,6 +43,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     READ_ONLY: '__readOnly'
   };
 
+  const COMPUTE_INFO = '__computeInfo';
+
   /**
    * @typedef {{
    * name: (string | undefined),
@@ -82,20 +84,23 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
    *
    * @param {Object} model Prototype or instance
    * @param {string} type Property effect type
+   * @param {boolean=} cloneArrays Clone any arrays mapped in the map
    * @return {Object} The own-property map of effects for the given type
    * @private
    */
-  function ensureOwnEffectMap(model, type) {
+  function ensureOwnEffectMap(model, type, cloneArrays) {
     let effects = model[type];
     if (!effects) {
       effects = model[type] = {};
     } else if (!model.hasOwnProperty(type)) {
       effects = model[type] = Object.create(model[type]);
-      for (let p in effects) {
-        let protoFx = effects[p];
-        let instFx = effects[p] = Array(protoFx.length);
-        for (let i=0; i<protoFx.length; i++) {
-          instFx[i] = protoFx[i];
+      if (cloneArrays) {
+        for (let p in effects) {
+          let protoFx = effects[p];
+          let instFx = effects[p] = Array(protoFx.length);
+          for (let i=0; i<protoFx.length; i++) {
+            instFx[i] = protoFx[i];
+          }
         }
       }
     }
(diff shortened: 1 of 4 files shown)


lib/mixins/property-effects.html
```diff
@@ -411,32 +416,122 @@
...
+  /**
+   * Returns whether any changed props match the effect trigger
+   *
+   * @param {!PropertyEffectsType} inst The instance to test
+   * @param {DataTrigger} trigger Descriptor
+   * @param {Object} changedProps Bag of current property changes
+   * @return {boolean} true when the property is dirty
+   * @private
+   */
+  function propertyIsDirty(inst, trigger, changedProps) {
```
TimvdLippe: I would put these functions before `runComputedEffect`. That makes reading these functions a bit easier, since `runComputedEffect` calls into these functions and I was confused which ones they were.

lib/mixins/property-effects.html
```diff
@@ -411,32 +416,122 @@
...
+   * @param {!PropertyEffectsType} inst The instance to test
+   * @param {?} info Computed effect metadata
+   * @return {Array<?>} Array of computed effect metadata for depenencies
+   * @private
+   */
+   function computedDependenciesFor(inst, info) {
+    let deps = info.computedDependencies;
+    if (!deps) {
+      deps = info.computedDependencies = [];
+      info.args.forEach(arg => {
```
TimvdLippe: I think this can be a `info.args.reduce` instead

lib/mixins/property-effects.html
```diff
@@ -411,32 +416,122 @@
         Object.assign(changedProps, inst.__dataPending);
         inputProps = inst.__dataPending;
         inst.__dataPending = null;
+        // Ensure all computed properties are de-duped against the same turn
+        dedupeId--;
       }
     }
   }
 
+  const TRANSITIVE_DEPENDENCY = '~transitive~dependency~';
```
TimvdLippe: Can we use a Symbol here to make sure we are always unique? (however unlikely someone names a property exactly this)

lib/mixins/property-effects.html
```diff
@@ -82,20 +84,23 @@
    *
    * @param {Object} model Prototype or instance
    * @param {string} type Property effect type
+   * @param {boolean=} cloneArrays Clone any arrays mapped in the map
```
TimvdLippe: Instead of adding a boolean parameter, I would rather see an `Object{cloneArrays:boolean}` per #4864

Issue #4757, opened 2017-07-28
Support data binding to promises


MaKleSoft:
This would be a killer feature imo, especially if used with computed bindings. Here is an example:

```html
<div>[[ _myAsyncValue(prop1, prop2) ]]</div>
```
```js
_myAsyncValue(prop1, prop2) {
  return new Promise((resolve, reject) => {
    // Something asynchronous, like doing an ajax call, spawning a web worker, etc.
  });
}
```
Thoughts?

rictic:
At the goals-level (before talking about implementation), I generally want to know and databind to three values per promise:

- a Boolean that's true while the promise is unfulfilled
- the resolution
- the rejection

(And I want resolution and rejection to be mutually exclusive, so at most one is non-null).

azakus:
Since you can already do something like this by binding to a property that is resolved by a promise, I don't think there is enough need to put this directly into the data binding system.

Example:
```js
static get properties() {
  return {
    asyncProperty: {
      type: Object
    }
}
//...
someFn() {
  new Promise((resolve, reject) => {
    //...
  }).then((result) => this.asyncProperty = result)
}

MaKleSoft:
Excuse me, but how is "it can already be done in a more roundabout way using significantly more boilerplate" a reason to not even consider this? By the same logic, you could say "Hey, we don't need promises at all, everything can be done with callbacks!" or "Arrow functions? Whats wrong with `.bind(this)`?" Hell, couldn't the same even be said for computed bindings? (Can be achieved using observers). 

With native Promises in ES2015 and `async/await` coming in ES2017, Promise are quickly becoming a first-class citizen in the JavaScript world and I think it would be an oversight to not treat them as such.

Now, can we please reopen this to continue the discussion instead of just dismissing it like this? I mentioned boilerplate as the most obvious argument but I see many more advantages to supporting Promises out of the box. Also, I'd like to hear a better argument against it than "it's not strictly needed".

dcbrwn:
Feature like this once was available in Angular 1.x but then was dismissed:
https://github.com/angular/angular.js/commit/5dc35b527b3c99f6544b8cb52e93c6510d3ac577

Even though I understand the desire to keep Polymer core as simple as possible, IMO @MaKleSoft is right. With promises literally everywhere it gets very annoying to do [this](https://github.com/Polymer/polymer/issues/4757#issuecomment-326157135) over and over again.

jasonwr:
[...]

PR #5146, opened 2018-03-11
Update property-effects.html - Add support for promises.

myuseringithub:
Adding support for promises in computed properties: 
A computed property returning a promise will now resolve and the data binding will render to the resolved value, instead of printing the un-resolved promise. Which will allow using async functions as values for properties.

**Before & After Example** 
For the element: 
```
class Element extends Polymer.Element {
    static get template() { return Polymer.html`<h1>[[localize('title')]]</h1>` }
    static get properties() {
        return {
            localize: {
                type: Function,
                value: () => async function(key) {
                    return await getIndexdbData(key) // Asynchronous call.
                }
            }
        } 
    }
}
```

_Before_ template would be rendered as: 
`<h1>{object Promise}</h1>`

_After_ changes template would render: 
`<h1>the resolved string data</h1>`

### DIFF
--- a/lib/mixins/property-effects.html
+++ b/lib/mixins/property-effects.html
@@ -576,6 +576,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
     } else {
       let value = info.evaluator._evaluateBinding(inst, part, path, props, oldProps, hasPaths);
+      if(Promise(value) == value) { // if 'value' is a promise.
+        // Resolve promise & propagate resolvedValue to child
+        value.then(resolvedValue => applyBindingValue(inst, node, binding, part, resolvedValue));
+        return;
+      }
       // Propagate value to child
       applyBindingValue(inst, node, binding, part, value);
     }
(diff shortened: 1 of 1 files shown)


TimvdLippe:
@myuseringithub thank you for your pull request. This feature request has come up before, as filed in #4757. However, per https://github.com/Polymer/polymer/issues/4757#issuecomment-326157135 we have closed that issue. This is primarily because of the nitty-gritty of the full implementation, where we have to handle the multiple states and provide meaningful data binding behavior for them. E.g. what to do when a promise fails? What should we render when a promise is still resolving? In this case, I would say `{object Promise}` is wrong, but leaving it empty maybe is as well.

Therefore, we prefer to leave this up to the user, even though we acknowledge that is more boilerplate. That way, you have full control over the workflow and render what you want, rather than us implementing one approach which is probably wrong for a large portion of the other use-cases.

Hopefully this makes clear why we are closing this PR. Thank you for your contribution nonetheless!

P.S. Note that, while this is a feature request for the Polymer data binding system, with [lit-html](https://github.com/Polymer/lit-html) you are already in a JavaScript context and therefore you have full control over the behavior. If you are heavily relying on promises in your components, `lit-html` is probably more appropriate for your use-case.

myuseringithub:
Thanks, that's right I will use lit-html then, with something like ` ${ (async () => await fetch())() } `.

PR #5153, opened 2018-03-15
Returning null in template should nullify parent template

### LINKED ISSUE #5148: null template should override inherited template
Currently, if an element extends a parent element that _has_ a template, there's no simple way to override that template and get an element with no template or shadow root.

```js
class ParentElement {
  static get template() { html`Haha! You're stuck with my template!`; }
}

class ChildElement extends ParentElement {
  // I don't want your template! Or anyone's template! Leave me alone!
  static get template() { return null; }
}
```

@kevinpschaaf suggested that an explicit falsey value like `null` should override the parent template. If the child template getter returns `undefined`, it will continue to inherit the parent template.

TimvdLippe:
The suspicion of @kevinpschaaf in https://github.com/Polymer/polymer/issues/5148#issuecomment-372491826 was correct, in that moving the assignment fixed the issue.
### Reference Issue
Fixes #5148

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -311,9 +311,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           } else {
             template = template.cloneNode(true);
           }
-          this.prototype._template = template;
         }
 
+        this.prototype._template = template;
       }
 
       /**
--- a/test/unit/inheritance.html
+++ b/test/unit/inheritance.html
@@ -113,6 +113,24 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     </script>
   </dom-module>
 
+  <dom-module id="parent-element-template-overriding">
+    <script>
+    HTMLImports.whenReady(function() {
+      class ParentElementWithTemplate extends Polymer.Element {
+        static get template() {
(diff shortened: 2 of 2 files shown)


test/unit/inheritance.html
```diff
@@ -237,6 +255,23 @@
...
+      el = document.createElement('child-with-no-template');
+      document.body.appendChild(el);
+    });
+
+    teardown(function() {
+      document.body.removeChild(el);
+    });
+
+    test('returning null nullifies the parent template', function() {
+      assert.isNotOk(el.shadowRoot);
```
azakus: Let's make this `assert.equal(el.shadowRoot, null)`

lib/mixins/element-mixin.html
```diff
@@ -311,9 +311,9 @@
           } else {
             template = template.cloneNode(true);
           }
-          this.prototype._template = template;
         }
 
+        this.prototype._template = template;
```
azakus: Nice easy fix!

arthurevans:
So, follow up: With this fix, does any falsy value work, as originally documented, or does it have to be `null`?

TimvdLippe:
Any falsy value should work.

arthurevans:
Thanks Tim!

PR #5154, opened 2018-03-16
Deduplicate style includes

### LINKED ISSUE #5139: Style module includes are not de-duplicated
### Description
Style modules includes are not de-duplicated. When the same module is imported multiple times within a shadowRoot, the styles are also inserted multiple times.

Currently the most common case where this will happen is when importing style module A which imports style module B, and then also import style module B yourself. Style module B is included twice. 

Another case where this will happen is when authoring templates in JS. Shared style templates could have the same style include which will cause style modules to be imported twice.

I know this isn't a documented use case of style modules, so it's not really a bug. But a developer probably expects the imports to be de-deplicated. Otherwise, they will need to know about internal implementation of consuming style module or elements.

#### Live Demo
https://jsbin.com/capowineda/1/edit?html,output

#### Steps to Reproduce
1. Create an element
2. Include style module A, which also includes style module B
3. Include style module B as well
4. Check the rendered shadowRoot, it will include module B twice

#### Expected Results
Module B is included once.

#### Actual Results
Module B is included twice.

### Browsers Affected
- A;;

### Versions
- Polymer: v2.x
- webcomponents: v1.x

TimvdLippe:
An easy `new Set()` deduplicates the style modules.
### Reference Issue
Fixes #5139

### DIFF
--- a/lib/utils/style-gather.html
+++ b/lib/utils/style-gather.html
@@ -123,7 +123,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           // other dom-modules that contain styling
           let include = e.getAttribute(INCLUDE_ATTR);
           if (include) {
-            styles.push(...this.stylesFromModules(include));
+            styles.push(...new Set(this.stylesFromModules(include)));
           }
           if (baseURI) {
             e.textContent = Polymer.ResolveUrl.resolveCss(e.textContent, baseURI);
--- a/lib/utils/style-gather.html
+++ b/lib/utils/style-gather.html
@@ -123,7 +123,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           // other dom-modules that contain styling
           let include = e.getAttribute(INCLUDE_ATTR);
           if (include) {
-            styles.push(...new Set(this.stylesFromModules(include)));
+            styles.push(...this.stylesFromModules(include).filter(function(item, index, self) {
+              return self.indexOf(item) === index;
+            }));
           }
           if (baseURI) {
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Tests are now passing again, so PTAL

TimvdLippe:
Hm I spoke too soon. It works in Firefox and Chrome, but some other untouched tests failed in Edge, IE and Safari. Chrome 41 actually failed on the newly added test. I have no clue why they would suddenly fail with the `new Set` :fearful:

azakus:
I don't think we polyfill `Set`, so that might explain the failures.

lib/utils/style-gather.html
```diff
@@ -123,7 +123,9 @@
           // other dom-modules that contain styling
           let include = e.getAttribute(INCLUDE_ATTR);
           if (include) {
-            styles.push(...new Set(this.stylesFromModules(include)));
+            styles.push(...this.stylesFromModules(include).filter(function(item, index, self) {
```
azakus: This looks O(n<sup>2</sup>) to me
I think this would be faster as
```js
let stylesFromModules = this.styleFromModules(include);
let uniqueStyles = Object.keys(this.stylesFromModules(include).reduce((acc, s) => {acc[s] = true; return acc}, {}));
styles.push(...uniqueStyles);
```

azakus: oh this wouldn't work because the array is of `<style>` elements.

TimvdLippe: Since most arrays will be small here, I think we can keep this for now. We can always revisit later (or just use Set someday...)

PR #5167, opened 2018-03-28
fix(path.html): split with undefined path

### LINKED ISSUE #5166: [bug] path.html split error with undefined path
### Description

We got error on `this.get(path)` when `path` is `undefined`

screen shot 2018-03-28 at 11 00 28 am

#### Live Demo
screen shot 2018-03-28 at 11 01 06 am

http://jsbin.com/fipavab/edit?html,console

AliMD:
We got error on `this.get(path)` when `path` is `undefined`
### Reference Issue
Fixes #5166

### DIFF
--- a/lib/utils/path.html
+++ b/lib/utils/path.html
@@ -191,7 +191,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (Array.isArray(path)) {
         return this.normalize(path).split('.');
       }
-      return path.toString().split('.');
+      else {
+        return String(path).split('.');
+      }
     },
 
     /**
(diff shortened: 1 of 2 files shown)


TimvdLippe:
Closing per https://github.com/Polymer/polymer/issues/5166#issuecomment-378350880

AliMD:
@TimvdLippe Could you please reopen this PR
I think using `path.toString()` is wrong and using String(path) is better for convert to string
Thank you in advance

TimvdLippe:
@AliMD Please see https://github.com/Polymer/polymer/issues/5166#issuecomment-378350880 for the explanation. In other words, `String(path)` ignores the core issue, while `path.toString()` fails fast. We want to make this as clear to the user as possible, as this API should NOT be called with either `undefined` or `null`.

PR #5170, opened 2018-03-30
Pass through fourth namespace param on attributeChangedCallback.

rictic:
This came up when doing some type checking of polymer core. `HTMLElement#attributeChanged` takes a fourth `namespace` param, so we should pass all four through.

### DIFF
--- a/lib/elements/dom-module.html
+++ b/lib/elements/dom-module.html
@@ -78,9 +78,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @param {string} name Name of attribute.
      * @param {?string} old Old value of attribute.
      * @param {?string} value Current value of attribute.
+     * @param {?string} namespace Attribute namespace.
      * @return {void}
      */
-    attributeChangedCallback(name, old, value) {
+    attributeChangedCallback(name, old, value, namespace) {
       if (old !== value) {
         this.register();
       }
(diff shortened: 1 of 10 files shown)


test/smoke/behavior-mixin.html
```diff
@@ -22,6 +22,7 @@
   </template>
 </dom-module>
 <script>
+  /** @polymerBehavior */
```
rictic: Oh, this slipped in from some other investigation. It's an improvement though so /shruggie

azakus:
Looks like you'll need a `/* eslint-disable no-unused-args */` and `/* eslint-enable no-unused-args */` around each of these.
Also, please run `npm run update-types`.

lib/elements/dom-module.html
```diff
@@ -78,9 +78,10 @@
      * @param {string} name Name of attribute.
      * @param {?string} old Old value of attribute.
      * @param {?string} value Current value of attribute.
+     * @param {?string} namespace Attribute namespace.
      * @return {void}
      */
-    attributeChangedCallback(name, old, value) {
+    attributeChangedCallback(name, old, value, namespace) {
```
justinfagnani: since dom-module isn't a mixin, why do we need to add the parameter? All the parameters should be optional from the implementation side, no?

rictic: I don't think this is necessary, but AIUI it's slightly nicer for the JS runtime. Happy to drop or not.

PR #5172, opened 2018-04-02
Add importMeta getter to derive importPath from modules. Fixes #5163

### LINKED ISSUE #5163: Improve ergonomics around defining importPath in modules
### Description
When defining elements in ES modules, the library cannot automatically find the import path to set `importPath` for data-binding use in templates (in HTML Imports the `ownerDocument` of the template was used to make this automatic).  As such, users must manually write a `static get importPath` getter in their module, with the notion was that they could just return `import.meta.url`.  However, `importPath` must be a folder URL ending in `/`, whereas `import.meta.url` returns a filename.

Ideally the library would accept a filename and take its folder name, or else we could consider accepting the `meta` object directly (via a differently named getter?) to allow more flexibility in the future.

### Versions
- Polymer: v3.x

kevinpschaaf:
In order to determine the import path for ES modules, rather than implement `static get importPath()` and return the _folder name_ of `import.meta.url`, users can implement the following getter (verbatim), and Polymer will use that when defined to determine the `importPath`:

```js
static get importMeta() { return import.meta; }
```

### Reference Issue
Fixes #5163 

cc: @arthurevans to roll into 3.0 docs

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -414,11 +414,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
             this._importPath = meta.url.slice(0, meta.url.lastIndexOf('/') + 1);
           } else {
             const module = Polymer.DomModule && Polymer.DomModule.import(/** @type {PolymerElementConstructor} */ (this).is);
-            if (module) {
-              this._importPath = module ? module.assetpath : '';
-            } else {
-              this._importPath = Object.getPrototypeOf(/** @type {PolymerElementConstructor}*/ (this).prototype).constructor.importPath;
-            }
+            this._importPath = (module && module.assetpath) || 
+              Object.getPrototypeOf(/** @type {PolymerElementConstructor}*/ (this).prototype).constructor.importPath;
           }
         }
         return this._importPath;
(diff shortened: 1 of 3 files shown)


lib/mixins/element-mixin.html
```diff
@@ -409,13 +409,33 @@
...
         if (!this.hasOwnProperty(JSCompiler_renameProperty('_importPath', this))) {
+          const meta = this.importMeta;
+          if (meta) {
+            this._importPath = meta.url.slice(0, meta.url.lastIndexOf('/') + 1);
+          } else {
             const module = Polymer.DomModule && Polymer.DomModule.import(/** @type {PolymerElementConstructor} */ (this).is);
-            this._importPath = module ? module.assetpath : '' ||
-            Object.getPrototypeOf(/** @type {PolymerElementConstructor}*/ (this).prototype).constructor.importPath;
+            if (module) {
+              this._importPath = module ? module.assetpath : '';
```
TimvdLippe: Can be just `module.assetpath`, as the condition already asserted for `module`

lib/mixins/element-mixin.html
```diff
@@ -409,13 +409,30 @@
        */
       static get importPath() {
         if (!this.hasOwnProperty(JSCompiler_renameProperty('_importPath', this))) {
+          const meta = this.importMeta;
+          if (meta) {
+            this._importPath = meta.url.slice(0, meta.url.lastIndexOf('/') + 1);
```
sorvell: use `resolveUrl.pathFromUrl`

PR #5176, opened 2018-04-03
Remove unnecessary local alias for Polymer.Element

aomarks:


TimvdLippe:
TypeScript says no about this change though:

```
class Element extends
     Polymer.ElementMixin(
-    HTMLElement) {
+    Polymer.Element) {
   }
```

I don't think `Element` should extend itself here?

aomarks:
I think we found another way to get this working with non-new type inference, closing.

PR #5177, opened 2018-04-03
Fix broken code highlighting in README.md

ExE-Boss:
This fixes broken code highlighting in [README.md](https://github.com/Polymer/polymer/blob/master/README.md).

## Before
Polymer-README-broken

## After
Polymer-README-fixed


TimvdLippe:
Closing this as it appears to be stale and we have updated the README in the mean time.

PR #5179, opened 2018-04-05
Add {}s to be more consistent with Google style

MatrixFrog:
Closure Compiler understands both forms, so this isn't blocking us. It's just tripping me up slightly.


TimvdLippe:
A similar change was made in #5176, but in both cases the TypeScript compilation of this mixin is now wrong (https://github.com/Polymer/polymer/pull/5176#issuecomment-378371070) This might be an issue in the analyzer/gen-typescript-declarations, or a wrong typing in this repository.

MatrixFrog:
Google is such a fascinating place. There is no such thing as a change so trivial it cannot somehow break something.

PR #5180, opened 2018-04-05
Add ability to define importMeta on hybrid elements. Fixes #5163

### LINKED ISSUE #5163: Improve ergonomics around defining importPath in modules

kevinpschaaf:
After this change, hybrid elements defined in ES modules using legacy API can ensure a correct `importPath` as follows:

```js
Polymer({
  ...
  importMeta: import.meta
});
```

This change also removes the default `static get importMeta()` implementation, for better compatibility with defining element metadata using (compiled) static class fields.

### Reference Issue
Fixes #5163

### DIFF
--- a/lib/legacy/legacy-element-mixin.html
+++ b/lib/legacy/legacy-element-mixin.html
@@ -87,6 +87,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
 
       /**
+       * Forwards `importMeta` from the prototype (i.e. from the info object
+       * passed to `Polymer({...})`) to the static API.
+       *
+       * @return {!Object} The `import.meta` object set on the prototype
+       */
+      static get importMeta() {
+        return this.prototype.importMeta;
+      }
+
+      /**
        * Legacy callback called during the `constructor`, for overriding
        * by the user.
        * @return {void}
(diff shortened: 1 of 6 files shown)


lib/legacy/legacy-element-mixin.html
```diff
@@ -86,6 +86,17 @@
         this._applyListeners();
       }
 
+      /**
+       * Forwards `importMeta` from the prototype (i.e. from the info object
+       * passed to `Polymer({...})`) to the static API.
+       *
+       * @return {!Object} The `import.meta` object set on the prototype
+       * @suppress {missingProperties}
```
azakus: Please add reasoning.

\`this\` is always in the instance in closure for some reason.

kevinpschaaf:
cc @aomarks @justinfagnani Support for `import.meta` injection in legacy API

PR #5184, opened 2018-04-09
update npm dependencies

azakus:


TimvdLippe:
These were updated in separate branches already.

PR #5193, opened 2018-04-16
3.x modularized

rictic:
This is proposing to merge the output of polymer-modularizer, plus some minor fixup commits, into the new `3.x` branch. 

If this PR is merged, it will mark the point at which we will stop running modulizer on Polymer core, and we'll fix the remaining issues via normal PRs against the `3.x` branch. For elements the plan of record is still that we'll be using modulizer.

As of now, the `3.x` branch has two commits on top of master:

03d85982 – Delete typings for now. (removes the `types/` directory)
527d2cdd – Rename HTML files to .js files to trick git's rename detection.

These are so that this PR's diff is easy to read and review. I'd recommend reviewing this with `?w=1`, or on the command line with `-w --word-diff`.

The tests run, at least in chrome, safari, and firefox, with the command line `wct --npm`. I fixed any issues that caused them to immediately and catastrophically fail, but I haven't started to dive deeply into the remaining failures. In Safari e.g. I'm seeing 3566 passing tests and 104 failing.

This was run with the https://github.com/Polymer/polymer-modulizer/pull/401 branch of modulizer.


rictic:
So far I haven't gotten `git merge` to work at merging existing PRs into this branch. I asked a question on stackoverflow to see if there's some git wizardry that can help: https://stackoverflow.com/questions/49867161/rename-detection-with-whitespace-ignored-in-git-merge

rictic:
Per stack overflow, doing the following kinda works:

```
  git checkout pr-branch
  git merge 527d2cd # merge with the commit that just renames .html files to .js
  git merge 3.x-modularized  -s recursive -Xignore-all-space -Xpatience # merge into the real branch
```

The result doesn't have super tight constraints on which lines changed, and the result has indentation issues. It seems like an autoformatter would be a big productivity boost here.

rictic:
Merging this into 3.x as it's an improvement over where 3.x is right now.

PR #5197, opened 2018-04-17
Add `@const` to elements

rslawik:
Adding `@const` helps JS compiler understand that `Polymer.DomIf` is an alias for the class defined just above that.

With this, the following code can compiled:
```javascript
const /** !Polymer.DomIf */ domIf = /** @type {!Polymer.DomIf} */ (this.$['dom-if']);
domIf.render();
```

Added `@const` to all elements under `lib/elements` to expose their types.


rslawik:
Not sure what happened with that. I exported the commits before configuring git properly. Can the CLA state be fixed for this pull request?

TimvdLippe:
I think if you force push with the correct git credentials it should be resolved.

PR #5199, opened 2018-04-17
Adding support for data uris in resolveUrl.

### LINKED ISSUE #5198: Polymer doesn't handle data uri import links properly.
### Description
Uncaught error when trying to create a link import element with a [data uri](https://en.wikipedia.org/wiki/Data_URI_scheme) as its href attribute.

#### Live Demo
https://jsbin.com/votetokeye/edit?html,output

#### Steps to Reproduce
1. Create a <link> element with a [data uri](https://en.wikipedia.org/wiki/Data_URI_scheme) as its href attribute.

#### Expected Results
No error thrown.

#### Actual Results
Opening the javascript console will reveal the following:
```
Uncaught TypeError: Failed to construct 'URL': Invalid URL
    at Object.resolveUrl (resolve-url.html:52)
    at HTMLElement.get assetpath [as assetpath] (dom-module.html:110)
    at Function.get importPath [as importPath] (element-mixin.html:425)
    at HTMLElement._initializeProperties (element-mixin.html:469)
    at HTMLElement._initializeProperties (legacy-element-mixin.html:188)
    at new PropertiesChanged (properties-changed.html:166)
    at new PropertyAccessors (property-accessors.html:118)
    at new TemplateStamp (template-stamp.html:121)
    at new PropertyEffects (property-effects.html:1138)
    at new PropertiesMixin (properties-mixin.html:110)
```

### Browsers Affected
All, this is a core issue.

### Versions
- Polymer: v2.6.0
- webcomponents: v1.2.0

nicolasnoble:
Fixes #5198.


test/unit/data-uri-link.html
```diff
@@ -0,0 +1,28 @@
...
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<html>
+<head>
+  <script src="../../../webcomponentsjs/webcomponents-loader.js"></script>
+  <script src="../../../web-component-tester/browser.js"></script>
+  <link rel="import" href="../../polymer.html">
+  <link rel="import" href="data:text/html,%3Cdom-module%20id%3D%22element-data-uri%22%3E%3Cscript%3EPolymer(%7Bis%3A%22element-data-uri%22%2Cproperties%3A%7Bexists%3A%7Btype%3ABoolean%2Cvalue%3Atrue%7D%7D%7D)%3C%2Fscript%3E%3C%2Fdom-module%3E">
```
TimvdLippe: Would it be possible to put this into code with the URLencode? That would make the test more readable and understandable.

nicolasnoble: Possibly, yes. I can try at least :-)

nicolasnoble: Updated; PTAL.

nicolasnoble: Actually, what I attempted ([visible here](https://github.com/nicolasnoble/polymer/commit/b624a30ac92822c8332cce64af09eca4f7cae203)) works locally, but seems to timeout on tests on your travis environment. I'm not sure how to do this otherwise.

nicolasnoble:
My guess is the onload trick isn't properly working everywhere for all browsers, but I am not sure how to solve it.

Worst case scenario, for readability, it might be relevant to just add a comment next to the link that shows what its contents are.

Also I would believe the test would be more relevant to use a Polymer component because that way it exercises the resolveURL codepath - your attempt doesn't.

I can rearrange my original PR to add the comment block if you want.

TimvdLippe:
Yes lets resort to the comment. Strange thing is that it works just fine on
Firefox and Chrome locally, but on CI it borks.

nicolasnoble:
Yep, I have the same experience: it works fine locally for me too, so I don't exactly know what's going on with the CI... Oh well :-)

I've updated the original PR to add a comment. PTAL.

nicolasnoble:
Sure. Is there going to be a planned Polymer 2.0 release with this fix possibly in there however ? I wouldn't like to have to upgrade my project immediately to Polymer 3.0 just to get this fix in there.

TimvdLippe:
Hm, I am actually not sure what happens in that case. I will discuss with the others and see what we can do!

nicolasnoble:
Okay, thanks. I mean, most of the time, substantials projects still get "maintenance fixes and releases" for a little while before abandoning the previous version, and Polymer 2.6.0 got released less than a month ago. It'd be quite harsh to drop support for it, forcing people to migrate to get fixes like this one in there.

I mean, I could still monkey-patch my Polymer 2.6.0 during postinstall to incorporate this fix instead of waiting on a hypothetical 2.6.1, but I'd rather not.

TimvdLippe:
Yeah I am definitely not saying we are dropping support, we simply havent decided yet what our plans are. We still support 1.x with bug fixes, so we will for certain do that with 2.x as well.

nicolasnoble:
Cool, thanks! :)

nicolasnoble:
Gentle ping. I would actually love to see this one backported to the older polymer 2.x branch too.

nicolasnoble:
Yeah it's stale alright.

PR #5200, opened 2018-04-17
3.x tests

e111077:
Chrome failing tests 56 -> 17. Remaining are assigned to Dan and we have a plan of action for each assigned to him already.

For info on what I fixed and what was wrong, see [this doc](https://docs.google.com/document/d/1wIW8GG3MjEo_K_yqGRBB9KB5sg1lX4fMqTg-FNvUb2Y/edit?usp=sharing) (visible only to googlers).

CC @azakus


test/unit/shady-unscoped-style.html
```diff
@@ -99,26 +99,12 @@
   var el1 = document.querySelector('my-element');
   var el2 = document.querySelector('my-element2');
 
-  test('unscoped styles apply', function() {
-    assertComputed(el1.$.fromStyle, '1px');
-    assertComputed(el1.$.fromImport, '2px');
-    assertComputed(el2.$.fromStyle, '1px');
-    assertComputed(el2.$.fromImport, '2px');
-  });
```
kevinpschaaf: Why are all these tests removed?  The `fromStyle` ones seem to just be style module-related...

e111077: good point.

e111077: The second set are deleted because there is nothing to dedupe since we don't have the css html import

e111077: relevant tests added back in

PR #5202, opened 2018-04-20
Fix ESLint gulp task for 3.x

TimvdLippe:
This makes `gulp-eslint` pass again on our source code. I started working on the Closure stuff, but that seems more complicated and warrants a separate PR.


kevinpschaaf:
Merging this into 3.x; there is a short list of remaining issues keeping tests from green.

PR #5204, opened 2018-04-24
Improve documentation and legibility in 3.x.

rictic:
- Did a pass for every place we were writing `Polymer.foo`.
- Basic cleanup of a number of places for legibility by both humans and polymer-analyzer.

### DIFF
--- a/lib/utils/async.js
+++ b/lib/utils/async.js
@@ -7,6 +7,19 @@ The complete set of contributors may be found at http://polymer.github.io/CONTRI
 Code distributed by Google as part of the polymer project is also
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 */
+
+/**
+ * @fileoverview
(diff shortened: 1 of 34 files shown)


rictic:
Oh, and I removed all the now incorrect `@memberof` annotations.

lib/utils/async.js
```diff
@@ -33,30 +46,18 @@ function microtaskFlush() {
...
-and this string can then be deleted`;
-
 /**
  * Async interface wrapper around `setTimeout`.
  *
+ * @namespace
  * @summary Async interface wrapper around `setTimeout`.
  */
-export const timeOut = {
+const timeOut = {
```
TimvdLippe: Why are these not directly exported but others are?

rictic:
Noting that this is part of https://github.com/Polymer/docs/issues/2379

PR #5209, opened 2018-04-27
Updated package-lock to grab latest version of 'request' npm package transitively through recent wct-sauce release.

usergenic:


azakus:
I'll just fix it in master

PR #5223, opened 2018-05-09
Remove importHref from 3.0

kevinpschaaf:
Remove importHref, as HTML Import support has been removed from 3.0.


beaudesigns:
The [Polymer 3.0 API Reference](https://www.polymer-project.org/3.0/docs/api/) still includes the page and information for `import-href`: https://www.polymer-project.org/3.0/docs/api/utils/import-href

It threw me off for a while during my upgrade.

PR #5230, opened 2018-05-10
Only disable tabs for "disablable" elements

### LINKED ISSUE #5190: click() doesn't invoke tap listener on a disabled element
JS:
```javascript
Polymer({
  is: 'x-foo',
  properties: {
    disabled: {
      type: Boolean,
      reflectToAttribute: true,
    },                  
  },
  listeners: {
    'tap': 'onTap',
  },
  onTap: () => {
    console.log('clicked!');
  }
});
```
html:
```html
  <x-foo></x-foo>
  <x-foo disabled></x-foo>
```

in v1:
```javascript
normal.click() // tapHandler is called
disabled.click() // tapHandler is caller
```

in v2:
```javascript
$normal.click() // tap Handler is called
$disabled.click() // tapHandler is *NOT* called
```

#### Live Demo
See minimal repro here (modify import to switch between v1/v2):
https://jsbin.com/zusabajelu/1/edit?html,console,output

probably related to #4685 and #5128

azakus:
Fixes #5190

### DIFF
--- a/lib/utils/gestures.html
+++ b/lib/utils/gestures.html
@@ -113,6 +113,19 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     'select': true
   };
 
+  /** @type {!Object<boolean>} */
+  const canBeDisabled = {
+    'button': true,
+    'command': true,
+    'fieldset': true,
+    'input': true,
+    'keygen': true,
+    'optgroup': true,
+    'option': true,
+    'select': true,
+    'textarea': true
+  };
+
   /**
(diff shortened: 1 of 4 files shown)


lib/utils/gestures.html
```diff
@@ -113,6 +113,19 @@
     'select': true
   };
 
+  /** @type {!Object<boolean>} */
```
TimvdLippe: Maybe add a link to MDN/standard to the list where this is specified?

kevinpschaaf:
Once you're ready, let's make sure to merge this into `master` also.

PR #5233, opened 2018-05-15
Update template docs.

Westbrook:
In support of #5222, this updated the `template()` documentation so that it no longer lists support for `typeof template === 'string'` types.

### Reference PR
Follow up: #5222


TimvdLippe:
Note for future self: while you thought this change was harmless, do not try to be smart again and merge without Travis passing. In 3.x we do not run the typescript compilation, but if we were this would have failed.

PR #5245, opened 2018-05-28
Add bower_components and dist to .npmignore.

### LINKED ISSUE #5244: bower_components is published to npm.
### Description
After running `npm install @polymer/polymer` you get everything from polymer 2 into `node_modules/@polymer/polymer/bower_components/`.

#### Steps to Reproduce

1. Execute `npm install @polymer/polymer`

#### Expected Results
`node_modules/@polymer/polymer` should not contain bower_components.

#### Actual Results
`node_modules/@polymer/polymer` is 16MB.  Of that 14MB is bower_components (polymer-2).  In addition a `dist` folder is included which appears to be from polymer 2.

### Versions
- Polymer: v3.0.2

coreyfarrell:
Fixes #5244

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,0 @@
-test/
-util/
-img/
-\..*
--- a/package.json
+++ b/package.json
@@ -64,5 +64,12 @@
   },
   "dependencies": {
     "@webcomponents/shadycss": "^1.2.0"
-  }
+  },
+  "files": [
+    "externs",
+    "lib",
+    "index.html",
+    "polymer-*.js",
+    "manifest.json"
+  ]
 }


coreyfarrell:
I've switched to using `files`, deleted `.npmignore`.  This cleans a couple more files that are development only (detailed in commit message).  Redundant entries are not listed, see [npmjs docs](https://docs.npmjs.com/files/package.json#files) for files that are automatically included.  You can test this using `npm pack`, then inspect the tarball that is created.

With this change `@polymer/polymer` will be 1022kb unpacked.  Of this `CHANGELOG.md` (automatically included by `npm pack`) is 324k, 31% of the total install.  Maybe CHANGELOG.md should be split so it only includes the current semver-major, have a CHANGELOG-ARCHIVE.md or similar containing entries for legacy versions, add an a link to the end of the current changelog?  Not included in this commit as decreasing from 14936kb to 1022kb is a huge improvement but I still feel it's worth considering not allowing the CHANGELOG.md included in the package to expand indefinitely.

PR #5254, opened 2018-06-11
Replaces an array spread of a NodeList with `Array.from` because NodeList isn't iterable in Safari 9.

bicknellr:


azakus:
Failure in Safari 11 is due to something unrelated in the 11.1 release.

e111077:
Heya, what's the status on this. Do you know when this is going to be released?

PR #5261, opened 2018-06-16
Ensure properties is only invoked once

### LINKED ISSUE #5252: properties is called twice for each class
_From @blikblum on May 11, 2018 20:18_

This occurs because the check to `constructor.properties` value at https://github.com/Polymer/polymer/blob/master/lib/mixins/properties-mixin.js#L89

_Copied from original issue: Polymer/lit-element#55_

TimvdLippe:
The `static get properties` was invoked twice because of an `if`-guard. Instead, invoke it once and reassign to a variable and use that in the `if`-guard instead.
### Reference Issue
Fixes #5252


TimvdLippe:
Test failure is in `dom-repeat` which seems unrelated. Not sure why this is suddenly breaking. Works just fine on master.

TimvdLippe:
This PR is ready for review again

PR #5266, opened 2018-06-21
[2.x] Add legacy-data-mixin as 1.x->2.x migration aide. Fixes #5262.

### LINKED ISSUE #5262: Add 2.x option to treat complex observers as they were in Polymer1

kevinpschaaf:
Add legacy-data-mixin as 1.x->2.x migration aide. 

Mixin to selectively add back Polymer 1.x's `undefined` rules governing when observers & computing functions run based on all arguments being defined (reference https://www.polymer-project.org/1.0/docs/devguide/observers#multi-property-observers).

When `polymer/lib/legacy/legacy-data-mixin.html` is loaded at the app level, all legacy elements (defined with `Polymer({...})`) will have the mixin applied. The mixin only restores legacy data handling if `_legacyUndefinedCheck: true` is set on the element's prototype.

This mixin is intended for use to help migration from Polymer 1.x to 2.x+ by allowing legacy code to work during the migration period, while identifying observers and computing functions that need undefined checks to work without the mixin in Polymer 2.  It is generally not intended for use in production.

### Reference Issue
Fixes #5262.
Supersedes and closes #5263


kevinpschaaf:
This is an alternate implementation of https://github.com/Polymer/polymer/pull/5263, based on same concept there, with the goal of making the code to implement the legacy behavior purely opt-in.  cc: @beckysiegel

kevinpschaaf:
Two affordances were made in core to enable writing such a mixin:
1. Allow `marshalArgs` to be overridable via mixin my moving it from a private function to a protected function
2. Extend the API for `Polymer.Class` to allow a mixin to be applied prior to desugaring the info object into subclasses of the base LegacyElementMixin.

These seem like generally good changes on their own.

TimvdLippe:
Also the tests appear to fail on IE11, likely related to how it handles exceptions. And the latest version of shadydom appears to be now failing some of our tests. We have to look into these regressions separately.

beckysiegel:
Thanks!

lib/mixins/property-effects.html
```diff
@@ -2180,6 +2130,56 @@
...
+       * The `path` and `value` arguments are used to fill in wildcard descriptor
+       * when the method is being called as a result of a path notification.
+       *
+       * @param {!Array<!MethodArg>} args Array of argument metadata
+       * @param {string} path Property/path name that triggered the method effect
+       * @param {Object} props Bag of current property changes
+       * @return {Array<*>} Array of argument values
+       * @private
+       */
+      _marshalArgs(args, path, props) {
```
samccone: this function has a non-trivial cyclical complexity score, can you please add unit tests to ensure the functionality is correct. Also it might be a sign that the logic should be split up into smaller more meaningful functions to maintain a higher degree of readability.

lib/mixins/property-effects.html
```diff
@@ -2180,6 +2130,56 @@
...
+      /**
+       * Gather the argument values for a method specified in the provided array
+       * of argument metadata.
+       *
+       * The `path` and `value` arguments are used to fill in wildcard descriptor
+       * when the method is being called as a result of a path notification.
+       *
+       * @param {!Array<!MethodArg>} args Array of argument metadata
+       * @param {string} path Property/path name that triggered the method effect
+       * @param {Object} props Bag of current property changes
```
samccone: what is a bag? perhaps restate as a "hash" or key value collection of ...

test/unit/legacy-data.html
```diff
@@ -0,0 +1,278 @@
...
+<head>
+  <meta charset="utf-8">
+  <script src="../../../webcomponentsjs/webcomponents-lite.js"></script>
+  <script src="../../../web-component-tester/browser.js"></script>
+  <link rel="import" href="../../polymer.html">
+  <link rel="import" href="../../lib/legacy/legacy-data-mixin.html">
+</head>
+<body>
+
+  <script>
```
samccone: can you drop this since it is empty?

test/unit/legacy-data.html
```diff
@@ -0,0 +1,278 @@
...
+        el.b = undefined;
+        el.c = undefined;
+        assert.equal(el.static.callCount, 1);
+        assert.equal(el.one.callCount, 2);
+        assert.equal(el.two.callCount, 3);
+        assert.equal(console.warn.callCount, 0);
+        document.body.removeChild(el);
+      });
+    });
+    // ------------------------------------------------
```
samccone: is this comment needed?

lib/legacy/legacy-data-mixin.html
```diff
@@ -0,0 +1,105 @@
...
+  const UndefinedArgumentError = class extends Error {};
+
+  /**
+   * Mixin to selectively add back Polymer 1.x's `undefined` rules
+   * governing when observers & computing functions run based
+   * on all arguments being defined (reference https://www.polymer-project.org/1.0/docs/devguide/observers#multi-property-observers).
+   * 
+   * When loaded, all legacy elements (defined with `Polymer({...})`)
+   * will have the mixin applied. The mixin only restores legacy data handling
+   * if `_legacyUndefinedCheck: true` is set on the element's prototype.
```
samccone: any reason why we are _ prefixing this prototype value?

test/unit/legacy-data.html
```diff
@@ -0,0 +1,278 @@
...
+        el.b = 'b';
+        el.c = 'c';
+        document.body.appendChild(el);
+        assert.equal(el.static.callCount, 1);
+        assert.equal(el.one.callCount, 1);
+        assert.equal(el.two.callCount, 1);
+        el.a = undefined;
+        el.b = undefined;
+        assert.equal(el.static.callCount, 1);
+        assert.equal(el.one.callCount, 2);
```
samccone: instead of callCount spy's can we sideeffect in the calls by incrementing a counter and then the tests are totally isolated without having to spy or mock on anything.


For example

```
foo(a, b) {
   this.fooCall++;
}
```

test/unit/legacy-data.html
```diff
@@ -0,0 +1,278 @@
...
+  <test-fixture id="declarative-two">
+    <x-data a="a" b="b"></x-data>
+  </test-fixture>
+
+  <test-fixture id="declarative-all">
+    <x-data a="a" b="b"></x-data>
+  </test-fixture>
+
+  <script>
+  suite('imperative', () => {
```
samccone: This suite of tests has a lot of duplication I think we could do with some test helpers to significantly clean up the test implementation to make the code easier to maintain and easier to reason about.

test/unit/shady-content.html
```diff
@@ -12,10 +12,9 @@
 <head>
   <meta charset="utf-8">
   <script>
-    if (!window.customElements) {
-      window.customElements = {};
+    if (window.customElements) {
+      customElements.forcePolyfill = true;
```
samccone: this change seems unrelated to the intent of the CL?

lib/legacy/legacy-data-mixin.html
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+    }
+
+    return LegacyDataMixin;
+
+  });
+
+  const Class = Polymer.Class;
+  Polymer.Class = info => Class(info, Polymer.LegacyDataMixin);
```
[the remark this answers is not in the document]
kevinpschaaf: Ah good catch.  The mixin option was a late addition.

test/unit/legacy-data.html
```diff
@@ -0,0 +1,278 @@
...
+      test('two arguments defined', () => {
+        const el = document.createElement('x-data');
+        el._legacyUndefinedCheck = true;
+        el.a = 'a';
+        el.b = 'b';
+        document.body.appendChild(el);
+        assert.equal(el.static.callCount, 1);
+        assert.equal(el.one.callCount, 1);
+        assert.equal(el.two.callCount, 0);
+        assert.isTrue(console.warn.calledOnce);
```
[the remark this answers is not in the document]
kevinpschaaf: Yup, thanks.

test/unit/legacy-data.html
kevinpschaaf: Did some refactoring

test/unit/shady-content.html
kevinpschaaf: Separated out to https://github.com/Polymer/polymer/pull/5268

test/unit/legacy-data.html
samccone: thanks, much nicer!

test/unit/shady-content.html
samccone: thank you

vdonich:
Looks like Computed bindings (https://www.polymer-project.org/1.0/docs/devguide/data-binding#annotated-computed) are not handled by this mixin.
Here's jsfiddle that demonstrates it: https://jsfiddle.net/gaj0umLs/13/

kevinpschaaf:
@vdonich Thanks for the report. I confirmed that `TemplateInstance`s created from `dom-if`/`dom-repeat` don't get the `LegacyDataMixin` applied, hence the `undefined` check isn't running.  Will look into a fix for that.

kevinpschaaf:
Addressed the `dom-if`/`dom-repeat` issue pointed out by @vdonich here: https://github.com/Polymer/polymer/pull/5266/commits/55e9dfd1517e40b8fb1be0e91c0578e6800ed019

vdonich:
I've hit another snag when using wildcards.

Observer is being called despite undefined property value.
Change record is valid, but all value-related properties such as changeRecord.base and changeRecord.value are undefined:

{base: undefined, path: "alwaysUndefinedObject", value: undefined}

Here's the jsfiddle with example:
https://jsfiddle.net/gaj0umLs/26/

PR #5270, opened 2018-06-25
Port Closure Compiler fixes to polymer-externs

azakus:


azakus:
Ugh, firefox test failures are a different problem.

PR #5280, opened 2018-06-30
Prevent loading duplicate of custom element

### LINKED ISSUE #5279: Duplicate loading of custom element should not crash the app
### Description
Loading a custom element that depends on other custom elements that are already imported by other custom elements in the app will prevent the page from loading. While trying to define a custom element with same name should be avoided, if the defining custom element is same as the defined custom element (i.e. duplicate definition) instead of crashing the app (or subset of the app), it should warn the developer that the custom element is already loaded and let app to continue loading the rest of the app.

#### Steps to Reproduce
1. Start new polymer starter kit project
```bash
polymer init
? Which starter template would you like to use? polymer-3-starter-kit
```
2. Install @polymer/paper-button
```bash
npm i @polymer/paper-button
```
3. Import @polymer/paper-button/paper-button.js to use in the app.
```js
import "@polymer/paper-button/paper-button.js";
```

#### Expected Results

Page should load properly.

#### Actual Results

Page does not load because an error gets thrown while trying to define new custom element.

This is because the starter project imports ```@polymer/paper-icon-button/paper-icon-button``` which defines ```<paper-ripple>```. When I import  ```@polymer/paper-button/paper-button.js``` this also tries to define ```<paper-ripple>``` which stops the page from loading.

##### Console Logs
Chrome
```
Uncaught (in promise) DOMException: Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry
    at Polymer (http://127.0.0.1:8081/node_modules/@polymer/polymer/lib/legacy/polymer-fn.js:43:18)
    at http://127.0.0.1:8081/node_modules/@polymer/paper-button/node_modules/@polymer/paper-ripple/paper-ripple.js:314:1
```
Firefox
```
Error: A custom element with name 'paper-ripple' has already been defined.
```
Safari
```
Unhandled Promise Rejection: NotSupportedError: Cannot define multiple custom elements with the same tag name
```
### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

### Versions

- Polymer: v3.0.0
- webcomponents: v2.0.0

dakotaJang:
### Reference Issue
Fixes #5279

### DIFF
--- a/lib/legacy/polymer-fn.js
+++ b/lib/legacy/polymer-fn.js
@@ -39,7 +39,11 @@ const Polymer = function(info) {
   } else {
     klass = Polymer.Class(info);
   }
-  customElements.define(klass.is, /** @type {!HTMLElement} */(klass));
+  if (!customElements.get(klass.is)){
+    customElements.define(klass.is, /** @type {!HTMLElement} */(klass));
+  } else {
+    console.log(`Already loaded custom element named ${klass.is}`)
+  }
   return klass;
 };
 


TimvdLippe:
This is native behavior and not specific to Polymer. There are standards suggestions to tackle this problem instead: https://github.com/w3c/webcomponents/issues/716 Therefore, we should not make such a fix here, rather update the standard to handle this in the proper way. Feel free to voice your opinion in the linked issue to participate in the discussion.

PR #5295, opened 2018-07-18
[2.x] Introduce strictTemplatePolicy

### LINKED ISSUE #5323: Introduce strictTemplatePolicy
### Description

kevinpschaaf:
Enabling `strictTemplatePolicy` by setting `Polymer = {setStrictTemplatePolicy: true};` prior to loading Polymer causes the following changes in behavior:
* Elements that do not have templates by design must explicitly indicate a null template to prevent template retrieval from injected `<dom-module>`s, otherwise they will throw an error

Fixes #5323

### DIFF
--- a/lib/elements/dom-bind.html
+++ b/lib/elements/dom-bind.html
@@ -56,6 +56,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       constructor() {
         super();
+        if (window.strictTemplatePolicy) {
+          throw new Error(`strictTemplatePolicy: dom-bind not allowed`);
+        }
         this.root = null;
         this.$ = null;
         this.__children = null;
--- a/lib/elements/dom-bind.html
+++ b/lib/elements/dom-bind.html
@@ -56,7 +56,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       constructor() {
         super();
-        if (window.strictTemplatePolicy) {
+        if (Polymer.strictTemplatePolicy) {
           throw new Error(`strictTemplatePolicy: dom-bind not allowed`);
         }
         this.root = null;
--- a/lib/elements/dom-bind.html
+++ b/lib/elements/dom-bind.html
@@ -24,6 +24,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * @implements {Polymer_PropertyEffects}
      * @implements {Polymer_OptionalMutableData}
      * @implements {Polymer_GestureEventListeners}
+     * @private
      */
     const domBindBase =
       Polymer.GestureEventListeners(
--- a/lib/elements/dom-if.html
+++ b/lib/elements/dom-if.html
@@ -255,8 +255,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         if (c$ && c$.length) {
           // use first child parent, for case when dom-if may have been detached
           let parent = c$[0].parentNode;
-          for (let i=0, n; (i<c$.length) && (n=c$[i]); i++) {
-            parent.removeChild(n);
+          if (parent) {
+            for (let i=0, n; (i<c$.length) && (n=c$[i]); i++) {
+              parent.removeChild(n);
+            }
           }
         }
         this.__instance = null;
--- a/lib/elements/dom-if.html
+++ b/lib/elements/dom-if.html
@@ -255,6 +255,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         if (c$ && c$.length) {
           // use first child parent, for case when dom-if may have been detached
           let parent = c$[0].parentNode;
+          // Instance children may be disconnected from parents when dom-if
+          // detaches if a tree was innerHTML'ed
           if (parent) {
             for (let i=0, n; (i<c$.length) && (n=c$[i]); i++) {
               parent.removeChild(n);
--- a/lib/elements/dom-module.html
+++ b/lib/elements/dom-module.html
@@ -15,6 +15,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   let modules = {};
   let lcModules = {};
+  function setModule(id, module) {
+    // store id separate from lowercased id so that
+    // in all cases mixedCase id will stored distinctly
+    // and lowercase version is a fallback
+    modules[id] = lcModules[id.toLowerCase()] = module;
+  }
   function findModule(id) {
     return modules[id] || lcModules[id.toLowerCase()];
   }
@@ -124,16 +130,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     register(id) {
       id = id || this.id;
       if (id) {
-        if (Polymer.strictTemplatePolicy && findModule(id)) {
-          modules[id] = lcModules[id.toLowerCase()] = null;
-          throw new Error(`strictTemplatePolicy: dom-module ${id} registered twice`);
+        // Under strictTemplatePolicy, reject and null out any re-registered
+        // dom-module since it is ambiguous whether first-in or last-in is trusted 
+        if (Polymer.strictTemplatePolicy && findModule(id) !== undefined) {
+          setModule(id, null);
+          throw new Error(`strictTemplatePolicy: dom-module ${id} re-registered`);
         }
         this.id = id;
-        // store id separate from lowercased id so that
-        // in all cases mixedCase id will stored distinctly
-        // and lowercase version is a fallback
-        modules[id] = this;
-        lcModules[id.toLowerCase()] = this;
+        setModule(id, this);
         styleOutsideTemplateCheck(this);
       }
     }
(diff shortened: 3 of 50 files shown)


test/unit/strict-template-policy.html
```diff
@@ -0,0 +1,240 @@
...
+          document.getElementById('target').innerHTML =
+            '<dom-module id="trusted-element">' +
+            '  <template>' +
+            '    <div id="injected"></div>'+
+            '  </template>`' +
+            '</dom-module>' +
+            '<trusted-element></trusted-element>';
+        }, /trusted-element registered twice/);
+        const el = document.querySelector('trusted-element');
+        assert.notOk(el && el.shadowRoot && el.shadowRoot.querySelector('#injected'));
```
sorvell: Remove `&& el.shadowRoot.querySelector('#injected')` as it's enough to assert that the element does not have a shadowRoot when the template gets nulled due to 2x register.

sorvell: Separate the creation of the `trusted-element` so an innerHTML error does not disrupt element creation.

test/unit/strict-template-policy.html
```diff
@@ -0,0 +1,240 @@
...
+
+      test('dom-module after registration (legacy)', function() {
+        assertThrows(function() {
+          document.getElementById('target').innerHTML =
+            '<dom-module id="trusted-element-legacy">' +
+            '  <template>' +
+            '    <div id="injected"></div>'+
+            '  </template>`' +
+            '</dom-module>' +
+            '<trusted-element-legacy></trusted-element-legacy>';
```
sorvell: Again, make element creation separate.

test/unit/strict-template-policy.html
```diff
@@ -0,0 +1,240 @@
...
+        }, /trusted-element-legacy registered twice/);
+        const el = document.querySelector('trusted-element-legacy');
+        assert.notOk(el && el.shadowRoot && el.shadowRoot.querySelector('#injected'));
```
sorvell: Again, only test that `el.shadowRoot` does not exist.

lib/mixins/element-mixin.html
```diff
@@ -386,16 +407,15 @@
...
-          if (Polymer.strictTemplatePolicy && this.is && !template) {
-            throw new Error(`strictTemplatePolicy: expecting dom-module or null template for ${this.is}`);
-          }
-          this._template = template;
+          this._template =
+            // Look in dom-module associated with this element's is
+            getTemplateFromDomModule(/** @type {PolymerElementConstructor}*/ (this).is) ||
+            // Next look for superclass template (call the super impl this
+            // way so that `this` points to the superclass)
+            Object.getPrototypeOf(/** @type {PolymerElementConstructor}*/ (this).prototype).constructor.template ||
```
sorvell: The "super" call should be last here

lib/elements/dom-module.html
```diff
@@ -124,16 +130,14 @@
     register(id) {
       id = id || this.id;
       if (id) {
-        if (Polymer.strictTemplatePolicy && findModule(id)) {
-          modules[id] = lcModules[id.toLowerCase()] = null;
-          throw new Error(`strictTemplatePolicy: dom-module ${id} registered twice`);
+        // Under strictTemplatePolicy, reject and null out any re-registered
+        // dom-module since it is ambiguous whether first-in or last-in is trusted 
+        if (Polymer.strictTemplatePolicy && findModule(id) !== undefined) {
```
sorvell: Add a test that tries to register the same module 3x.

lib/legacy/class.html
```diff
@@ -373,6 +350,11 @@
         Polymer.LegacyElementMixin(HTMLElement));
       // decorate klass with registration info
       klass.is = info.is;
+      // if user provided template on info, make sure the static _template
+      // is set so the static template getter uses it
+      if (info._template !== undefined) {
```
sorvell: Add a test that sets a template on a behavior.

lib/elements/dom-module.html
kevinpschaaf: Done.

lib/legacy/class.html
kevinpschaaf: Done, in test/unit/behaviors.html.

lib/elements/dom-if.html
```diff
@@ -255,8 +255,10 @@
         if (c$ && c$.length) {
           // use first child parent, for case when dom-if may have been detached
           let parent = c$[0].parentNode;
-          for (let i=0, n; (i<c$.length) && (n=c$[i]); i++) {
-            parent.removeChild(n);
+          if (parent) {
```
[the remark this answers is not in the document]
kevinpschaaf: Done.

lib/elements/dom-module.html
```diff
@@ -124,6 +124,10 @@
     register(id) {
       id = id || this.id;
       if (id) {
+        if (Polymer.strictTemplatePolicy && findModule(id)) {
```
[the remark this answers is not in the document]
kevinpschaaf: Done.

lib/legacy/class.html
```diff
@@ -157,7 +157,10 @@
          */
         static get template() {
           // get template first from any imperative set in `info._template`
-          return info._template ||
+          if (info._template !== undefined) {
+            return info._template;
+          }
+          const template = info._template ||
```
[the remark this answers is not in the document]
kevinpschaaf: Removed template getter here altogether, in favor of element-mixin template getter.

lib/mixins/element-mixin.html
```diff
@@ -386,12 +386,16 @@
        */
       static get template() {
         if (!this.hasOwnProperty(JSCompiler_renameProperty('_template', this))) {
-          this._template = Polymer.DomModule && Polymer.DomModule.import(
+          const template = Polymer.DomModule && Polymer.DomModule.import(
```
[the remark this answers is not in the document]
kevinpschaaf: Done, and factored out into `getTemplateFromDomModule`.

lib/mixins/element-mixin.html
```diff
@@ -386,12 +386,16 @@
...
             /** @type {PolymerElementConstructor}*/ (this).is, 'template') ||
             // note: implemented so a subclass can retrieve the super
             // template; call the super impl this way so that `this` points
             // to the superclass.
             Object.getPrototypeOf(/** @type {PolymerElementConstructor}*/ (this).prototype).constructor.template;
+          if (Polymer.strictTemplatePolicy && this.is && !template) {
```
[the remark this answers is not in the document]
kevinpschaaf: Done, refactored.

lib/utils/templatize.html
```diff
@@ -500,6 +500,9 @@
        * @suppress {invalidCasts}
        */
       templatize(template, owner, options) {
+        if (Polymer.strictTemplatePolicy && !owner._methodHost) {
```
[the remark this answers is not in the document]
kevinpschaaf: Done.

test/unit/strict-template-policy.html
kevinpschaaf: Done & done.

test/unit/strict-template-policy.html
kevinpschaaf: Done.

test/unit/strict-template-policy.html
kevinpschaaf: Done.

lib/mixins/element-mixin.html
kevinpschaaf: Done.

test/unit/behaviors.html
```diff
@@ -505,6 +593,37 @@
...
+suite('templates from behaviors', function() {
+
+  test('template from registered callback', function() {
+    var el = fixture('from-registered');
+    assert.ok(el.shadowRoot.querySelector('#from-registered'));
+  });
+
+  test('template from base', function() {
+    var el = fixture('from-base');
+    assert.ok(el.shadowRoot.querySelector('#from-base'));
```
sorvell: The note inside this template says "should not be used" so it's confusing it's used here.

Since the behavior should set `_template` on the prototype, why isn't that preferred over the `dom-module`?

lib/elements/dom-bind.html
```diff
@@ -56,6 +56,9 @@
 
       constructor() {
         super();
+        if (Polymer.strictTemplatePolicy) {
```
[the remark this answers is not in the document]
kevinpschaaf: Confirmed that on CE polyfill (both v0 and v1) element upgrading halts when a constructor throws.

In V1 it actually try/catch'es, but [re-throws the error](https://github.com/webcomponents/custom-elements/blob/master/src/CustomElementInternals.js#L285).  Not sure why.  Filed issue here: https://github.com/webcomponents/custom-elements/issues/176

test/unit/behaviors.html
kevinpschaaf: Good catch. Made a change to `Polymer.Class` to ensure `_template` from behaviors take precedence over dom-module lookup for `is`, which both fixes a latent bug w.r.t. original Polymer 1.x behavior, and ensures that legacy elements that intend to get their template from a behavior are not subject to dom-module injection.

PR #5297, opened 2018-07-20
Port externs update from internal changes

dfreedm:

### DIFF
--- a/externs/polymer-externs.js
+++ b/externs/polymer-externs.js
@@ -118,3 +118,34 @@ PolymerElement.prototype.registered = function() {};
 PolymerElement.prototype.attached = function() {};
 /** On detached from the DOM callback. */
 PolymerElement.prototype.detached = function() {};
+
+/**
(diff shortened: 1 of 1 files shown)


TimvdLippe:
These changes have been applied in https://github.com/Polymer/polymer/pull/5305/files#diff-76242aa10cb2457011adacb2abe35784R138 which was already merged. I think we should instead perform the changes of #5305 into the `2.x` branch as well. @aomarks do you mind doing that?

aomarks:
> These changes have been applied in https://github.com/Polymer/polymer/pull/5305/files#diff-76242aa10cb2457011adacb2abe35784R138 which was already merged. I think we should instead perform the changes of #5305 into the 2.x branch as well. @aomarks do you mind doing that?

Those changes are mostly some incremental improvements to closure compatibility, I'm not sure we are planning to backport all changes to 2.0 at this point. @azakus do we have a policy on that?

PR #5298, opened 2018-07-20
Port externs updates from 2.x

azakus:


azakus:
Looks like this actually passed, but `wct` didn't stop running.

TimvdLippe:
Superseded by #5305

PR #5303, opened 2018-07-24
[1.x] Introduce strictTemplatePolicy

### LINKED ISSUE #5323: Introduce strictTemplatePolicy

kevinpschaaf:
Enabling `strictTemplatePolicy` by setting `Polymer = {setStrictTemplatePolicy: true};` prior to loading Polymer causes the following changes in behavior:
* Elements that do not have templates by design must explicitly indicate a null template to prevent template retrieval from inject `<dom-module>`s, otherwise they will throw an error

Fixes #5323

### DIFF
--- a/src/lib/template/dom-bind.html
+++ b/src/lib/template/dom-bind.html
@@ -78,6 +78,9 @@ elements to the template itself as the binding scope.
     _template: null,
 
     created: function() {
+      if (Polymer.Settings.strictTemplatePolicy) {
+        throw new Error(`strictTemplatePolicy: dom-bind not allowed`);
+      }
       // Ensure dom-bind doesn't stamp until all possible dependencies
       // have resolved
       var self = this;
--- a/src/lib/template/dom-bind.html
+++ b/src/lib/template/dom-bind.html
@@ -79,7 +79,7 @@ elements to the template itself as the binding scope.
 
     created: function() {
       if (Polymer.Settings.strictTemplatePolicy) {
-        throw new Error(`strictTemplatePolicy: dom-bind not allowed`);
+        throw new Error('strictTemplatePolicy: dom-bind not allowed');
       }
       // Ensure dom-bind doesn't stamp until all possible dependencies
       // have resolved
--- a/src/mini/template.html
+++ b/src/mini/template.html
@@ -29,7 +29,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var module;
       if (this._template === undefined) {
         module = Polymer.DomModule.import(this.is);
-        const template = module && module.querySelector('template');
+        var template = module && module.querySelector('template');
         if (Polymer.Settings.strictTemplatePolicy && !template) {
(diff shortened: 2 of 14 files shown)


src/lib/template/dom-bind.html
```diff
@@ -78,6 +78,9 @@
     _template: null,
 
     created: function() {
+      if (Polymer.Settings.strictTemplatePolicy) {
```
sorvell: Test this throws as expected using the custom elements polyfill (and doesn't halt other elements from booting up)

test/unit/strict-template-policy.html
```diff
@@ -0,0 +1,146 @@
...
+        assertThrows(function() {
+          document.getElementById('target').innerHTML =
+            '<template is="dom-repeat" items="[0]">' +
+            '  <div id="injected"></div>'+
+            '</template>`';
+        }, /template owner not trusted/);
+        assert.notOk(document.getElementById('injected'));
+      });
+
+      test('dom-module after registration', function() {
```
sorvell: Add a test that registers the same module name 3x.

kevinpschaaf: Added: https://github.com/Polymer/polymer/pull/5303/commits/1d529ab9512745457aa3873c6795624083658cc6#diff-e2ce14a791fe5ac94beadd7de8fcdc95R142

src/lib/template/dom-bind.html
kevinpschaaf: Confirmed that on CE polyfill (both v0 and v1) element upgrading halts when a constructor throws.

In V1 it actually try/catch'es, but [re-throws the error](https://github.com/webcomponents/custom-elements/blob/master/src/CustomElementInternals.js#L285).  Not sure why.  Filed issue here: https://github.com/webcomponents/custom-elements/issues/176

src/standard/annotations.html
```diff
@@ -236,7 +236,7 @@
 
     // push configuration references at configure time
     _configureAnnotationReferences: function() {
-      var notes = this._notes;
+      var notes = this._notes || [];
```
sorvell: Why is this changed needed?

src/mini/template.html
```diff
@@ -29,7 +29,11 @@
       var module;
       if (this._template === undefined) {
         module = Polymer.DomModule.import(this.is);
-        this._template = module && module.querySelector('template');
+        var template = module && module.querySelector('template');
```
sorvell: Isn't it incorrect to do this if `this._template` is explicitly `null`?

kevinpschaaf: Correct, this is all skipped when `this._template === undefined`, including when template is explicitly set to null.

src/standard/annotations.html
kevinpschaaf: Good question.  When `_prepTemplate` throws under a `strictTemplatePolicy` violation, it leaves `_notes` uninitialized, which was previously not a case that could ever occur since `_prepAnnotations` comes after that and unconditionally sets `_notes`.  So this just provides a good fallback behavior under the violation case.

PR #5304, opened 2018-07-24
[3.x] Introduce strictTemplatePolicy

### LINKED ISSUE #5323: Introduce strictTemplatePolicy

kevinpschaaf:
Enabling `strictTemplatePolicy` by calling `setStrictTemplatePolicy(true)` setting causes the following changes in behavior:
* Template retrieval from `<dom-module>` is disabled (since templates should be defined inline in JS via `static get template()` using the `html` tag function) by default when `strictTemplatePolicy` is enabled
* Legacy `<dom-module>` template lookup can be re-enabled via `setAllowTemplateFromDomModule(true)` setting; under that setting, the following additional behavior will apply:
  * Elements that do not have templates by design must explicitly indicate a null template to prevent template retrieval from inject `<dom-module>`s, otherwise they will throw an error

Fixes #5323


TimvdLippe:
I have resolved the merge conflicts on this branch

PR #5312, opened 2018-08-02
Make Polymer gestures library safe for Closure property renaming.

aomarks:
The issue here was that we were assuming that the method for a given gesture would always match the string name of that gesture, e.g. that if the gesture name was "mousedown" that we could call
gestureDefinition.mousedown. In actuality, gestureDefinition.mousedown could be renamed by Closure, which would throw with "undefined is not a function".

The fix is to move the GesturesRecognizer type into our externs, which prevents those methods from being renamed.


aomarks:
Actually holding off for now, need to test somrthing first.

PR #5317, opened 2018-08-09
Export the current Polymer version in polymer-element.js

### LINKED ISSUE #5211: Add `version` info to 3.x branch
There's currently no replacement for `Polymer.version` in the 3.x branch. 

Perhaps we can put this as a static property on the `PolymerElement` class.

TimvdLippe:
Potential solution for exposing the Polymer version in `polymer-element.js`. Not sure if this is the desired solution, but it seemed like the best option we have.
### Reference Issue
Fixes #5211

### DIFF
--- a/polymer-element.js
+++ b/polymer-element.js
@@ -28,3 +28,8 @@ export { html } from './lib/utils/html-tag.js';
  */
 export const PolymerElement = ElementMixin(HTMLElement);
 
+/**
+ * Current Polymer version in Semver notation.
+ * @type {String} Semver notation of the current version of Polymer.
+ */
+export const version = '3.0.5';
--- a/polymer-element.js
+++ b/polymer-element.js
@@ -9,6 +9,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 */
 
 import { ElementMixin } from './lib/mixins/element-mixin.js';
+export { version } from './lib/mixins/element-mixin.js';
 export { html } from './lib/utils/html-tag.js';
 
 /**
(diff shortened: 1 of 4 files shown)


polymer-element.js
```diff
@@ -28,3 +28,8 @@ export { html } from './lib/utils/html-tag.js';
  */
 export const PolymerElement = ElementMixin(HTMLElement);
 
+/**
+ * Current Polymer version in Semver notation.
+ * @type {String} Semver notation of the current version of Polymer.
+ */
+export const version = '3.0.5';
```
kevinpschaaf: Although this change makes introspecting the version from a running page possible, it is fairly cumbersome:
```js
import('./need-to-know-path-to-polymer/polymer-element.js').then(module => console.log(module.version));
```
And that only works if a build tool hasn't compiled modules into bundles or out altogether.

I think it would be pretty useful to have e.g. a static getter on the `ElementMixin` return the version (in addition to a plain module export), so that it's more straightforward to introspect the version being used on a running page.

e.g. from an element in the inspector
```js
console.log($0.constructor.polymerElementVersion);
```

or e.g. from the registry:
```js
console.log(customElements.get('my-app').polymerElementVersion);
```
So perhaps the version can be exported from `element-mixin.js` instead of here, and returned from a new `static get polymerElementVersion() { return version; }` getter on the `ElementMixin`?  It could also be re-exported from `polymer-element.js` for convenience.

TimvdLippe: @kevinpschaaf Good idea! Updated.

TimvdLippe:
@kevinpschaaf This is ready for review again

PR #5319, opened 2018-08-09
Add a template setter to ElementMixin

aomarks:
Currently, to assign a template to a Polymer element, you need to override the setter, or call `Object.defineProperty`.

This PR also allows users to simply assign the template onto their constructor.

This is especially useful for TypeScript, which compiles `static template = foo` to `MyClass.template = foo`.


justinfagnani:
Needs a test

aomarks:
> Needs a test

Added a test.

PR #5320, opened 2018-08-10
Update TemplateStamp event listen param types from Node to EventTarget.

aomarks:
Allows passing e.g. `window`, which I assume is valid here?

### DIFF
--- a/lib/utils/gestures.js
+++ b/lib/utils/gestures.js
@@ -635,7 +635,7 @@ function _findRecognizerByEvent(evName) {
  * @return {void}
  */
 export function setTouchAction(node, value) {
-  if (HAS_NATIVE_TA && node instanceof Node) {
+  if (HAS_NATIVE_TA && node instanceof HTMLElement) {
     // NOTE: add touchAction async so that events can be added in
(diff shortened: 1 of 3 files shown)


aomarks:
> Fixed. Glad we're testing the types compile in CI now!

Had to fix gestures.js too. We set `.style` in that lib, which window doesn't have. I added an `instanceof` check. @azakus can you confirm this make sense?

lib/utils/gestures.js
```diff
@@ -630,12 +630,12 @@ function _findRecognizerByEvent(evName) {
...
  * adding event listeners.
  *
- * @param {!Node} node Node to set touch action setting on
+ * @param {!EventTarget} node Node to set touch action setting on
  * @param {string} value Touch action value
  * @return {void}
  */
 export function setTouchAction(node, value) {
-  if (HAS_NATIVE_TA) {
+  if (HAS_NATIVE_TA && node instanceof Node) {
```
azakus: This should be a check for `Element` to ensure the style property.

aomarks: Done (HTMLElement actually)

PR #5329, opened 2018-08-16
 Fix a grab bag of closure compiler warnings

rictic:
no-op at runtime, just helping jscompiler (and humans!) understand the code

### DIFF
--- a/lib/utils/style-gather.js
+++ b/lib/utils/style-gather.js
@@ -7,18 +7,27 @@ The complete set of contributors may be found at http://polymer.github.io/CONTRI
 Code distributed by Google as part of the polymer project is also
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 */
+
+/**
+ * Module with utilities for collection CSS text from `<templates>`, external
+ * stylesheets, and `dom-module`s.
+ *
+ * @summary Module with utilities for collection CSS text from various sources.
+ */
+
+import { DomModule } from '../elements/dom-module.js';
 import { resolveCss } from './resolve-url.js';
 
 const MODULE_STYLE_LINK_SELECTOR = 'link[rel=import][type~=css]';
 const INCLUDE_ATTR = 'include';
 const SHADY_UNSCOPED_ATTR = 'shady-unscoped';
 
+/**
+ * @param {string} moduleId .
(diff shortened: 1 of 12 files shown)


lib/utils/style-gather.js
```diff
@@ -7,18 +7,27 @@ The complete set of contributors may be found at http://polymer.github.io/CONTRI
...
+ * @param {string} moduleId .
+ * @return {?DomModule} .
+ */
 function importModule(moduleId) {
-  const /** DomModule */ PolymerDomModule = customElements.get('dom-module');
-  if (!PolymerDomModule) {
-    return null;
-  }
-  return PolymerDomModule.import(moduleId);
+  return /** @type {?DomModule} */(DomModule.import(moduleId));
```
kevinpschaaf: Any recollection why the original code used `customElements.get` to get the DomModule class?  It was changed in this PR that doesn't have any detail: https://github.com/Polymer/polymer/pull/4748/files

rictic:
Still about ten files left. I'm waiting on a couple other internal-specific cleanup changes to land so I can tackle them, but I hope to have another PR in the next week or so taking care of the rest, then we should totally enforce no warnings in travis!

I don't think we need an updated closure compiler version, but it never hurts. These changes were tested against this week's release

lib/utils/style-gather.js
rictic: Yeah, Steve mentioned that ideally utils wouldn't depend on elements, and the `customElements.get` was there to avoid having that dependency, I guess in case someone wanted to use targeted imports to not load in dom-module.js.

I tried to preserve that by declaring some interfaces, but I'm worried it would be fragile, both to changes in DomModule, and in the face of aggressive compiler optimizations. And in practice essentially everyone who loads style-gather is going to load dom-module anyways...

kevinpschaaf: Seems ok for now.  We've recently discussed one way to make Polymer double-import safe would be to have <dom-module> only register if not already registered (first in wins), and then have _all_ `DomModule.import` call sites get `DomModule` from the CE registry like the old code here did.  But if we wanted to do that it would be a more sweeping change, and can do it separately.  Steve agreed this change should be fine for now.

PR #5330, opened 2018-08-17
Cleanup element-mixin leftovers from modulizer

### LINKED ISSUE #5315: Erratum in element-mixin.js ? - error lint/constant-condition
Hi everyone!

I'm reading element-mixin.js code and in [line 713](https://github.com/Polymer/polymer/blob/master/lib/mixins/element-mixin.js#L713) have a rare code:
```javascript
export function register(prototype) {
  registrations.push(prototype);
  undefined && _regLog(prototype);
}
```
Any idea why write "undefined" in that condition? because `_regLog()` always is executed.

TimvdLippe:
This was some leftover output from modulizer that we can remove.
### Reference Issue
Closes #5315

### DIFF
--- a/lib/mixins/element-mixin.js
+++ b/lib/mixins/element-mixin.js
@@ -721,17 +721,6 @@ export const ElementMixin = dedupingMixin(base => {
 });
 
 /**
- * Provides basic tracking of element definitions (registrations) and
- * instance counts.
- *
- * @summary Provides basic tracking of element definitions (registrations) and
- * instance counts.
- */
-`TODO(modulizer): A namespace named Polymer.telemetry was
-declared here. The surrounding comments should be reviewed,
-and this string can then be deleted`;
-
-/**
  * Total number of Polymer element instances created.
  * @type {number}
  */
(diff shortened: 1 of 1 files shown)


lib/mixins/element-mixin.js
```diff
@@ -760,7 +749,7 @@ function _regLog(prototype) {
  */
 export function register(prototype) {
   registrations.push(prototype);
-  undefined && _regLog(prototype);
```
rictic: I don't think this is right. In 2.x this only logged if `Polymer.log` was set. Modulizer made it never log, which isn't great, but after this change we'll always log, which we also don't want. 

IMO we should invent a global name that, when truthy, means that we'll log here.

TimvdLippe: @rictic Ah crap, you are correct! Reinvent the `Polymer.log` setting? 😂

kevinpschaaf: I'd say let's just remove the registration logging altogether in 3.x. I don't know that this feature was ever [documented](https://www.polymer-project.org/2.0/docs/devguide/settings), and the whole registration tracking system is fairly dubious now that WC V1 has a registry API.

Will make a new PR to remove that line; `dumpRegistrations` can remain.

kevinpschaaf: https://github.com/Polymer/polymer/pull/5331

PR #5334, opened 2018-08-20
The return type of mixinBehaviors is unknown

rictic:
It's more than just the input class. Leaving it unknown allows the user to add annotations like:

```
 * @constructor
 * @extends {Polymer.Element}
 * @implements {IronValidatableBehaviorInterface}
```

### DIFF
--- a/lib/legacy/class.js
+++ b/lib/legacy/class.js
@@ -33,7 +33,7 @@ let metaProps = {
  * @template T
  * @param {!Object|!Array<!Object>} behaviors Behavior object or array of behaviors.
  * @param {function(new:T)} klass Element class.
- * @return {function(new:T)} Returns a new Element class extended by the
+ * @return {?} Returns a new Element class extended by the
  * passed in `behaviors` and also by `LegacyElementMixin`.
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Would we classify this as a breaking change in terms of the Typescript generated types?

rictic:
Practically any change to types could break someone's build (this is why TypeScript doesn't really follow semver). IMO we should define a breaking change solely in terms of runtime behavior.

That said, I think we have a good sense of how to type this from the TypeScript side, we just need a way to express overrides. Filed https://github.com/Polymer/tools/issues/658

TimvdLippe:
> Practically any change to types could break someone's build (this is why TypeScript doesn't really follow semver). IMO we should define a breaking change solely in terms of runtime behavior.

Yes agreed. It was more of a theoretical question rather than an actual blocker on this PR. Maybe we should document this stance officially somewhere in our types, to make users aware of this process.

kevinpschaaf:
PR build is broken due to lint error merged in https://github.com/Polymer/polymer/pull/5335, being fixed in https://github.com/Polymer/polymer/pull/5337.

PR #5339, opened 2018-08-21
[3.x] Add legacy-data-mixin as 1.x->2.x/3.x migration aide. Fixes #5262.

### LINKED ISSUE #5262: Add 2.x option to treat complex observers as they were in Polymer1

kevinpschaaf:
Add legacy-data-mixin as 1.x->2.x/3.x migration aide. 

Mixin to selectively add back Polymer 1.x's `undefined` rules governing when observers & computing functions run based on all arguments being defined (reference https://www.polymer-project.org/1.0/docs/devguide/observers#multi-property-observers).

When `polymer/lib/legacy/legacy-data-mixin.js` is loaded at the app level, all legacy elements (defined with `Polymer({...})`) will have the mixin applied. The mixin only restores legacy data handling if `_legacyUndefinedCheck: true` is set on the element's prototype.

This mixin is intended for use to help migration from Polymer 1.x to 2.x+ by allowing legacy code to work during the migration period, while identifying observers and computing functions that need undefined checks to work without the mixin in Polymer 2/3.  It is generally not intended for use in production.

### Reference Issue
Fixes #5262.
Supersedes and closes #5263


lib/legacy/legacy-data-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+
+import { Class } from './class.js';
+import { Polymer } from '../../polymer-legacy.js';
+import { dedupingMixin } from '../utils/mixin.js';
+
+const UndefinedArgumentError = class extends Error {
+  constructor(message) {
+    super(message);
+    this.name = this.constructor.name;
+    // Affordances for ensuring instanceof works after babel ES5 compilation
```
sorvell: Surprised this is needed. Verify?

lib/legacy/legacy-data-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+export const LegacyDataMixin = dedupingMixin(superClass => {
+
+  /**
+   * @constructor
+   * @extends {superClass}
+   * @unrestricted
+   * @private   */
+  class LegacyDataMixin extends superClass {
+    /**
+     * Overrides `Polyer.PropertyEffects` to add `undefined` argument
```
sorvell: s/Polyer/Polymer

lib/legacy/legacy-data-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+     * 
+     * @param {!Array<!MethodArg>} args Array of argument metadata
+     * @param {string} path Property/path name that triggered the method effect
+     * @param {Object} props Bag of current property changes
+     * @return {Array<*>} Array of argument values
+     * @private
+     */
+    _marshalArgs(args, path, props) {
+      const vals = super._marshalArgs(args, path, props);
+      if (this._legacyUndefinedCheck && args.length > 1) {
```
sorvell: Add comment briefly explaining the rules so it's clear that single argument observers are always called.

lib/legacy/legacy-data-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+        for (let i=0; i<args.length; i++) {
+          if (vals[i] === undefined) {
+            // Break out of effect's control flow; will be caught in
+            // wrapped property effect function below
+            throw new UndefinedArgumentError(`argument '${args[i].name}' is undefined; ensure it has an undefined check`);
```
sorvell: Suggest making the error use sentence casing and periods.

lib/legacy/legacy-data-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+     * @param {!Array<!MethodArg>} args Array of argument metadata
+     * @param {string} path Property/path name that triggered the method effect
+     * @param {Object} props Bag of current property changes
+     * @return {Array<*>} Array of argument values
+     * @private
+     */
+    _marshalArgs(args, path, props) {
+      const vals = super._marshalArgs(args, path, props);
+      if (this._legacyUndefinedCheck && args.length > 1) {
+        for (let i=0; i<args.length; i++) {
```
sorvell: Seems like it would be more clear to loop over `vals`?

lib/legacy/legacy-data-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+     * @param {Object=} effect Effect metadata object
+     * @return {void}
+     * @protected
+     */
+    _addPropertyEffect(property, type, effect) {
+      if (effect && effect.fn) {
+        const fn = effect.fn;
+        effect.fn = function() {
+          try {
+            fn.apply(this, arguments);
```
sorvell: Can we just check the arguments here and avoid the need to try catch the error? Then you don't even need to expose `_marshalArgs`.

test/unit/legacy-data.html
```diff
@@ -0,0 +1,229 @@
...
+      Polymer({
+        is: 'x-data',
+        properties: {
+          singleProp: String,
+          multiProp1: String,
+          multiProp2: String
+        },
+        _legacyUndefinedCheck: true,
+        observers: [
+          'staticObserver("staticObserver")',
```
sorvell: Shouldn't we also test inline template expressions that have multiple arguments?

lib/legacy/legacy-data-mixin.js
kevinpschaaf: The version of babel we use in CLI doesn't handle subclassing builtins correctly; the latest one does, so will add TODO to remove once CLI updates to latest babel.

lib/legacy/legacy-data-mixin.js
kevinpschaaf: Done.

lib/legacy/legacy-data-mixin.js
kevinpschaaf: Done.

lib/legacy/legacy-data-mixin.js
kevinpschaaf: Done.

lib/legacy/legacy-data-mixin.js
kevinpschaaf: Done.

lib/legacy/legacy-data-mixin.js
kevinpschaaf: No, needs to also abort computed properties & inline computing functions. Will add tests for those.

test/unit/legacy-data.html
kevinpschaaf: Added.

lib/legacy/legacy-data-mixin.js
web-padawan: Note: should be fixed by Polymer/tools#670

lib/legacy/legacy-data-mixin.js
```diff
@@ -0,0 +1,150 @@
...
+      return super._addTemplatePropertyEffect(templateInfo, prop, wrapEffect(effect));
+    }
+
+  }
+
+  return LegacyDataMixin;
+
+});
+
+Polymer.Class = (info, mixin) => Class(info,
```
sorvell: Why not just put `LegacyDataMixin` on top of the user class?

lib/legacy/legacy-data-mixin.js
```diff
@@ -0,0 +1,150 @@
...
+     * Overrides `Polymer.PropertyEffects` to add `undefined` argument
+     * checking to match Polymer 1.x style rules
+     * 
+     * @param {!Array<!MethodArg>} args Array of argument metadata
+     * @param {string} path Property/path name that triggered the method effect
+     * @param {Object} props Bag of current property changes
+     * @return {Array<*>} Array of argument values
+     * @private
+     */
+    _marshalArgs(args, path, props) {
```
sorvell: Did you consider altering this method to get the info needed to present a good error? I'm concerned that since the intension is to warn, but the code throws here, we're depending on catching that where it's called.

sorvell: Ok, disregard that. It's clear that we need to abort the effect of marshaling the arguments and throwing let's us do this.

lib/legacy/legacy-data-mixin.js
sorvell: Ok, the answer is no because the mixin customizes how properties are created on the element prototype. Since a subclass only creates properties that it defines, it will not modify properties created by its superclass.

PR #5340, opened 2018-08-22
Minor changes to formatting, jsdoc, and setting initialization.

kevinpschaaf:
Upstreamed from internal codebase.


polymer-element.js
```diff
@@ -25,6 +25,7 @@ export { html } from './lib/utils/html-tag.js';
  * @summary Custom element base class that provides the core API for Polymer's
  *   key meta-programming features including template stamping, data-binding,
  *   attribute deserialization, and property change observation
+ * @const
```
rictic: I'd prefer to revert this internally instead, as it's already assigning to a const. This was due to a temporary hack we had.

kevinpschaaf:
Still struggling with undiagnosable FF timeouts on travis.  Merging.

PR #5349, opened 2018-08-29
Added export for LiteralString

SergeShkurko:
### Reference Issue

Fixes [this](https://github.com/Polymer/tools/pull/661)

### DIFF
--- a/lib/utils/html-tag.js
+++ b/lib/utils/html-tag.js
@@ -14,7 +14,7 @@ import './boot.js';
  * strings by asseting that they have been created via this class. The
  * `value` property returns the string passed to the constructor.
  */
-class LiteralString {
+export class LiteralString {
   constructor(string) {
     /** @type {string} */
     this.value = string.toString();


TimvdLippe:
Tagging @justinfagnani as he is the one most knowledgeable about these kind of changes.

kevinpschaaf:
Exposing this would bypass the guarantee that `html` only interpolates values that originate from string literals in script.  Exporting this would encourage e.g. `const notSafe = new LiteralString(untrustedString);` which could be interpolated into `html`.

`html` uses `innerHTML` to generate a template, and is thus not safe to interpolate untrusted strings. For your use case (interpolating strings from an import), the thing you import needs to have been generated using `html` or `htmlLiteral`; perhaps this could be done by passing your pug file through a webpack transform that wraps with `html`.

We will not support interpolation of arbitrary strings (i.e., not verifiably trusted strings) into Polymer's `html` tag function, since we want the default that ships with Polymer to be verifiably safe.

Note however, it is not a requirement that you use the `html` tag function that ships with Polymer in your static template getter.  It only needs to return an `HTMLTemplateElement`.  You can use your own javascript to make a template however you like, and it should be trivial for you to create a function that `innerHTML`'s a template from whatever sources you want -- we're just not going to write that code for you since it's you (the author) who needs to verify it's safe for your use case.  Hope that helps.

PR #5355, opened 2018-09-11
Add cast for compilation

rictic:
this._target is an Element, getFlattenedNodes accepts an HTMLElement

Upcoming version of closure compiler type checks this better, and caught this

### DIFF
--- a/lib/legacy/polymer.dom.js
+++ b/lib/legacy/polymer.dom.js
@@ -54,13 +54,13 @@ export class DomApi {
    * Returns an instance of `FlattenedNodesObserver` that
    * listens for node changes on this element.
    *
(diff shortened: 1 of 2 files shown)


rictic:
I agree that it's not a breaking change, but I wasn't sure if we had another reason for preferring Element there instead of HTMLElement. In some of our APIs I think we have classes with methods that are only safe to call with some values but not others (maybe?)

TimvdLippe:
I am not aware of these cases, but probably @azakus knows more about that?

azakus:
It is very likely no one thought too hard about what should be `Element` and what should be `HTMLElement`.

rictic:
Updated the types a bit, PTAL

PR #5356, opened 2018-09-13
Add @export

rictic:
CustomStyle#getStyle is referred to as customStyle['getStyle'] in a few places, so it needs to be exported

Externalizing cl/212875987 from @samccone


samccone:
nice thank you.

PR #5357, opened 2018-09-14
Improve typings for legacy elements

### LINKED ISSUE #5350: Type issues with Polymer 3.0.5 and legacy elements
I am having issues with the types for legacy elements:

#### Steps to reproduce
Declare a legacy polymer element via TypeScript:
```typescript
import { Polymer, html } from '@polymer/polymer/polymer-legacy';

Polymer({
  _template: html`<span>hello</span>`,
  is: 'hello-element',
});
```

#### Expected Results
There are no type errors.

#### Actual Results
The element renders fine but the following type error is shown:
```Object literal may only specify known properties, but '_template' does not exist in type 'PolymerInit'. Did you mean to write 'template'?```

### Versions
- Polymer: v3.0.5
- TypeScript: v3.0.1

### PR DESCRIPTION
**Pull request by a-xin**
Fixes #5350

### DIFF
--- a/interfaces.d.ts
+++ b/interfaces.d.ts
@@ -26,9 +26,21 @@ export interface PolymerInit {
   extends?: string;
   properties?: PolymerElementProperties;
   observers?: string[];
-  template?: HTMLTemplateElement|string;
+  _template?: HTMLTemplateElement;
   hostAttributes?: {[key: string]: any};
   listeners?: {[key: string]: string};
+  behaviors?: PolymerInit | PolymerInit[];
+
+  // Lifecycle methods
+  registered?(): void;
+  created?(): void;
+  attached?(): void;
+  detached?(): void;
+  ready?(): void;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: interfaces.d.ts [hunk 1]
```diff
@@ -26,11 +26,28 @@ export interface PolymerInit {
...
   listeners?: {[key: string]: string};
+  behaviors?: BehaviorInit | BehaviorInit[];
+
+  // Lifecycle methods
+  registered?(): void;
+  created?(): void;
+  attached?(): void;
+  detached?(): void;
+  ready?(): void;
+  attributeChanged?(name: string, old?: string, value?: string);
```
TimvdLippe: This lacks a return type. Please see the CI output for the full error. You can run `npm run generate-types` to locally run the same check.

a-xin: Thanks!

aomarks:
Looks good, running some tests will ping when done.

PR #5359, opened 2018-09-15
Add TypeScript types for observer parameters.

aomarks:
- PolymerDeepPropertyChange for .* observers.
- PolymerSpliceChange and PolymerSplice for .splices observers.
- Parameterized for the property/path types.
- Matches the name of the interfaces in polymer-externs.js.

### DIFF
--- a/interfaces.d.ts
+++ b/interfaces.d.ts
@@ -31,6 +31,47 @@ export interface PolymerInit {
   listeners?: {[key: string]: string};
 }
 
+/**
+ * The object passed to ".*" wildcard obsevers. A record of a change made to an
+ * object.
+ * @template V The type of values at the path.
+ * @template B The type of the base property.
+ */
+export interface PolymerDeepPropertyChange<V, B> {
+  /** Path to the property that changed. */
+  path: string;
+  /** New value of the path that changed. */
+  value: V;
+  /** The object matching the non-wildcard portion of the path. */
+  base: B;
+}
+
+/**
+ * A record of changes made to an array.
+ * @template T The type of object contained by the array.
+ */
+export interface PolymerSplice<T> {
+  /** Position where the splice started. */
+  index: number;
+  /** Array of removed items. */
+  removed: T[];
+  /** Number of new items inserted at index. */
+  addedCount: number;
+  /** A reference to the array in question. */
+  object: T[];
+  /** The string literal 'splice'. */
+  type: 'splice';
+}
+
+/**
+ * The object passed to ".splices" observers. A set of mutations made to the
+ * array.
+ * @template T The type of object contained by the array.
+ */
+export interface PolymerSpliceChange<T> {
+  indexSplices: Array<PolymerSplice<T>>;
+}
+
 // Types from "externs/polymer-internal-shared-types.js"
 
 export interface StampedTemplate extends DocumentFragment {


interfaces.d.ts
```diff
@@ -31,6 +31,47 @@ export interface PolymerInit {
...
+ * @template B The object matching the non-wildcard portion of the path.
+ * @template V Additional types that could be set at the path.
+ */
+export interface PolymerDeepPropertyChange<B, V> {
+  /** Path to the property that changed. */
+  path: string;
+  /** The object matching the non-wildcard portion of the path. */
+  base: B;
+  /** New value of the path that changed. */
+  value: B|V;
```
rictic: Why B|V and not just V?

interfaces.d.ts
```diff
@@ -31,6 +31,47 @@ export interface PolymerInit {
...
+}
+
+/**
+ * A record of changes made to an array.
+ * @template T The type of the array being observed.
+ */
+export interface PolymerSplice<T extends Array<{}>> {
```
rictic: Prefer `unknown` over `{}` as `{}` is non-nullable. `T extends Array<unknown>`

interfaces.d.ts
```diff
@@ -31,6 +31,47 @@ export interface PolymerInit {
...
+  /** The string literal 'splice'. */
+  type: 'splice';
+}
+
+/**
+ * The object passed to ".splices" observers. A set of mutations made to the
+ * array.
+ * @template T The type of the array being observed.
+ */
+export interface PolymerSpliceChange<T extends Array<{}>> {
```
rictic: Prefer `unknown`

interfaces.d.ts
aomarks: `unknown` isn't in g3 yet, how about `{}|null|undefined` for now? I think `any` is actually safe here too, because I think in this context (constraining by supertype), the `any` wouldn't leak anywhere (but I would avoid it just so linters don't complain)?

interfaces.d.ts
aomarks: In the change record, `base` will always be the base of the wildcard path (e.g.if you observe `foo.bar.*` then `base` will always be `foo.bar`), while `value` could be the base, or any other path nested below it (e.g. `foo.bar`, then `foo.bar.baz`). So the idea is you specify the type of the base path, and also a union of whatever other types you know you could encounter deeply within it:

`fooChanged(changeRecord: PolymerDeepPropertyChanged<MyElement['foo'], string|undefined>)`

aomarks:
> What do you think about putting these types in an importable file? Or, are these types importable? I guess they're using `export` so they shouldn't be global by default. How would you get these types, as a Polymer user?

It's a module so it's importable, no globals here. Users import from `@polymer/polymer/interfaces`.

interfaces.d.ts
aomarks: To more directly answer your question, it's `B|V` because you always get one notification with the base in the `value` field, and again every time the base is re-set.

interfaces.d.ts
rictic: {}|null|undefined is good.

You're right about `any`, I'd just like to avoid using it as much as possible.

interfaces.d.ts
rictic: I see. That's really hard to typecheck!

aomarks:
> LGTM. Just prefer using the unknown-like `{}|null|undefined` type

Done.

interfaces.d.ts
aomarks: See other thread.

interfaces.d.ts
aomarks: Done.

PR #5360, opened 2018-09-18
Update gen-closure-declarations to 0.5.0

### LINKED ISSUE #5212: Fix generate-externs for module compatibility in 3.x
### Description
`gulp generate-externs` currently outputs a bad externs file, and needs to be updated to work with Polymer 3.x's module-based source.

#### Steps to Reproduce
`gulp generate-externs`

#### Expected Results
No errors, outputs correct externs.

#### Actual Results
Externs file is tiny and contains lots of `undefined`.

TimvdLippe:
This fixes the externs generation for our modules

### Reference Issue
Fixes #5212


aomarks:
> `polymer-externs.js` was not a generated file, and I'm not sure we can delete that just yet. @aomarks is this file still necessary?

Yes, we can't delete that.

PR #5370, opened 2018-10-08
implement native toggleAttribute signature

### LINKED ISSUE #5364: Typescript 3.1 - compile error. toggleAttribute type conflict in  LegacyElementMixin
### Description

The following error is seen in typescript 3.1 when using any element with the LegacyElementMixin:  

` error TS2320: Interface 'AppDrawerElement' cannot simultaneously extend types 'LegacyElementMixin' and 'HTMLElement'.
  Named property 'toggleAttribute' of types 'LegacyElementMixin' and 'HTMLElement' are not identical.`

It appears a new API for [toggleAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute) was introduced with a signature ` toggleAttribute(qualifiedName: string, force?: boolean): boolean;`  this is in conflict with the existing method toggleAttribute in LegacyElementMixin with a signature of `toggleAttribute(name: string, bool?: boolean, node?: Element|null): void;`.   The result is compilation errors when using Typescript 3.1 with Polymer. 

#### Live Demo
none

### Versions
Polymer 2/3 typescript defs.

43081j:
Fixes #5364.

As long as we return bool, we implement the same interface that was natively introduced recently. While keeping our third parameter..

### DIFF
--- a/lib/legacy/legacy-element-mixin.js
+++ b/lib/legacy/legacy-element-mixin.js
@@ -817,7 +817,7 @@ export const LegacyElementMixin = dedupingMixin((base) => {
      * @param {boolean=} bool Boolean to force the attribute on or off.
      *    When unspecified, the state of the attribute will be reversed.
      * @param {Element=} node Node to target.  Defaults to `this`.
-     * @return {void}
+     * @return {boolean}
      */
     toggleAttribute(name, bool, node) {
       node = /** @type {Element} */ (node || this);
@@ -826,8 +826,10 @@ export const LegacyElementMixin = dedupingMixin((base) => {
       }
       if (bool) {
         node.setAttribute(name, '');
+        return true;
       } else {
         node.removeAttribute(name);
+        return false;
       }
     }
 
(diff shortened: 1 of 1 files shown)


43081j:
No worries, tried it locally on a project too and it resolves the problem 👍

lib/legacy/legacy-element-mixin.js
```diff
@@ -817,7 +817,7 @@ export const LegacyElementMixin = dedupingMixin((base) => {
      * @param {boolean=} bool Boolean to force the attribute on or off.
      *    When unspecified, the state of the attribute will be reversed.
      * @param {Element=} node Node to target.  Defaults to `this`.
-     * @return {void}
+     * @return {boolean}
```
TimvdLippe:
```
/home/travis/build/Polymer/polymer/lib/legacy/legacy-element-mixin.js
  813:5  error  Missing JSDoc return description  valid-jsdoc
```

PR #5371, opened 2018-10-08
Add toggleAttribute to excluded identifiers

keanulee:
TypeScript 3.1 added `Element.toggleAttribute(name, force?)` to its DOM typings which is compatible but not identical to `LegacyElementMixin.toggleAttribute(name, force?, element?)`

#5370 fixes the compatibility issue, but tsc still errors:

```
node_modules/@polymer/app-layout/app-drawer/app-drawer.d.ts:67:11 - error TS2320: Interface 'AppDrawerElement' cannot simultaneously extend types 'LegacyElementMixin' and 'HTMLElement'.
  Named property 'toggleAttribute' of types 'LegacyElementMixin' and 'HTMLElement' are not identical.

67 interface AppDrawerElement extends LegacyElementMixin, HTMLElement {
             ~~~~~~~~~~~~~~~~
```

This patch removes the custom typing of `toggleAttribute()`. We did this before with `scroll()` in iron-scroll-target-behavior (https://github.com/PolymerElements/iron-scroll-target-behavior/pull/48).

### DIFF
--- a/gen-tsd.json
+++ b/gen-tsd.json
@@ -35,5 +35,8 @@
     "./lib/utils/debounce.js": [
       "Debouncer"
     ]
-  }
+  },
+  "excludeIdentifiers": [
+    "toggleAttribute"
+  ]
 }


TimvdLippe:
This is a breaking change for users on TypeScript 2. I am confused though, as @43081j indicated he saw the issue was resolved. @43081j do you mind elaborating upon this PR?

43081j:
Interesting, in the project i tried it on, I was using a few `paper-*` elements and they seemed to compile fine.

I'll take a look and get back to it

keanulee:
@43081j Specifically, I had issues with `npm run generate-types` on [app-layout](https://github.com/PolymerElements/app-layout) after bumping to TS 3.1. First error:

```
Compilation failed
app-box/app-box.d.ts:93:11 - error TS2320: Interface 'AppBoxElement' cannot simultaneously extend types 'LegacyElementMixin' and 'HTMLElement'.
  Named property 'toggleAttribute' of types 'LegacyElementMixin' and 'HTMLElement' are not identical.

93 interface AppBoxElement extends AppScrollEffectsBehavior, IronResizableBehavior, LegacyElementMixin, HTMLElement {
             ~~~~~~~~~~~~~
```

43081j:
yup this is reproducible [here](http://www.typescriptlang.org/play/#src=interface%20Foo%20%7B%0D%0A%20%20%20%20toggleAttribute(name%3A%20string%2C%20bool%3F%3A%20boolean%2C%20node%3F%3A%20Element%20%7C%20null)%3A%20boolean%3B%0D%0A%7D%0D%0A%0D%0Ainterface%20Bar%20extends%20Foo%2C%20HTMLElement%20%7B%0D%0A%7D).

strangely my local project which uses paper-dialog builds fine with the type change (from void to bool).

still investigating why mine builds and the TS repro doesn't.

43081j:
🙈 

@TimvdLippe  it turns out `skipLibCheck` had slipped into my tsconfig. I can now reproduce this locally too.

Is it possible we could output the built in signature but still call it in JS with the extra parameter? It sounds hacky... but can't think of a better solution

TimvdLippe:
I think that is the way to go right now :(

PR #5372, opened 2018-10-11
make toggleAttribute match with native signature

### LINKED ISSUE #5364: Typescript 3.1 - compile error. toggleAttribute type conflict in  LegacyElementMixin

43081j:
Fixes #5364, again.

This solution is needed instead of #5371 as that would break TS definitions for consumers using TS <3.1.

I had to use `arguments` (unfortunately) because otherwise the function signature does not match the JSDoc, which seems more confusing than a a little `arguments` check.

This produces correct types which are identical to the native `toggleAttribute`, while still supporting our third argument behind the scenes **in JS**.

cc @TimvdLippe 
Supersedes and closes #5371

### DIFF
--- a/lib/legacy/legacy-element-mixin.js
+++ b/lib/legacy/legacy-element-mixin.js
@@ -816,11 +816,13 @@ export const LegacyElementMixin = dedupingMixin((base) => {
      * @param {string} name HTML attribute name
      * @param {boolean=} bool Boolean to force the attribute on or off.
      *    When unspecified, the state of the attribute will be reversed.
-     * @param {Element=} node Node to target.  Defaults to `this`.
      * @return {boolean} true if the attribute now exists
      */
-    toggleAttribute(name, bool, node) {
-      node = /** @type {Element} */ (node || this);
+    toggleAttribute(name, bool) {
+      let node = /** @type {Element} */ this;
+      if (arguments.length === 3) {
+        node = /** @type {Element} */ arguments[2];
+      }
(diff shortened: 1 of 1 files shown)


keanulee:
This seems fine to me. @azakus you think this will be okay for internal users?

43081j:
Any chance we can get this merged?

Also, does anyone know what the plan is for next release so i have a rough idea of when we'll get this? (the next time we tag & npm publish it)

keanulee:
cc @rictic @aomarks do you think this will be okay for internal users?

PR #5405, opened 2018-10-14
2.x

JaCkJeNnIfFeR:
### Reference Issue

The master app is better than the test version as it lacks in many components


TimvdLippe:
I am not sure what the intention of this PR is. The description does not clarify either why we would want to pursue this. Therefore, I am closing this PR.

PR #5413, opened 2018-10-25
Hybrid compatibility for PolymerDomApi and Polymer.Iconset types.

aomarks:
Polymer V1 applications are compiled with hand-written externs at https://github.com/google/closure-compiler/blob/master/contrib/externs/polymer-1.0.js. These externs contain some types that do not exist in the V2 code.

PolymerDomApi is the type returned by the V1 Polymer.dom API, but in V2 it returns Polymer.DomApi. This adds the PolymerDomApi interface to the V2 externs, and annotates that Polymer.DomApi implements it. This allows both V1 and V2 code to use the PolymerDomApi type. Similar story for
PolymerDomApi.ObserveCallback.

Polymer.Iconset would ideally live in the iron-iconset repo, but many packages reference the type without actually depending on that library, so its simpler to just include it here, similar to how it worked for
Polymer V1.

### DIFF
--- a/externs/polymer-dom-api-externs.js
+++ b/externs/polymer-dom-api-externs.js
@@ -1,0 +1,179 @@
+/**
+ * @externs
+ * @fileoverview Externs for PolymerDomApi for backwards compatibility with
+ * the Polymer 1 externs.
+ */
+
+/**
+ * A Polymer DOM API for manipulating DOM such that local DOM and light DOM
+ * trees are properly maintained.
+ *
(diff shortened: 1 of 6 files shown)


azakus:
Looks like the type generator didn't like this.

aomarks:
> Looks like the type generator didn't like this.

We don't need these types in TypeScript, so I just updated the generator config to map those types back to their original names. PTAL.

PR #5416, opened 2018-10-31
Add explicit null template for array-selector

### LINKED ISSUE #5415: array-selector is incompatible with strictTemplatePolicy
### Description
Enabling strictTemplatePolicy causes an error to be thrown when array-selector used because it doesn't set its template to null.

#### Steps to Reproduce
* Enable strictTemplatePolicy
* Instantiate an iron-list (which uses array-selector)

#### Expected Results
No error is logged.

#### Actual Results
Error logged to console
`strictTemplatePolicy: expecting dom-module or null template for array-selector`

### Browsers Affected
Probably all, only tested on Chrome and Firefox.

### Versions
- Polymer: 3.0.5-0d3543

arobins:
This makes the element compatible with strictTemplatePolicy.

Fixes #5415.

### DIFF
--- a/lib/elements/array-selector.js
+++ b/lib/elements/array-selector.js
@@ -426,6 +426,7 @@ class ArraySelector extends baseArraySelector {
   // Not needed to find template; can be removed once the analyzer
   // can find the tag name from customElements.define call
   static get is() { return 'array-selector'; }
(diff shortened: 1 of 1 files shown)


kevinpschaaf:
Thanks!

PR #5418, opened 2018-10-31
2.x - Performance optimizations geared towards legacy `Polymer({...})` use

sorvell:
* Behaviors:
  * Now mixed in lazily rather than at class time. This change means behaviors are no longer in the element's prototype chain.
  * A memoized list of behavior lifecycle methods is constructed at finalization time and then used at runtime.
* Adds `Polymer.legacyOptimizations` flag which avoids copying element templates (needed for subclassing) and turns on `stripWhitespace` for all templates.
* TemplateStamp: A TreeWalker is now used to crawl through template nodes. This should be slightly faster when the ShadyDOM polyfill is used.
* `Polymer.telemetry`: now properly records registered elements, even if they have not been created


lib/mixins/element-mixin.html
```diff
@@ -245,31 +252,33 @@
...
-        let s = stylesWithImports[i];
-        let templateStyle = templateStyles[templateStyleIndex];
-        // if the style is not in this template, it's been "included" and
-        // we put a clone of it in the template before the style that included it
-        if (templateStyle !== s) {
-          s = s.cloneNode(true);
-          templateStyle.parentNode.insertBefore(s, templateStyle);
-        } else {
-          templateStyleIndex++;
+      if (!window.skipStyleIncludesAndUrls) {
```
TimvdLippe: Rather than putting this on `window`, shouldn't we put this in the `Polymer.Settings` object?

test/unit/disable-upgrade.html
```diff
@@ -238,30 +242,39 @@ <h2 id="element">[[prop]]</h2>
...
         assert.notOk(el.$.disabledBoundEl.hasCreated);
         assert.notOk(el.$.disabledBoundEl.enabled);
         assert.notOk(el.$.disabledBoundEl.$);
+        el.$.disabledBoundEl.fooHandler = sinon.spy();
+        el.$.disabledBoundEl.fire('foo');
+        assert.equal(el.$.disabledBoundEl.fooHandler.callCount, 0);
       });
 
       test('elements upgrade when `disable-upgrade` removed', function() {
-        assert.notOk(el.$.disabledEl.hasCreated);
```
TimvdLippe: Can we not remove these assertions or are they regressions? Are they a breaking change for our users?

test/unit/mixin-behaviors.html
```diff
@@ -552,7 +551,13 @@
   });
 
   test('nested-behavior dedups', function() {
-    assert.equal(el.behaviors.length, 4);
+    assert.equal(el.behaviors.length, 2);
```
TimvdLippe: This also seems like a regression to me. Are we sure users will not run into any breaking changes?

lib/mixins/element-mixin.html
sorvell: Oops, thanks. Going to get this `window.ShadyCSS.cssBuild` being set (which will shortly be added to `ShadyCSS`.

test/unit/disable-upgrade.html
sorvell: These were actually duplicate assertions from the previous test (https://github.com/Polymer/polymer/pull/5418/files#diff-ed6f0a354a50488549e9b8ea17788d86R240) that were incorrectly being tested again.

test/unit/mixin-behaviors.html
sorvell: It's not a regression since the test changed, but the behavior is weird and I think we should change it.

We didn't have a test where `Polymer.mixinBehaviors` was called on a legacy element with behaviors (since this is pretty exotic) and this test was just updated to do that.

When that happens, the behaviors list is set to the subclasses' list of behaviors. However, the "active" set of behaviors on the element is the de-duplicated flat list of all behaviors on anything in the prototype chain. Given that, it seems like `.behaviors` should be this de-duplicated flattened list of "active" behaviors. We can make that change and update the test.

lib/legacy/class.html
```diff
@@ -96,15 +95,26 @@
...
     // (again same as 1.x)
-    function _mixinBehaviors(behaviors, klass) {
-      for (let i=0; i<behaviors.length; i++) {
-        let b = behaviors[i];
-        if (b) {
-          klass = Array.isArray(b) ? _mixinBehaviors(b, klass) :
-            GenerateClassFromInfo(b, klass);
+    function copyBehaviorProperties(behaviors, klass) {
+      const meta = {};
+      const superMeta = klass.prototype.__behaviorMetaProps;
```
kevinpschaaf: Since these are all lifecycle-related, maybe `__behaviorCallbacks` would be better than something with "props" in it, since that's sort of overloaded with all the data stuff.

lib/legacy/class.html
```diff
@@ -133,31 +143,103 @@
...
         static get properties() {
-          return info.properties;
+          const properties = {};
+          if (this.prototype.__behaviors) {
+            for (let i=0, b; i < this.prototype.__behaviors.length; i++) {
+              b = this.prototype.__behaviors[i];
+              mergeProperties(properties, b.properties);
+            }
+          }
+          mergeProperties(properties, info.properties);
```
kevinpschaaf: It seemed like a possible optimization to skip the `mergeProperties` step from the base info object into an empty object; it seems superfulous for the initial class -- but then I realized this was required for the `computed` to `readOnly` coersion and the shorthand type expansion.  So that was not intuitive, maybe due to the naming?  Suggest different name or at least a comment.

lib/legacy/class.html
```diff
@@ -96,15 +95,26 @@
...
+        klass.prototype.__behaviorMetaProps = meta;
+        for (let i=0; i<behaviors.length; i++) {
+          copyProperties(behaviors[i], klass.prototype);
+          memoizeBehaviorMetaProps(meta, behaviors[i], superMeta);
+        }
+      }
+      klass.prototype.__behaviorMetaProps = meta;
+    }
+
+    function memoizeBehaviorMetaProps(meta, behavior, superMeta) {
```
kevinpschaaf: There's a bug here where the full flattened list of lifecycle callbacks is called once per call to `GenerateClassFromInfo`, which happens each time `mixinBehaviors` is called.

Instead, store the list of callbacks in the closure for one set of behaviors associated with each call to `GenerateClassFromInfo`.

lib/legacy/class.html
```diff
@@ -166,28 +248,57 @@
...
-          super._registered();
           /* NOTE: `beforeRegister` is called here for bc, but the behavior
            is different than in 1.x. In 1.0, the method was called *after*
            mixing prototypes together but *before* processing of meta-objects.
            However, dynamic effects can still be set here and can be done either
            in `beforeRegister` or `registered`. It is no longer possible to set
            `is` in `beforeRegister` as you could in 1.x.
           */
+          const proto = this;
+          if (proto.hasOwnProperty('__behaviors')) {
```
kevinpschaaf: This is just what's in the closure, i.e.
`if (behaviors) {`

lib/legacy/class.html
```diff
@@ -14,18 +14,38 @@
...
       attributeChanged: true,
-      // meta objects
-      behaviors: true
     };
 
+    const noBehaviorCopyProps = Object.assign({
+      behaviors: true
+    }, metaProps);
+
+    const memoizedProps = Object.assign({
```
kevinpschaaf: Naming of these is confusing.  Consider "memoized" --> "filtered"

test/unit/mixin-behaviors.html
```diff
@@ -256,11 +256,10 @@
     <script>
     HTMLImports.whenReady(function() {
       customElements.define('nested-behaviors',
-        class extends Polymer.mixinBehaviors(
+        class extends Polymer.mixinBehaviors([window.BehaviorD, window.LifeCycleBehavior1], Polymer.mixinBehaviors(
```
kevinpschaaf: Add assertions for BehaviorD?

kevinpschaaf:
Also maybe add a test like
```js
const b1 = { ready() {..} };
const b2 = { ready() {..} };
const b3 = { ready() {..} };
const b4 = { ready() {..} };
const sup = Polymer({is: 'sup', behaviors: [b1, b2], ready() {..} })
class sub extends mixinBehaviors([b3,b4], sup) { ready() {..} }
```
And verify that the lifecycle order goes:
```
sub, b4, b3, sup, b2, b1
```

lib/legacy/class.html
sorvell: removed.

lib/legacy/class.html
```diff
@@ -133,31 +143,103 @@
...
+          }
+          return observers;
         }
 
         /**
          * @return {void}
          */
         created() {
-          super.created();
+          const list = this.__behaviorMetaProps.created;
```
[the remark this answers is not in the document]
sorvell: Done.

lib/legacy/class.html
sorvell: Added comment.

lib/legacy/class.html
sorvell: Fixed and test added.

lib/legacy/class.html
sorvell: Done.

lib/legacy/class.html
sorvell: Done.

PR #5421, opened 2018-11-01
3.x - Performance optimizations geared towards legacy `Polymer({...})` use

sorvell:
* Behaviors:
  * Now mixed in lazily rather than at class time. This change means behaviors are no longer in the element's prototype chain.
  * A memoized list of behavior lifecycle methods is constructed at finalization time and then used at runtime.
* Adds `Polymer.legacyOptimizations` flag which avoids copying element templates (needed for subclassing) and turns on `stripWhitespace` for all templates.
* TemplateStamp: A TreeWalker is now used to crawl through template nodes. This should be slightly faster when the ShadyDOM polyfill is used.
* `Polymer.telemetry`: now properly records registered elements, even if they have not been created


LarsDenBakker:
What kind of speedup have you guys seen with these optimizations?

kevinpschaaf:
@LarsDenBakker It very much depends on the profile of the app.  These optimizations could net 5-10% on startup time for an app that heavily utilizes behaviors, and that registers a lot more elements than are actually created for the first render.

LarsDenBakker:
Sounds like my app. Thanks for the info, I will test it out!

PR #5424, opened 2018-11-04
Get Polymer compiling clean under closure recommended flags

rictic:
With these changes we have zero errors and zero warnings with `RECOMMENDED_FLAGS`!

Most of the changes were adding `@override` for methods and properties in mixins. Apparently if you implement an interface you need to say `@override` for each method or property on the interface. This combines with our mixin strategy to the tune of needing to add `@override` on every non-private method and property.

I'm not sure this is intended behavior of the compiler. Filed https://github.com/google/closure-compiler/issues/3137 to see if it is.

Downstreamed as cl/220013205

### DIFF
--- a/lib/legacy/legacy-element-mixin.js
+++ b/lib/legacy/legacy-element-mixin.js
@@ -38,13 +38,12 @@ let styleInterface = window.ShadyCSS;
  * @summary Element class mixin that provides Polymer's "legacy" API
  */
 export const LegacyElementMixin = dedupingMixin((base) => {
-
   /**
    * @constructor
-   * @extends {base}
    * @implements {Polymer_ElementMixin}
    * @implements {Polymer_GestureEventListeners}
    * @implements {Polymer_DirMixin}
+   * @extends {HTMLElement}
    * @private
    */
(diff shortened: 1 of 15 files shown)


lib/legacy/legacy-element-mixin.js
```diff
@@ -816,22 +869,19 @@ export const LegacyElementMixin = dedupingMixin((base) => {
...
-        node = /** @type {Element} */ arguments[2];
-      }
+    toggleAttribute(name, bool, node) {
+      node = /** @type {Element} */ (node || this);
       if (arguments.length == 1) {
         bool = !node.hasAttribute(name);
       }
       if (bool) {
         node.setAttribute(name, '');
-        return true;
```
TimvdLippe: I think this would break our TS compilation. See recent PRs to tackle this.

rictic: Ack! Good catch, that was a merge conflict that I misresolved. PTAL

rictic:
@azakus got the typescript types compiling too, PTAL

PR #5426, opened 2018-11-05
Prefer _template from behavior over dom-module from is.

kevinpschaaf:
See discussion at https://github.com/Polymer/polymer/pull/5295#discussion_r230224477.  Brings master in line with 2.x changes in #5295.

### Reference Issue
#5295


kevinpschaaf:
Closed in favor of https://github.com/Polymer/polymer/pull/5421

Issue #5434, opened 2018-11-15
Is Polymer 3.0 ES6-bundle supported in iOS version 10.3.3 for iPad?


shawkivai:
### Description

We have build a polymer 3.0 project but the project is not executed in Safari 10.3.3 for iPad and also MAC. But in polymer Browser support overview it seems that ES6 supports Safari version 10+. So is there any dependency for iPad Safari version 10.3.3 for supporting ES6-bundle! Please give me a feedback as soon as possible.

Polymer Browser support:
https://polymer-library.polymer-project.org/3.0/docs/browsers

Thanks.

### Browsers Affected

- [ ] Safari 10.3.3

### Versions

- Polymer: v3.x
- webcomponents: vX.X.X 
****

moderndeveloperllc:
@shawkivai Polymer 3 is supported for iOS 10.x, and it would [seem that ES6 is too](http://kangax.github.io/compat-table/es6/). You can do[ remote debugging of Safari](https://blog.idrsolutions.com/2015/02/remote-debugging-ios-safari-on-os-x-windows-and-linux/) to see what the actual error is. If you deal with lots of older systems, you may wish to just transpile to ES5 and see if that solves your issues too.

shawkivai:
@moderndeveloperllc  Thanks for your reply but i am only focusing on the iPad version 10.3.3. For only this specific version ES6 is not supported. Please let me know if Safari 10.3.3 on iPad has some dependency to run ES6!! In iPhone the project is working well. 
I have also got a useful link on this issue : https://apple.stackexchange.com/questions/322467/is-there-a-modern-browser-for-ios-10-3-3-capable-of-supporting-es6-javascript

PR #5435, opened 2018-11-22
Make Polymer great again.

tyloo:
Let's get back to some seriousness, shall we ?

Fixed issues from #1 to #5434

:trollface:


web-padawan:
That wasn't a good joke. Vue has its own problems and limitations.

TimvdLippe:
Let's refrain from naming and shaming with unproductive conversations as a result.

PR #5440, opened 2018-11-29
Avoid initializing properties unnecessarily (perf benefit). [3.x]

kevinpschaaf:
Due to the megamorphic nature of code in base classes with many extensions such as these, we saw these initializations _costing_ more than helping with "object shaping."

### DIFF
--- a/lib/mixins/dir-mixin.js
+++ b/lib/mixins/dir-mixin.js
@@ -136,7 +136,7 @@ export const DirMixin = dedupingMixin((base) => {
     constructor() {
       super();
       /** @type {boolean} */
-      this.__autoDirOptOut = false;
+          this.__autoDirOptOut;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lib/mixins/dir-mixin.js
```diff
@@ -136,7 +136,7 @@ export const DirMixin = dedupingMixin((base) => {
     constructor() {
       super();
       /** @type {boolean} */
-      this.__autoDirOptOut = false;
+          this.__autoDirOptOut;
```
azakus: Weird indent

kevinpschaaf: Thanks. 

Could you also check my closure types?  I sort of just spitballed them in there. Am I correct that removing the initial value means we need to put some sort of type on there to keep closure happy?

PR #5450, opened 2018-12-06
JSDoc fixes for API doc generation.

arthurevans:
Tweaks to get API docs to generate for the Async module in 3.x. Previously nothing was generated for this file. This required removing the "namespace" tags. LMK if that messes with something else.

Also tried to suppress weird extra classes in lib/legacy/legacy-data-mixin.js, without success.

Updated generated API docs in PR here:

https://github.com/Polymer/polymer-library-docs/pull/17

Staged:
https://api-doc-update-dot-polymer-library.appspot.com/3.0/api/utils/async [minimal async doc better than none?]
https://api-doc-update-dot-polymer-library.appspot.com/3.0/api/legacy/legacy-data-mixin [extra classes in docs, dunno why.]

### DIFF
--- a/lib/legacy/legacy-data-mixin.js
+++ b/lib/legacy/legacy-data-mixin.js
@@ -13,6 +13,10 @@ import { Polymer } from '../../polymer-legacy.js';
 import { dedupingMixin } from '../utils/mixin.js';
 import { templatize } from '../utils/templatize.js';
 
+/**
+ * @class UndefinedArgumentError
+ * @private
+ */
 const UndefinedArgumentError = class extends Error {
   constructor(message, arg) {
     super(message);
@@ -76,6 +80,8 @@ export const LegacyDataMixin = dedupingMixin(superClass => {
    * @constructor
    * @extends {superClass}
    * @unrestricted
+   * @polymer
+   * @mixinClass
    * @private   */
   class LegacyDataMixin extends superClass {
     /**
@@ -141,17 +147,24 @@ export const LegacyDataMixin = dedupingMixin(superClass => {
 
 });
 
-// LegacyDataMixin is applied to base class _before_ metaprogramming, to
-// ensure override of _addPropertyEffect et.al. are used by metaprogramming
-// performed in _finalizeClass
+/**
+ * LegacyDataMixin is applied to base class _before_ metaprogramming, to
+ * ensure override of _addPropertyEffect et.al. are used by metaprogramming
+ * performed in _finalizeClass
+ * @private
+ */
 Polymer.Class = (info, mixin) => Class(info, 
   superClass => mixin ? 
     mixin(LegacyDataMixin(superClass)) : 
     LegacyDataMixin(superClass)
 );
 
-// Apply LegacyDataMixin to Templatizer instances as well, and defer
-// runtime switch to the root's host (_methodHost)
+/** 
+ * Apply LegacyDataMixin to Templatizer instances as well, and defer
+ * runtime switch to the root's host (_methodHost)
+ * @private
+ * 
+ */
 templatize.mixin = 
   dedupingMixin(superClass => class extends LegacyDataMixin(superClass) {
     get _legacyUndefinedCheck() {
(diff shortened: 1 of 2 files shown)


arthurevans:
cc: @aomarks who I think has been doing some checkins for typescript declarations...

arthurevans:
@aomarks Yeah, I wasn't sure if the @namespace annotations were meaningful for TypeScript. I don't know how to generate or compare the typescript definitions. If we don't check them in, where do they exist? Are they a build artifact we publish to npm?

arthurevans:
@aomarks OK, I think I figured it out, and here's what I'm seeing.

If we have the `@namespace` tag:

```
/**
 * Async interface wrapper around `requestIdleCallback`.  Falls back to
 * `setTimeout` on browsers that do not support `requestIdleCallback`.
 *
 * @namespace
 * @summary Async interface wrapper around `requestIdleCallback`.
 */
const idlePeriod = {
```
Then the generated `.d.ts` file includes the comment:

```
/**
 * Async interface wrapper around `requestIdleCallback`.  Falls back to
 * `setTimeout` on browsers that do not support `requestIdleCallback`.
 */
declare namespace idlePeriod {
```

However, it also prevents the docs from showing up. (More on that later.)

If we're missing the @namespace tag, we get:

```
declare namespace animationFrame {
```

Bare declaration with no comment (the comments on the namespace members are in  there, though).

In this case, we get the minimal doc shown here: https://api-doc-update-dot-polymer-library.appspot.com/3.0/api/utils/async#microTask.cancel

Note that in the namespace case, the namespaces are included in the analysis output--it's just that our iron-doc-viewer elements have no notion that a namespace could occur inside a module, so they don't get rendered. That's fixable, but I'm not sure it's worth any of our time to do so.

arthurevans:
Let me know what you think of that, @aomarks ...

I'm going to have to roll back changes to legacy-data-mixin.js. They're causing eslint errors and they didn't work, anyway, because apparently we ignore @private and I can't figure out any way to prevent our docs from displaying the bogus classes, anyway.

aomarks:
So with @namespace, the d.ts file has a comment on the namespace, but is otherwise identical? And without @namespace, the generated docs don't show up? Weird! But I don't think having a comment on the d.ts namespace is that important (fixing this for both cases probably requires analyzer changes).

Can you confirm that there are no other d.ts changes?

arthurevans:
Ah, nope. It looks like the `export` statements are missing, too. I'm guessing those are more significant. Here's the complete diff between my generated typings and one from a clean npm install of Polymer 3.1.0:

```
arthure-macbookpro2:updated arthure$ diff lib/utils/async.d.ts ~/work/polymer/ts_decs/node_modules/@polymer/polymer/lib/utils/async.d.ts
10a11,13
> /**
>  * Async interface wrapper around `setTimeout`.
>  */
36a40,44
> export {timeOut};
> 
> /**
>  * Async interface wrapper around `requestAnimationFrame`.
>  */
53a62,67
> export {animationFrame};
> 
> /**
>  * Async interface wrapper around `requestIdleCallback`.  Falls back to
>  * `setTimeout` on browsers that do not support `requestIdleCallback`.
>  */
70a85,95
> export {idlePeriod};
> 
> /**
>  * Async interface for enqueuing callbacks that run at microtask timing.
>  *
>  * Note that microtask timing is achieved via a single `MutationObserver`,
>  * and thus callbacks enqueued with this API will all run in a single
>  * batch, and not interleaved with other microtasks such as promises.
>  * Promises are avoided as an implementation choice for the time being
>  * due to Safari bugs that cause Promises to lack microtask guarantees.
>  */
87a113,114
> export {microTask};
> 
```

aomarks:
Yeah, ok, that will break users. Maybe we just need to update the documentation elements to discover these namespace objects correctly.

arthurevans:
Not sure that's going to happen, but I'll close this PR and open an issue on iron-doc-viewer.

aomarks:
> Not sure that's going to happen, but I'll close this PR and open an issue on iron-doc-viewer.

I can try to take a look sometime, will add it to my todo list.

PR #5460, opened 2018-12-16
Fix case map 

### LINKED ISSUE #5459: casemap bug
### Description

For strings that have already been casemapped (and cashed under caseMap object in https://github.com/Polymer/polymer/blob/master/lib/utils/case-map.js), calling either `dashToCamelCase` or `camelToDashCase` will convert to the other notation, whatever the initial notation is.

#### Steps to Reproduce
```js
Polymer.CaseMap.dashToCamelCase('i-should-stay-dash') => "iShouldStayDash" (o.k.)
Polymer.CaseMap.camelToDashCase('i-should-stay-dash') => "iShouldStayDash" (not o.k. - it should return 'i-should-stay-dash')
```

#### Expected Results
calling camelToDashCase on a 'dashed-string' should keep the sting in a dashed mode.

christophe-g:
### Reference Issue
Fixes #5459 

I branched against 2.7.0 - hence the list of commits. You should be able to sherry-pick 
fc1e370 because it is the only relevent change.

### DIFF
--- a/lib/utils/case-map.html
+++ b/lib/utils/case-map.html
@@ -13,6 +13,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   'use strict';
 
   const caseMap = {};
+  const camelCaseMap = {};
   const DASH_TO_CAMEL = /-[a-z]/g;
   const CAMEL_TO_DASH = /([A-Z])/g;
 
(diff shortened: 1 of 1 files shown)


christophe-g:
@TimvdLippe  https://github.com/Polymer/polymer/pull/5461

PR #5461, opened 2018-12-17
regression test against #5460 and #5459

christophe-g:


TimvdLippe:
Could you add this test to your other PR? That way we can verify your fix works and we won't break Travis along the way.

TimvdLippe:
Closing per above comment. Please add the test to #5460 instead.

PR #5462, opened 2018-12-20
Suppress upcoming jscompiler errors.

rictic:
These errors don't exist yet, but we need to suppress them in the files that will have them so that we're not broken when they're turned on.


TimvdLippe:
@rictic You still want to do this?

rictic:
Ah, yes, merging

PR #5464, opened 2019-01-09
Allow `Polymer({})` calls with ES6 class

azakus:
External copy of cl/228511716


azakus:
Yes, but the test will be funky because this situation is produced by a typescript decorator.

azakus:
@TimvdLippe found a way to add tests :)

azakus:
Chrome 41 failures are in master as well

PR #5472, opened 2019-01-30
Update several string references which can be renamed by closure-compiler

ChadKillingsworth:
I stumbled across a few path references in Polymer that were not reflected. For users who are not using the `EXPORT_ALL` policy of closure compiler, the properties would have been renamed inconsistently.

Once https://github.com/google/closure-compiler/pull/3184 is merged, the compiler will be able to handle reflection of observer and computed properties paths. However references to paths within other code, such as those to `notifyPath`, will still need to be manually reflected.

### DIFF
--- a/lib/elements/array-selector.html
+++ b/lib/elements/array-selector.html
@@ -199,13 +199,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
           let sidx = 0;
           this.__selectedMap.forEach(idx => {
             if (idx >= 0) {
-              this.linkPaths(`${JSCompiler_renameProperty('items', this)}.` + idx, `${JSCompiler_renameProperty('selected', this)}.` + sidx++);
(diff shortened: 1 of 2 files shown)


lib/elements/array-selector.html
```diff
@@ -260,9 +260,9 @@
       }
 
       __selectedIndexForItemIndex(idx) {
-        let selected = this.__dataLinkedPaths['items.' + idx];
+        let selected = this.__dataLinkedPaths[`${JSCompiler_renameProperty('items', this)}.` + idx];
```
TimvdLippe: Nit: `idx` can be in the other string.

ChadKillingsworth: Fixed

ChadKillingsworth:
I did search the entire code base for these type of usages, including:

 - `.notifyPath`
 - `.notifySplices`
 - `.get`
 - `.set`
 - `.push`
 - `.shift`
 - `.unshift`
 - `.splice`

This was all of them.

PR #5474, opened 2019-01-30
Legacy warnings 3.x

### LINKED ISSUE #5468: Overridden computed property functions do not correctly observe dependency properties
### Description

When a behavior overrides a computed property from an earlier behavior, the correct computing function is called, but the arguments are incorrect, it looks like it's being called with the args for the first definition of the computed function.

#### Live Demo

https://github.com/ReadingPlus/polymer-computed-bug

#### Steps to Reproduce
1. clone above repo
2. polymer install
3. polymer test

#### Expected Results
Tests should pass

#### Actual Results
Test fails because computed function called with wrong args

### Browsers Affected

- [X] Chrome
- [X] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

### Versions

- Polymer: v2.7.0
- webcomponents: 1.3.0

This does NOT affect polymer 1.0. I didn't put together a polymer 1 test case because I was trying to upgrade my app from polymer 1 to 2 when I ran into this, so trust me, it works on Polymer 1. Should be easy enough to adapt this case to polymer 1 to prove it.

### LINKED ISSUE #5473: Add warning for undeclared property in template binding
### Description
In Polymer 1.x, any property bound in the template was automatically added to the list of attributes that are deserialized to properties.

In Polymer 2.x this changed, such that only properties declared in `properties` are added to `observedAttributes` and deserialized. This is because template parsing is lazy (deferred to first element instance) while `observedAttributes` must be present at registration time.  That said, properties are used in templates still get setters and work as normal in terms of dynamic updates based on property changes; they just do not have corresponding observed attributes.

Since this is a silent difference between Polymer 1.x and 2.x, we propose adding a warning to the `legacyOptimizations` flag mode (to err on the conservative side, since users may have taken advantage of the terseness of not declaring properties and may not consider it a "bug" per se).

### Versions
- Polymer: v2.x and 3.x

kevinpschaaf:
Adds warnings for a couple of subtle changes in behavior between 1.x and 2.x:
* Warn when re-declaring a computed property (this is no longer allowed in Polymer 2.x; see https://github.com/Polymer/polymer/issues/5468#issuecomment-459038151). Fixes #5468.
* Warn when using an undeclared property in a template binding (only when `legacyOptimizations` flag is enabled). Fixes #5473

### Reference Issue
Fixes #5473 and #5468.

### DIFF
--- a/lib/mixins/property-effects.js
+++ b/lib/mixins/property-effects.js
@@ -2390,15 +2390,6 @@ export const PropertyEffects = dedupingMixin(superClass => {
      * @protected
      */
     static _addTemplatePropertyEffect(templateInfo, prop, effect) {
-      // `dynamicFns` is the flattened property list, so we can use that to
-      // detect non-declared properties.  Properties must be listed in
-      // `properties` to be included in `observedAttributes` since CE V1
-      // reads that at registration time, and we want to keep template parsing
-      // lazy
-      if (legacyOptimizations && !(prop in templateInfo.dynamicFns)) {
-        console.warn(`Property '${prop}' used in template but not declared in 'properties'; ` +
-          `attribute will not be observed.`);
-      }
       let hostProps = templateInfo.hostProps = templateInfo.hostProps || {};
       hostProps[prop] = true;
       let effects = templateInfo.propertyEffects = templateInfo.propertyEffects || {};
--- a/lib/mixins/property-effects.js
+++ b/lib/mixins/property-effects.js
@@ -18,7 +18,7 @@ import { camelToDashCase, dashToCamelCase } from '../utils/case-map.js';
 import { PropertyAccessors } from './property-accessors.js';
 /* for annotated effects */
 import { TemplateStamp } from './template-stamp.js';
-import { sanitizeDOMValue, legacyOptimizations } from '../utils/settings.js';
+import { sanitizeDOMValue } from '../utils/settings.js';
 
 // Monotonically increasing unique ID used for de-duping effects triggered
 // from multiple properties in the same turn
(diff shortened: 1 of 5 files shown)


lib/mixins/property-effects.js
```diff
@@ -2390,6 +2390,9 @@ export const PropertyEffects = dedupingMixin(superClass => {
      * @protected
      */
     static _addTemplatePropertyEffect(templateInfo, prop, effect) {
+      if (legacyOptimizations && !(prop in templateInfo.dynamicFns)) {
+        console.warn(`Property '${prop}' used in template but not declared in 'properties'.`);
```
sorvell:
```suggestion
        console.warn(`Property '${prop}' used in template but not declared in 'properties'. Setting an attribute will not set the corresponding property unless the property is declared in 'properties'.`;
```

lib/mixins/property-effects.js
```diff
@@ -2390,6 +2390,15 @@ export const PropertyEffects = dedupingMixin(superClass => {
      * @protected
      */
     static _addTemplatePropertyEffect(templateInfo, prop, effect) {
+      // `dynamicFns` is the flattened property list, so we can use that to
+      // detect non-declared properties.  Properties must be listed in
+      // `properties` to be included in `observedAttributes` since CE V1
+      // reads that at registration time, and we want to keep template parsing
+      // lazy
+      if (legacyOptimizations && !(prop in templateInfo.dynamicFns)) {
```
sorvell: Using `dynamicFns` here seems a bit abusive since there's no guarantee this is connected to `observedAttributes`. One alternative would be to override this in `element-mixin` and use `this._properties`.

kevinpschaaf: Moved to property-effects.

PR #5476, opened 2019-01-31
Ensure that marshalArgs pulls wildcard info value from __data

### LINKED ISSUE #5475: Wildcard observer value wrong with reentrant change
### Description
When a wildcard observer is queued to run, and a different batched property re-sets a dependency to the observer, the nested call to the observer returns the correct information, but the initially queued observer runs with a stale `value` in the `info` argument.

#### Live Demo
Setup: The `trigger` dependency causes `computeArray` to be run with an uninitialized size (defaults to 0); at this point the `arrayChanged` observer is queued to run, but before it does the `trigger` observer runs and sets `size` to 1. This causes a nested invocation of `computeArray` and `arrayChanged` to be run first (with the correct `info` information reflecting the new array size of 1).  However, when the processing unrolls back to the initially queued invocation of `arrayChanged`, it runs with stale `info` information.

3.x: https://jsbin.com/babeyaz/1/edit?html,console,output (wrong)
1.x: https://jsbin.com/vojode/edit?html,console,output (correct)

#### Expected Results
`info.value.length` always matches `this.array.length` at the time the observer runs.

#### Actual Results
The initial observer invocation (which runs last, due to a re-entrant change after it was enqueued) returns `info.value.length == 0` while `this.array.length == 1`.

Note that in 3.x, the observer runs twice, while in 1.x it only runs once.  This is expected, since in 1.x the initial call to `computeArray` is aborted because of the "undefined" rule (the `size` argument is yet undefined); however in 2.x `comptueArray` runs first due to `trigger`, and subsequently due to `size`.  That is WAI, however the `info` should always reflect the current state of the world

### Browsers Affected
All
### Versions
- Polymer: v3.x, v2.x

kevinpschaaf:
It currently pulls the value from `changedProps` rather than __data, meaning it could provide stale data for re-entrant changes.

### Reference Issue
Fixes #5475

### DIFF
--- a/lib/mixins/property-effects.js
+++ b/lib/mixins/property-effects.js
@@ -964,17 +964,13 @@ function parseArg(rawArg) {
   return a;
 }
 
-function getArgValue(data, props, structured, path) {
-  if (structured) {
-    let value = get(data, path);
-    // when data is not stored e.g. `splices`, get the changed value
-    if (value === undefined) {
-      value = props[path];
-    }
-    return value;
-  } else {
-    return data[path];
+function getArgValue(data, props, path) {
+  let value = get(data, path);
+  // when data is not stored e.g. `splices`, get the changed value
+  if (value === undefined) {
+    value = props[path];
   }
+  return value;
 }
 
 // data api
@@ -992,8 +988,7 @@ function getArgValue(data, props, structured, path) {
  * @private
  */
 function notifySplices(inst, array, path, splices) {
-  let splicesPath = path + '.splices';
-  inst.notifyPath(splicesPath, { indexSplices: splices });
+  inst.notifyPath(path + '.splices', { indexSplices: splices });
   inst.notifyPath(path + '.length', array.length);
 }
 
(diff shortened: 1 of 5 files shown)


lib/mixins/property-effects.js
```diff
@@ -964,6 +964,19 @@ function parseArg(rawArg) {
   return a;
 }
 
+function getArgValue(data, props, structured, path) {
+  if (structured) {
+    let value = get(data, path);
+    // when data is not stored e.g. `splices`, get the changed value
+    if (value === undefined) {
```
sorvell: So we want to pull the live value here, but we fall back to the stored value in `props` for `splices`. Adding a specific check for `splices` would be best but might be a performance hit. What if we take the value in `props` only if it is not `=== undefined`? This allows the live value to be undefined which it sometimes can be.

kevinpschaaf: Added skipped tests and TODO to address this in a separate change.

lib/mixins/property-effects.js
```diff
@@ -964,6 +964,15 @@ function parseArg(rawArg) {
   return a;
 }
 
+function getArgValue(data, props, path) {
+  let value = get(data, path);
+  // when data is not stored e.g. `splices`, get the changed value
```
sorvell: Let's add a comment and/or link to an issue here about this being technically incorrect since it can cause an `undefined` path to return a stale value.

PR #5484, opened 2019-02-12
Fix a couple of minor typos.

kevinpschaaf:
### Reference Issue
Fix typos discovered during 3.x --> 2.x sync.


lib/legacy/polymer.dom.js
```diff
@@ -197,7 +197,7 @@ class DomApiNative {
    * For shadow roots, returns the currently focused element within this
    * shadow root.
    *
-   * return {Node|undefined} Currently focused element
+   * @return {Node|undefined} Currently focused element
```
azakus: This breaks builds internally, I'd say we just remove this line.

PR #5489, opened 2019-02-15
Add `classList` to `Polymer.dom` when `ShadyDOM.noPatch` is used

### LINKED ISSUE #5488: Polymer.dom(el).classList should be supported when ShadyDOM.noPatch is enabled
Currently, this is not available on `Polymer.dom` when `ShadyDOM.noPatch` is used.

sorvell:
Fixes #5488

### DIFF
--- a/lib/legacy/polymer.dom.js
+++ b/lib/legacy/polymer.dom.js
@@ -395,6 +395,10 @@ if (window['ShadyDOM'] && window['ShadyDOM']['inUse'] && window['ShadyDOM']['noP
     }
   });
 
+  forwardReadOnlyProperties(Wrapper.prototype, [
+    'classList'
+  ]);
+
   DomApiImpl = Wrapper;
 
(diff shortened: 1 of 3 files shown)


lib/legacy/polymer.dom.js
```diff
@@ -395,6 +395,10 @@ if (window['ShadyDOM'] && window['ShadyDOM']['inUse'] && window['ShadyDOM']['noP
     }
   });
 
+  forwardReadOnlyProperties(Wrapper.prototype, [
```
kevinpschaaf: Can you add some backstory in a comment about why this appears as such a special case?

It's not a made-up `Polymer.dom` thing, but it also never needed a Shady DOM patch?  Why was it on 1.x's `Polymer.dom` then?

sorvell: Done.

PR #5494, opened 2019-02-19
Avoid over-warning on templatizer props and "static" dynamicFns.

### LINKED ISSUE #5493: New warnings output under `legacyOptimizations` spammy
### Description
The warnings added for binding to undeclared properties in #5473 miss a few different cases where valid properties can be bound:
* templatizer "instance properties" (such as the `item` property in a `dom-repeat`)
* methods used in template with no arguments or only literal arguments (e.g. `[[noDeps()]]` or `[[onlyLiterals('foo', 42)]]`
### Versions
- Polymer: v3.x (master)

kevinpschaaf:
### Reference Issue
Fixes #5493

### DIFF
--- a/lib/mixins/element-mixin.js
+++ b/lib/mixins/element-mixin.js
@@ -11,7 +11,7 @@
  */
 import '../utils/boot.js';
 
-import { rootPath, strictTemplatePolicy, allowTemplateFromDomModule, legacyOptimizations, syncInitialRender } from '../utils/settings.js';
+import { rootPath, strictTemplatePolicy, allowTemplateFromDomModule, legacyOptimizations, legacyWarnings, syncInitialRender} from '../utils/settings.js';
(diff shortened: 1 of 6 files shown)


lib/mixins/element-mixin.js
```diff
@@ -777,7 +777,9 @@ export const ElementMixin = dedupingMixin(base => {
       // The warning is only enabled in `legacyOptimizations` mode, since
       // we don't want to spam existing users who might have adopted the
       // shorthand when attribute deserialization is not important.
-      if (legacyOptimizations && !(prop in this._properties)) {
+      if (legacyOptimizations && !(prop in this._properties) &&
```
sorvell: Let's add some comments about these special cases.

kevinpschaaf: Done

PR #5495, opened 2019-02-20
Add property reflection to notify path and friends calls

ChadKillingsworth:
This is a forward port of #5472 and #5491.

There were no linting errors, but that's as much as I can test. I don't currently have a robust project to test Polymer 3 against with closure-compiler renaming.

### DIFF
--- a/lib/elements/custom-style.js
+++ b/lib/elements/custom-style.js
@@ -75,7 +75,7 @@ export class CustomStyle extends HTMLElement {
    * @export
    * @return {HTMLStyleElement} This element's light-DOM `<style>`
    */
-  ['getStyle']() {
+  getStyle() {
     if (this._style) {
       return this._style;
     }
--- a/lib/elements/custom-style.js
+++ b/lib/elements/custom-style.js
@@ -75,7 +75,7 @@ export class CustomStyle extends HTMLElement {
    * @export
    * @return {HTMLStyleElement} This element's light-DOM `<style>`
    */
-  getStyle() {
+  'getStyle'() {
     if (this._style) {
       return this._style;
     }
--- a/lib/elements/custom-style.js
+++ b/lib/elements/custom-style.js
@@ -75,7 +75,7 @@ export class CustomStyle extends HTMLElement {
    * @export
    * @return {HTMLStyleElement} This element's light-DOM `<style>`
    */
-  'getStyle'() {
+  ['getStyle']() {
     if (this._style) {
       return this._style;
     }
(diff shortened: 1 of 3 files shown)


lib/elements/custom-style.js
```diff
@@ -75,7 +75,7 @@ export class CustomStyle extends HTMLElement {
    * @export
    * @return {HTMLStyleElement} This element's light-DOM `<style>`
    */
-  getStyle() {
+  'getStyle'() {
```
aomarks: Was this change intentional? It causes an error `@export is not supported on this expression.`

lib/elements/custom-style.js
```diff
@@ -75,7 +75,7 @@ export class CustomStyle extends HTMLElement {
    * @export
    * @return {HTMLStyleElement} This element's light-DOM `<style>`
    */
-  'getStyle'() {
+  ['getStyle']() {
```
aomarks: But why do we need this at all? What's wrong wrong with `getStyle()` and the `@export` we already have?

lib/elements/custom-style.js
kevinpschaaf: Good point, export should be sufficient.  Will revert back to `getStyle()`

lib/elements/custom-style.js
kevinpschaaf: Good point, export should be sufficient.  Will revert back to `getStyle()`

ChadKillingsworth: `@export` requires both closure-library and the `--generate_exports` flag to be used with the compiler making it a not-great solution.

There are two options:

 1. Remove the `@export` annotation and quote the property directly.
 2. Manually export the method with a `CustomStyle.prototype['getStyle'] = CustomStyle.prototype.getStyle` statement right below the class definition.

aomarks: All of our internal users have `--generate_exports` enabled. We're also advocating for it to become enabled by default. So I think that's our preferred solution. Any reason you don't want to turn on the flag?

ChadKillingsworth: Yes it requires closure library. That's a non starter for most external projects.

aomarks: Hm, seems odd. I thought it was just a JS compiler pass. Maybe it's like JSCompiler_renameProperty and just requires any implementation of the functions that it generates, since the compiler replaces the call anyway?

ChadKillingsworth: It's very odd. The compiler pass literally adds `goog.exportSymbol` calls but requires the definition of that method to be included as source. Unlike property reflection, the source actually is present in the final output. 

The definition for that method could be injected like a polyfill, but nobody has done that work.

PR #5497, opened 2019-02-20
[3.x] legacy flags (undefined, noBatch, notifyOrder, warnings)

kevinpschaaf:
Adds `legacyUndefined` setting, which reverts to Polymer 1.x rules around `undefined` in multi-property observers & computing functions.

Adds `legacyNoBatch` setting, which reverts to Polymer 1.x behavior around non-batched accessors; each property is initialized and its effects are run one-by one (should always be used in conjunction with `legacyUndefined` to avoid running multi-prop observers at startup more than once), and property sets are flushed immediately during downward binding.

Adds `legacyNotifyOrder` setting, which reverts to Polymer 1.x order for when notifying properties fire change events with respect to other effects.  In Polymer 1.x they fire before observers; in 2.x they fire after all other effect types.

Adds `legacyWarnings` setting which enables warnings useful when migrating from Polymer 1.x to 2.x, specifically warning when non-declared properties are used in a template, since in 2.x these will not automatically be observed attributes.


lib/mixins/property-effects.js
```diff
@@ -1472,6 +1493,9 @@ export const PropertyEffects = dedupingMixin(superClass => {
           this.__dataToNotify = this.__dataToNotify || {};
           this.__dataToNotify[property] = shouldNotify;
         }
+        if (legacyNoBatch) {
+          this._invalidateProperties();
```
sorvell: Add `return false` here since no extra work is needed after this.

lib/mixins/property-effects.js
```diff
@@ -1533,7 +1557,30 @@ export const PropertyEffects = dedupingMixin(superClass => {
      */
     _flushProperties() {
       this.__dataCounter++;
-      super._flushProperties();
+      if (legacyNoBatch && !this.__dataReady) {
+        // The first time this element is being flushed, propagate pending
+        // data down the tree first; this approximates 1.x's distributeConfig
+        this._propagatePropertyChanges(this.__dataPending, {}, false);
+        // Flushing clients causes this element to become __dataReady
```
sorvell: add  " and also triggers `_flushProperties` and `_propagatePropertyChanges` down the tree."

lib/mixins/property-effects.js
```diff
@@ -1533,7 +1557,30 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
+        this._flushClients();
+        // Capture element data and flush properties one-by one to defeat
+        // Polymer 2.x's batched _propertiesChanged scheme; this approximates
+        // 1.x's applyConfig
```
sorvell: Add a note about how  all property getters will return undefined by virtue of resetting `__data`

lib/mixins/property-effects.js
```diff
@@ -1533,7 +1557,30 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
+        const changedProps = this.__data;
+        const notify = this.__dataToNotify;
+        this.__data = {};
+        this.__dataPending = this.__dataToNotify = null;
+        for (let prop in changedProps) {
```
sorvell: In 1.x notifications where handled after the "config." We could do this here by capturing `__data` before propagating and playing through those props first and then the notifying props.

lib/mixins/property-effects.js
```diff
@@ -1676,12 +1723,21 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
-      // Clear notify properties prior to possible reentry (propagate, observe),
-      // but after computing effects have a chance to add to them
-      let notifyProps = this.__dataToNotify;
-      this.__dataToNotify = null;
+      if (legacyNoBatch) {
+        // When not batching, computed effects may re-enter, so capture
+        // notifying properties early
+        notifyProps = this.__dataToNotify;
+        this.__dataToNotify = null;
+        runEffects(this, this[TYPES.COMPUTE], changedProps, oldProps, hasPaths);
```
sorvell: Add a comment about why we're not using `runComputedEffects` here.

lib/mixins/property-accessors.js
```diff
@@ -289,7 +290,14 @@ export const PropertyAccessors = dedupingMixin(superClass => {
      * @override
      */
     _definePropertyAccessor(property, readOnly) {
-      saveAccessorValue(this, property);
+      if (legacyNoBatch) {
+        const ready = this.__dataReady;
```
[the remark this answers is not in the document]
kevinpschaaf: Done

lib/mixins/property-effects.js

lib/mixins/property-effects.js

lib/mixins/property-effects.js

lib/mixins/property-effects.js
kevinpschaaf: Will defer

lib/mixins/property-effects.js

sorvell:
We'll avoid merging this for now. These settings are being tested by some users but we're not sure if this should land in master yet.

bicknellr:
Given that https://github.com/Polymer/polymer/pull/5644 is ready, does this PR need anything before being merged? "DO NOT MERGE" is in the title and description but I'm not sure if that was mainly because it hadn't been reviewed for breaking changes until recently. Does it need a general review?

PR #5499, opened 2019-02-27
[3.x] Use set and clear debouncer upon completion. Fixes #5250.

### LINKED ISSUE #5250: dom-if memory leak
### Description
`dom-if` with restamp causes memory to be leaked until the window is refreshed. This does not work well with SPAs.

### Expected outcome
Memory to be cleaned up when the garbage collector runs after a page has been left.

### Actual outcome
Memory is leaked until the page is refreshed

### Steps to reproduce
Use the polymer-2 starter kit. Modify my-app.html by replacing the normal view code in `iron-pages` with
```javascript
<template is="dom-if" if="[[isView1]]" restamp>
    <my-view1 name="view1"></my-view1>
</template>
<template is="dom-if" if="[[isView2]]" restamp>
    <my-view2 name="view2"></my-view2>
</template>
<template is="dom-if" if="[[isView3]]" restamp>
    <my-view3 name="view3"></my-view3>
</template>
```
Inside of _routePageChanged(page): add this code after `this.page = page || 'view';`
```javascript
if (page === 'view1') {
    this.isView1 = true;
    this.isView2 = false;
    this.isView3 = false;
} else if (page === 'view2') {
    this.isView1 = false;
    this.isView2 = true;
    this.isView3 = false;
} else if (page === 'view3') {
    this.isView1 = false;
    this.isView2 = false;
    this.isView3 = true;
}
```
Adding this code to my-view2.html makes it easy to see the memory leak.
```javascript
connectedCallback() {
    super.connectedCallback();
    this.hugeArray = [];
    for (var i = 0; i < 1000 * 100; i++) {
        this.hugeArray.push(Math.ceil(Math.random() * 100))
    }
}
```
This was taken after switching between the views and forcing a garbage collect after landing back on view1.
image

### Browsers Affected

- [x] Chrome
- [ ] Edge
- [ ] Firefox
- [ ] IE 11
- [ ] Safari 8
- [ ] Safari 9

### Versions
- Polymer: v2.6, v3.0.2
- webcomponents: v1.2

kevinpschaaf:
### Description
The queue used to allow debouncers to be flushable did not have a mechanism to be cleared outside of calling `flush()`.

This change switches the queue to use a `Set` instead of an `Array` to allow for O(1) removals, and ensures debouncers are cleared from the queue once they run.

### Reference Issue
Fixes #5250

### DIFF
--- a/lib/utils/debounce.js
+++ b/lib/utils/debounce.js
@@ -136,8 +136,6 @@ export const enqueueDebouncer = function(debouncer) {
   debouncerQueue.add(debouncer);
 };
 
-window.debouncerQueue = debouncerQueue;
-
 /**
  * Flushes any enqueued debouncers
  *
--- a/lib/utils/debounce.js
+++ b/lib/utils/debounce.js
@@ -123,6 +123,8 @@ let debouncerQueue = new Set();
  * @return {void}
  */
 export const enqueueDebouncer = function(debouncer) {
+  // Re-enqueued debouncers are put at the end of the queue; for Set, this
+  // means removing and re-adding, since forEach traverses insertion order
   if (debouncerQueue.has(debouncer)) {
     debouncerQueue.delete(debouncer);
   }
@@ -136,6 +138,8 @@ export const enqueueDebouncer = function(debouncer) {
  */
 export const flushDebouncers = function() {
   const didFlush = Boolean(debouncerQueue.size);
+  // If new debouncers are added while flushing, Set.forEach will ensure
+  // newly added ones are also flushed
   debouncerQueue.forEach(debouncer => {
     try {
       debouncer.flush();
--- a/lib/utils/debounce.js
+++ b/lib/utils/debounce.js
@@ -47,6 +47,7 @@ export class Debouncer {
     if (this.isActive()) {
       this._asyncModule.cancel(/** @type {number} */(this._timer));
       this._timer = null;
+      debouncerQueue.delete(this);
     }
   }
   /**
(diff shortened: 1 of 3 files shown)


lib/utils/debounce.js
```diff
@@ -112,3 +113,42 @@ export class Debouncer {
...
+/**
+ * Adds a `Debouncer` to a list of globally flushable tasks.
+ *
+ * @param {!Debouncer} debouncer Debouncer to enqueue
+ * @return {void}
+ */
+export const enqueueDebouncer = function(debouncer) {
+  // Re-enqueued debouncers are put at the end of the queue; for Set, this
+  // means removing and re-adding, since forEach traverses insertion order
+  if (debouncerQueue.has(debouncer)) {
```
sorvell: The expected behavior is FIFO so to match the way the array was processed previously, this shouldn't be deleted.

test/unit/debounce.html
```diff
@@ -209,6 +210,59 @@
     });
 
   });
+
+  suite('enqueueDebouncer & flush', function() {
```
sorvell: Let's add a test that checks flush order when debouncers are re-enqueued.

lib/utils/debounce.js
kevinpschaaf: Updated

test/unit/debounce.html
kevinpschaaf: Done

lib/utils/debounce.js
```diff
@@ -45,10 +46,19 @@ export class Debouncer {
    */
   cancel() {
     if (this.isActive()) {
-      this._asyncModule.cancel(/** @type {number} */(this._timer));
+      this._cancelAsync();
       this._timer = null;
+      debouncerQueue.delete(this);
```
sorvell: Add comment about what this does to flush ordering.

kevinpschaaf: Done

lib/utils/debounce.js
```diff
@@ -104,11 +113,49 @@ export class Debouncer {
...
+ * Adds a `Debouncer` to a list of globally flushable tasks.
+ *
+ * @param {!Debouncer} debouncer Debouncer to enqueue
+ * @return {void}
+ */
+export const enqueueDebouncer = function(debouncer) {
+  debouncerQueue.add(debouncer);
+};
+
+window.debouncerQueue = debouncerQueue;
```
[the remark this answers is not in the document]
kevinpschaaf: Done

PR #5504, opened 2019-03-06
Allow value to merge from previous behavior property declaration. Fixes #5503

### LINKED ISSUE #5503: Default value from behavior lost when re-defining property on element
### Description
If a property is declared on a behavior to include a default value, and the property is also declared e.g. on the element (or any subsequent behavior) where the default value is not specified, the property will not be initialized.  This is an unintentional compatibility change from Polymer 1.x.

#### Live Demo
3.x: https://jsbin.com/kuxoqig/edit?html,console,output
1.x: https://jsbin.com/puvowih/edit?html,console,output

#### Expected Results
Console logs `"prop default value: 42"`

#### Actual Results
Console logs `"prop default value: undefined"`

### Browsers Affected
All

### Versions
- Polymer: v2.x-3.x

kevinpschaaf:
### Description
Ensure the `value` field from an earlier behavior's property declaration is carried forward if any successive declaration for the same property omits `value`.  This was the [1.x behavior](https://github.com/Polymer/polymer/blob/a6660bd72521cb9918753d14f7216820abbc35c4/src/standard/configure.html#L105-L118), which overwrote the configuration value when present for each successive behavior followed by the element.

### Reference Issue
Fixes #5503

### DIFF
--- a/lib/legacy/class.js
+++ b/lib/legacy/class.js
@@ -155,6 +155,27 @@ function flattenBehaviors(behaviors, list, exclude) {
   return list;
 }
 
+/**
+ * Copies property descriptors from source to target, overwriting all fields
+ * of any previous descriptor for a property *except* for `value`, which is
+ * merged in from the target if it does not exist on the source.
+ * 
+ * @param {*} target Target properties object
+ * @param {*} source Source properties object
+ */
+function mergeProperties(target, source) {
+  for (const p in source) {
+    const targetInfo = target[p];
+    const value = targetInfo && targetInfo.value;
+    const sourceInfo = source[p];
+    if (sourceInfo.value === undefined && value !== undefined) {
+      target[p] = Object.assign({value}, sourceInfo);
+    } else {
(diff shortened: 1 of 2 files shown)


lib/legacy/class.js
```diff
@@ -155,6 +155,27 @@ function flattenBehaviors(behaviors, list, exclude) {
...
+ * 
+ * @param {*} target Target properties object
+ * @param {*} source Source properties object
+ */
+function mergeProperties(target, source) {
+  for (const p in source) {
+    const targetInfo = target[p];
+    const value = targetInfo && targetInfo.value;
+    const sourceInfo = source[p];
+    if (sourceInfo.value === undefined && value !== undefined) {
```
sorvell: Why not use an `in` check instead of treating `undefined` specially.

kevinpschaaf: Yeah, can do.

kevinpschaaf: Done

PR #5508, opened 2019-03-07
Ensure the debouncer is not already canceled before canceling.

### LINKED ISSUE #5250: dom-if memory leak

kevinpschaaf:
The initial fix for #5250 had a flaw that could lead to an async task being canceled twice (found in internal testing).

### Reference Issue
Fixes #5250

### DIFF
--- a/test/unit/debounce.html
+++ b/test/unit/debounce.html
@@ -40,7 +40,7 @@ suite('enqueueDebouncer & flush', function() {
 
   // NOTE: This is a regression test; the bug it fixed only occured if the
   // debouncer was flushed before any microtasks run, hence it should be
-  // first in this file.
+  // first in this file
   test('re-enqueue canceled debouncer', function() {
     const cb = sinon.spy();
     let db;
@@ -51,6 +51,8 @@ suite('enqueueDebouncer & flush', function() {
     assert.equal(cb.callCount, 0);
     db = Debouncer.debounce(db, microTask, cb);
     enqueueDebouncer(db);
+    flush();
+    assert.isTrue(cb.calledOnce);
   });
 
   const testEnqueue = (shouldFlush, done) => {
--- a/test/unit/debounce.html
+++ b/test/unit/debounce.html
@@ -36,6 +36,59 @@ import { microTask, timeOut, animationFrame, idlePeriod } from '../../lib/utils/
 import { enqueueDebouncer, flush } from '../../lib/utils/flush.js';
 Polymer({is: 'x-basic'});
 
+suite('enqueueDebouncer & flush', function() {
+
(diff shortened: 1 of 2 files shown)


test/unit/debounce.html
```diff
@@ -36,6 +36,59 @@
...
+  test('re-enqueue canceled debouncer', function() {
+    const cb = sinon.spy();
+    let db;
+    db = Debouncer.debounce(null, microTask, cb);
+    enqueueDebouncer(db);
+    db.cancel();
+    assert.equal(db.isActive(), false);
+    assert.equal(cb.callCount, 0);
+    db = Debouncer.debounce(db, microTask, cb);
+    enqueueDebouncer(db);
```
sorvell: Add a comment that part of the test is that this doesn't throw. Also assert that this runs.

kevinpschaaf: Added assert.

PR #5517, opened 2019-04-01
Fix class name bindings

sorvell:
Puts `className` on `Polymer.dom` and ensures any sets to `className` in the binding system, go through that wrapper.

Note, depends on https://github.com/webcomponents/shadydom/pull/326.


sorvell:
Closing in favor of https://github.com/Polymer/polymer/pull/5518

PR #5518, opened 2019-04-02
Ensure `className` bindings work correctly when `ShadyDOM.noPatch` is…

sorvell:
… used.

Puts `className` on `Polymer.dom` and ensures any sets to `className` in the binding system, go through that wrapper.


azakus:
Looks like some legit className failure in Chrome 41 and Safari 9

PR #5519, opened 2019-04-09
Fix type of template getter in element-mixin.html

### LINKED ISSUE #5443: Closure compiler ERROR - The types of the getter and setter for property 'template' do not match.
### Description

Hi, im trying out polymer 2.x in gerrit (using https://gerrit-review.googlesource.com/c/gerrit/+/203832 and https://gerrit-review.googlesource.com/c/gerrit/+/202532)

When i checked out 1b5849e5f4b528d5932722bb8c9229a28f1a8c62 `bazel build release` started to fail with:

```
bazel-out/darwin-fastbuild/bin/polygerrit-ui/app/gr-app.js:6136: ERROR - The types of the getter and setter for property 'template' do not match.
      static set template(value) {
                 ^
  ProTip: "CONFLICTING_GETTER_SETTER_TYPE" can be added to the `suppress` attribute of:
  //polygerrit-ui/app:polygerrit_ui_closure_lib

1 error(s), 1 warning(s)
```

This was working on commit 79de83cee042e4f0532804da8b3951b3159bca16 though.

Caused by https://github.com/Polymer/polymer/commit/8882c3db248634403453e2969d1e9e02f5333b48

#### Live Demo

https://jsbin.com/luhaxab/edit

https://glitch.com/edit/#!/polymer-repro?path=my-element.html:2:0

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

### Versions

- Polymer: 2.x (from the 2.x branch)
- webcomponents: v1.3.0

bro314:
Fix type of template getter in element-mixin.html

Fixes #5443

### DIFF
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -408,7 +408,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        *     }
        *   }
        *
-       * @return {!HTMLTemplateElement|string} Template to be stamped
+       * @return {string|!HTMLTemplateElement|null|undefined} Template to be stamped
        */
       static get template() {
         // Explanation of template-related properties:
@@ -442,7 +442,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       /**
        * Set the template.
        *
-       * @param {!HTMLTemplateElement|string} value Template to set.
+       * @param {string|!HTMLTemplateElement|null|undefined} value Template to set.
        */
       static set template(value) {
         this._template = value;
--- a/lib/mixins/element-mixin.html
+++ b/lib/mixins/element-mixin.html
@@ -408,7 +408,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        *     }
        *   }
        *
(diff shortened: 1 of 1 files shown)


arthurevans:
Just to note, the docs currently recommend returning `null` in one specific (probably very rare) circumstance--if you're extending an element with a template, but you don't want your element to have one. https://polymer-library.polymer-project.org/3.0/docs/devguide/dom-template#elements-with-no-shadow-dom

If the strict template changes changed this behavior, we should update the docs, but pinging @kevinpschaaf here in case he has input.

rictic:
As I read it, this method can return either undefined or null.

undefined if the mixed-in element doesn't have a template, either as a dom-module or inline.

null if the mixed-in element explicitly declares a null template, as we recommend that you do if it's an element with no template (like e.g. iron-ajax)

I'd prefer:

`{string|!HTMLTemplateElement|null|undefined}`

bro314:
Done. PTAL

bro314:
HTMLTemplateElement without ! implicitly means nullable, so I think just removing the ! from the setter parameter type comment is the way to go. Changed both to {HTMLTemplateElement|string}. PTAL

PR #5520, opened 2019-04-10
Avoid upgrading template if no hostProps, for better perf.

kevinpschaaf:
When e.g. a `dom-repeat` only binds to local instance properties (like `item` or `index`), it is unnecessary to "upgrade" the associated `<template>` as a PropertyEffects client, since there will be no host properties to forward to it.  This condition can be detected based on the count of host properties in the template.

### DIFF
--- a/lib/utils/templatize.js
+++ b/lib/utils/templatize.js
@@ -360,6 +360,9 @@ function createTemplatizerClass(template, templateInfo, options) {
 }
 
 /**
+ * Adds propagate effects from the template to the template instance for
+ * properties that the host binds to the template using the `_host_` prefix.
+ * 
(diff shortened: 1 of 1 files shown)


lib/utils/templatize.js
```diff
@@ -364,7 +364,7 @@ function createTemplatizerClass(template, templateInfo, options) {
  */
 function addPropagateEffects(template, templateInfo, options) {
   let userForwardHostProp = options.forwardHostProp;
-  if (userForwardHostProp) {
+  if (userForwardHostProp && templateInfo.hasHostProps) {
```
sorvell: Let's add a comment here or perhaps just in the doc for the function explaining what this does so it's clear why it's only related to `hostProps`.

PR #5523, opened 2019-04-12
[legacy-data-mixin 2.x] Ensure wildcard arguments get undefined treatment. Fixes #5428

### LINKED ISSUE #5428: legacy-data-mixin doesn't work with wildcards observers
Observer is being called despite undefined property value.
Change record is valid, but all value-related properties such as changeRecord.base and changeRecord.value are undefined:
```
{base: undefined, path: "alwaysUndefinedObject", value: undefined}
```
Here's the jsfiddle with example:
https://jsfiddle.net/gaj0umLs/26/

_Originally posted by @vdonich in https://github.com/Polymer/polymer/pull/5266#issuecomment-435533956_

kevinpschaaf:
When checking the results of `_marshalArgs` for `undefined` values, a wildcard argument would normally always have an object and thus would not trigger a warning even if the underlying base object was `undefined`.  In those cases check wildcard argument's base object for `undefined`, which would match Polymer 1's undefined rule behavior.

### Reference Issue
Fixes #5428.

### DIFF
--- a/lib/legacy/legacy-data-mixin.html
+++ b/lib/legacy/legacy-data-mixin.html
@@ -99,7 +99,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         // undefined or not. Multi-property observers must have all arguments defined
         if (this._legacyUndefinedCheck && vals.length > 1) {
           for (let i=0; i<vals.length; i++) {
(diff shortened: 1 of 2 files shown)


kevinpschaaf:
LGTM from sorvell based on #5522 review.

PR #5528, opened 2019-04-22
Upstream compilation fixes

rictic:


rictic:
@ChadKillingsworth FYI, this includes workarounds for https://github.com/google/closure-compiler/issues/3240

stephanwlee:
Would it be possible to port this change over to 2.x and create a patch release for Polymer 2.8? Our project uses Polymer2 (interim; we have plans to upgrade to 3.x ultimately) with Closure compiler and we had to use a weird hack to get around the issue.

PR #5531, opened 2019-05-01
Check directly for // in resolveUrl because it isn't a valid URL

aomarks:
Upstreaming cl/245302268


aomarks:
> Can we add a test please? Thanks.

Done

Issue #5289, opened 2018-07-15
Polymer Gestures preventing programmed clicks on mobile


myfrom:
### Description

When trying to simulate a click event on input (`input.click()`) from other event listener, Polymer Gestures stops that click event.

#### Live Demo

https://jsfiddle.net/myfrom/rypqaxck/18/

#### Steps to Reproduce

1. Open a page on phone or simulate mobile device with DevTools
2. Create a button and an input (best with `<input type="file">` as it opens a dialog when clicked)
3. Add click event listener to that button and call `click()` function on the input

#### Expected Results

Input file is clicked; in case of file input, dialog opens

#### Actual Results

Nothing happends

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

I could check only on Chrome but I think it should happen on other mobile browsers too.

### Versions

- Polymer: v2.6.0
- webcomponents: v1.2.2 

### Additional information

I did a bit of digging while trying to debug it and found out that it's caused by `mouseCanceller()` function ([here](https://github.com/Polymer/polymer/blob/fbbbb19308df2c5e07aa6cc906267879eb8de247/lib/utils/gestures.html#L153)).

However, **I found a workaround:**
Instead of calling `input.click()`, sending non-bubbling event will work:
````javascript
input.dispatchEvent(new MouseEvent('click', { bubbles: false }))
````
  
I'm not sure if it can be fixed without breaking something else but I wanted to post it at least to share the workaround if someone encounters the same problem :wink:

LarsDenBakker:
Having this issue as well. The gestures module is very intrusive in our app, canceling events and modifying native browser behavior. Is there a way to disable it globally?

chrisolof:
Also experiencing this on a project where we utilize one polymer-reliant web component.  Very intrusive and breaks other JS in the project.  I haven't found a way to disable this globally.  Our work-around in this case will be swapping out the polymer-reliant component for a polymer-free component.

LarsDenBakker:
I've created a stackblitz which demonstrates the issue: https://stackblitz.com/edit/lit-element-example-tfopfl?embed=1&file=index.js

I am loading the gestures module, then creating a div which does a programmatic click on a button. On desktop it works, on mobile it doesn't work. This is really a major bug as we can't use a standard browser feature. Is there a way we can at least make this behavior configurable?

PR #5533, opened 2019-05-03
Allow configuring cancelling synthetic click behavior

LarsDenBakker:
Fixes https://github.com/Polymer/polymer/issues/5289

The polymer gestures module cancels synthetic mouse clicks fired by old mobile browsers. This is no longer necessary (see https://developers.google.com/web/updates/2013/12/300ms-tap-delay-gone-away). 

The cancellation detection incorrectly cancels programmatic clicks from javascript directly following a touch action. I added a failing test which demonstrates this behavior. If you run the test in chrome with device simulation enabled, the test would fail if the synthetic click events option is set to true.

This PR adds an option to allow configuring this behavior, leaving the default behavior intact to avoid breaking changes.

If this change is accepted, could we backport this to the Polymer 2 branch as well?

### DIFF
--- a/test/unit/gestures.html
+++ b/test/unit/gestures.html
@@ -24,6 +24,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   <script type="module">
 import './gestures-elements.js';
 import { afterNextRender } from '../../lib/utils/render-status.js';
+import { setCancelSyntheticClickEvents } from '../../lib/utils/settings.js';
 import { resetMouseCanceller, deepTargetFind, recognizers, addListener, removeListener } from '../../lib/utils/gestures.js';
 
 suite('simulate events', function() {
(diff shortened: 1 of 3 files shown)


test/unit/gestures.html
```diff
@@ -104,6 +105,31 @@
...
+    const buttonA = document.createElement('button');
+    const buttonB = document.createElement('button');
+    document.body.appendChild(buttonA);
+    document.body.appendChild(buttonB);
+    buttonB.addEventListener('click', () => {
+      buttonB.parentElement.removeChild(buttonA);
+      buttonB.parentElement.removeChild(buttonB);
+      done();
+    });
+    // simulate a user cick on a button on a touch device
```
TimvdLippe: Not: `cick` -> `click`

LarsDenBakker: Fixed ;)

LarsDenBakker:
@TimvdLippe Thanks! I also created the 2.x version: https://github.com/Polymer/polymer/pull/5535

LarsDenBakker:
@sorvell @kevinpschaaf @azakus could you take a look at this?

LarsDenBakker:
@azakus is it possible to release this as well as https://github.com/Polymer/polymer/pull/5536 ? :)

PR #5534, opened 2019-05-03
Fix class$ bindings for ShadyDOM.noPatch mode

azakus:


azakus:
> Let's add a test. This should be something that binds to a class with an undefined value.

Test added

PR #5537, opened 2019-05-08
Implement scopeSubtree for ShadyDOM noPatch mode

azakus:
Add scopeSubtree as a util function, and use in LegacyElement

### DIFF
--- a/lib/utils/scope-subtree.js
+++ b/lib/utils/scope-subtree.js
@@ -9,13 +9,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 */
 
 import './boot.js';
-import { wrap } from './wrap.js';
 
 const ShadyDOM = window.ShadyDOM;
 const ShadyCSS = window.ShadyCSS;
 
 /**
  * Ensure that elements in a ShadowDOM container are scoped correctly.
+ * This function is only needed when ShadyDOM is used and unpatched DOM APIs are used in third party code.
+ * This can happen in noPatch mode or when specialized APIs like ranges or tables are used to mutate DOM.
  *
  * @param  {!Element} container Container element to scope
  * @param  {boolean=} shouldObserve if true, start a mutation observer for added nodes to the container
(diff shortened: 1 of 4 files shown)


lib/utils/scope-subtree.js
```diff
@@ -0,0 +1,76 @@
...
+*/
+
+import './boot.js';
+import { wrap } from './wrap.js';
+
+const ShadyDOM = window.ShadyDOM;
+const ShadyCSS = window.ShadyCSS;
+
+/**
+ * Ensure that elements in a ShadowDOM container are scoped correctly.
```
sorvell: Let's add a note about when this is needed: only when ShadyDOM is used and unpatched DOM APIs are used in third party code. This can happen in noPatch mode or when specialized APIs like ranges or tables are used to mutate DOM.

lib/utils/scope-subtree.js
```diff
@@ -0,0 +1,76 @@
...
+ * @param  {boolean=} shouldObserve if true, start a mutation observer for added nodes to the container
+ * @return {?MutationObserver} Returns a new MutationObserver on `container` if `shouldObserve` is true.
+ */
+export function scopeSubtree(container, shouldObserve = false) {
+  // If using native ShadowDOM, abort
+  if (!ShadyDOM || !ShadyCSS) {
+    return null;
+  }
+  // ShadyCSS handles DOM mutations in patched mode
+  if (!ShadyDOM['noPatch']) {
```
sorvell: Let's remove this. Then we can handle cases like Ranges where the DOM is unpatched in ShadyDOM.

lib/utils/scope-subtree.js
```diff
@@ -0,0 +1,76 @@
...
+  const ScopingShim = ShadyCSS['ScopingShim'];
+  // if ScopingShim is not available, abort
+  if (!ScopingShim) {
+    return null;
+  }
+  // capture correct scope for container
+  const containerScope = ScopingShim['scopeForNode'](container);
+
+  const scopify = (node) => {
+    const wrappedNode = wrap(node);
```
sorvell: To match Polymer 1 behavior, we should use native querySelector: `ShadyDOM.nativeMethods.querySelectorAll.call(node, ...)`

lib/utils/scope-subtree.js
```diff
@@ -0,0 +1,76 @@
...
+  scopify(container);
+
+  if (shouldObserve) {
+    const mo = new MutationObserver((mxns) => {
+      for (let i = 0; i < mxns.length; i++) {
+        const mxn = mxns[i];
+        for (let j = 0; j < mxn.addedNodes.length; j++) {
+          const addedNode = mxn.addedNodes[j];
+          if (addedNode.nodeType === Node.ELEMENT_NODE) {
+            scopify(addedNode);
```
sorvell: This will not scope the node itself, only its subtree.

lib/utils/scope-subtree.js
azakus: Done.

PR #5540, opened 2019-05-10
Adds support for binding to slots' `name` attribute.

### LINKED ISSUE #5539: Polymer 1 Slotting does not play nice in Firefox & Safari
### Description
Slot is greedy and taking non-slotted elements in Firefox
#### Live Demo
https://codepen.io/anon/pen/joWRpr?editors=1010

expectation: each page has different text saying that it's being shown. this happens in chrome
reality: firefox shows all content on the first page, and it's a mess

### Browsers Affected

- [ ] Chrome
- [x] Firefox
- [?] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

Don't have access to Edge, sorry.
### Versions

- Polymer: v1.X.X

bicknellr:
Previously bindings to the `name` attribute of slots were not supported because, when slots were converted to content elements, the `name$` attribute was not converted to `select$` in the same way the `name` attribute was converted to `select`. Also, changes to a content element's `select` attribute did not trigger redistribution of its root's host element.

Fixes #5539.

### DIFF
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -305,7 +305,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       var name$Value = slot.getAttribute('name$');
       if (name$Value) {
         // See the comment on `_select$Attr`.
-        const select$Attr = this._select$Attr.cloneNode();
+        var select$Attr = this._select$Attr.cloneNode();
         select$Attr.value = '[slot=\'' + name$Value + '\']';
         content.attributes.setNamedItem(select$Attr);
       }
--- a/src/lib/annotations/annotations.html
+++ b/src/lib/annotations/annotations.html
@@ -273,6 +273,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       }
     },
 
+    _select$Attr: (function () {
+      var div = document.createElement('div');
+      div.innerHTML = "<div select$>";
+      return div.childNodes[0].attributes.getNamedItem("select$");
+    })(),
+
     _replaceSlotWithContent: function(slot) {
       var content = slot.ownerDocument.createElement('content');
       while (slot.firstChild) {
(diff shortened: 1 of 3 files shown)


src/lib/annotations/annotations.html
```diff
@@ -273,20 +273,42 @@
...
     _replaceSlotWithContent: function(slot) {
       var content = slot.ownerDocument.createElement('content');
       while (slot.firstChild) {
         content.appendChild(slot.firstChild);
       }
       var attrs = slot.attributes;
       for (var i=0; i<attrs.length; i++) {
-        var attr = attrs[i];
-        content.setAttribute(attr.name, attr.value);
+        content.attributes.setNamedItem(attrs[i].cloneNode());
```
sorvell: Let's add a comment here explaining that cloning the attribute node like this allows attributes with invalid characters to be moved.

src/lib/annotations/annotations.html
```diff
@@ -273,20 +273,42 @@
...
       }
       var name = slot.getAttribute('name');
       if (name) {
         content.setAttribute('select', '[slot=\'' + name + '\']');
       }
+      var name$Value = slot.getAttribute('name$');
```
sorvell: Doesn't the content have a `name$` attribute also in this case? It seems like we should remove that. It also seems like we should remove the `name` attribute in the case above.

src/lib/annotations/annotations.html
```diff
@@ -273,20 +273,42 @@
...
+      if (name$Value) {
+        // See the comment on `_select$Attr`.
+        var select$Attr = this._select$Attr.cloneNode();
+        select$Attr.value = '[slot=\'' + name$Value + '\']';
```
sorvell: It'd be nice not to duplicate the fixup here with the same above.

src/lib/annotations/annotations.html
bicknellr: Added.

src/lib/annotations/annotations.html
bicknellr: I moved these into the loop and do them instead of the copy if the attribute name is `'name'` or `'name$'`.

src/lib/annotations/annotations.html
bicknellr: I moved this into a separate function.

PR #5546, opened 2019-05-22
Sync closure compiler annotations

azakus:
Synced from cl/245073668


azakus:
Firefox 67 seems to be having some issues on Travis, but works locally.

PR #5560, opened 2019-06-19
Optimizations to template elements - fastDomIf & removeNestedTemplates

kevinpschaaf:
New settings:
* `fastDomIf` - uses host's runtime template stamping API rather than templatizer as optimization (lighter-weight, since dom-if doesn't introduce a new scope
* `removeNestedTemplates` - removes the `<template>` element from a `dom-if` or `dom-repeat`, and instead moves the host bindings to the parent


lib/elements/dom-if.js
```diff
@@ -283,7 +322,14 @@ export class DomIf extends PolymerElement {
   _showHideChildren() {
     let hidden = this.__hideTemplateChildren__ || !this.if;
     if (this.__instance) {
-      this.__instance._showHideChildren(hidden);
+      if (fastDomIf) {
+        showHideChildren(hidden, this.__instance.templateInfo.childNodes);
+      } else {
+        this.__instance._showHideChildren(hidden);
```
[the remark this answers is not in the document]
kevinpschaaf: It's not strictly needed now that the functionality is pulled out to exported function `showHideChildren`; just seems safer not to change the `TemplateInstance` API so I left it where it was.

lib/elements/dom-repeat.js
```diff
@@ -339,7 +339,10 @@ export class DomRepeat extends domRepeatBase {
     // until ready, since won't have its template content handed back to
     // it until then
     if (!this.__ctor) {
-      let template = this.template = /** @type {HTMLTemplateElement} */(this.querySelector('template'));
+      // When `removeNestedTemplates` is true, the "template" is the element
+      // itself, which has been given a `_templateInfo` property
+      let template = this.template = this._templateInfo ? this :
```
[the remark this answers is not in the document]
kevinpschaaf: Broadened `template` argument type of `modelForElement` to `HTMLElement`.

lib/mixins/property-effects.js
```diff
@@ -2682,14 +2694,23 @@ export const PropertyEffects = dedupingMixin(superClass => {
         // and link into list of templates if necessary
         templateInfo = /** @type {!TemplateInfo} */(Object.create(templateInfo));
         templateInfo.wasPreBound = wasPreBound;
-        if (!wasPreBound && this.__templateInfo) {
-          let last = this.__templateInfoLast || this.__templateInfo;
-          this.__templateInfoLast = last.nextTemplateInfo = templateInfo;
-          templateInfo.previousTemplateInfo = last;
-          return templateInfo;
+        if (wasPreBound || !this.__templateInfo) {
```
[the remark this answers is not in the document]
kevinpschaaf: This method is called in a few different ways:

* During finalization (`instanceBinding==false`, `wasPreBound==false`):
  * `_bindTemplate(false)` called directly during finalization - parses the template (for the first time), and then assigns that _prototypical_ template info to `__templateInfo` _on the prototype_; note in this case `wasPreBound` is false; this is the first time we're binding it, thus we create accessors.
* During first stamping (`instanceBinding==true`, `wasPreBound==true`):
  * `_stampTemplate` calls `_bindTemplate(true)` - the `templateInfo` returned matches the prototypical one, and so this is `wasPreBound == true` state; thus we _skip_ creating accessors, but _do_ create a sub-instance of the template info to serve as the start of our linked list (needs to be an instance, not the prototypical one, so that we can chain runtime-stamped template infos off of it). Note that this shadows the prototypical `__templateInfo`, such that subsequent stampings of the same template will not be considered "prebound"
* During subsequent runtime stamping (`instanceBinding==true`, `wasPreBound==false`):
  * `_stampTemplate` calls `_bindTemplate(true)` - here `templateInfo` is guaranteed to _not_ match the prototypical one, because it was either a different template altogether, or even if it was the same template, the step above created a sub-instance of the info; in this case `wasPreBound == false`, so we _do_ create accessors, _and_ link a sub-instance into the linked list.

So in short...
* `instanceBinding` determines whether to link a sub-instance of the template info into the linked list (at the root `__templateInfo` if it doesn't exist yet), or set just set it onto the prototype
* `wasPreBound` determines whether we create accessors on the instance or not

So...
* No `wasPreBound` can't return early since it still needs to put a sub-instance of its `templateInfo` on as the root of the linked list.
* And no, `instanceBinding` determines whether you set `__templateInfo` to `templateInfo` or `Object.create(templateInfo)`

However, I think perhaps one improvement to make it less confusing might be to not use the same prototypical property name as the instance property name, since they're basically used differently.  I'll change the prototypical one to `__preBoundTemplateInfo`, whereas the instance-time linked list would stay `__templateInfo`

lib/mixins/property-effects.js
```diff
@@ -2734,13 +2755,13 @@ export const PropertyEffects = dedupingMixin(superClass => {
      * @override
      * @protected
      */
-    _stampTemplate(template) {
+    _stampTemplate(template, templateInfo) {
+      templateInfo =  templateInfo || /** @type {!TemplateInfo} */(this._bindTemplate(template, true));
```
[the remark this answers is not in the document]
kevinpschaaf: Yeah, I can't think of a reason it should matter.

lib/elements/dom-if.js
```diff
@@ -181,8 +183,11 @@ export class DomIf extends PolymerElement {
     let parentNode = wrap(this).parentNode;
     // Guard against element being detached while render was queued
     if (parentNode) {
-      if (!this.__ctor) {
-        let template = /** @type {HTMLTemplateElement} */(wrap(this).querySelector('template'));
+      if (!this.__template) {
```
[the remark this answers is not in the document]
kevinpschaaf: Refactored into subclasses.

lib/elements/dom-if.js
```diff
@@ -242,33 +273,41 @@ export class DomIf extends PolymerElement {
   }
 
   __syncHostProperties() {
-    let props = this.__invalidProps;
+    let props = this.__syncProps;
```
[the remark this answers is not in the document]
kevinpschaaf: Refactored into subclasses.

lib/mixins/property-effects.js
```diff
@@ -2682,14 +2694,23 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
         templateInfo.wasPreBound = wasPreBound;
-        if (!wasPreBound && this.__templateInfo) {
-          let last = this.__templateInfoLast || this.__templateInfo;
-          this.__templateInfoLast = last.nextTemplateInfo = templateInfo;
-          templateInfo.previousTemplateInfo = last;
-          return templateInfo;
+        if (wasPreBound || !this.__templateInfo) {
+          this.__templateInfo = templateInfo;
+        } else {
+          const parent = templateInfo.parent || this.__templateInfo;
```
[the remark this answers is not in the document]
kevinpschaaf: Added comment.

lib/mixins/property-effects.js
```diff
@@ -2780,23 +2808,22 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
       // Unlink template info
       let templateInfo = dom.templateInfo;
-      if (templateInfo.previousTemplateInfo) {
-        templateInfo.previousTemplateInfo.nextTemplateInfo =
-          templateInfo.nextTemplateInfo;
-      }
-      if (templateInfo.nextTemplateInfo) {
-        templateInfo.nextTemplateInfo.previousTemplateInfo =
-          templateInfo.previousTemplateInfo;
+      const {previousSibling, nextSibling, parent} = templateInfo;
```
sorvell: Let's add a comment about how it's important that the parent is not unlinked.

lib/mixins/property-effects.js
```diff
@@ -2925,12 +2952,35 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
         this, node, templateInfo, nodeInfo);
+      const parent = node.parentNode;
+      const nestedTemplateInfo = nodeInfo.templateInfo;
+      const isDomIf = parent.localName === 'dom-if';
+      const isDomRepeat = parent.localName === 'dom-repeat';
+      // Remove nested template and redirect its host bindings & templateInfo
+      // onto the parent (dom-if/repeat element)'s nodeInfo
+      if (removeNestedTemplates && (isDomIf || isDomRepeat)) {
+        parent.removeChild(node);
+        nodeInfo.parentInfo.templateInfo = nestedTemplateInfo;
```
sorvell: So, is this making the `<dom-if>` have a `templateInfo` pointing to this template's info? I think this is probably complex enough that seeding comments step-by-step here, might be good.

lib/mixins/property-effects.js
```diff
@@ -2925,12 +2952,35 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
+        noted = false;
+      }
       // Merge host props into outer template and add bindings
-      let hostProps = nodeInfo.templateInfo.hostProps;
-      let mode = '{';
-      for (let source in hostProps) {
-        let parts = [{ mode, source, dependencies: [source], hostProp: true }];
-        addBinding(this, templateInfo, nodeInfo, 'property', '_host_' + source, parts);
+      let hostProps = nestedTemplateInfo.hostProps;
+      if (fastDomIf && isDomIf) {
```
sorvell: Add a note about why this is only done for `dom-if`?

sorvell: Having 2 settings seems like this is making this pretty complex. Maybe have only one setting?

lib/mixins/property-effects.js
```diff
@@ -2925,12 +2952,35 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
+        if (hostProps) {
+          if (!removeNestedTemplates) {
+            nodeInfo.parentInfo.noted = true;
```
sorvell: Why is this done?

lib/mixins/property-effects.js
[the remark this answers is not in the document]
sorvell: Where is the `parent` originally set?

lib/mixins/template-stamp.js
```diff
@@ -258,7 +260,7 @@ export const TemplateStamp = dedupingMixin(
       if (element.hasAttributes && element.hasAttributes()) {
         noted = this._parseTemplateNodeAttributes(element, templateInfo, nodeInfo) || noted;
       }
-      return noted;
+      return noted || nodeInfo.noted;
```
sorvell: When can `nodeInfo.noted` be set?

lib/mixins/template-stamp.js
```diff
@@ -437,16 +439,19 @@ export const TemplateStamp = dedupingMixin(
...
      */
-    _stampTemplate(template) {
+    _stampTemplate(template, templateInfo) {
       // Polyfill support: bootstrap the template if it has not already been
       if (template && !template.content &&
           window.HTMLTemplateElement && HTMLTemplateElement.decorate) {
         HTMLTemplateElement.decorate(template);
       }
-      let templateInfo = this.constructor._parseTemplate(template);
+      templateInfo = templateInfo || this.constructor._parseTemplate(template);
```
sorvell: I thought `_parseTemplate` memo-ized so why do we have to pass this in?

lib/utils/settings.js
```diff
@@ -23,7 +23,8 @@ export const useNativeCustomElements = !(window.customElements.polyfillWrapFlush
  * `rootPath` to provide a stable application mount path when
  * using client side routing.
  */
-export let rootPath = pathFromUrl(document.baseURI || window.location.href);
+export let rootPath = window.Polymer && window.Polymer.rootPath ||
```
sorvell: Why are these changes necessary?

lib/utils/settings.js
```diff
@@ -263,3 +275,44 @@ export let cancelSyntheticClickEvents = true;
...
+ * type-extended templates as a result of the V1 specification changes.
+ */
+export let removeNestedTemplates =
+  window.Polymer && window.Polymer.removeNestedTemplates || false;
+
+/**
+ * Sets `removeNestedTemplates` globally, to eliminate nested templates
+ * inside `dom-if` and `dom-repeat` as part of template parsing.
+ *
+ * @param {boolean} useCemoveNestedTemplates enable or disable removing nested
```
sorvell: Fix typo `useCemoveNestedTemplates` => `useRemoveNestedTemplates`

lib/utils/templatize.js
```diff
@@ -107,6 +107,48 @@ const templateInstanceBase = PropertyEffects(
...
+        if (hide) {
+          n.__polymerTextContent__ = n.textContent;
+          n.textContent = '';
+        } else {
+          n.textContent = n.__polymerTextContent__;
+        }
+      // remove and replace slot
+      } else if (n.localName === 'slot') {
+        if (hide) {
+          n.__polymerReplaced__ = document.createComment('hidden-slot');
```
sorvell: Bookmark. This seems kinda of nuts v. just doing the show/hide logic for the assignedNodes. Remember why we did it this way?

lib/utils/templatize.js
```diff
@@ -375,20 +379,32 @@ function createTemplatizerClass(template, templateInfo, options) {
...
-      /** @private */
-      klass = templateInfo.templatizeTemplateClass =
-        class TemplatizedTemplate extends templatizedBase {};
+      if (isTemplate) {
+        let templatizedBase = options.mutableData ? MutableDataTemplate : DataTemplate;
+        /** @private */
+        klass = templateInfo.templatizeTemplateClass =
+          class TemplatizedTemplate extends templatizedBase {};
+      } else {
+        // Create a cached subclass of the base custom element class onto which
```
sorvell: How is this cached?

sorvell: Could this also be `Object.create(Object.getPrototypeOf(target)`? It's not just because we're otherwise expecting a class here?

lib/utils/templatize.js
```diff
@@ -402,19 +418,36 @@ function addPropagateEffects(template, templateInfo, options, methodHost) {
...
+      target.__dataOld = null;
+      target._enableProperties();
+    } else {
+      // Swizzle the cached subclass prototype onto the custom element
+      Object.setPrototypeOf(target, klass.prototype);
+      // Check for any pre-bound instance host properties, and do the
+      // instance property delete/assign dance for those (directly into data;
+      // not need to go through accessor since they are pulled at instance time)
+      const hostProps = templateInfo.hostProps;
+      for (let prop in hostProps) {
```
sorvell: When is this necessary?

lib/utils/settings.js
sorvell: Offline discussion: unrelated but a good fix.

lib/utils/templatize.js
sorvell: Doing that could get out of sync so let's leave this as is.

lib/utils/templatize.js
sorvell: Bookmark: it seems like we could potentially not create accessors for these and instead mark them as having accessors so `setPendingProperty` is used and then we wouldn't have to remove the value here.

lib/utils/templatize.js
sorvell: Fix the caching by setting `templateInfo.templatizeTemplateClass = klass`

lib/mixins/property-effects.js
```diff
@@ -1978,11 +1978,23 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
       }
-      let templateInfo = this.__templateInfo;
-      while (templateInfo) {
+      if (this.__templateInfo) {
+        this._runEffectsForTemplate(this.__templateInfo, changedProps, oldProps, hasPaths);
+      }
+    }
+
+    _runEffectsForTemplate(templateInfo, changedProps, oldProps, hasPaths) {
+      const baseRunEffects = (changedProps) => {
```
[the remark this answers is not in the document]
kevinpschaaf: Tested this vs. copying the code between the if/else, and couldn't measure a difference on `template-heavy`.

lib/mixins/property-effects.js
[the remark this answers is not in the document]
kevinpschaaf: Beefed up comments.

lib/mixins/property-effects.js
kevinpschaaf: Done.

lib/mixins/property-effects.js
kevinpschaaf: Simplified the code and added more comments.

lib/mixins/property-effects.js
kevinpschaaf: Done.

lib/mixins/property-effects.js
kevinpschaaf: Added comment.

lib/mixins/template-stamp.js
kevinpschaaf: Added comment.

lib/mixins/template-stamp.js
kevinpschaaf: Added comment

lib/utils/settings.js
kevinpschaaf: Done

lib/utils/templatize.js
kevinpschaaf: Done

lib/utils/templatize.js
kevinpschaaf: I tested a modification to dom-if/dom-repeat that avoided creating forwarding accessors for host properties, and I couldn't measure a difference on `template-heavy`.

For posterity, here's the gambit I tested:
> I fudged `__dataAccessors` which is both what we use to avoid (re-)creating accessors, as well as to know whether to call `_setPendingProperty` or do assignments during binding.  However, this has to be done at a very critical time, after the dom-if/repeat receives its `_templateInfo` for its template (which contains the `hostProps` it needs to care about), but before bindings to it run.  The most sure thing was to make an accessor for `_templateInfo` on dom-if/repeat to trap when it is being set by the host to modify `__dataHasAccessor`:

```js
  set _templateInfo(info) {
    this.__templateInfo = info;
    const hostProps = info.hostProps;
    if (hostProps) {
      const accessors = this.__dataHasAccessor;
      for (const p in hostProps) {
        accessors['_host_' + p] = true;
      }
    }
  }
  get _templateInfo() {
    return this.__templateInfo;
  }
```

> After that, it was just a matter of ensuring a couple places (templatizer property initialization and event handling) read from `__data` rather than via the getter.  

The change avoided creating any accessors for host properties, and allowed me to delete the code here that does the instance property swizzle.  Again, not measurable.

lib/elements/dom-if.js
```diff
@@ -177,81 +255,305 @@ export class DomIf extends PolymerElement {
...
+  // but closure compiler is finnicky
+  /* eslint-disable valid-jsdoc */
+  /**
+   * Abstract API to be implemented by subclass: Returns true if a template
+   * instance has been created and inserted.
+   *
+   * @protected
+   * @return {boolean} True when an instance has been created.
+   */
+  __hasInstance() { }
```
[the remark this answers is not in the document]
kevinpschaaf: The `__instance` prop is of different types in each subclass and thus not something the super class can implement.

lib/elements/dom-if.js
```diff
@@ -267,26 +569,74 @@ export class DomIf extends PolymerElement {
...
+   *
+   * @override
    * @protected
+   * @return {void}
    * @suppress {visibility}
    */
   _showHideChildren() {
-    let hidden = this.__hideTemplateChildren__ || !this.if;
-    if (this.__instance) {
+    const hidden = this.__hideTemplateChildren__ || !this.if;
```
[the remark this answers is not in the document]
kevinpschaaf: The `__instance` prop is of different types in each subclass and thus not something the super class can implement.

lib/elements/dom-if.js
```diff
@@ -177,81 +255,305 @@ export class DomIf extends PolymerElement {
...
+   *
+   * @return {void}
+   */
   __syncHostProperties() {
-    let props = this.__invalidProps;
+    const props = this.__invalidProps;
     if (props) {
-      for (let prop in props) {
-        this.__instance._setPendingProperty(prop, this.__dataHost[prop]);
+      this.__squelchedRunEffects(this.__invalidProps);
```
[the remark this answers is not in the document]
kevinpschaaf: Updated to store all info necessary to sync (`runEffects`, `changedProps`, `hasPaths`) in a single object that gets nulled.

lib/elements/dom-if.js
```diff
@@ -177,81 +255,305 @@ export class DomIf extends PolymerElement {
...
+          this._showHideChildren();
+          changedProps = Object.assign(invalidProps, changedProps);
         }
+        runEffects(changedProps);
+      } else {
+        // Store off any values changed while `if` was false, along with the
+        // runEffects method to sync them, so that we can replay them once `if`
+        // becomes true
+        this.__invalidProps = Object.assign(this.__invalidProps || {}, changedProps);
+        this.__squelchedRunEffects = runEffects;
```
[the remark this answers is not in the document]
kevinpschaaf: Good catch. `hasPaths` also needs to be captured. Updated to store all info necessary to sync (`runEffects`, `changedProps`, `hasPaths`) in a single object, and updated test to ensure hasPaths is correctly used.

lib/elements/dom-if.js
```diff
@@ -143,31 +133,119 @@ export class DomIf extends PolymerElement {
...
         return;
       }
-      this._showHideChildren();
     } else if (this.restamp) {
       this.__teardownInstance();
     }
-    if (!this.restamp && this.__instance) {
-      this._showHideChildren();
-    }
+    this._showHideChildren();
```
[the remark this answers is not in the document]
kevinpschaaf: The superclass can't know about `__instance`, so it's now the child's responsibility. The implementations now have built-in dirty-checking so that it's not expensive to call it unconditionally.

lib/mixins/property-effects.js
```diff
@@ -2685,17 +2691,38 @@ export const PropertyEffects = dedupingMixin(superClass => {
...
+          this.__templateInfo = templateInfo;
+        } else {
+          // Append this template info onto the end of its parent template's
+          // list, which will determine the tree structure via which property
+          // effects are run; if this template was not nested in another
+          // template, use the root template (the first stamped one) as the
+          // parent. Note, `parent` is the `templateInfo` instance for this
+          // template's parent (containing) template, which was set up in
+          // `applyTemplateContent`.  While a given template's `parent` is set
+          // apriori, it is only added to the parent list at the point that it
```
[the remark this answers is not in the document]
kevinpschaaf: Updated to "parent's child list"

lib/elements/dom-if.js
```diff
@@ -177,81 +254,313 @@ export class DomIf extends PolymerElement {
...
+   */
   __syncHostProperties() {
-    let props = this.__invalidProps;
-    if (props) {
-      for (let prop in props) {
-        this.__instance._setPendingProperty(prop, this.__dataHost[prop]);
+    const syncInfo = this.__syncInfo;
+    if (syncInfo) {
+      syncInfo.runEffects(syncInfo.changedProps, syncInfo.hasPaths);
+      this.__syncInfo = null;
```
sorvell: Is it safer to set this first?

kevinpschaaf: Yeah, wouldn't hurt.

kevinpschaaf: Reordered.

PR #5569, opened 2019-07-09
Only consider a class "registered" when it is the top-most prototype

### LINKED ISSUE #5567: Behaviors may not run `registered` if legacy element is subclassed
Root cause of https://github.com/PolymerElements/iron-a11y-keys-behavior/issues/93

In this scenario:
```html
<script type="module">
  import {Polymer} from '@polymer/polymer/polymer-legacy.js';
  const behavior = {
    registered() {
      this.fizz = {};
    }
  };
  Polymer({
    is: 'x-foo',
    behaviors: [behavior]
  });

  class XBar extends customElements.get('x-foo') {}
  customElements.define(XBar.is, XBar);
</script>

<x-bar></x-bar>

<script>
  addEventListener('load', () => {
    const el = document.createElement('x-foo');
    document.body.appendChild(el);
  })
</script>
```

`x-bar` will have `registered` called on it, and `x-foo` will not.
This means that `fizz` will only be set on `x-bar`, and not on `x-foo`'s class, which is causing the error in iron-a11y-keys-behavior.

azakus:
Fixes #5567

### DIFF
--- a/lib/legacy/class.js
+++ b/lib/legacy/class.js
@@ -302,17 +302,23 @@ function GenerateClassFromInfo(info, Base, behaviors) {
       // only proceed if the generated class' prototype has not been registered.
       const generatedProto = PolymerGenerated.prototype;
       if (!generatedProto.hasOwnProperty('__hasRegisterFinished')) {
-        generatedProto.__hasRegisterFinished = true;
-        // ensure superclass is registered first.
-        super._registered();
-        // copy properties onto the generated class lazily if we're optimizing,
-        if (legacyOptimizations) {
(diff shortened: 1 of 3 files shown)


balloob:
A recent PR has caused merge conflicts and this PR needs rebasing.

balloob:
This is still a problem.

balloob:
We're locally applying a patch now with Home Assistant, still would be nice to see it fixed upstream

steverep:
This came in July 2019... why was it left out of 3.4 and 3.5? 😞

PR #5570, opened 2019-07-12
Legacy undefined no batch disable upgrade

sorvell:
Adds `disable-upgrade` support to legacy elements by default.

### DIFF
--- a/lib/legacy/class.js
+++ b/lib/legacy/class.js
@@ -238,6 +238,24 @@ function GenerateClassFromInfo(info, Base, behaviors) {
       return observedAttributesGetter.call(this).concat(DISABLED_ATTR);
     }
 
+    // Prevent element from initializing properties when it's upgrade disabled.
+    /** @override */
+    _initializeProperties() {
+      if (!this.hasAttribute(DISABLED_ATTR)) {
+        super._initializeProperties();
+      } else {
+        this.__isUpgradeDisabled = true;
+      }
+    }
+
+    // Prevent element from enabling properties when it's upgrade disabled.
+    /** @override */
+    _enableProperties() {
+      if (!this.__isUpgradeDisabled) {
+        super._enableProperties();
+      }
+    }
+
     /**
      * @override
      * @param {string} name Attribute name.
@@ -248,8 +266,14 @@ function GenerateClassFromInfo(info, Base, behaviors) {
      */
     attributeChangedCallback(name, old, value, namespace) {
       if (name == DISABLED_ATTR) {
-        if (!this.__dataEnabled && value == null && this.isConnected) {
-          super.connectedCallback();
+        // When disable-upgrade is removed, intialize properties and
+        // provoke connectedCallback if the element is already connected.
+        if (!this.__dataEnabled && value == null) {
+          super._initializeProperties();
+          this.__isUpgradeDisabled = false;
+          if (this.isConnected) {
+            super.connectedCallback();
+          }
         }
       } else {
         super.attributeChangedCallback(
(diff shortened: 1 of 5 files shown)


lib/legacy/class.js
```diff
@@ -211,9 +215,89 @@ function GenerateClassFromInfo(info, Base, behaviors) {
...
+        this.__isUpgradeDisabled = true;
+      } else {
+        super._initializeProperties();
+      }
+    }
+
+    // Prevent element from enabling properties when it's upgrade disabled.
+    /** @override */
+    _enableProperties() {
+      if (!this.__isUpgradeDisabled) {
```
kevinpschaaf: Normally we would never get here, since this is normally called from `connectedCallback` and that also has a gate. However `Templatizer` boots up trees of elements by forcing `_enableUpgrade` on them, even if they're not connected (since booting elements out of the tree has shown to be performance positive). Worth a comment.

lib/legacy/class.js
```diff
@@ -211,9 +215,89 @@ function GenerateClassFromInfo(info, Base, behaviors) {
...
+     * @param {?string} old The previous value for the attribute.
+     * @param {?string} value The new value for the attribute.
+     * @param {?string} namespace The XML namespace for the attribute.
+     * @return {void}
+     */
+    attributeChangedCallback(name, old, value, namespace) {
+      if (name == DISABLED_ATTR) {
+        // When disable-upgrade is removed, intialize properties and
+        // provoke connectedCallback if the element is already connected.
+        if (!this.__dataEnabled && value == null) {
```
kevinpschaaf: Using `this. __isUpgradeDisabled` might read more clearly.

lib/legacy/class.js
```diff
@@ -211,9 +215,89 @@ function GenerateClassFromInfo(info, Base, behaviors) {
...
+          super._initializeProperties();
+          this.__isUpgradeDisabled = false;
+          if (this.isConnected) {
```
kevinpschaaf: Need a `wrap` and test for removing `disable-upgrade` out of the tree.

lib/legacy/legacy-element-mixin.js
```diff
@@ -47,7 +53,8 @@ export const LegacyElementMixin = dedupingMixin((base) => {
    * @extends {HTMLElement}
    * @private
    */
-  const legacyElementBase = DirMixin(GestureEventListeners(ElementMixin(base)));
+  const legacyElementBase = builtCSS ? GesturesElement :
```
kevinpschaaf: The unconditional `@implements {Polymer_DirMixin}` annotation _should_ be fine here?

lib/mixins/disable-upgrade-mixin.js
```diff
@@ -15,6 +15,8 @@ import { dedupingMixin } from '../utils/mixin.js';
 
 const DISABLED_ATTR = 'disable-upgrade';
 
+let observedAttributesGetter;
```
kevinpschaaf: This is a lot of duplicated code; would be ideal to see if just using the mixin in `GenerateClassFromInfo` affects perf, and if not factor it so it can be used.

lib/mixins/element-mixin.js
```diff
@@ -466,7 +470,8 @@ export const ElementMixin = dedupingMixin(base => {
           this.prototype.hasOwnProperty(JSCompiler_renameProperty('_template', this.prototype)) ?
           this.prototype._template :
           // Look in dom-module associated with this element's is
-          (getTemplateFromDomModule(/** @type {PolymerElementConstructor}*/ (this).is) ||
+          ((this.hasOwnProperty(JSCompiler_renameProperty('is', this)) &&
```
sorvell: Why is this change needed? It may be better in general, but it's unclear why it's required.

sorvell: OK, it's so that an `is` below this prototype cannot override `_template` below this prototype.

PR #5577, opened 2019-07-30
Workaround bindings to textarea.placeholder in IE

azakus:
Textareas have a very weird bug in IE, where the text of the placeholder
is copied to the content of the textarea when set.

This a creates two bugs:
1. An unintended binding is made to the textContent of the textarea's
   text child node, meaning updates to the `placeholder` will be an
   unnecessary binding process in the best case, or an exception thrown
   when updating the text child node in the worst case.
2. When `legacyOptimizations` is enabled, the child node of the text
   area is removed when the binding for `placeholder` is processed and
   removed, leaving a binding to a `null` node, and throwing exceptions.

Therefore, when we detect this placeholder behavior, we will remove the
textnode before template processing, preventing both bugs.

### DIFF
--- a/lib/mixins/template-stamp.js
+++ b/lib/mixins/template-stamp.js
@@ -28,6 +28,7 @@ let placeholderBug = false;
 
 function hasPlaceholderBug() {
   if (!placeholderBugDetect) {
+    placeholderBugDetect = true;
     const t = document.createElement('textarea');
     t.placeholder = 'a';
     placeholderBug = t.placeholder === t.textContent;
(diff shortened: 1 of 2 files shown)


sorvell:
For the second problem, this is caused because when `textarea` has a `placeholder` in IE, it generates `textContent` of the same value. Polymer sees this and creates a binding to the `textarea.textContent`. Polymer removes the `placeholder` attribute and on IE this removes the `textContent` of the textarea. Therefore, when the template is stamped, there is no node in the `textarea` and Polymer cannot find the node to which the `textarea.textContent` binding should point.

lib/mixins/template-stamp.js
```diff
@@ -22,6 +22,49 @@ const templateExtensions = {
...
+}
+
+/**
+ * Some browsers have a bug with textarea, where placeholder text is copied as
+ * a textnode child of the textarea.
+ *
+ * If the placeholder is a binding, this can break template stamping in two
+ * ways.
+ *
+ * One issue is that when the `placeholder` binding is removed, the textnode
```
sorvell: `placeholder` attribute is removed when the binding is processed

lib/mixins/template-stamp.js
```diff
@@ -22,6 +22,49 @@ const templateExtensions = {
...
+ * child of the textarea is deleted, and the template info tries to bind into
+ * that node.
+ *
+ * When `legacyOptimizations` is enabled, the node is removed from the textarea
```
sorvell: When the template is stamped and the `textarea.textContent` binding is processed, no corresponding node is found because it was removed during parsing. An exception is generated when this binding is updated.

lib/mixins/template-stamp.js
```diff
@@ -22,6 +22,49 @@ const templateExtensions = {
...
+ * when the `placeholder` binding is processed, leaving an "undefined" cell in
+ * the binding metadata object.
+ *
+ * When `legacyOptimizations` is disabled, the template is cloned before
```
sorvell: When `legacyOptimizations` is not used, the template is cloned before processing and this changes the above behavior. The cloned template also has a `value` property set to the `placeholder` and `textContent`. This prevents the removal of the `textContent` when the `placeholder` attribute is removed. Therefore the exception does not occur. There's an extra unnecessary binding.

lib/mixins/template-stamp.js
```diff
@@ -22,6 +22,49 @@ const templateExtensions = {
   'dom-if': true,
   'dom-repeat': true
 };
+
+let placeholderBugDetect = false;
+let placeholderBug = false;
+
+function hasPlaceholderBug() {
+  if (!placeholderBugDetect) {
+    const t = document.createElement('textarea');
```
sorvell: `placeholderBugDetect = true`

lib/mixins/template-stamp.js
```diff
@@ -22,6 +22,49 @@ const templateExtensions = {
...
+ * When `legacyOptimizations` is disabled, the template is cloned before
+ * processing, and has an extra binding to the textContent of the text node
+ * child of the textarea. This at best is an extra binding to process that has
+ * no useful effect, and at worst throws exceptions trying to update the text
+ * node.
+ *
+ * @param {!Node} node Check node for placeholder bug
+ * @return {boolean} True if placeholder is bugged
+ */
+function shouldFixPlaceholder(node) {
```
sorvell: Suggest refactoring to `fixPlaceholder` and apply the fix.

PR #5585, opened 2019-08-27
Add support for ShadyDOM on-demand patching

sorvell:
Do not merge until on https://github.com/webcomponents/polyfills/pull/189.

### DIFF
--- a/externs/polymer-externs.js
+++ b/externs/polymer-externs.js
@@ -139,13 +139,7 @@ Polymer.syncInitialRender;
 Polymer.legacyUndefined;
 
 /** @type {boolean} */
-Polymer.legacyNoBatch;
-
-/** @type {boolean} */
 Polymer.legacyWarnings;
-
-/** @type {boolean} */
-Polymer.legacyNotifyOrder;
 
 /** @type {boolean} */
 Polymer.orderedComputed;
@@ -158,6 +152,9 @@ Polymer.removeNestedTemplates;
 
 /** @type {boolean} */
 Polymer.suppressTemplateNotifications;
+
+/** @type {boolean} */
+Polymer.lazyDefine;
 
 // nb. This is explicitly 'var', as Closure Compiler checks that this is the case.
 /**


lib/elements/dom-repeat.js
```diff
@@ -347,11 +347,11 @@ export class DomRepeat extends domRepeatBase {
           /** @type {!HTMLElement} */ (this));
       let template = this.template = thisAsTemplate._templateInfo ?
           thisAsTemplate :
-          /** @type {!HTMLTemplateElement} */ (this.querySelector('template'));
+          /** @type {!HTMLTemplateElement} */ (wrap(this).querySelector('template'));
```
kevinpschaaf: nbd, but nothing should ever be in a `dom-repeat` or `dom-if` except for a template (i.e. no nested elements with SR's that would need to be filtered out), so this is basically fine to always be native qs

lib/utils/templatize.js
```diff
@@ -121,11 +121,11 @@ export function showHideChildren(hide, children) {
       } else if (n.localName === 'slot') {
         if (hide) {
           n.__polymerReplaced__ = document.createComment('hidden-slot');
-          wrap(wrap(n).parentNode).replaceChild(n.__polymerReplaced__, n);
+          wrap(wrapIfNeeded(n).parentNode).replaceChild(n.__polymerReplaced__, n);
         } else {
           const replace = n.__polymerReplaced__;
           if (replace) {
-            wrap(wrap(replace).parentNode).replaceChild(n, replace);
+            wrap(wrapIfNeeded(replace).parentNode).replaceChild(n, replace);
```
kevinpschaaf: Not totally clear why the outer `wrap` can't be `wrapIfNeeded`.  Adding a `<slot>` to any random spot in a shadow root obviously wouldn't be automatically on-demand patched, but this is only putting a `<slot>` _back_ into a node that already had it (which I think means it is guaranteed to be patched)?

lib/mixins/property-effects.js
```diff
@@ -926,7 +926,7 @@ function setupCompoundStorage(node, binding) {
       // We may also want to consider doing this for `textContent` and
       // `innerHTML`.
       if (target === 'className') {
-        node = wrap(node);
+        node = wrapIfNeeded(node);
```
kevinpschaaf: Seems like setting `className` on an otherwise uninteresting element in a SR (i.e. didn't need to have inside/outside patched) needs to always be wrapped so we don't lose scoping.

Might explain the test failure here? https://travis-ci.org/Polymer/polymer/builds/580988656#L780 (although unclear why this would only fail on Chrome 41?)

lib/utils/wrap.js
```diff
@@ -22,3 +22,7 @@ export const wrap = (window['ShadyDOM'] && window['ShadyDOM']['noPatch'] && wind
   window['ShadyDOM']['wrap'] :
   (window['ShadyDOM'] ? (n) => ShadyDOM['patch'](n) : (n) => n);
 
+// Wrapping a local operation (e.g. appendChild) is only needed iff noPatch is
+// explicitly true, and on-demand patching is not used.
+export const wrapIfNeeded = (window['ShadyDOM'] && window['ShadyDOM']['noPatch'] === true && window['ShadyDOM']['wrap']) ?
```
kevinpschaaf: Let's add comment to list known cases where `wrap` and not `wrapIfNeeded` is required:
* When moving (`appendChild`/`insertBefore`) a node without an explicit `removeChild` to an unpatched location: in this case it needs to be unlinked from logical tree, have its className unscoped, trigger `slotchange` and, trigger distribution if it was a `<slot>`
* When setting `className` to an unpatched node: if it was in a SR it needs to go through ShadyCSS to avoid losing its scoping class
* When calling `querySelector` on an unpatched node: it won't return scoped results
* When getting `textContent` or `innerHTML` on an unpatched node: it won't return scoped results

Also, now wondering if `wrap` and `wrapAlways` might be better naming (and less annoying to boot)?

lib/elements/dom-repeat.js
sorvell: Dome

lib/mixins/property-effects.js
sorvell: Fixed. The issue is that `patch` needs to be the fallback to wrapping so that this works when instance patching is required.

lib/utils/templatize.js
sorvell: Done.

lib/utils/wrap.js
sorvell: Added.

test/unit/lazy-define.html
```diff
@@ -0,0 +1,96 @@
...
+    test('Polymer function does return class when `_needsCtor` is set', function() {
+      assert.ok(window.XNeedsCtor);
+    });
+
+    test('paser generated elements upgrade', function() {
+      let e = document.querySelector('x-lazy-declarative');
+      assert.ok(e.root.querySelector('span'));
+    });
+
+    test('imperatively generated elements upgrade and customElements.get works after instance created', function() {
```
kevinpschaaf: Add a test for something like 
```js
Polymer.telemetry.registrations.filter(r => r.is == 'x-lazy-imperative').length === 0
```
?

PR #5587, opened 2019-08-29
Ensure parent is linked to child templateInfo. Fixes fastDomIf unstamping issue.

kevinpschaaf:
`_stampTemplate` was failing to link the `parent` field of the `templateInfo` linked tree. This resulted in the `lastChild` not being correctly unlinked if it was unstamped, causing subsequent children to be linked off of the removed `lastChild`. Since the host iterates the tree from `firstChild`, it would never see the added templates and their effects (updates subsequent to first render) would not be run.

### DIFF
--- a/lib/mixins/property-effects.js
+++ b/lib/mixins/property-effects.js
@@ -2866,7 +2866,7 @@ export const PropertyEffects = dedupingMixin(superClass => {
       // parent list, a template's `parent` reference is never removed, since
       // this is is determined by the tree structure and applied at
       // `applyTemplateInfo` time.
-      let templateInfo = dom.templateInfo;
+      const templateInfo = dom.templateInfo;
       const {previousSibling, nextSibling, parent} = templateInfo;
       if (previousSibling) {
         previousSibling.nextSibling = nextSibling;
@@ -2878,6 +2878,7 @@ export const PropertyEffects = dedupingMixin(superClass => {
       } else if (parent) {
         parent.lastChild = previousSibling;
       }
+      templateInfo.nextSibling = templateInfo.previousSibling = null;
       // Remove stamped nodes
       let nodes = templateInfo.childNodes;
       for (let i=0; i<nodes.length; i++) {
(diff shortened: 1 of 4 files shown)


lib/mixins/property-effects.js
```diff
@@ -2756,6 +2756,7 @@ export const PropertyEffects = dedupingMixin(superClass => {
           // once).
           const parent = template._parentTemplateInfo || this.__templateInfo;
           const previous = parent.lastChild;
+          templateInfo.parent = parent;
```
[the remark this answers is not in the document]
kevinpschaaf: It may not have had a template in `applyTemplateInfo`, so we always need to set the `parent` calculated on line 2757 to `templateInfo.parent` here.  So while we could potentially change `applyTemplateInfo`, we couldn't remove this code here. Seems not worth it.

kevinpschaaf:
Added more tests to catch issues that would arise from not setting `parent` reference.

test/unit/dom-if-elements.js
```diff
@@ -126,9 +126,15 @@ Polymer({
       <x-foo prop="{{prop1}}"></x-foo>
       <template is="dom-if" id="if-2" if="{{shouldStamp2}}">
         <x-foo prop="{{prop2}}"></x-foo>
-        <template is="dom-if" id="if-3" if="{{shouldStamp3}}">
+        <template is="dom-if" id="if-3" if="{{shouldStamp3}}" restamp>
```
sorvell: Let's name these to reflect the nesting.

lib/mixins/property-effects.js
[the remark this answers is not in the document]
sorvell: OK, makes sense.

test/unit/property-effects-template.html
```diff
@@ -162,6 +162,30 @@
 </script>
 </dom-module>
 
+<script type="module">
+  import { PolymerElement, html } from '../../polymer-element.js';
+  customElements.define('x-runtime-nested', class extends PolymerElement {
+    static get template() {
+      return html`<template id="t1">
```
sorvell: Let's rename these to reflect the nesting.

kevinpschaaf: Done.

PR #5591, opened 2019-09-12
Applies micro-optimizations and removes obsolete settings

sorvell:
Applies micro-optimizations that were found to improve element creation benchmarks by 5-10%, and removes obsolete settings:

* Removed `legacyNoBatch` and `legacyNotifyOrder` settings.
* dom-if/repeat: `dom-change` and `renderedCount` no longer fire with `legacyOptimizations` set.
* `LegacyElementMixin`: `isAttached` now set before calling `connectedCallback` so it is batched with initial rendering.
* `PropertiesChanged`: property accessor code now inlined for efficiency rather than calling `_get/_setProperty`. The `__dataCounter` tracking flag has been moved here to avoid the need to override `_flushProperties` in `PropertyEffects`.
* `PropertyEffects`: inlined `runEffectsForProperty` into `runEffects` for efficiency. Removed wrapping around sending data events.
* `async`: In the microtask scheduler, now only provoke a DOM mutation if needed.

### DIFF
--- a/lib/mixins/property-effects.js
+++ b/lib/mixins/property-effects.js
@@ -1362,6 +1362,8 @@ export const PropertyEffects = dedupingMixin(superClass => {
       this.__readOnly;
       /** @type {!TemplateInfo} */
       this.__templateInfo;
+      /** @type {boolean} */
+      this._overrideLegacyUndefined;
     }
 
     get PROPERTY_EFFECT_TYPES() {
--- a/lib/mixins/property-effects.js
+++ b/lib/mixins/property-effects.js
@@ -320,6 +320,11 @@ function dispatchNotifyEvent(inst, eventName, value, path) {
   if (path) {
     detail.path = path;
   }
+  // As a performance optimization, we could elide the wrap here since notifying
+  // events are non-bubbling and shouldn't need retargeting. However, a very
+  // small number of internal tests failed in obscure ways, which may indicate
+  // user code relied on timing differences resulting from ShadyDOM flushing
+  // as a result of the wrapped `dispatchEvent`.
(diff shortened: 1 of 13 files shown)


lib/utils/settings.js
```diff
@@ -278,3 +278,22 @@ export let fastDomIf = window.Polymer && window.Polymer.fastDomIf || false;
...
+ * Setting to disable `dom-change` and `rendered-item-count` events from
+ * `dom-if` and `dom-repeat`. Users can opt back into `dom-change` events by
+ * setting the `notify-dom-change` attribute (`notifyDomChange: true` property)
+ * to `dom-if`/`don-repeat` instances.
+ */
+export let suppressTemplateNotifications =
+  window.Polymer && window.Polymer.suppressTemplateNotifications || false;
+
+/**
+ * Sets `fastDomIf` globally, to put `dom-if` in a performance-optimized mode.
```
sorvell: Fix docs, not `fastDomIF`

lib/mixins/property-effects.js
```diff
@@ -320,8 +320,7 @@ function dispatchNotifyEvent(inst, eventName, value, path) {
   if (path) {
     detail.path = path;
   }
-  // Note, since this does not bubble, it should not need to be wrapped.
-  inst.dispatchEvent(new CustomEvent(eventName, { detail }));
+  wrap(/** @type {!HTMLElement} */(inst)).dispatchEvent(new CustomEvent(eventName, { detail }));
```
sorvell: Add comment about why this needs wrap (flushing)

lib/utils/settings.js
kevinpschaaf: Done

lib/mixins/property-effects.js

PR #5594, opened 2019-10-04
Add module field

rslawik:
https://unpkg.com/@polymer/polymer@3.3.0?module fails to load saying that the package is not an ES module. Adding `module` field should allow it to load (see: https://github.com/mjackson/unpkg/blob/4bb6cf424a6f7e803fc98c13662ee4d04bfe4af2/modules/middleware/validateFilename.js#L7)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "3.3.0",
   "description": "The Polymer library makes it easy to create your own web components. Give your element some markup and properties, and then use it on a site. Polymer provides features like dynamic templates and data binding to reduce the amount of boilerplate you need to write",
   "main": "polymer-element.js",
+  "module": "polymer-element.js",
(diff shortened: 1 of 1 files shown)


justinfagnani:
@azakus @kevinpschaaf how do we maintain the CHANGELOG in Polymer? Is it generated or updated by hand?

kevinpschaaf:
Yeah, the changelog in the polymer repro predates when we made changelogs per PR; we run a script at release time to gather the commit log.

rslawik:
If the changelog is updated automatically, can we merge this PR in?

rslawik:
Ping. It seems that the CHANGELOG update is not a blocker to merge this.
@justinfagnani @azakus 

Once merged, when is it expected to be published to npm?

PR #5595, opened 2019-10-09
Adds basic support for `adoptedStyleSheets`

sorvell:
Uses `shadowRoot.adoptedStyleSheets` for element styling when supported (currently Chrome) and only if a css build is used. The build restriction is in place in order to support `@apply` via ShadyCSS. Currently ShadyCSS works only with a style element in the element template.

### DIFF
--- a/lib/utils/settings.js
+++ b/lib/utils/settings.js
@@ -8,15 +8,26 @@ Code distributed by Google as part of the polymer project is also
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 */
 import './boot.js';
-
 import { pathFromUrl } from './resolve-url.js';
 export const useShadow = !(window.ShadyDOM) || !(window.ShadyDOM.inUse);
 export const useNativeCSSProperties = Boolean(!window.ShadyCSS || window.ShadyCSS.nativeCss);
 export const useNativeCustomElements = !(window.customElements.polyfillWrapFlushCallback);
 export const supportsAdoptingStyleSheets = useShadow &&
     ('adoptedStyleSheets' in Document.prototype) &&
-    ('replaceSync' in CSSStyleSheet.prototype);
-
+    ('replaceSync' in CSSStyleSheet.prototype) &&
+    // Since space may change, feature detect exact API we need
+    (() => {
+      try {
+        const sheet = new CSSStyleSheet();
+        sheet.replaceSync('');
+        const host = document.createElement('div');
+        host.attachShadow({mode: 'open'});
+        host.shadowRoot.adoptedStyleSheets = [sheet];
+        return (host.shadowRoot.adoptedStyleSheets[0] === sheet);
+      } catch(e) {
+        return false;
+      }
+    })();
 
 /**
  * Globally settable property that is automatically assigned to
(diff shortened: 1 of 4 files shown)


lib/utils/settings.js
```diff
@@ -10,9 +10,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 import './boot.js';
 
 import { pathFromUrl } from './resolve-url.js';
-export const useShadow = !(window.ShadyDOM);
+export const useShadow = !(window.ShadyDOM) || !(window.ShadyDOM.inUse);
 export const useNativeCSSProperties = Boolean(!window.ShadyCSS || window.ShadyCSS.nativeCss);
 export const useNativeCustomElements = !(window.customElements.polyfillWrapFlushCallback);
+export const supportsAdoptingStyleSheets = useShadow &&
+    ('adoptedStyleSheets' in Document.prototype) &&
+    ('replaceSync' in CSSStyleSheet.prototype);
```
kevinpschaaf: Given that the spec is now somewhat in question re: frozen array vs. mutation APIs, would it make sense to try and make this test as defensive as possible in case they end up changing it and need to deprecate the ability to assign an array?  I don't know exactly what it'd be but maybe something like 
```js
const testSheet = new ConstructableStylesheet();
try { document.adoptedStylesheets = [testSheet]; } catch (e) { }
supportsAdoptingStyleSheets = supportsAdoptingStyleSheets && 
  document.adoptedStylesheets[0] === testSheet;
```
(maybe made safe to allow having set some before)?

If so, might also consider making this change to LitElement's detection.

test/unit/styling-build-adopted-stylesheets.html
```diff
@@ -0,0 +1,1027 @@
+<!doctype html>
```
kevinpschaaf: Is this just `styling-scoped.html`, with the flag turned on and `@apply` removed?

sorvell: Yes, with the addition of one test that checks `adoptedStyleSheets`

lib/utils/settings.js
sorvell: Done.

lib/utils/settings.js
```diff
@@ -8,12 +8,26 @@ Code distributed by Google as part of the polymer project is also
...
 import { pathFromUrl } from './resolve-url.js';
-export const useShadow = !(window.ShadyDOM);
+export const useShadow = !(window.ShadyDOM) || !(window.ShadyDOM.inUse);
 export const useNativeCSSProperties = Boolean(!window.ShadyCSS || window.ShadyCSS.nativeCss);
 export const useNativeCustomElements = !(window.customElements.polyfillWrapFlushCallback);
-
+export const supportsAdoptingStyleSheets = useShadow &&
+    ('adoptedStyleSheets' in Document.prototype) &&
+    ('replaceSync' in CSSStyleSheet.prototype) &&
+    // Since space may change, feature detect exact API we need
```
kevinpschaaf:
```suggestion
    // Since spec may change, feature detect exact API we need
```

PR #5598, opened 2019-10-18
Add node field to PolymerDomApi

rictic:
And keep DomApiNative type compatible with PolymerDomApi.

Upstreaming cl/275091781

### DIFF
--- a/externs/polymer-dom-api-externs.js
+++ b/externs/polymer-dom-api-externs.js
@@ -14,7 +14,7 @@
  *
  * @interface
  */
-var PolymerDomApi = function() {};
+let PolymerDomApi = function() {};
 
 /**
  * @param {?Node} node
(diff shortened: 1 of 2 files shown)


aomarks:
I think this should go to `legacy-undefined-noBatch`, since that's what Google tracks?

kevinpschaaf:
Either place... we're regularly merging `master` into `legacy-undefined-noBatch`.

kevinpschaaf:
Curious what was relying on `node`?  It's basically an implementation detail of the PolymerDomAPI class (despite it not being prefixed), so it smells odd that something would reference it.

rictic:
It looks like it's from a desire to unwrap a PolymerDomApi value back to its underlying dom value in order to pass it to a basic DOM handling API. Might have been totally based on a type error from the compiler and it may have worked fine at runtime, in which case maybe they should just cast the PolymerDomApi as a Node instead.

To check out the code, see http://cs/setPolymerNodeContent

kevinpschaaf:
@sorvell Can you review?

PR #5600, opened 2019-10-22
Remove TimvdLippe from CODEOWNERS

TimvdLippe:
I am no longer an active maintainer of the Polymer repository, but I am still receiving review requests for PRs. While I could review PRs, it would be better for active maintainers to review community PRs. I am still available for reviews where my domain knowledge would be required, but I don't think I should be added a reviewer to every PR.


masonlouchart:
We already miss you @TimvdLippe 💔 😭

TimvdLippe:
I am still active in the web components community, so feel free to reach out to me if you need to!

PR #5605, opened 2019-11-14
Update the polyfills; fix a type generation error.

bicknellr:
`base = class extends base {};` causes a type `declare class base extends base {}` to be emitted but this fails to verify in the next step. Also, I didn't want to ignore everything called `base` so I renamed it.

### DIFF
--- a/gen-tsd.json
+++ b/gen-tsd.json
@@ -8,6 +8,7 @@
     "index.html"
   ],
   "excludeIdentifiers": [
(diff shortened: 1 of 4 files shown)


bicknellr:
Lerna didn't update the dependencies in webcomponentsjs during the last publish because the package itself didn't have changes. Fixing in https://github.com/webcomponents/polyfills/pull/244 and then I'll bump here.

PR #5606, opened 2019-11-16
Update test browser list.

bicknellr:


bicknellr:
Nevermind, the plan is to update test runner so that the custom elements polyfill is not forced instead.

PR #5622, opened 2020-01-22
Allow `undefined` in legacy _template field to fall-through to normal lookup path

kevinpschaaf:
Allow `undefined` in legacy _template field to fall-through to normal lookup path.

Allows a user to unconditionally set `_template` to a value and have build-time controls to change the value between a template and `undefined` to allow build-time experimentation with transitioning from `<dom-module>` templates to inlined templates in JS.

### DIFF
--- a/test/unit/inheritance.html
+++ b/test/unit/inheritance.html
@@ -17,6 +17,18 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   <script type="module" src="../../polymer-legacy.js"></script>
 <body>
 
+  <dom-module id="uses-underscore-template">
+    <template><div>error</div></template>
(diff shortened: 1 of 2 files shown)


test/unit/inheritance.html
```diff
@@ -17,6 +17,18 @@
...
+  <dom-module id="uses-underscore-template">
+    <template><div>error</div></template>
+  </dom-module>
+
+  <dom-module id="uses-dom-module">
+    <template><div>from-dom-module</div></template>
+  </dom-module>
+
+  <dom-module id="no-template">
+    <template><div>error</div>></template>
```
sorvell:
```suggestion
    <template><div>error</div></template>
```

PR #5630, opened 2020-02-13
Store syncInfo on instance and don't sync paths. Fixes #5629

### LINKED ISSUE #5629: Falsey fastDomIf can lead to invalid batched paths
### Description
When `fastDomIf` is enabled (`legacy-undefined-noBatch` branch), path notifications will batch in `syncInfo` until the `dom-if` becomes truthy again, at which point all batched path notifications will be included in `changedProps`.  Polymer 2/3 has a rule that paths must not be batched, because the property effect de-duping mechanism will only present the _first_ matching path to an observer, which is unexpected and can cause program errors.

Although there is a fundamental flaw in Polymer 2/3's "falsey dom-if squelchinng" related to path notifications (#4818), `fastDomIf` mistakenly batches them rather than dropping them like the non-fastDomIf case, and keeps them batched across re-stamped instances, which even when using the workarounds for #4818 (`restamp` or `mutable-data`) can still cause errors.  The goal for this issue is to match the non-fastDomIf behavior & limitations.

### Versions
- Polymer: unreleased (`legacy-undefined-noBatch` branch)

kevinpschaaf:
* Previously the `syncInfo` used to store changed properties while the dom-if was false was stored on the dom-if. This means that even under `restamp`, any stored properties from a previous instance would be applied to the next instance. This change moves the `syncInfo` storage to the instance, so it naturally goes away when the instance is discarded.  Any new instance will take current values from the host.
* Due to limitations described in #4818, it is bad/illegal to allow paths to be batched and played through `runEffects`, since effect de-duping occurs by _root property_, so effects will only be running once for the first path matching an effect, with all other paths being dropped on the ground.  This can be particularly bad if the user e.g. `set` a path to an object, and then subsequently nulled the object; the observer would then be acting on a path that is no longer valid.  This change only stores the root property & value for any paths that come in, which matches the non-`fastDomIf` behavior with only storing `root(prop)` in `__invalidProps`.

### Reference Issue
Fixes #5629

### DIFF
--- a/lib/elements/dom-if.js
+++ b/lib/elements/dom-if.js
@@ -252,7 +252,7 @@ class DomIfBase extends PolymerElement {
       this.__teardownInstance();
     }
     this._showHideChildren();
-    if ((!suppressTemplateNotifications || this.notifyDomChange) 
+    if ((!suppressTemplateNotifications || this.notifyDomChange)
         && this.if != this._lastIf) {
       this.dispatchEvent(new CustomEvent('dom-change', {
         bubbles: true,
@@ -339,7 +339,6 @@ class DomIfFast extends DomIfBase {
   constructor() {
     super();
     this.__instance = null;
-    this.__syncInfo = null;
   }
 
   /**
@@ -386,7 +385,8 @@ class DomIfFast extends DomIfBase {
     // Install runEffects hook that prevents running property effects
     // (and any nested template effects) when the `if` is false
     templateInfo.runEffects = (runEffects, changedProps, hasPaths) => {
-      const syncInfo = this.__syncInfo;
+      const instance = this.__instance;
+      let syncInfo = instance && instance.__syncInfo;
       if (this.if) {
         // Mix any props that changed while the `if` was false into `changedProps`
         if (syncInfo) {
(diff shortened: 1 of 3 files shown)


lib/elements/dom-if.js
```diff
@@ -397,25 +397,34 @@ class DomIfFast extends DomIfBase {
...
-          syncInfo.hasPaths = syncInfo.hasPaths || hasPaths;
-          Object.assign(syncInfo.changedProps, changedProps);
-        } else {
-          this.__syncInfo = {
-            runEffects,
-            changedProps: Object.assign({}, changedProps),
-            hasPaths
-          };
+        if (instance) {
+          if (!syncInfo) {
```
sorvell: Think it's more readable to put the positive case first.

lib/elements/dom-if.js
```diff
@@ -397,25 +397,34 @@ class DomIfFast extends DomIfBase {
...
+            syncInfo = instance.__syncInfo = { runEffects, changedProps: {} };
```
sorvell: Why is it ok to drop the passed in `changedProps` here? (The next section is only for `hasPaths`.)

lib/elements/dom-if.js
```diff
@@ -386,7 +385,8 @@ class DomIfFast extends DomIfBase {
     // Install runEffects hook that prevents running property effects
     // (and any nested template effects) when the `if` is false
     templateInfo.runEffects = (runEffects, changedProps, hasPaths) => {
-      const syncInfo = this.__syncInfo;
+      const instance = this.__instance;
+      let syncInfo = instance && instance.__syncInfo;
```
sorvell: The non-fast case stores `__invalidProps` on the dom-if. Why is that ok if this must be changed?

kevinpschaaf: Yeah, for the `restamp` case, it should probably be on the instance as well.  This change isn't exactly _required_ for correctness once the other part of the bug (not storing paths) is fixed; it really just prevented `restamp` from being a good workaround for the bug.  But probably good to move it.

lib/elements/dom-if.js
kevinpschaaf: They're added in the `Object.assign` below;  just a refactoring of the code.

lib/elements/dom-if.js
sorvell: sorry, this is the only case, ignore.

kevinpschaaf: It's actually not an else case.

lib/elements/dom-if.js
kevinpschaaf: Ended up nulling __syncInfo at teardown, to match what we do with __invalidProps

PR #5644, opened 2020-03-30
v3.4.0 release notes and commit list

bicknellr:
Adds release notes for all the new and changed behavior I could find in the `legacy-undefined-noBatch` branch. ~~I still need to collect some more info about the feasibility of new behavior around constructable stylesheets.~~ *(This behavior is now behind a flag: `useAdoptedStyleSheetsWithBuiltCSS`.)*

- [x] Merge #5645 and update this PR
- [x] Release Polymer
- [x] Add commit list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 ## Unreleased
 
-### New settings
+### New global behaviors
+
+#### Built Shady CSS styles automatically use constructable stylesheet objects
+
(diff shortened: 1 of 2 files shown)


bicknellr:
I added a section on `useAdoptedStyleSheetsWithBuiltCSS`, PTAL

bicknellr:
I was just trying out `./util/gen-changelog.sh` and it puts commit messages in a list above the old changelog content. I'd like to put these above the raw list of commits, so I wonder if this PR would be better if it went into master as part of the typical post-release changelog commit rather than editing the changelog as part of this branch?

bicknellr:
Is that fix enabled by default? (i.e. unrelated to if `reuseChunkedInstances` is set)

kevinpschaaf:
Right, it's an unconditional fix to a behavior that regressed in Polymer 2/3 vs. Polymer 1.

bicknellr:
Ok, added, PTAL. Also, what do you think about the PR-moving situation I mentioned earlier?

bicknellr:
Ok, I'll wait until https://github.com/Polymer/polymer/pull/5497 is released and then add the commit log to this PR.

PR #5645, opened 2020-04-03
legacy-undefined-noBatch: Put `adoptedStyleSheets` use behind a global setting

bicknellr:
Currently, the `legacy-undefined-noBatch` branch always uses `adoptedStyleSheets` to share styles between component instances if supported by the browser and provided with built CSS. This PR puts this feature behind a new global setting (`shareBuiltCSSWithAdoptedStyleSheets`).


bicknellr:
Ok, I updated the name, WDYT?

PR #5666, opened 2020-06-23
Loosen _attachDom return signature.

### LINKED ISSUE #5665: _attachDom type signature is too restrictive
### Description

`_attachDom`'s docs say that it can be overriden to support any type of node that is not just only a shadow root. The return types though seem to be `ShadowRoot | null` which means that you cannot return `this` as an acceptable root.

https://github.com/Polymer/polymer/blob/76c71e186ecc605294c3575dd31ac7983a8b3ae3/lib/mixins/element-mixin.js#L734

https://github.com/Polymer/polymer/blob/36fb37e9d37195065e035d8c82f57789d95e670b/lib/mixins/element-mixin.d.ts#L221

#### Steps to Reproduce

1. override _attachDom
2. place the snippet below
3. attempt to compile with closure or TS

```js
_attachDom(dom) {
  this.appendChild(dom);
  return this;
}
```

#### Expected Results

no compilation error

#### Actual Results

compilation error

### Browsers Affected

N/A

### Versions

- Polymer: v3.4.1

tomoueda:


test/unit/polymer.element.html
```diff
@@ -80,6 +80,11 @@
     this._calledReady++;
   }
 
+  _attachDom(dom) {
+    this.appendChild(dom);
+    return this;
+  }
```
rictic, line 86: Was this change necessary, or is it demonstrating usage?

With this change, the test element doesn't have a shadow root I think. Could we assert for that in this test?

rictic:
Fixes #5665

rictic:
Please also update the changelog

test/unit/polymer.element.html
tomoueda: Not necessary. In the Contributor docs, it says every change has to be accompanied by some change in tests.

I'll try adding an assert.

PR #5668, opened 2020-07-01
Cancel chunking when disconnected. Fixes #5667

### LINKED ISSUE #5667: dom-repeat keeps running asynchronous code after detached if chunking is enabled.
### Description
When `dom-repeat` is used with chunked rendering, it schedules a `requestAnimationFrame` which is not cancelled when the element is disconnected from the DOM. As a result, even after the component is detached from the DOM, it keeps running code.

#### Live Demo
https://jsbin.com/sirelojane/1/edit?html,console,output

#### Steps to Reproduce
Run the script, look at the console. 

#### Expected Results
No `RAF` should have been printed after the component is detached

#### Actual Results
There is at least a couple more.

### Browsers Affected
- [X] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

I only tested on Chrome, but looking at the implementation of `dom-repeat` it should affect all of them.

### Versions
- Polymer: 2.8.0 and 3.4.1
- webcomponents: v1

### Potential fix
Adding `cancelAnimationFrame(this.__chunkingId);` to `disconnectedCallback` in dom-repeat.js seems to resolve the issue, but I don't know what other consequences that could have.

kevinpschaaf:
If there was a pending `requestAnimationFrame` for chunking when the `dom-repeat` was disconnected, cancel the rAF so that we're not spending extra cycles creating DOM that no one will see, especially when tearing down/discarding views.  `__chunkingId` retains its handle upon disconnection as a signal that a render is required to kick chunking back off if the `dom-repeat` is subsequently re-connected.

Note that this still isn't a guarantee that rendering/chunking can't restart while the `dom-repeat` is disconnected (e.g. if any render-causing properties change while disconnected, such as `items`), since we don't gate all of `dom-repeat`'s work on actually being connected; that would be a larger change with the potential to break users relying on the current ability to render while disconnected.

### Reference Issue

Fixes #5667

### DIFF
--- a/test/unit/dom-repeat.html
+++ b/test/unit/dom-repeat.html
@@ -3935,7 +3935,7 @@ suite('chunked rendering', function() {
   let chunked;
   let notifyChange;
   let targetCount;
-  let handleChange = () => {
+  const handleChange = () => {
     if (!targetCount || chunked.$.repeater.renderedItemCount >= targetCount) {
       notifyChange(Array.from(chunked.root.querySelectorAll('*:not(template):not(dom-repeat)')));
     }
--- a/test/unit/dom-repeat.html
+++ b/test/unit/dom-repeat.html
@@ -3933,16 +3933,24 @@ suite('limit', function() {
 suite('chunked rendering', function() {
 
   let chunked;
-  let resolve;
+  let notifyChange;
   let targetCount;
-  const handleChange = () => {
+  let handleChange = () => {
     if (!targetCount || chunked.$.repeater.renderedItemCount >= targetCount) {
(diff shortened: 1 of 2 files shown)


test/unit/dom-repeat.html
```diff
@@ -3933,16 +3933,24 @@ <h4>x-repeat-limit</h4>
 suite('chunked rendering', function() {
 
   let chunked;
-  let resolve;
+  let notifyChange;
   let targetCount;
-  const handleChange = () => {
+  let handleChange = () => {
```
sorvell, line 3938: Change back to `const`?

kevinpschaaf: Done

PR #5687, opened 2021-05-13
Make sure that throwing an exception in an observer does not leave the PropertiesChanged mixin in an invalid state

balint777:
…e PropertiesChanged mixin in an invalid state


balint777:
@googlebot I signed it!

PR #5690, opened 2021-07-10
Fix `Polymer.dom(el).attachShadow/shadowRoot`

sorvell:
Polymer itself does not need this since it directly uses `ShadyDOM.wrap`, but since this is a foot gun omission, `attachShadow` and `shadowRoot` is added here to the DOM wrapper returned by Polymer.dom.


kevinpschaaf (approved):
LGTM

PR #5701, opened 2022-06-01
Adds Trusted Types support for reflected boolean properties.

bicknellr:
Like <https://github.com/Polymer/polymer/commit/8582dd6449822e003313dee3e6792aa0a8f98989>, this is also to work around <https://crbug.com/993268>. Given that [this bug should be fixed now](https://bugs.chromium.org/p/chromium/issues/detail?id=993268#c19), I'm not sure if we want to upstream this (since we're already including a workaround for it elsewhere) or leave it as an internal only patch.

### DIFF
--- a/lib/mixins/properties-changed.js
+++ b/lib/mixins/properties-changed.js
@@ -531,7 +531,13 @@ export const PropertiesChanged = dedupingMixin(
       if (str === undefined) {
         node.removeAttribute(attribute);
       } else {
-        node.setAttribute(attribute, str);
+        node.setAttribute(
+            attribute,
+            // Closure's type for `setAttribute`'s second parameter incorrectly
+            // excludes `TrustedScript`.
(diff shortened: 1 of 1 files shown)


kevinpschaaf (approved):
LGTM

rictic (approved):

rictic:
Oh yay, the on* bug was fixed!

bicknellr:
SGTM. Also, I just realized that I didn't update the types, so I'll work on that and post again once I've pushed a fix.

bicknellr:
Closure's type for `setAttribute` seems to be wrong, so I had to cast `.emptyScript` with `?`. Also, I moved this into the same method as the `setAttribute` call so that I didn't have to update `_serializeValue`'s return value or its callers. PTAL

rictic (approved):

PR #5705, opened 2023-01-12
Update support notice

arthurevans:
Just an update to the README.


arthurevans:
Left a comment on https://github.com/Polymer/polymer-library-docs/pull/61 ... same thing applies here. You can totally replace this language with something different, but you should definitely update the support notice.

PR #5756, opened 2026-07-01
reject prototype-polluting path segments in Path.set

zayeem06:
### Reference Issue
N/A

`Path.set` walks a dotted path and assigns straight through each segment, so a path whose segments include `__proto__`, `prototype`, or `constructor` (for example `__proto__.polluted` or `constructor.prototype.polluted`) writes onto `Object.prototype` rather than the target object. This helper backs the element `set()`/`notifyPath` public API and the two-way binding change path, so an app that forwards an externally influenced path into `set` ends up polluting every object. The fix rejects those segments inside `set` and returns without writing, which keeps the pollution guard in the shared helper instead of at each call site. Legitimate paths are unaffected, and I added a regression case to `test/unit/path.html` that asserts the magic keys no-op while ordinary nested and top-level sets still work.


zayeem06:
any update?

PR #5757, opened 2026-07-15
Harden template trust and prototype mutation handling

M0nd0R:
## What changed
- hardens template trust checks so only framework-created nested templates are accepted by `html`
- blocks prototype-mutation paths in `path.js` and `PropertyEffects` public path APIs
- routes legacy reflection helpers through `sanitizeDOMValue`
- closes strict-template-policy bypasses involving `dom-module` assetpath/style/custom-style lookups
- hardens stamped ID maps and global `dom-module` registries against `__proto__` special-property mutation

## Why
These flows allowed attacker-controlled markup or path input to bypass trust boundaries, reintroduce unsanitized DOM writes, or mutate framework-owned object prototypes.

## Impact
This reduces exposure to template trust bypasses, strict-template-policy bypasses, legacy sanitizer bypasses, and prototype-pollution primitives in Polymer applications.

## Validation
- `git diff --check`
- targeted jsdom repro and verification harnesses for the template trust, path mutation, sanitization, strict-template-policy, stamped ID map, and `dom-module` registry cases
- full browser `polymer test` suite was not run here because this environment does not provide a local Chrome or Chromium binary


M0nd0R:
Closing.

